PR #1, opened 2019-01-23
Fix missing entires in MAIFEST.in

pazembrz:
Drop python2 support


drjova:
could you please re-push? We need that for travis to pickup this PR

PR #5, opened 2019-02-04
records: add files support in records - INSPIR-1866

pazembrz:
Added support for Files in base record
Added support for Documents and Figures in Literature record
Added fixture which adds default location for files

Added tests for 100% coverage of new code (except cli/fixture)

Requires: https://github.com/inspirehep/inspire-schemas/pull/353


inspirehep/records/api/base.py
```diff
@@ -166,3 +179,228 @@ def hard_delete(self):
...
+            )
+
+        bucket = (
+            RecordsBuckets.query.join(Bucket)
+            .filter(
+                RecordsBuckets.record_id == record_id,
+                Bucket.default_storage_class == storage_class,
+                Bucket.default_location == location_obj.id,
+            )
+            .one_or_none()
```
drjova: isn't it enough:

```python
 records_buckets = RecordsBuckets.query.filter_by(
            record_id=self.id).first()
```

inspirehep/records/api/base.py
```diff
@@ -166,3 +179,228 @@ def hard_delete(self):
...
+        stream = fsopen(url, mode="rb")
+        # TODO: change to stream.read() when fs will be updated to > 2.0
+        # As HTTPOpener is not working with size = -1
+        # (and read() method sets this size as default)
+        # This is workaround until we will update to fs >2.0
+        data = stream._f.wrapped_file.read()
+        key = hashlib.sha1(data).hexdigest()
+        if key not in self.files.keys:
+            self.files[key] = BytesIO(data)
+        return key
```
drjova: what will happen if the download fails? It's better to through a custom error

inspirehep/records/api/base.py
```diff
@@ -166,3 +179,228 @@ def hard_delete(self):
...
+            url (str): Local url which starts with /api/files/
+
+        Returns (str): key(sha-1) of downloaded file
+        Example:
+            >>> url = '/api/files/261926f6-4923-458e-adb0/207611e7bf8a83f0739bb2e'
+            >>> self._download_file_from_lcoal_storage(url)
+                '207611e7bf8a83f0739bb2e'
+        """
+        url_splited = url.split("/")
+        file = ObjectVersion.get(bucket=url_splited[-2], key=url_splited[-1])
```
drjova: I'd move the url splitting in a separated, maybe ``static`` method, so we can easily test it.

inspirehep/records/api/base.py
```diff
@@ -166,3 +179,228 @@ def hard_delete(self):
...
+        if not file:
+            raise FileNotFoundError(f"{url} is not a valid file in local storage!")
+        file_stream = file.file.storage().open()
+        data = file_stream.read()
+        key = hashlib.sha1(data).hexdigest()
```
drjova: it'd be better to move the hashing part to a separate method so we don't have duplicated code and we can easily change in the future https://github.com/inspirehep/inspirehep/pull/5/files#diff-b05b21cfe41c0c87d42e71a8841ba419R277

```python
@staticmethod
def hash_data(data):
   ...
```

inspirehep/records/api/base.py
```diff
@@ -166,3 +179,228 @@ def hard_delete(self):
...
+        """
+        stream = fsopen(url, mode="rb")
+        # TODO: change to stream.read() when fs will be updated to > 2.0
+        # As HTTPOpener is not working with size = -1
+        # (and read() method sets this size as default)
+        # This is workaround until we will update to fs >2.0
+        data = stream._f.wrapped_file.read()
+        key = hashlib.sha1(data).hexdigest()
+        if key not in self.files.keys:
+            self.files[key] = BytesIO(data)
```
drjova: is this enough to create a new file?

inspirehep/records/api/base.py
```diff
@@ -166,3 +179,228 @@ def hard_delete(self):
...
+    def _find_and_add_file(self, url, original_url=None):
+        """Finds proper url (url or original_url) and method to download file.
+        Args:
+            url (str): Local or remote path to a file
+            original_url (str): Local or remote path to a file
+
+        Returns: Key of downloaded file, or None if file was not found
+
+        """
+        urls = [_url for _url in [url, original_url] if _url]
```
drjova: it's hard to understand what it does here 😕

inspirehep/records/api/base.py
pazembrz: It's safer this way. With old code I was able to create two buckets for record, and it was making problems. It's safer to be sure that there is only one bucket and we are not saving files to second bucket which might be invisible for the system in some use-cases.

inspirehep/records/api/base.py
pazembrz: If download fails we won't be in this part of the code. Exception `ResourceNotFoundError` will be thrown and in `_find_and_add_file` empty key will be returned which won't allow to add this file to documents but won't break other files in the list. Later on in workflow or other place  which will be calling add_files or set_files there should be check if all files was added correctly. May be I should look again in `create` and `update` methods to somehow inform about missing files.

inspirehep/records/api/base.py
pazembrz: ok

inspirehep/records/api/base.py
pazembrz: ok

inspirehep/records/api/base.py
```diff
@@ -166,3 +179,228 @@ def hard_delete(self):
...
+        # As HTTPOpener is not working with size = -1
+        # (and read() method sets this size as default)
+        # This is workaround until we will update to fs >2.0
+        data = stream._f.wrapped_file.read()
+        key = hashlib.sha1(data).hexdigest()
+        if key not in self.files.keys:
+            self.files[key] = BytesIO(data)
+        return key
+
+    def _download_file_from_lcoal_storage(self, url, **kwargs):
```
drjova:
```suggestion
    def _download_file_from_local_storage(self, url, **kwargs):
```
typo ``lcoal``

inspirehep/records/api/base.py
```diff
@@ -166,3 +179,228 @@ def hard_delete(self):
...
+
+        metadata = kwargs
+        metadata["key"] = key
+
+        if not original_url:
+            metadata["original_url"] = url
+        else:
+            metadata["original_url"] = original_url
+
+        metadata["filename"] = filename
```
drjova: filename can be ``''`` here no? https://github.com/inspirehep/inspirehep/pull/5/files#diff-b05b21cfe41c0c87d42e71a8841ba419R386

inspirehep/records/cli.py
```diff
@@ -0,0 +1,42 @@
...
+def fixtures():
+    """Command related to records in inspire"""
+
+
+@fixtures.command()
+@with_appcontext
+def init_records_store_paths():
+    """Init default file and records store location."""
+    init_default_storage_path()
+    init_records_files_storage_path()
```
drjova: we can move it to ``cli/fixutres.py`` so we can register each ``group`` independently

inspirehep/records/api/base.py
pazembrz: Filename shouldn't be empty. In schemas I allowed it to be empty for back compatibility with inspire-next, but for inspirehep it shouldn't happen so in worst case scenario (filename not provided, not in key and not in url) filename will be same as key.

inspirehep/records/api/literature.py
```diff
@@ -33,6 +41,98 @@ class LiteratureRecord(InspireRecord):
 
     pid_type = "lit"
 
+    @classmethod
+    def create(cls, data, **kwargs):
+
```
drjova: we don't need the empty space :)

tests/integration/conftest.py
```diff
@@ -147,6 +154,45 @@ def _create_pidstore(object_uuid, pid_type, pid_value):
...
 
 
+@pytest.fixture(scope="function")
+def init_files_db(db):
+    init_default_storage_path()
+    init_records_files_storage_path()
+
+
+@pytest.fixture(scope="function")
+def fsopen_mock():
```
drjova: shall we move this fixture inside ``integration/records``? It seems related to these test cases only

tests/unit/records/test_base.py
```diff
@@ -22,3 +22,106 @@ def test_strip_empty_values():
     assert "empty_string" not in data_stripped
     assert "empty_array" not in data_stripped
     assert "empty_dict" not in data_stripped
+
+
+def test_split_urls():
```
drjova: isn't it cleaner to use something like ``@pytest.mark.parametrize``

pazembrz: Actually in my opinion, tests with more complex `expected values` are not good for pytest.mark.parametrize as it's making things less readable/it creates bigger mess. But this is only my opinion and if you are not agreeing with me i can change it.

inspirehep/records/api/literature.py
```diff
@@ -18,6 +20,97 @@ class LiteratureRecord(InspireRecord):
...
     @staticmethod
     def mint(record_uuid, data):
         PidStoreLiterature.mint(record_uuid, data)
+
+    def delete(self):
+        self.set_files(force=True)
+        super().delete()
+
+    def set_files(self, documents=None, figures=None, force=False):
+        """Sets new documents and figures for record.
```
ammirate: newline :P

inspirehep/records/api/literature.py
```diff
@@ -18,6 +20,97 @@ class LiteratureRecord(InspireRecord):
...
+        Every figure or document not listed in arguments will be removed from record.
+        If you want to only add new documents, use `add_files`
+        Args:
+            documents (list): List of documents which should be set to this record
+            figures (list): List of figures which should be set to this record
+
+            Documents and figures are lists of dicts.
```
ammirate: You can just add to the doc above `documents (List[dict]): ...`

inspirehep/records/fixtures.py
```diff
@@ -0,0 +1,56 @@
...
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with INSPIRE. If not, see <http://www.gnu.org/licenses/>.
+#
+# In applying this license, CERN does not waive the privileges and immunities
+# granted to it by virtue of its status as an Intergovernmental Organization
+# or submit itself to any jurisdiction.
```
ammirate: License is different from other files :eyes:

tests/integration/records/conftest.py
```diff
@@ -0,0 +1,55 @@
...
+            )
+        elif url == "http://figure_url.cern.ch/figure2.pdf":
+            file = open(
+                f"{dir_path}/test_data/test_figure1.pdf", mode=kwargs.get("mode", "rb")
+            )
+        elif url == "http://figure_url.cern.ch/some_strange_path":
+            file = open(
+                f"{dir_path}/test_data/test_figure1.pdf", mode=kwargs.get("mode", "rb")
+            )
+        else:
```
ammirate: What about using a dictionary?
```
urls = {
    'http://document_url.cern.ch/file.pdf': 'test_data/test_document.pdf',
    'http://figure_url.cern.ch/file.png': 'test_data/test_figure2.png'
    . . .
}
```

tests/integration/records/test_api_base.py
```diff
@@ -296,3 +306,385 @@ def test_get_linked_records_in_field_with_different_pid_types(
...
+
+def test_get_bucket_multiple_same_parameters(
+    base_app, db, create_record, init_files_db
+):
+    record_factory = create_record("lit")
+    record = InspireRecord.get_record(record_factory.id)
+
+    bucket = record.get_bucket()
+    bucket_second = record.get_bucket()
+    bucket_copy = record.get_bucket()
```
ammirate: sorry but I don't get this test :(

tests/integration/records/conftest.py
pazembrz: I thought about it but i forgot. ;)

pazembrz:
It's failing only because of isort. I cannot do anything about this...

PR #16, opened 2019-02-13
Citations link

harunurhan:


ui/src/tracker.js
```diff
@@ -16,7 +19,9 @@ export function injectTrackerToHistory(history) {
 }
 
 export function trackEvent(...args) {
-  Piwik.push(['trackEvent', ...args]);
+  if (isTrackerConfigured()) { 
+    Piwik.push(['trackEvent', ...args]);
```
drjova: this was failing before?

drjova:
what about the detail view?

PR #20, opened 2019-02-15
ui: change link-color and affiliations' color

harunurhan:


ui/src/__tests__/App.test.jsx
```diff
@@ -231,7 +231,7 @@ describe('App', () => {
     expect(wrapper.find(Errors)).toExist();
   });
 
-  it('redirects to Errors when /anythingElse', () => {
+  xit('redirects to Errors when /anythingElse', () => {
```
drjova: ?

ui/src/middlewares/statePersister.js
```diff
@@ -0,0 +1,40 @@
+import { fromJS } from 'immutable';
```
drjova: ``persistentState.js`` sounds better (?)

ui/src/common/layouts/Header/Header.scss
```diff
@@ -2,15 +2,16 @@
...
+          color: white;
+          cursor: pointer;
+        }
       }
+      padding: 0 12px;
     }
-    padding: 0 12px;
   }
-}
+}
```
drjova: end of line

ui/src/middlewares/statePersister.js
harunurhan: i don't know, we have already `searchDispatcher`.

ui/src/__tests__/App.test.jsx
harunurhan: didn't have time to fix this but it is working. For some reasons started rendering `<Redirect to="/errors" .../>` instead of `<Errors .../>` after `App` is connected to store :/

PR #28, opened 2019-02-22
setup prettier with precommit

harunurhan:
Also removes prettier setup from /ui


drjova:
missing from sdist:
  .prettierrc.json
suggested MANIFEST.in rules:
  include *.json
The command "pipenv run check-manifest --ignore ".travis-*,ui/*,ui"" exited with 1.

PR #53, opened 2019-03-04
search: do not return buckets with doc_count 0

harunurhan:
* Filters out buckets of author_count aggregation where doc_count = 0
so that it they don't show up on the UI leads users to empty searches.

JIRA:INSPIR-1983

### DIFF
--- a/tests/integration/records/test_views.py
+++ b/tests/integration/records/test_views.py
@@ -186,6 +186,35 @@ def test_literature_facets(api_client, db, create_record):
     assert expected_facet_keys == response_data_facet_keys
 
 
+def test_literature_facets_author_count_does_not_have_empty_bucket(
+    api_client, db, create_record
+):
+    response = api_client.get("/literature/facets/")
+    response_data = json.loads(response.data)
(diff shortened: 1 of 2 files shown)


tests/integration/records/test_views.py
```diff
@@ -186,6 +186,16 @@ def test_literature_facets(api_client, db, create_record):
     assert expected_facet_keys == response_data_facet_keys
 
 
+def test_literature_facets_author_count_does_not_have_empty_bucket(
+    api_client, db, create_record
+):
+    response = api_client.get("/literature/facets/")
+    response_data = json.loads(response.data)
+    response_status_code = response.status_code
```
monaawi: you should assert on the status code

tests/integration/records/test_views.py
```diff
@@ -186,6 +186,16 @@ def test_literature_facets(api_client, db, create_record):
...
+    author_count_agg = response_data.get("aggregations")["author_count"]
+    assert author_count_agg["buckets"] == []
+
```
monaawi: maybe you could also add a test in which the author_count is more than 0 and check that the bucket is there?

PR #56, opened 2019-03-04
Track user role

harunurhan:


michamos:
Won't this approach only affect the custom events and not the regular page visits through history? wouldn't it be better to use [custom variables](https://matomo.org/docs/custom-variables/) instead (I know this should have been discussed during the refinement, but I wasn't there, so couldn't give my opinion).

harunurhan:
Yes, this will only group custom events that we track based on user's role. There was a mention this feature during refinement as far as I remember but this approach gives easier and more obvious reporting for custom events, that's why we have refined this way.

harunurhan:
superseded by #57

PR #64, opened 2019-03-06
[INSPIR-1878] Add Indexing 

pazembrz:


inspirehep/records/indexer/tasks.py
```diff
@@ -0,0 +1,122 @@
...
+            f"Cannot pull record {pid_type}:{pid_value}in version {record_version}."
+            f"Current version: {record.model.version_id}."
+        )
+        raise StaleDataError()
+    return record
+
+
+@shared_task(ignore_result=False, bind=True)
+def process_references_for_record(
+    pid_type=None, pid_value=None, record_version=None, record=None
```
ammirate: This signature needs more checking if you want to keep everything nullable. Otherwise You should change it :boom:

inspirehep/records/indexer/tasks.py
```diff
@@ -0,0 +1,122 @@
...
+    )
+    try:
+        record = get_record(pid_type, pid_value, record_version)
+    except (NoResultFound, StaleDataError) as e:
+        logger.warn(
+            f"Record {pid_type}:{pid_value} not yet at version {record_version} on DB"
+        )
+        backoff = 2 ** (self.request.retries + 1)
+        if self.max_retries < self.request.retries + 1:
+            logger.warn(f"({pid_value}) - Failing - too many retries")
```
ammirate: Should we `raise` again in this case? Not sure tough :)

inspirehep/records/api/base.py
```diff
@@ -684,3 +722,159 @@ def _add_file(self, url, original_url=None, filename=None, **kwargs):
...
+        """
+        arguments_for_indexer = {
+            "pid_value": record["control_number"],
+            "pid_type": cls.pid_type,
+            "deleted": deleted or record.get("deleted", False),
+        }
+        return arguments_for_indexer
+
+    @property
+    def _previous_version(self):
```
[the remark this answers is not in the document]
pazembrz: Because it's used only in specified method. Shouldn't be used outside without any special reason.

inspirehep/records/api/base.py
```diff
@@ -684,3 +722,159 @@ def _add_file(self, url, original_url=None, filename=None, **kwargs):
...
+        try:
+            ids_oldest = set(
+                self._previous_version.get_records_pid_from_field("references.record")
+            )
+        except AttributeError:
+            return []
+
+        return set.symmetric_difference(ids_latest, ids_oldest)
+
+    def _dump_for_es(self):
```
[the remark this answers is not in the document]
pazembrz: Not really. It's usefull if you would like to check/modify/add dictionary sent to serializer (before serialization) which is usefull in dump_for_es in literature where we are adding `_ui_display` field for es.

inspirehep/records/api/receivers.py
```diff
@@ -0,0 +1,44 @@
...
+    """
+    for model_instance, change in changes:
+        logger.debug("index_after_commit hook")
+        if isinstance(model_instance, RecordMetadata):
+            logger.debug(
+                f"Model instance ({model_instance.id}) is correct. Processing..."
+            )
+            if change in ("insert", "update", "delete"):
+                # InspireRecord.get_record(model_instance.id).index()
+                logger.debug(f"Change type is {change}.")
```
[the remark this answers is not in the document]
pazembrz: Why? By default we are not logging debug, and debug is very useful when debugging ;) But i should change f"" string to not process variables when debug is not used.

inspirehep/records/indexer/base.py
```diff
@@ -0,0 +1,103 @@
...
+        """
+        _index, _doc_type = self.record_to_index(record)
+        if not doc_type:
+            doc_type = _doc_type
+        if not index:
+            index = _index
+        return {
+            "_op_type": "index",
+            "_index": index,
+            "_type": doc_type,
```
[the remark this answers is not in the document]
pazembrz: I would prefer not to put logic in return statement

inspirehep/records/indexer/base.py
```diff
@@ -0,0 +1,103 @@
...
+            records_uuids(list[str): List of strings which are UUID's of records
+                to reindex
+            request_timeout(int): Maximum time after which   es will throw an exception
+
+        Returns:
+            dict: dict with success count and failure list
+                (with uuids of failed records)
+
+        """
+        from inspirehep.records.api import InspireRecord
```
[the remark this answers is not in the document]
pazembrz: Yup. Indexer itself is imported in records/base.py and this import here is used only in bulk reindex (circular imports)

inspirehep/records/indexer/tasks.py
pazembrz: I think that ending this with exception is more natural than failing silently.

pazembrz:
Why tests failed on travis if exit code was 0? `The command "pipenv run py.test --random-order tests/unit" exited with 0.`

Pipfile
```diff
@@ -26,21 +28,23 @@ uwsgi = ">=2.0"
...
+flask-sqlalchemy = {git = "https://github.com/inspirehep/flask-sqlalchemy.git"}
 node-semver = "<0.2.0,>=0.1.1"
 pylint = "*"
 pybtex = "*"
+pytest = ">=3.3.1,<4.0.0"
 pytest-cov = "==2.6.1"
 remote-pdb = "*"
 cryptography = "<2.6"
-
+#  msgpack = "*"  #  We want to use it but it might mess with invenio-indexer which is forcing msgpack-python
```
harunurhan: Should be either `FIXME` if we want it eventually or just be removed.

inspirehep/cli/indexer.py
```diff
@@ -0,0 +1,222 @@
...
+    )
+    return query
+
+
+def _prepare_logdir(log_path):
+    if not path.exists(path.dirname(log_path)):
+        makedirs(path.dirname(log_path))
+
+
+# def _dump_errors_to_file(errors, log_file_path, tasks_uuids, msg='Check errors in log file'):
```
harunurhan: should be remove eventually, just putting this a reminder.

inspirehep/records/api/receivers.py
```diff
@@ -0,0 +1,44 @@
...
+@models_committed.connect
+def index_after_commit(sender, changes):
+    """Index a record in ES after it was committed to the DB.
+
+    This cannot happen in an ``after_record_commit`` receiver from Invenio-Records
+    because, despite the name, at that point we are not yet sure whether the record
+    has been really committed to the DB.
+    """
+    for model_instance, change in changes:
+        logger.debug("index_after_commit hook")
```
harunurhan: `.debug` logs are too many, I don't think we need them all.

inspirehep/records/ext.py
```diff
@@ -19,3 +19,4 @@ def __init__(self, app=None):
     def init_app(self, app):
         """Flask application initialization."""
         app.extensions["inspirehep-records"] = self
+        from inspirehep.records.indexer.base import InspireRecordIndexer  # noqa: F401
```
harunurhan: is this necessary, as far as I can see this file has only the class definition, nothing is executed when you import.

inspirehep/records/marshmallow/literature/base.py
```diff
@@ -124,3 +129,163 @@ def get_ui_display(self, data):
...
+    author_count = fields.Method("populate_author_count")
+    authors = fields.Method("preprocess_authors")
+    bookautocomplete = fields.Method("populate_bookautocomplete")
+    earliest_date = fields.Method("populate_earliest_date")
+    facet_inspire_doc_type = fields.Method("populate_inspire_document_type")
+    facet_author_name = fields.Method("populate_facet_author_name")
+    id = fields.UUID(dump_only=True)
+    thesis_info = fields.Nested(ThesisInfoSchemaForESV1, dump_only=True)
+
+    def populate_abstract_source_suggest(self, record):
```
harunurhan: All those methods should not be called populate anymore as it returns the value now rather than directly updating the record.

Could be prefixed with `get`

inspirehep/records/marshmallow/literature/base.py
```diff
@@ -124,3 +129,163 @@ def get_ui_display(self, data):
...
+        """Prepares record for ``abstract_source_suggest`` field."""
+        abstracts = record.get("abstracts", [])
+
+        for abstract in abstracts:
+            source = abstract.get("source")
+            if source:
+                abstract.update({"abstract_source_suggest": {"input": source}})
+        return abstracts
+
+    def populate_earliest_date(self, record):
```
harunurhan: ditto

inspirehep/records/marshmallow/literature/base.py
```diff
@@ -124,3 +129,163 @@ def get_ui_display(self, data):
...
+                [force_list(record.get_value(path)) for path in date_paths]
+            )
+        ]
+        if dates:
+            result = earliest_date(dates)
+            if result:
+                earliest_date_found = result
+        return earliest_date_found
+
+    def populate_author_count(self, record):
```
harunurhan: ditto
(naming)

inspirehep/records/marshmallow/literature/base.py
```diff
@@ -124,3 +129,163 @@ def get_ui_display(self, data):
...
+        authors = record.get("authors", [])
+
+        authors_excluding_supervisors = [
+            author
+            for author in authors
+            if "supervisor" not in author.get("inspire_roles", [])
+        ]
+        return len(authors_excluding_supervisors)
+
+    def preprocess_authors(self, record):
```
harunurhan: ditto
(needs its own schema)

inspirehep/records/marshmallow/literature/base.py
```diff
@@ -124,3 +129,163 @@ def get_ui_display(self, data):
...
         except json.JSONDecodeError:
             return {}
+
+
+class LiteratureESEnhancementV1(LiteratureMetadataSchemaV1):
+    """Elasticsearch serialzier"""
+
+    _created = fields.DateTime(dump_only=True, attribute="created")
+    _updated = fields.DateTime(dump_only=True, attribute="updated")
+    abstract_source_suggest = fields.Method("populate_abstract_source_suggest")
```
harunurhan: must be `abstracts` instead of `abstract_source_suggest`

inspirehep/records/marshmallow/literature/base.py
```diff
@@ -124,3 +129,163 @@ def get_ui_display(self, data):
...
+        field and generating name variations"""
+        processed_authors = []
+        for index, author in enumerate(record.get("authors", [])):
+            author = self.prepare_author_full_name_unicode_normalized(author)
+            author = self.populate_name_variations_for_author(author)
+            processed_authors.append(author)
+        return processed_authors
+
+    @staticmethod
+    def prepare_author_full_name_unicode_normalized(author):
```
harunurhan: ditto

inspirehep/records/marshmallow/literature/base.py
```diff
@@ -124,3 +129,163 @@ def get_ui_display(self, data):
...
+        """Prepares data for ``author.full_name_normalized`` field."""
+        full_name = str(author["full_name"])
+        author.update(
+            {"full_name_unicode_normalized": normalize("NFKC", full_name).lower()}
+        )
+        return author
+
+    def populate_inspire_document_type(self, record):
```
harunurhan: ditto

inspirehep/records/marshmallow/literature/base.py
```diff
@@ -124,3 +129,163 @@ def get_ui_display(self, data):
...
+        dates = [
+            str(el)
+            for el in chain.from_iterable(
+                [force_list(record.get_value(path)) for path in date_paths]
+            )
+        ]
+        if dates:
+            result = earliest_date(dates)
+            if result:
+                earliest_date_found = result
```
harunurhan: `return result`

inspirehep/records/marshmallow/literature/base.py
```diff
@@ -124,3 +129,163 @@ def get_ui_display(self, data):
...
+        return earliest_date_found
```
harunurhan: `return missing`

inspirehep/records/marshmallow/literature/base.py
```diff
@@ -124,3 +129,163 @@ def get_ui_display(self, data):
...
+    facet_inspire_doc_type = fields.Method("populate_inspire_document_type")
+    facet_author_name = fields.Method("populate_facet_author_name")
+    id = fields.UUID(dump_only=True)
+    thesis_info = fields.Nested(ThesisInfoSchemaForESV1, dump_only=True)
+
+    def populate_abstract_source_suggest(self, record):
+        """Prepares record for ``abstract_source_suggest`` field."""
+        abstracts = record.get("abstracts", [])
+
+        for abstract in abstracts:
```
harunurhan: it's not OK to go nested inside `field.Method`, you should create a new schema instead. For example for this case:

`AbstractSchema` with all fields `fields.Raw` and `abstract_source_suggest = fields.Method()` and then use it up there `abstracts: fields.Nested(AbstractSchema, many=True)`

tests/conftest.py
```diff
@@ -6,6 +6,4 @@
 # the terms of the MIT License; see LICENSE file for more details.
 
 """INSPIRE module that adds more fun to the platform."""
-
-
```
harunurhan: irrelevant

tests/helpers/providers/record_provider.py
```diff
@@ -51,7 +51,23 @@ def author_record():
             "_collections": ["Authors"],
         }
 
-    def record(self, record_type, data=None, with_control_number=False):
+    @staticmethod
+    def add_citations(citation_records=[]):
```
harunurhan: misleading optional value, it does nothing and useless if nothing passed, maybe it's better to just fail.

harunurhan: something like would be better `get_list_of_refereces_from_record_ids`

tests/helpers/providers/record_provider.py
```diff
@@ -51,7 +51,23 @@ def author_record():
...
+        data = []
+        for record in citation_records:
+            data.append(
+                {"record": {"$ref": f"http://localhost:5000/api/literature/{record}"}}
+            )
+        return {"references": data}
```
harunurhan: and return only `data` here.

tests/helpers/providers/record_provider.py
```diff
@@ -51,7 +51,23 @@ def author_record():
...
+
+    def record(
+        self,
+        record_type,
+        data=None,
+        with_control_number=False,
+        citations=[],
```
harunurhan: `recids_to_cite`

inspirehep/records/ext.py
pazembrz: Ouch, this is my mistake. I should import receivers here to register them properly.

inspirehep/records/marshmallow/literature/base.py
pazembrz: Not really, earliest_date_found keeps default data pulled from record['earliest_date'] and should be used when result is empty. I could jus `return result or earliest_date_found` but I'm trying not to put logic inside return statement.

inspirehep/records/marshmallow/literature/base.py
pazembrz: ?

inspirehep/records/api/base.py
```diff
@@ -224,8 +233,24 @@ def get_uuid_from_pid_value(cls, pid_value, pid_type=None):
...
+        for _cls in cls.__subclasses__():
+            records_map[_cls.pid_type] = _cls
+        return records_map
+
+    @classmethod
+    def get_record(cls, id_, with_deleted=False):
+        record = super().get_record(str(id_), with_deleted)
+        type_from_schema = PidStoreBase.get_pid_type_from_schema(record["$schema"])
+        if record.pid_type is None or record.pid_type != type_from_schema:
+            return cls.get_subclasses()[type_from_schema].get_record(id_, with_deleted)
```
michamos: Is the point of all this to get a `LiteratureRecord` when I do `InspireRecord.get_record(...)`? doesn't this require the subclass to be declared in a file that happens to have been read? what happens if I do `LiteratureRecord.get_record(...)`, does it still work?

inspirehep/records/api/base.py
```diff
@@ -250,12 +275,28 @@ def create(cls, data, **kwargs):
             record = super().create(data, id_=id_, **kwargs)
         return record
 
-    @classmethod
-    def get_linked_records_in_field(cls, data, path):
```
michamos: `get_linked_pids_in_field` for consistency?

inspirehep/records/api/base.py
```diff
@@ -250,12 +275,28 @@ def create(cls, data, **kwargs):
...
+    def get_records_pid_from_field(self, path):
+        """Returns list of tuples (pid_type, pid_value) of all linked records in field
```
michamos: `Return a list of (pid_type, pid_value) tuples for all records referenced in the field at the given path`.

I know it's written in indicative mood in the whole file, but the first line of the docstring should be in imperative mood to conform with PEP 257 (I get really annoyed by these small things :smile_cat: ).

inspirehep/records/api/base.py
```diff
@@ -250,12 +275,28 @@ def create(cls, data, **kwargs):
...
+        from specified path
+        Args:
+            path (str): the path of the linked records.
```
michamos: maybe explain that `path` needs to be understood as a `get_value` argument? we probably need a specific name for this thing.

inspirehep/records/api/base.py
```diff
@@ -250,12 +275,28 @@ def create(cls, data, **kwargs):
...
+        Returns:
+            list: tuples containing (pid_type, pid_value) of the linked records
+
+        """
+        full_path = ".".join([path, "$ref"])
+        pids = force_list(
```
michamos: what's the `force_list` for, given that its argument is already a list? (I know you didn't write this).

inspirehep/records/api/base.py
```diff
@@ -684,3 +720,173 @@ def _add_file(self, url, original_url=None, filename=None, **kwargs):
 
         metadata["url"] = file_path
         return metadata
+
+    def get_enhanced_data(self, serializer=None):
```
michamos: Wait, I thought we shouldn't say "enhancement" any longer?

inspirehep/records/api/base.py
```diff
@@ -684,3 +720,173 @@ def _add_file(self, url, original_url=None, filename=None, **kwargs):
...
+        """Prepares serialized record for elasticsearch
+        Args:
+            serializer(Schema): Schema which should be used to serialize/enchance
```
michamos: typo `enhance`.

inspirehep/records/api/base.py
```diff
@@ -684,3 +720,173 @@ def _add_file(self, url, original_url=None, filename=None, **kwargs):
...
+        Returns:
+            dict: Data serialized/enhanced by serializer.
+                Do no return data which is not specified in Schema
+
+        """
+        if not self.es_serializer and not serializer:
+            return None
```
michamos: shouldn't this be an exception?

inspirehep/records/api/base.py
```diff
@@ -684,3 +720,173 @@ def _add_file(self, url, original_url=None, filename=None, **kwargs):
...
         metadata["url"] = file_path
         return metadata
+
+    def get_enhanced_data(self, serializer=None):
+        """Prepares serialized record for elasticsearch
+        Args:
+            serializer(Schema): Schema which should be used to serialize/enchance
+        Returns:
+            dict: Data serialized/enhanced by serializer.
+                Do no return data which is not specified in Schema
```
michamos: `It does not return ...`

inspirehep/records/api/base.py
```diff
@@ -684,3 +720,173 @@ def _add_file(self, url, original_url=None, filename=None, **kwargs):
...
+            dict: Enhanced record data
+        """
+        if not self.ui_serializer and not serializer:
+            return None
+        if not serializer:
+            serializer_module = importlib.import_module(
+                self.__module__.replace("api", "marshmallow")
+            )
+            serializer = getattr(serializer_module, self.ui_serializer)
+        return serializer().dump(self).data
```
michamos: isn't this exactly the same function as the previous one, except for the use of `self.ui_serializer` instead of of `self.es_serializer`?

inspirehep/records/api/base.py
```diff
@@ -684,3 +720,173 @@ def _add_file(self, url, original_url=None, filename=None, **kwargs):
...
+
+    def _index(self, delete=None):
+        """Runs index in current process.
+            This is main logic of indexing called inside worker
+            if called from self.index() but when called from self.index()
+            it will run in a worker.
```
michamos: This sentence doesn't make sense to me. Maybe simply "This is the same logic as in ``:ref:`InspireRecord.index` which additionally dispatches indexing to a Celery worker``?

inspirehep/records/api/base.py
```diff
@@ -684,3 +720,173 @@ def _add_file(self, url, original_url=None, filename=None, **kwargs):
...
+            return None
+        if not serializer:
+            serializer_module = importlib.import_module(
+                self.__module__.replace("api", "marshmallow")
+            )
+            serializer = getattr(serializer_module, self.ui_serializer)
+        return serializer().dump(self).data
+
+    def _index(self, delete=None):
+        """Runs index in current process.
```
michamos: `Index the current record in the current process` + newline after.

inspirehep/records/api/base.py
```diff
@@ -684,3 +720,173 @@ def _add_file(self, url, original_url=None, filename=None, **kwargs):
...
+        Args:
+            delete: set to True if record is deleted,
+                If not set, tries to determine this automagically
+        Returns:
+            celery.result.AsyncResult: Task itself
+        """
+        logger.error(f"Indexing record {self.id}")
+        indexing_args = self._record_index(self, deleted=delete)
+        indexing_args["record_version"] = self.model.version_id
+        task = index_record.delay(**indexing_args)
```
michamos: instead of doing local imports in the `index_record` task, which is not very nice, you can call the task by name here, using [`send_task`](http://docs.celeryproject.org/en/latest/reference/celery.html#celery.Celery.send_task)

inspirehep/records/api/base.py
```diff
@@ -684,3 +720,173 @@ def _add_file(self, url, original_url=None, filename=None, **kwargs):
...
+        logger.info(f"Record {self.id} send for indexing")
+        return task
+
+    @classmethod
+    def _record_index(cls, record, _id=None, deleted=None):
```
michamos: do you ever need to call `_record_index` on something that's not an `InspireRecord`? in any case it should be a `staticmethod` at most, as `cls` is not used.

inspirehep/records/api/base.py
```diff
@@ -684,3 +720,173 @@ def _add_file(self, url, original_url=None, filename=None, **kwargs):
...
+        Returns:
+            Set[Tuple[str, int]]: pids of references changed from the previous
+            version.
+        """
+        try:
+            prev_version = self._previous_version
+        except AttributeError:
+            prev_version = {}
+
+        changed_deleted_status = self.get("deleted", False) ^ prev_version.get(
```
michamos: I would use `and` here instead of `^`

Pipfile
```diff
@@ -4,6 +4,8 @@ verify_ssl = true
 name = "pypi"
 
 [packages]
+celery = "<=4.2.1"
+click = "<7.0.0"  # Click in version 7 have broken progressbar in python 3
```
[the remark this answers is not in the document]
pazembrz: Yes, there is also a PR already in click repo, but it's not yet merged: https://github.com/pallets/click/pull/1163. I will ad this also to the code as a comment.

inspirehep/records/api/literature.py
```diff
@@ -80,7 +85,8 @@ def add_files(self, documents=None, figures=None):
         """Public method for adding documents and figures
 
         Args:
-            documents (list[dict]): List of documents which should be added to this record
+            documents (list[dict]): List of documents which should be added to this
```
michamos: `list` -> `List` as `list == List[Any]`.

Pipfile
```diff
@@ -26,21 +28,24 @@ uwsgi = ">=2.0"
 marshmallow = "<3.0.0"
 # do-json :( needs flask v0.x
 flask-shell-ipython = "~=0.3,<0.4.0" # version 0.4.0 requires `flask >= 1.0`
+flask-sqlalchemy = {git = "https://github.com/inspirehep/flask-sqlalchemy.git"}
 node-semver = "<0.2.0,>=0.1.1"
 pylint = "*"
 pybtex = "*"
+pytest = ">=3.3.1,<4.0.0"
```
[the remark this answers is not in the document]
pazembrz: Pytest 4.x breaks integration-async tests as old version of cellery (to which we are forced to pin) is not compatible with pytest 4.x (in which api was slightly changed). 
Do we need it on production? I do not think so but pytest-cov is also there and it installs pytest so I was forced to pin pytest also in there.

inspirehep/records/api/literature.py
```diff
@@ -117,3 +123,8 @@ def add_files(self, documents=None, figures=None):
                 files.append(metadata)
         super().update(builder.record)
         return files
+
+    def _dump_for_es(self):
+        serialized_data = super()._dump_for_es()
+        serialized_data["_ui_display"] = json.dumps(self.get_ui_data())
```
michamos: why isn't this part of the upstream class, if you're not redefining `get_ui_data` here? all formats will eventually have a `_ui_display` field.

inspirehep/celery.py
```diff
@@ -5,8 +5,12 @@
...
+import logging
 
 from flask_celeryext import create_celery_app
 from invenio_app.factory import create_api
 
-celery = create_celery_app(create_api())
+celery = create_celery_app(create_api(LOGGING_SENTRY_CELERY=True))
+
+# We don't want to log to Sentry backoff errors?
+logging.getLogger("backoff").propagate = 0
```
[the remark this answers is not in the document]
pazembrz: not really. I copied it from next without checking. I will remove it.

inspirehep/records/api/receivers.py
```diff
@@ -0,0 +1,44 @@
...
+        if isinstance(model_instance, RecordMetadata):
+            logger.debug(
```
michamos: I would take this whole thing out and put it in a separate function/method, to leave only something like
```
if isinstance(model_instance, RecordMetadata):
    `after_record_metadata_commit(model_instance, change)`
```

inspirehep/cli/indexer.py
```diff
@@ -0,0 +1,183 @@
...
+from invenio_db import db
+from invenio_pidstore.models import PersistentIdentifier, PIDStatus
+from invenio_records.models import RecordMetadata
+from sqlalchemy import or_, not_, type_coerce, cast, String
+from sqlalchemy.dialects.postgresql import JSONB
+
+from inspirehep.records.indexer.tasks import batch_index
+
+
+logger = logging.getLogger()
```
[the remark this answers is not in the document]
pazembrz: Not in here. This is a cli command and this should be treated as a root logger.

inspirehep/records/indexer/base.py
```diff
@@ -0,0 +1,113 @@
...
+                * internal
+                * external
+                * external_gt
+                * external_gte
+
+        Returns:
+            dict: dict with preprocessed record for bulk indexing
+
+        """
+        _index, _doc_type = self.record_to_index(record)
```
michamos: `index_from_record`, `doc_type_from_record`?

inspirehep/records/indexer/base.py
```diff
@@ -0,0 +1,113 @@
...
+    ):
+        """Process basic data required for indexing record during bulk indexing
+
+        Args:
+            record(InspireRecord): Proper inspire record record object
+            version_type(str): Proper ES versioning type:
+                * internal
+                * external
+                * external_gt
+                * external_gte
```
michamos: missing `args` (there should be an automated check for this). We should move to something a bit more automatic. Maybe https://github.com/agronholm/sphinx-autodoc-typehints ?

inspirehep/records/indexer/base.py
```diff
@@ -0,0 +1,113 @@
...
+        def actions():
+            for record_uuid in records_uuids:
+                try:
+                    record = InspireRecord.get_record(record_uuid)
+                    record.validate()
+                    if record.get("deleted", False):
+                        continue
+                    yield self._process_bulk_record_for_index(record)
+                except NoResultFound:
+                    logger.warning(f"Record {record_uuid} failed to load!")
```
michamos: Isn't this an `error`?

inspirehep/records/indexer/base.py
```diff
@@ -0,0 +1,113 @@
...
+                to reindex
+            request_timeout(int): Maximum time after which   es will throw an exception
+
+        Returns:
+            dict: dict with success count and failure list
+                (with uuids of failed records)
+
+        """
+        from inspirehep.records.api import InspireRecord
+        def actions():
```
michamos: I would unnest this and have a `_bulk_action(record_uuid)` in the indexer. This can to be a function from a record to an action and the iteration is trivial, so I would do the iteration in the caller with a simple generator/list comprehension.

inspirehep/records/indexer/base.py
```diff
@@ -0,0 +1,113 @@
...
+            es,
+            actions(),
+            request_timeout=request_timeout,
+            raise_on_error=False,
+            raise_on_exception=False,
+        )
+
+        return {
+            "success": success,
+            "failures": [failure for failure in failures or []],
```
michamos: so failures is supposed to be a `list` of `list`s?

inspirehep/records/indexer/tasks.py
```diff
@@ -0,0 +1,111 @@
...
+from sqlalchemy import tuple_
+from sqlalchemy.orm.exc import StaleDataError, NoResultFound
+
+from inspirehep.records.errors import MissingCitedRecordError, MissingArgumentError
+from inspirehep.records.indexer.base import InspireRecordIndexer
+
+logger = logging.getLogger(__name__)
+
+
+def index_record_task(record):
```
michamos: I'm a bit confused by this. It's called `task` and in the `tasks` file, but it's not a task, it's something that calls a task. Do you really need this here? maybe put it somewhere else?

inspirehep/records/indexer/tasks.py
```diff
@@ -0,0 +1,111 @@
...
+    uuid = record.id
+
+    index_record.delay(
+        uuid=uuid,
+        record_version=record.model.version_id,
+        deleted=record.get("deleted", False),
```
michamos: isn't this the automatically determined value if you don't pass anything for `deleted` anyway?

inspirehep/records/indexer/tasks.py
```diff
@@ -0,0 +1,111 @@
...
+        logger.info(
+            f"Cannot pull record {uuid}in version {record_version}."
+            f"Current version: {record.model.version_id}."
+        )
+        raise StaleDataError()
+    return record
+
+
+@shared_task(ignore_result=False, bind=True)
+def process_references_for_record(uuid=None, record_version=None, record=None):
```
michamos: docstring?

inspirehep/records/indexer/tasks.py
```diff
@@ -0,0 +1,111 @@
...
+        return batch_index(uuids)
+
+    else:
+        raise MissingCitedRecordError(
+            f"Cited records to reindex not found:\nuuids: {uuids}"
+        )
+
+
+@shared_task(ignore_result=False, bind=True, max_retries=6)
+def index_record(self, uuid, record_version=None, deleted=None):
```
michamos: docstring?

inspirehep/records/marshmallow/literature/base.py
```diff
@@ -124,3 +131,152 @@ def get_ui_display(self, data):
...
+        dates = [
+            str(el)
+            for el in chain.from_iterable(
+                [force_list(record.get_value(path)) for path in date_paths]
+            )
+        ]
+        if dates:
+            result = earliest_date(dates)
+
+        earliest_date_found = result or record.get("earliest_date", missing)
```
michamos: `earliest_date` is not in the schema, so how can it possibly be in the record?

inspirehep/records/marshmallow/literature/base.py
```diff
@@ -124,3 +131,152 @@ def get_ui_display(self, data):
...
+            "thesis_info.defense_date",
+            "publication_info.year",
+            "legacy_creation_date",
+            "imprints.date",
+        ]
+
+        dates = [
+            str(el)
+            for el in chain.from_iterable(
+                [force_list(record.get_value(path)) for path in date_paths]
```
michamos: you don't need the `[` and `]` as `chain.from_iterable` takes an iterable.

inspirehep/records/api/base.py
pazembrz: I will ask another question. Is there any point of receiving InspireRecord object at all if all the logic is in subclasses?
If you will import anything from records.api it will automatically import everything as all subclasses are imported in records.__init__ already.
Yes, it works correctly, there are even tests for this :)

inspirehep/cli/indexer.py
```diff
@@ -0,0 +1,183 @@
...
+            PersistentIdentifier.pid_type.in_(pid_types),
+            PersistentIdentifier.object_type == "rec",
+            PersistentIdentifier.status == PIDStatus.REGISTERED,
+            or_(
+                not_(type_coerce(RecordMetadata.json, JSONB).has_key("deleted")),
+                RecordMetadata.json["deleted"] == cast(False, JSONB),
+            ),
+        )
+    )
+    return query
```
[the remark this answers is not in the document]
pazembrz: I tried to do this with query builder `InspireRecord.query_builder().filter(PersistentIdentifier.status == PIDStatus.REGISTERED).not_deleted().no_duplicates().query()` (I'm not sure if this is 100% same query) but it took ~3 minutes on 1000 records on my local machine, the original query took less than a second. So there might be something wrong with query builder or with my query.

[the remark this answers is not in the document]
michamos: I think the `no_duplicates` is not present in the current version of the query, and on the other hand `pid_types` filter is. You can always `print` the query objects to compare.

inspirehep/records/api/base.py
pazembrz: Yest it's very similar.

inspirehep/records/api/base.py
```diff
@@ -684,3 +720,173 @@ def _add_file(self, url, original_url=None, filename=None, **kwargs):
...
+        Args:
+            delete: set to True if record is deleted
+
+        Returns:
+            dict: ES info about indexing
+
+        """
+        if self._schema_type != self.pid_type:
+            logger.warning(
+                f"Record {self.id} is wrapped in {self.__name__} class {self.pid_type} "
```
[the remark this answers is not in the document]
pazembrz: Yes. currently we have:
InspireRecord.pid_type = None
LiteratureRecord.pid_type = 'lit'
AuthorsRecord.pid_type = 'aut'

inspirehep/records/api/base.py
```diff
@@ -684,3 +720,173 @@ def _add_file(self, url, original_url=None, filename=None, **kwargs):
...
+                f"type, but $schema says that this is {self._schema_type} type object!"
```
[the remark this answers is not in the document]
pazembrz: self._schema_type is a property, it just runs `PidStoreBase.get_pid_type_from_schema(self["$schema"])`

inspirehep/records/api/base.py
```diff
@@ -684,3 +720,173 @@ def _add_file(self, url, original_url=None, filename=None, **kwargs):
...
+            )
+        if delete:
+            result = InspireRecordIndexer().delete(self)
+        else:
+            result = InspireRecordIndexer().index(self)
+        logger.info(f"Indexing finished: {result}")
+        return result
+
+    def index(self, delete=None):
+        """Reindexing task sent to `index_record`
```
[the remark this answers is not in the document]
pazembrz: That was already changed. I think that you were reviewing old commit.

inspirehep/records/api/base.py
pazembrz: I'm unable to call task with `send_task` inside tests, `Task of kind 'index_record' never registered, please make sure it's imported.`... I will investigate this

michamos: you need the fully qualified name though, something like `inspirehep.indexer.tasks.index_record`. You also need to set up the app properly in the tests so it knows about this task.

inspirehep/records/api/receivers.py
pazembrz: But this is already like a separate function. It's only doing this one thing.

inspirehep/records/api/literature.py
pazembrz: [Sphinx Example Google Style Python Docstrings](https://sphinx.readthedocs.io/en/master/usage/extensions/example_google.html):
`
    param3 (list(str)): Description of 'param3'.
`

inspirehep/records/indexer/base.py
pazembrz: :+1:

inspirehep/records/indexer/base.py
pazembrz: Right. Changed to error message.

inspirehep/records/indexer/base.py
pazembrz: It's list of strings usually but maybe I should just return failures itself.

inspirehep/cli/indexer.py
```diff
@@ -0,0 +1,183 @@
...
+        pid_type (List[str]): array of PID types, allowed: lit, con, exp, jou,
+            aut, job, ins
+        batch_size (int): number of documents per batch sent to workers.
+        queue_name (str): name of the celery queue
+        log_path (str): path of the indexing logs
+
+    Returns:
+        None
+    """
+    if log_path:
```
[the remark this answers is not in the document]
pazembrz: log_path can be set as empty manually
`
$ python test.py -f " "

`

inspirehep/records/api/base.py
pazembrz: but this will give different result. We want to check if state of deleted field has changed between versions.
`
>True ^ True  -> False
>True ^ False  -> True
>False ^ True -> True
>False ^ False -> False
`

`
>True and True -> True
>True and False -> False
>False and True -> False
>False and False -> False
`

I could use `!=` instead of `^` maybe, but does it matter?

inspirehep/records/indexer/tasks.py
pazembrz: It was leftover. Removed.

inspirehep/records/marshmallow/literature/base.py
pazembrz: I don't see how I could populate this field with nested. It requires access to `document_type`, `publication_type` and `refereed` fields from record itself.

tests/helpers/providers/record_provider.py
pazembrz: Value was changed, now it's not misleading.

inspirehep/records/api/base.py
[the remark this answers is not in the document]
michamos: It's a text improvement suggestion, add `with` :smiley:

inspirehep/records/api/base.py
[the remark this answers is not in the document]
michamos: also text improvement suggestion.

Pipfile
```diff
@@ -42,7 +53,7 @@ coverage = ">=4.4.1"
 isort = ">=4.3"
 mock = ">=2.0.0"
 pydocstyle = ">=2.0.0"
-pytest = ">=3.3.1"
+pytest = ">=3.3.1,<4.0.0"
```
drjova: why we have pytest twice? :)

inspirehep/records/marshmallow/literature/base.py
```diff
@@ -124,3 +135,106 @@ def get_ui_display(self, data):
...
+    _created = fields.DateTime(dump_only=True, attribute="created")
+    _updated = fields.DateTime(dump_only=True, attribute="updated")
+    abstracts = fields.Nested(AbstractSource, dump_only=True, many=True)
+    author_count = fields.Method("get_author_count")
+    authors = fields.Nested(AuthosInfoSchemaForES, dump_only=True, many=True)
+    bookautocomplete = fields.Method("get_bookautocomplete")
+    earliest_date = fields.Method("get_earliest_date")
+    facet_inspire_doc_type = fields.Method("get_inspire_document_type")
+    facet_author_name = fields.Method("get_facet_author_name")
+    id = fields.UUID(dump_only=True)
```
drjova: we shouldn't use build-ins

inspirehep/records/marshmallow/literature/common/author.py
```diff
@@ -39,3 +42,31 @@ def get_last_name(self, data):
             return names[0] or missing
 
         return missing
+
+
+class AuthorAutocompleteSchema(Schema):
+    input = fields.Method("generate_name_variations")
```
drjova: we shouldn't use build-ins

inspirehep/records/utils.py
```diff
@@ -0,0 +1,39 @@
...
+#
+# inspirehep is free software; you can redistribute it and/or modify it under
+# the terms of the MIT License; see LICENSE file for more details.
+
+"""INSPIRE module that adds more fun to the platform."""
+from inspire_utils.name import ParsedName
+from inspire_utils.record import get_value, get_values_for_schema
+
+
+def get_pid_from_record_uri(record_uri):
```
drjova: https://github.com/inspirehep/inspirehep/blob/master/inspirehep/pidstore/api/base.py#L104

tests/integration/records/test_api_literature.py
```diff
@@ -454,3 +453,40 @@ def test_update_record_files(fsopen_mock, base_app, db, init_files_db):
...
+
+def test_get_record_from_db_depending_on_its_pid_type(base_app, db):
+    data = faker.record("lit")
+    record = InspireRecord.create(data)
+    record_from_db = InspireRecord.get_record(record.id)
+    assert type(record_from_db) == LiteratureRecord
+
+
+def test_dump_for_es(base_app, db):
+    data = faker.record("lit")
```
drjova:
```
data = faker.record('lit', data=MORE_FIELDS)
```

tests/unit/records/indexer/base.py
```diff
@@ -0,0 +1,86 @@
+#!/usr/bin/env bash
```
drjova: leftover?

inspirehep/records/utils.py
```diff
@@ -0,0 +1,39 @@
...
+    try:
+        pid_type = parts[-2][:3]
+        pid_value = parts[-1]
+    except IndexError:
+        return None
+
+    return pid_type, pid_value
+
+
+def get_author_with_record_facet_author_name(author):
```
drjova: aren't these dups https://github.com/inspirehep/inspirehep/blob/master/inspirehep/records/marshmallow/authors/common/facet_author_name.py

Pipfile
pazembrz: pytest-cov which is specified in this section installs pytest, if pytest is not duplicated in [packages] section,  then Pipenv will install version >=4.0 which is too new. 
Why pytest-cov is duplicated in both sections, I am not sure as it's your commit. :)

inspirehep/records/utils.py
pazembrz: Thanks... That was leftover, PidstoreBase.get_pid_from_record_uri() is used in the code.

inspirehep/records/utils.py
pazembrz: Fixed.

inspirehep/records/indexer/tasks.py
```diff
@@ -0,0 +1,127 @@
...
+
+
+def get_record(uuid, record_version=None):
+    logger.debug("Pulling record %s on version %s", uuid, record_version)
+
+    record = InspireRecord.get_record(uuid, with_deleted=True)
+
+    if record_version and record.model.version_id < record_version:
+        logger.info(
+            f"Cannot pull record {uuid}in version {record_version}."
```
drjova:
```suggestion
            f"Cannot pull record {uuid} in version {record_version}."
```

inspirehep/records/marshmallow/literature/common/author.py
```diff
@@ -39,3 +42,31 @@ def get_last_name(self, data):
             return names[0] or missing
 
         return missing
+
+
+class AuthorAutocompleteSchema(Schema):
+    input_field = fields.Method("generate_name_variations", dump_to="input")
```
drjova: I'd put in all fields ``dump_only=True`` everywhere as we have in all serializers e

inspirehep/config.py
```diff
@@ -422,3 +422,14 @@ def _(x):
 }
 
 APP_HEALTH_BLUEPRINT_ENABLED = True
+
+# Indexer
+# =======
+
+INDEXER_DEFAULT_INDEX = "records-hep"
+INDEXER_DEFAULT_DOC_TYPE = "hep"
+INDEXER_BULK_REQUEST_TIMEOUT = float(900)
```
drjova: why float?

inspirehep/records/api/literature.py
```diff
@@ -6,19 +6,26 @@
...
+logger = logging.getLogger(__name__)
+
 
 class LiteratureRecord(InspireRecord):
     """Literature Record."""
 
     pid_type = "lit"
 
+    es_serializer = "LiteratureESEnhancementV1"
+    ui_serializer = "LiteratureMetadataSchemaV1"
```
drjova: string reference? can we just pass the classes?

inspirehep/records/indexer/tasks.py
```diff
@@ -0,0 +1,127 @@
...
+from sqlalchemy.orm.exc import NoResultFound, StaleDataError
+
+from inspirehep.records.api import InspireRecord
+from inspirehep.records.errors import MissingArgumentError, MissingCitedRecordError
+from inspirehep.records.indexer.base import InspireRecordIndexer
+
+logger = logging.getLogger(__name__)
+
+
+def get_record(uuid, record_version=None):
```
drjova: why we keep this here? can we just move the logic inside the InspireRecord?

inspirehep/records/utils.py
```diff
@@ -0,0 +1,27 @@
...
+from inspire_utils.record import get_value, get_values_for_schema
+
+
+def get_author_with_record_facet_author_name(author):
+    author_ids = author.get("ids", [])
+    author_bai = get_values_for_schema(author_ids, "INSPIRE BAI")
+    bai = author_bai[0] if author_bai else "BAI"
+    author_preferred_name = get_value(author, "name.preferred_name")
+    if author_preferred_name:
+        return u"{}_{}".format(bai, author_preferred_name)
```
drjova: I know are copied over but we don't need this anymore :)

drjova: I mean the ``u``

tests/integration-async/records/conftest.py
```diff
@@ -0,0 +1,181 @@
...
+    app_config["DEBUG"] = True
+    app_config["JSONSCHEMAS_HOST"] = "localhost:5000"
+    app_config["SEARCH_ELASTIC_HOSTS"] = "localhost:9200"
+    app_config[
+        "SQLALCHEMY_DATABASE_URI"
+    ] = "postgresql+psycopg2://inspirehep:inspirehep@localhost:5432/inspirehep"
+    app_config["SQLALCHEMY_TRACK_MODIFICATIONS"] = True
+    app_config["TESTING"] = True
+    app.config.update(app_config)
+    return app
```
drjova: do you need to configure all these? most of them are configured like the fixture in integration/conftest

inspirehep/records/indexer/tasks.py
```diff
@@ -0,0 +1,127 @@
...
+
+    """
+    if not record and not uuid:
+        raise MissingArgumentError("uuid or record has to be provided")
+    if not record:
+        record = get_record(uuid, record_version)
+    pids = record.get_modified_references()
+
+    if not pids:
+        logger.debug("No references change for record %s", uuid)
```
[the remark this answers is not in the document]
pazembrz: I intentionally skipped f"" here not to resolve full string if debug will be turned off.

inspirehep/records/indexer/tasks.py
```diff
@@ -0,0 +1,127 @@
...
+        deleted = record.get("deleted", False)
+    if force_delete or deleted:
+        try:
+            record._index(force_delete=force_delete)
+            logger.debug("Record %s removed from ES", uuid)
+        except NotFoundError:
+            logger.warning(f"During removal, record {uuid} not found in ES!")
+    else:
+        record._index()
+        logger.debug("Record '%s' successfully indexed on ES", uuid)
```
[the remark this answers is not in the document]
pazembrz: I intentionally skipped f"" here not to resolve full string if debug will be turned off.

inspirehep/records/marshmallow/literature/base.py
```diff
@@ -124,3 +132,106 @@ def get_ui_display(self, data):
...
+            "imprints.date",
+        ]
+
+        dates = [
+            str(el)
+            for el in chain.from_iterable(
+                force_list(record.get_value(path)) for path in date_paths
+            )
+        ]
+        if dates:
```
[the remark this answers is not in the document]
pazembrz: I do not see any change here... :/

tests/integration-async/records/indexer/test_indexer_cli.py
```diff
@@ -0,0 +1,208 @@
+# -*- coding: utf-8 -*-
+#
+# Copyright (C) 2019 CERN.
+#
+# inspirehep is free software; you can redistribute it and/or modify it under
+# the terms of the MIT License; see LICENSE file for more details.
+
+from __future__ import absolute_import, division, print_function
```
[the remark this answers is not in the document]
pazembrz: Right, I missed it.

inspirehep/records/marshmallow/literature/common/author.py
pazembrz: Then why for example `AuthorSchemaV1` is missing it? :)

inspirehep/config.py
pazembrz: Fixed

inspirehep/records/marshmallow/literature/base.py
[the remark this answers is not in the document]
drjova: it’s more readable this way :)

inspirehep/records/marshmallow/literature/common/author.py
drjova: That’s a mistake thrn, we have to add it everywhere. We’ll add it in a new PR will create a US

inspirehep/records/api/literature.py
pazembrz: Unfortunately not. Circular dependencies ;)

inspirehep/records/indexer/tasks.py
pazembrz: This function is used only by tasks. If any other part of the code will start using this, then we can do this, but for now adding another method inside InspireRecord which already looks messy is overkill and over complication in my opinion.

inspirehep/records/utils.py
pazembrz: Right ;)

pazembrz: I removed it now from every place where we had it in the code.

tests/integration-async/records/conftest.py
pazembrz: Here we do not have this fixture, and still this configuration has to be configured somewhere. :)

tests/integration-async/records/conftest.py
```diff
@@ -0,0 +1,181 @@
...
+        db.drop_all()
+        db.init_app(app)
+        db.create_all()
+        _es = app.extensions["invenio-search"]
+        list(_es.delete(ignore=[404]))
+        list(_es.create(ignore=[400]))
+        init_default_storage_path()
+        init_records_files_storage_path()
+        es.indices.refresh("records-hep")
+        es.indices.refresh("records-authors")
```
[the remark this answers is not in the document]
pazembrz: Unfortunately not (explained on zulip)

inspirehep/records/marshmallow/literature/common/author.py
```diff
@@ -39,3 +42,35 @@ def get_last_name(self, data):
...
+    input_field = fields.Method(
+        "generate_name_variations", dump_to="input", dump_only=True
+    )
+
+    def generate_name_variations(self, full_name):
+        name_variations = generate_name_variations(full_name)
+        return [variation for variation in name_variations if variation]
+
+
+class AuthosInfoSchemaForES(AuthorSchemaV1):
```
drjova:
```suggestion
class AuthorsInfoSchemaForES(AuthorSchemaV1):
```

inspirehep/records/utils.py
drjova: hmm how is this fixed? :/

tests/unit/records/marshmallow/literature/common/test_accelerator_experiment.py
```diff
@@ -115,35 +90,27 @@ def test_returns_legacy_name_as_name_if_experiment_missing(
...
-)
-def test_returns_none_as_name_if_empty_present(mock_get_linked_records_in_field):
+@mock.patch("inspirehep.records.api.base.InspireRecord.get_records_by_pids")
+def test_returns_none_as_name_if_empty_present(get_records_mock):
     schema = AcceleratorExperimentSchemaV1()
     dump = {}
     expected = {"name": None}
 
-    result = schema.dumps(dump).data
+    record = LiteratureRecord(data=dump)
```
drjova:
```suggestion
    record = LiteratureRecord(dump)
```

tests/unit/records/marshmallow/literature/common/test_accelerator_experiment.py
```diff
@@ -154,6 +121,7 @@ def test_returns_dashed_institution_accelerator_experiment_as_name_with_unicode(
     }
     expected = {"name": "PSI, Villigen-PSI πM1 beam line-MUSE"}
 
-    result = schema.dumps(dump).data
+    record = LiteratureRecord(data=dump)
```
drjova:
```suggestion
    record = LiteratureRecord(dump)
```

tests/unit/records/marshmallow/literature/common/test_reference_item.py
```diff
@@ -58,36 +53,25 @@ def test_returns_non_empty_fields(mock_get_linked_records_in_field):
         "collaborations": [{"value": "CMS"}],
         "collaborations_with_suffix": [{"value": "ATLAS Team"}],
     }
-
-    result = schema.dumps(dump).data
+    record = LiteratureRecord(data=dump)
```
drjova:
```suggestion
    record = LiteratureRecord(dump)
```

tests/unit/records/marshmallow/literature/common/test_reference_item.py
```diff
@@ -98,68 +82,50 @@ class TestSchema(Schema):
         "references": [{"collaborations": [{"value": "CMS"}, {"value": "LHCb"}]}]
     }
 
-    result = schema.dumps(dump).data
+    record = LiteratureRecord(data=dump)
```
drjova:
```suggestion
    record = LiteratureRecord(dump)
```

tests/unit/records/marshmallow/literature/common/test_reference_item.py
```diff
@@ -98,68 +82,50 @@ class TestSchema(Schema):
...
-    mock_get_linked_records_in_field
-):
+@mock.patch("inspirehep.records.api.base.InspireRecord.get_records_by_pids")
+def test_returns_empty_if_no_reference_or_record_field(get_records_mock):
     schema = ReferenceItemSchemaV1()
     dump = {}
     expected = {}
 
-    result = schema.dumps(dump).data
+    record = LiteratureRecord(data=dump)
```
drjova:
```suggestion
    record = LiteratureRecord(dump)
```

tests/unit/records/marshmallow/literature/common/test_reference_item.py
```diff
@@ -98,68 +82,50 @@ class TestSchema(Schema):
...
-    mock_get_linked_records_in_field
-):
+@mock.patch("inspirehep.records.api.base.InspireRecord.get_records_by_pids")
+def test_returns_empty_if_empty_reference_or_record_field(get_records_mock):
     schema = ReferenceItemSchemaV1()
     dump = {"record": {}, "reference": {}}
     expected = {}
 
-    result = schema.dumps(dump).data
+    record = LiteratureRecord(data=dump)
```
drjova:
```suggestion
    record = LiteratureRecord(dump)
```

tests/unit/records/marshmallow/literature/common/test_reference_item.py
```diff
@@ -98,68 +82,50 @@ class TestSchema(Schema):
...
-    mock_get_linked_records_in_field
-):
+@mock.patch("inspirehep.records.api.base.InspireRecord.get_records_by_pids")
+def test_returns_non_empty_fields_if_some_fields_missing(get_records_mock):
     schema = ReferenceItemSchemaV1()
     dump = {"reference": {"label": "123", "control_number": 123}}
     expected = {"label": "123", "control_number": 123}
 
-    result = schema.dumps(dump).data
+    record = LiteratureRecord(data=dump)
```
drjova:
```suggestion
    record = LiteratureRecord(dump)
```

tests/unit/records/marshmallow/literature/common/test_reference_item.py
```diff
@@ -168,14 +134,16 @@ def test_returns_no_misc_if_title_persent(mock_get_linked_records_in_field):
 
     assert validate([dump], subschema) is None
 
-    result = schema.dumps(dump).data
+    record = LiteratureRecord(data=dump)
```
drjova:
```suggestion
    record = LiteratureRecord(dump)
```

tests/unit/records/marshmallow/literature/common/test_reference_item.py
```diff
@@ -228,14 +191,16 @@ def test_returns_only_first_misc(mock_get_linked_records_in_field):
     expected = {"label": "123", "misc": "A Misc"}
     assert validate([dump], subschema) is None
 
-    result = schema.dumps(dump).data
+    record = LiteratureRecord(data=dump)
```
drjova:
```suggestion
    record = LiteratureRecord(dump)
```

inspirehep/records/utils.py
pazembrz: It was fixed in opposite way :P now it is fixed. 
`get_author_with_record_facet_author_name` form utils removed. Using staticmethod from `AuthorsMetadataSchemaV1`. 
Also same thing with `get_author_display_name`.

inspirehep/records/marshmallow/literature/common/author.py
pazembrz: :+1:

PR #65, opened 2019-03-07
Add oauth client config for session authentication

monaawi:
INSPIR-1992


drjova:
we don't need this any more

PR #71, opened 2019-03-12
e2e: initial setup

harunurhan:


drjova:
@harunurhan what about travis?

PR #76, opened 2019-03-14
Updated integration tests for submissions

monaawi:
Updated integration tests for submissions to assert data sent to inspire-next
INSPIR-2152

### DIFF
--- a/tests/integration/submissions/test_submissions_views.py
+++ b/tests/integration/submissions/test_submissions_views.py
@@ -8,14 +8,13 @@
 import json
 
 import pytest
-from flask import jsonify
 from invenio_accounts.testutils import login_user_via_session
 from mock import patch
 
 from inspirehep.submissions.views import AuthorSubmissionsResource
 
 
-def test_new_author_submit_without_authentication_post(api_client):
+def test_author_submit_requires_authentication(api_client):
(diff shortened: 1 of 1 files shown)


tests/integration/submissions/test_submissions_views.py
```diff
@@ -76,13 +74,41 @@ def test_new_author_submit_with_required_fields(
                 }
             }
         ),
-        headers=headers,
     )
     assert response.status_code == 200
 
+    mock_requests_post.assert_called_once()
```
ammirate: This is not actually how you should mock a request :)
Rather than mocking the request library, you should 'intercept' the request and return a fake response.
Give a look to https://requests-mock.readthedocs.io/en/latest/mocker.html

michamos: don't we want to use something like `vcrpy` instead?

ammirate: I think is a bit overkill to use `vcr` just to reply a json with and ID

michamos: The point is we will use `vcr` eventually for other stuff (orcid for example), so instead of doing the same thing in a different way, we might as well use it here too.

PR #79, opened 2019-03-18
E2e authors

harunurhan:


e2e/tests/submissions/author.test.js
```diff
@@ -0,0 +1,39 @@
...
+    const workflowResponse = await page.goto(
+      `${routes.HOLDINGPEN_API}/${submitResponseJson.workflow_object_id}`
+    );
+    const workflowJson = await workflowResponse.json();
+
+    expect(workflowJson.metadata.name).toEqual({
+      preferred_name: 'Diego Martínez',
+      value: 'Martínez Santos, Diego',
+    });
+    expect(workflowJson.metadata._collections).toEqual(['Authors']);
```
drjova: we have to assert more:

* ``metadata.acquisition_source.email``
* ``metadata.acquisition_source.method``
* ``_workflow.workflow_name``

e2e/setupTests.js
```diff
@@ -1,5 +1,5 @@
 const { toMatchImageSnapshot } = require('jest-image-snapshot');
 
-jest.setTimeout(60000);
+jest.setTimeout(120000);
```
drjova: this is for what reason?

harunurhan: It's because too many services are involved and right after they're up, they take a bit time to respond to requests.

We can probably set it lower even back to 1min but this isn't really important since it is just MAX allowed time per test case.

ui/docker/Dockerfile-ui
```diff
@@ -21,6 +21,7 @@ WORKDIR /app
 
 COPY docker/conf.d /etc/nginx/conf.d
 
+ENV REACT_APP_ENABLE_LOCAL_LOGIN YES
```
pazembrz: Is this variable required during build? If not isn't better to set it in docker-compose.yml file?

drjova: yes we need this during build

ui/docker/conf.d/default.conf
```diff
@@ -23,16 +23,33 @@ server {
...
+  location /workflows {
+    proxy_pass         http://web-next:5000/workflows;
+    proxy_set_header Host localhost:5000;
+    proxy_http_version 1.1;
+  }
+
   location /api {
-    proxy_set_header Host localhost:8081;
+    proxy_set_header Host localhost:8000;
+    proxy_set_header X-Forwarded-For $proxy_add_x_forwarded_for;
```
drjova: do we need that? we don't have it in the others

PR #80, opened 2019-03-19
Remove Orcid Login

monaawi:
INSPIR-2079 Removed Orcid Login


ui/src/__tests__/App.test.jsx
```diff
@@ -121,7 +121,7 @@ describe('App', () => {
         </MemoryRouter>
       </Provider>
     );
-    expect(wrapper.find(Literature)).not.toExist();
+    expect(wrapper.find(Literature)).toExist();
   });
 
   it('navigates to Authors when /authors if user is logged in', () => {
```
harunurhan: I think you can remove those now

PR #82, opened 2019-03-20
INSPIRE-2209 - Files flag

pazembrz:
Add support for files feature flag
Add fixture to enable/disable files feature flag
Update tests
Add new tests

### DIFF
--- a/inspirehep/config.py
+++ b/inspirehep/config.py
@@ -423,3 +423,9 @@ INDEXER_REPLACE_REFS = False
 
 
 CELERY_IMPORTS = ["inspirehep.records.indexer.tasks"]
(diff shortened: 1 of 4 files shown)


inspirehep/config.py
```diff
@@ -433,3 +433,9 @@ def _(x):
 
 
 CELERY_IMPORTS = ["inspirehep.records.indexer.tasks"]
+
+
+# Feature flags
+# =============
+
+FEATURE_FLAG_FILES_ENABLED = False
```
drjova:
```suggestion
FEATURE_FLAG_ENABLE_FILES = False
```

PR #91, opened 2019-03-25
Orcid check for submission form

harunurhan:


e2e/utils/dom.js
```diff
@@ -8,6 +8,7 @@ async function submitForm(page, formData) {
   for (const fieldName of fieldNames) {
     await page.type(`[name=${fieldName}]`, formData[fieldName]); // eslint-disable-line no-await-in-loop
   }
+  await page.click(`[name=${fieldNames[0]}]`); // to trigger blur form validation before submit
```
drjova: shouldn't this be ``body`` or similar?

inspirehep/records/marshmallow/authors/base.py
```diff
@@ -91,3 +91,11 @@ def strip_empty(self, data):
 
 class AuthorsSchemaV1(RecordSchemaJSONV1):
     metadata = fields.Nested(AuthorsMetadataSchemaV1, dump_only=True)
+
+
+class AuthorsMetadataOnlyControlNumberSchemaV1(Schema):
```
drjova: I'd share this with other types, i.e. literature. It'll be helpful in the future for ``arXiv``,``doi`` etc. It's up to you though

e2e/utils/dom.js
harunurhan: let's see if that works

PR #92, opened 2019-03-25
records: addition of cli

drjova:


inspirehep/records/cli.py
```diff
@@ -0,0 +1,103 @@
...
+@click.option(
+    "-f",
+    "--files",
+    multiple=True,
+    default=[],
+    type=click.File("rb"),
+    help="Path to a JSON file, example: ``data/records/literature/999108.json``.",
+)
+@with_appcontext
+def records(urls, directory, files):
```
harunurhan: could you split logic foreach option into a separate function?

drjova:
@harunurhan updated

e2e/utils/user.js
```diff
@@ -4,7 +4,9 @@ async function submitLoginForm(page, username, password) {
   await page.type('[data-test-id=email]', username);
   await page.type('[data-test-id=password]', password);
   await page.click('[data-test-id=login]');
-  await page.waitFor(() => !document.querySelector('[data-test-id=login]'));
+  await page.waitFor(() => !document.querySelector('[data-test-id=login]'), {
+    timeout: 45000,
```
harunurhan: `// TODO: remove this when health checks are in place`

inspirehep/records/cli.py
```diff
@@ -0,0 +1,113 @@
...
+
+def _create_record(data):
+    control_number = data["control_number"]
+    click.echo(f"Creating record {control_number}.")
+    # ``earliest_date``` is not part of the schema and fails on
+    # validation. We are adding it in the serializers, hence
+    # it's part of our API responses.
+    data.pop("earliest_date", None)
+
+    # FIXME: move the type creation logic in ``InspireRecord.create``
```
harunurhan: just not to forget, do we have a JIRA issue for this?

scripts/server
```diff
@@ -16,6 +16,8 @@ FLASK_ENV=development
 # Start Worker and Server. Whatchdog detects any change to the .py files and restarts celery
 pipenv run watchmedo auto-restart -d . -p '*.py' -- celery worker -E -B -A inspirehep.celery -l INFO & pid_celery=$!
 
+pipenv run gunicorn --reload -b 0.0.0.0:5000  --access-logfile "-" --error-logfile - inspirehep.wsgi:application & pid_server=$!
```
harunurhan: is `server` script even relevant? I thought we were supposed to use `docker` for dev?

e2e/utils/user.js
harunurhan: if we have to do it.

scripts/server
harunurhan: I realized that this is used as a command for `worker` service. And now it runs `gunicorn` in worker service too. This might be the issue why `e2e` needed more timeout!

drjova: Good one, let me check this

inspirehep/records/cli.py
drjova: YEP we have INSPIR-2210

scripts/server
drjova: it worked, I refactored the scripts

README.md
```diff
@@ -14,11 +14,74 @@ $ pip install pipenv
...
 These commands allow you to spin up a contenarized environment that reproduces the Inspire-HEP and Inspire-Next applications.
 
+#### Docker
+
 ```bash
 $ ./scripts/bootstrap
 $ ./docker-inspire up -d
 $ ./docker-inspire run --rm web ./scripts/setup
 $ ./docker-inspire run --rm web-next scripts/setup_inspire_next
+$ firefox localhost:8000
```
harunurhan: `8081`

README.md
```diff
@@ -14,11 +14,74 @@ $ pip install pipenv
...
+
+```bash
+$ ./scripts/bootstrap
+$ ./docker-inspire up -d
+$ ./scripts/setup
+$ ./scripts/server
+$ firefox localhost:5000
+```
+
+#### Without Only The UI
```
harunurhan: `Without docker`

PR #94, opened 2019-03-26
records: add all pid_types

monaawi:
* Added jobs pid_type, record_Class, item and search endpoints 

* INSPIR-2145

### DIFF
--- a/inspirehep/config.py
+++ b/inspirehep/config.py
@@ -67,13 +67,21 @@ PID_TYPES_TO_ENDPOINTS = {
     "aut": "authors",
     "job": "jobs",
     "jou": "journals",
+    "exp": "experiments",
 }
-SCHEMA_TO_PID_TYPES = {"hep": "lit", "authors": "aut", "jobs": "job", "journals": "jou"}
+SCHEMA_TO_PID_TYPES = {
+    "hep": "lit",
+    "authors": "aut",
+    "jobs": "job",
+    "journals": "jou",
+    "experiments": "exp",
+}
 PID_TYPE_TO_INDEX = {
     "lit": "records-hep",
     "aut": "records-authors",
     "job": "records-jobs",
     "jou": "records-journals",
+    "exp": "records-experiments",
 }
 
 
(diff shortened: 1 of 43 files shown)


inspirehep/config.py
```diff
@@ -59,7 +59,7 @@ def _(x):
 
 # Deal with inconcistency :puke:
 PID_TYPES_TO_ENDPOINTS = {"lit": "literature", "aut": "authors"}
```
drjova:
```suggestion
PID_TYPES_TO_ENDPOINTS = {"lit": "literature", "aut": "authors", "job": "jobs"}
```

inspirehep/config.py
```diff
@@ -59,7 +59,7 @@ def _(x):
...
-PID_TYPES_TO_SCHEMA = {"hep": "lit", "authors": "aut"}
+PID_TYPES_TO_SCHEMA = {"hep": "lit", "authors": "aut", "jobs": "job"}
```
drjova:
```suggestion
PID_TYPES_TO_SCHEMA = {"hep": "lit", "authors": "aut", "job": "jobs"}
```

tests/integration/conftest.py
```diff
@@ -105,7 +105,11 @@ def test_with_record(base_app, create_record)
 
     def _create_record(record_type, data=None, with_pid=True, with_indexing=False):
         # FIXME: find a better location
-        MAP_PID_TYPE_TO_INDEX = {"lit": "records-hep", "aut": "records-authors"}
+        MAP_PID_TYPE_TO_INDEX = {
```
drjova: It's time to move this in ``config.py``

inspirehep/records/api/jobs.py
```diff
@@ -0,0 +1,21 @@
...
+from ...pidstore.api import PidStoreJobs
+from .base import InspireRecord
+
+
+class JobsRecord(InspireRecord):
+    """Authors Record."""
+
+    pid_type = "job"
+
+    es_serializer = "RecordMetadataSchemaJSONV1"
```
drjova: This is wrong, we should add ``None`` for the moment to be explicit
```suggestion
    es_serializer = None
```

drjova:
@monaawi about the commit message, should have a bit more details about the minter, REST endpoints etc

inspirehep/config.py
monaawi: Actually here the field should be renamed to SCHEMA_TO_PID_TYPES and it should remain "jobs": "job"

drjova: Indeed, it's only used here https://github.com/inspirehep/inspirehep/blob/2588356b81ffa8a92ebb8a65d4d54b15dcc42e9b/inspirehep/pidstore/api/base.py#L70 could you please also refactor this part? it's just a small renaming.

monaawi: yes, of course

PR #95, opened 2019-03-26
records: custom facets for author's publications

harunurhan:


inspirehep/search/facets.py
```diff
@@ -28,3 +29,22 @@ def inner(values):
         return Q("bool", must=filters)
 
     return inner
+
+
+def hep_author_publications():
+    current_app.config["HEP_COMMON_FILTERS"]
```
drjova: left over?

PR #98, opened 2019-03-27
submissions: add author emails field

harunurhan:
* Update public_emails field to accept both emails with "hidden" check

* INSPIR-2157


tests/unit/submissions/test_authors.py
```diff
@@ -554,22 +554,34 @@ def test_load_author_project_membership():
...
+        **DEFAULT_DATA_TO_DUMP,
+        "emails": [{"value": "email@email.com", "hidden": True}],
+    }
 
     result = Author().load(data).data
-    expected = {**DEFAULT_DATA_LOAD, "email_addresses": [{"value": "email@email.com"}]}
+    expected = {
+        **DEFAULT_DATA_LOAD,
+        "email_addresses": [{"value": "email@email.com", "hidden": True}],
+    }
```
drjova: don't we need to add a case without ``hidden`` field? 🤔

harunurhan: we're not doing anything special for it, but sure.

PR #101, opened 2019-03-29
ui: addition of ``UnclickableTag`` component

drjova:
* Repalces ``Tag`` with ``UnclickableTag`` component which
  doesn't show the ``pointer`` cursor for links.

* INSPIR-2067


ui/src/common/components/UnclickableTag/UnclickableTag.jsx
```diff
@@ -0,0 +1,27 @@
+import React, { Component } from 'react';
+import PropTypes from 'prop-types';
+import { Tag } from 'antd';
+import classnames from 'classnames';
+
+import './UnclickableTag.scss';
+
+class UnclickableTag extends Component {
+  render() {
+    const { children, className, ...otherProps } = this.props;
```
harunurhan: you can set `children` like `...otherProps` too.

ui/src/common/components/UnclickableTag/UnclickableTag.jsx
```diff
@@ -0,0 +1,27 @@
...
+
+    return (
+      <Tag
+        className={classnames('__UnclickableTag__', className)}
```
harunurhan:
```
className={`${className} __UnclickableTag__`}
```

would be better, `classnames` is a bit of an overkill when there is no conditional class to apply.

ui/src/common/components/UnclickableTag/UnclickableTag.jsx
```diff
@@ -0,0 +1,27 @@
...
+        {...otherProps}
+      >
+        {children}
+      </Tag>
+    );
+  }
+}
+
+UnclickableTag.propTypes = {
+  children: PropTypes.node.isRequired,
```
harunurhan: now no propType is needed for this one.

ui/src/common/components/UnclickableTag/UnclickableTag.jsx
drjova: do you mean ``otherProps.children``?

ui/src/common/components/UnclickableTag/UnclickableTag.jsx
drjova: I don't like it this way, because we should always have ``defaultProps`` and ``UnclickableTag`` always will have a ``className`` prop

```
    -       <UnclickableTag>
    +       <UnclickableTag
    +         className=""
    +       >
```

ui/src/common/components/UnclickableTag/UnclickableTag.jsx
```diff
@@ -0,0 +1,22 @@
...
+class UnclickableTag extends Component {
+  render() {
+    const { className, ...otherProps } = this.props;
+
+    return (
+      <Tag
+        className={classnames('__UnclickableTag__', className)}
+        {...otherProps}
+      >
+        {otherProps.children}
```
harunurhan: this is unnecessary

drjova: 👍 done

PR #102, opened 2019-03-29
submissions: add acquisition source for literature

harunurhan:
* INSPIR-2266

### DIFF
--- a/inspirehep/submissions/views.py
+++ b/inspirehep/submissions/views.py
@@ -23,20 +23,48 @@ from .utils import get_record_from_legacy
 blueprint = Blueprint("inspirehep_submissions", __name__, url_prefix="/submissions")
 
 
-def send_post_request_to_inspire_next(endpoint, data):
-    headers = {
-        "content-type": "application/json",
-        "Authorization": f"Bearer {current_app.config['AUTHENTICATION_TOKEN']}",
-    }
-    response = requests.post(
-        f"{current_app.config['INSPIRE_NEXT_URL']}{endpoint}",
-        data=json.dumps(data),
-        headers=headers,
-    )
-    return response
+class BaseSubmissionsResource(MethodView):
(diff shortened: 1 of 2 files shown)


inspirehep/submissions/views.py
```diff
@@ -23,20 +23,48 @@
...
+            internal_uid=int(current_user.get_id()),
+        )
+
+        orcid = self._get_user_orcid()
+        if orcid:
+            acquisition_source["orcid"] = orcid
+
+        return acquisition_source
+
+    def _get_user_orcid(self):
```
drjova: we don't need ``_`` here since we moved it

harunurhan: it's still bit a of private thing though.

drjova: it's similar to ``get_acquisition_source``

tests/integration/submissions/test_submissions_views.py
```diff
@@ -262,6 +261,8 @@ def test_new_literature_submit(app, api_client, create_user_and_token, requests_
         == history.headers["Authorization"]
     )
     assert history.url == f"{app.config['INSPIRE_NEXT_URL']}/workflows/literature"
+    assert "datetime" in post_data["data"]["acquisition_source"]
+    del post_data["data"]["acquisition_source"]["datetime"]
```
drjova: why we delete that?

drjova: https://github.com/inspirehep/inspirehep/blob/d3968f24b810a91ef1711daf4600976a7af6dcf7/tests/integration/records/serializers/test_latex.py#L11

harunurhan: alternative would be to to mock `time` (not sure if we have it) and test MOCKED time. I don't think this has a lot of value compared to effort needed. So checking if it's there or not it's fair IMO.

PR #103, opened 2019-03-29
submission: add former_name field to the author form

ammirate:
* INSPIR-2156


inspirehep/submissions/marshmallow/author.py
```diff
@@ -13,6 +13,7 @@
 class Author(Schema):
     display_name = fields.Raw()
     family_name = fields.Raw()
+    former_name = fields.Raw()
```
harunurhan: This the first step, you also need to add this to `@post_load` and `@pre_dump`

ui/src/submissions/authors/components/AuthorForm.jsx
```diff
@@ -69,6 +69,12 @@ class AuthorForm extends Component {
               placeholder="For non-Latin names e.g. 麦迪娜 or Эдгар Бугаев"
               component={TextField}
             />
+            <Field
+              name="former_name"
+              label="Alternate/former name"
```
harunurhan: @StellaCh can we go for only `Former name` here?

ui/src/submissions/authors/schemas/author.js
```diff
@@ -28,6 +28,7 @@ const authorSchema = object().shape({
     .trim()
     .required()
     .label('Display Name'),
+  former_name: string().label('Alternate/former name'),
```
harunurhan: `label` is used only in validation error messages, since `string()` without extra stuff like `required` won't cause any error ever, it's not necessary here.

ui/src/submissions/authors/components/AuthorForm.jsx
ammirate: We'll change it to "Alternate name" only.

inspirehep/records/marshmallow/literature/common/author.py
```diff
@@ -13,7 +13,7 @@
 
 class AuthorSchemaV1(Schema):
     affiliations = fields.Raw()
-    alternative_names = fields.Raw()
```
harunurhan: actually if you check hep schema `references.reference` part, this is `alternative_names`.

inspirehep/submissions/marshmallow/author.py
```diff
@@ -123,6 +127,8 @@ def build_author(self, data):
         given_name = data.get("given_name")
         family_name = data.get("family_name")
         full_name = self.get_full_name(family_name, given_name)
+        alternate_name = data.get("alternate_name")
```
harunurhan: can you move those 2 lines under `set_full_name`?

inspirehep/records/marshmallow/literature/common/author.py
ammirate: But this is below `name`, not `references.reference`

PR #104, opened 2019-04-01
e2e: add tests for literature submissions

harunurhan:
* INSPIR-2039

### DIFF
--- a/e2e/utils/dom.js
+++ b/e2e/utils/dom.js
@@ -2,10 +2,20 @@ const ID_ATTRIBUTE = 'data-test-id';
 const TYPE_ATTRIBUTE = 'data-test-type';
 
 async function selectFromSelectBox(page, selectId, value) {
-  await page.click(`[${ID_ATTRIBUTE}=${selectId}]`);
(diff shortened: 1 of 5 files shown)


e2e/utils/dom.js
```diff
@@ -2,10 +2,20 @@ const ID_ATTRIBUTE = 'data-test-id';
...
+  await page.click(selectSelector);
+
+  const openSelectSelector = `.ant-select-open>${selectSelector}`;
+  await page.waitFor(openSelectSelector);
+
   await page.click(`[${ID_ATTRIBUTE}=${selectId}-option-${value}]`);
-  await page.click('body');
-  await page.waitFor('.ant-select-dropdown-hidden');
+
+  await page.click('label'); // FIXME: won't work if there is no label on the page
```
drjova: what is the ``label`` here?

harunurhan: As you know we need to close the select dropdown by clicking to sth else and`click('body')` is sometimes clicking another option on the dropdown, then closes it.

When this breaks we can replace it with new element like `data-test-id="page-blank-space"`

drjova: @harunurhan thank you for the explanation

PR #113, opened 2019-04-02
ui: fix flaky lit submission ui-tests

harunurhan:
* Improve selectDocType logic to be more deterministic similar to the
one for e2e tests (INSPIR-2280)


drjova:
It'd be nice if we can share this code somehow, it seems the same.

PR #116, opened 2019-04-04
serializers: fix serializers so serialized data won't break indexer

pazembrz:
fix serializers so serialized data won't break indexer anymore

Fix `id` field in serializer (use `control_number` instead `uuid`)
Fix `abstracts` field, change field name `value` -> `input`
Fix `resolve_expreiment_records` to work properly when `many` is `True`


inspirehep/records/marshmallow/literature/common/accelerator_experiment.py
```diff
@@ -16,16 +16,22 @@ class AcceleratorExperimentSchemaV1(Schema):
...
-            data
-        )
         if not many:
+            experiment_records_map = self.get_control_numbers_to_resolved_experiments_map(
+                data
+            )
             return self.get_resolved_record_or_experiment(experiment_records_map, data)
+        results = []
+        for record in data:
+            experiment_records_map = self.get_control_numbers_to_resolved_experiments_map(
```
drjova: it doesn't worth to calculate this for each record, ``get_linked_records_from_field`` yields all the linked records. We can either modify ``get_linked_records_from_field`` to return an ``InspireRecord`` obj or modify ``get_control_numbers_to_resolved_experiments_map`` to handle this case. The cleaner way is to modify ``get_linked_records_from_field`` to return ``InspireRecords``, this will allow us to remove https://github.com/inspirehep/inspirehep/blob/296d4af998f932dc54aea8b3faf98f247328ceab/inspirehep/records/marshmallow/literature/common/reference_item.py#L84-L85 as well

tests/unit/records/marshmallow/literature/common/test_accelerator_experiment.py
```diff
@@ -124,3 +124,27 @@ def test_returns_dashed_institution_accelerator_experiment_as_name_with_unicode(
...
+        "legacy_name": "LEGACY-EXP2",
+        "institutions": [{"value": "INS"}],
+        "accelerator": {"value": "ACC"},
+        "experiment": {"value": "EXP2"},
+    }
+    expected = [{"name": "INS-ACC-EXP1"}, {"name": "INS-ACC-EXP2"}]
+
+    record1 = LiteratureRecord(dump1)
+    record2 = LiteratureRecord(dump2)
+    result = schema.dumps([record1, record2], many=True).data
```
drjova: I'd separate this two cases and add them as integration tests

tests/integration/records/test_api_literature.py
```diff
@@ -492,7 +492,7 @@ def test_dump_for_es(base_app, db):
     assert "control_number" in dump
     assert record["control_number"] == dump["control_number"]
     assert "id" in dump
-    assert str(record.id) == dump["id"]
+    assert record["control_number"] == dump["id"]
```
drjova: We should add a regression test for indexing, we can create a record and make sure is accessible in the search

tests/unit/records/marshmallow/literature/common/test_abstract_source.py
```diff
@@ -54,20 +67,32 @@ def test_abstract_source_one_missing_source():
...
 
 
-def test_abstract_source_one_ONLY():
+@mock.patch(
+    (
+        "inspirehep.records.marshmallow.literature.common.reference_item"
+        ".InspireRecord._get_linked_records_from_field"
+    )
+)
+def test_abstract_source_one_ONLY(mock_get_linked_records_from_field):
```
drjova:
```suggestion
def test_abstract_source_one_only(mock_get_linked_records_from_field):
```

tests/unit/records/marshmallow/literature/common/test_accelerator_experiment.py
```diff
@@ -124,3 +125,25 @@ def test_returns_dashed_institution_accelerator_experiment_as_name_with_unicode(
     result = schema.dumps(record).data
 
     assert expected == json.loads(result)
+
+
+@mock.patch("inspirehep.records.api.base.InspireRecord.get_records_by_pids")
+def test_accelerator_experiment_many(get_records_mock):
```
drjova: I'd move this to an integration test :)

inspirehep/records/api/base.py
```diff
@@ -299,10 +301,61 @@ def get_linked_pids_from_field(self, path):
...
+                ]
+            }
+            >>>  record = InspireRecord(data)
+            >>>  records = record.get_linked_records_from_field("references.record")
+
+                """
+        pids = cls._get_linked_pids_from_field(data, path)
+        return cls.get_records_by_pids(pids)
+
+    def get_linked_pids_from_field(self, path):
```
drjova: it's a bit confusing the name is ``get_linked_pids_from_field`` but actually returns ``records``, why not ``get_linked_records_from_field``? WDYT?

tests/integration/conftest.py
```diff
@@ -184,3 +184,12 @@ def _create_user():
...
     return _create_user
+
+
+@pytest.fixture(scope="function")
+def elastic(app):
+    with app.app_context():
+        _es = app.extensions["invenio-search"]
+        list(_es.delete(ignore=[404]))
+        list(_es.create(ignore=[400]))
+        list(_es.put_templates(ignore=[400]))
```
drjova: why we need this fixture? we have ``es_clear`` https://pytest-invenio.readthedocs.io/en/latest/api.html#pytest_invenio.fixtures.es_clear :)

tests/integration/records/test_indexer.py
```diff
@@ -0,0 +1,64 @@
...
+
+    response = es.search("records-conferences")
+    expected_count = 1
+    expected_control_number = 1203206
+    assert len(response["hits"]["hits"]) == expected_count
+    assert (
+        response["hits"]["hits"][0]["_source"]["control_number"]
+        == expected_control_number
+    )
+
```
drjova: IMHO it'd better to compare the whole record ``metadata`` instead of numbers (this applies in all test cases)

tests/unit/records/marshmallow/literature/common/test_accelerator_experiment.py
pazembrz: But why? Only thing which is tested here is one serializer, no real record is required for this, no integration with anything.

inspirehep/records/api/base.py
pazembrz: But it's not returning a record.
```
"""Return a list of (pid_type, pid_value) tuples for all records referenced
        in the field at the given path"""
```

tests/unit/records/marshmallow/literature/common/test_accelerator_experiment.py
drjova: this serializer also use db to retrieve experiments (if present), we have to actual test this case as well

inspirehep/records/api/base.py
```diff
@@ -299,10 +301,61 @@ def get_linked_pids_from_field(self, path):
         full_path = ".".join([path, "$ref"])
         pids = [
             PidStoreBase.get_pid_from_record_uri(rec)
-            for rec in self.get_value(full_path, [])
+            for rec in get_value(data, full_path, [])
```
drjova: do we even use ``self.get_value``? can we just remove it?

inspirehep/records/api/base.py
drjova: 👍 I misread

inspirehep/records/api/base.py
```diff
@@ -273,11 +273,13 @@ def create(cls, data, **kwargs):
             record = super().create(data, id_=id_, **kwargs)
         return record
 
-    def get_linked_pids_from_field(self, path):
+    @classmethod
+    def _get_linked_pids_from_field(cls, data, path):
```
drjova: TBH it's very weird to have  a ``classmethod`` starting with ``_`` and use it everywhere

inspirehep/records/api/base.py
```diff
@@ -299,10 +301,61 @@ def get_linked_pids_from_field(self, path):
...
+                        'record': {
+                            '$ref': 'http://localhost/literature/1'
+                        }
+                    }
+                ]
+            }
+            >>>  record = InspireRecord(data)
+            >>>  records = record.get_linked_records_from_field("references.record")
+
+                """
```
drjova: The ```"""``` indentation  is wrong :P

tests/integration/conftest.py
pazembrz: Because `es_clear` requires `es` and `es` requires `appctx` which requires `base_app` which requires `create_app` which should be scope 'module' but for async tests our app need to be `session` scoped. It's a little mess with this es_clear...

inspirehep/records/api/base.py
drjova: do we even have a use case for this method? shouldn't it be a ``static`` ☺️

pazembrz: It's used in many places:
for example in:
inspirehep/records/api/base.py:939, 941, 944

inspirehep/records/api/base.py
pazembrz: It's used in many places, for now we have 14 places where it's used in the code... Using it this way it's more natural than passing data to some util function...

inspirehep/records/api/base.py
pazembrz: Actually this one is used only in the base class, but `_get_linked_records_from_field` is used in serializers, so propably I should change it's name....

tests/integration/conftest.py
drjova: but this is not the ``async`` tests :/

tests/unit/records/marshmallow/literature/common/test_accelerator_experiment.py
pazembrz: done

tests/integration/conftest.py
pazembrz: You are right... My mistake.

tests/integration/records/test_indexer.py
pazembrz: fixed

tests/integration/records/serializers/test_json.py
```diff
@@ -565,3 +567,50 @@ def test_institutions_default_json_v1_response_search(
...
+        {"name": "HESS"},
+        {"name": "LOFAR"},
+        {"name": "HAWC"},
+        {"name": "AUGER"},
+        {"name": "ALMA"},
+        {"name": "VLBI"},
+    ]
+
+    create_record("exp", data=experiment_data)
+    #  Create experiments:
```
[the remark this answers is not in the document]
pazembrz: Why?

tests/integration/records/serializers/test_json.py
```diff
@@ -565,3 +567,50 @@ def test_institutions_default_json_v1_response_search(
...
+    for experiment in data["accelerator_experiments"]:
+        if "record" in experiment:
+            control_number = experiment["record"]["$ref"].split("/")[-1]
+            if "1118472" == control_number:
```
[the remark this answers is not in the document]
pazembrz: As you said before, we need to test serializer for accelerator experiments which is pulling data form db, so we need to create this experiments if they have to be pulled from DB during literature serialization.

tests/integration/records/serializers/test_json.py
```diff
@@ -565,3 +567,50 @@ def test_institutions_default_json_v1_response_search(
...
+                continue
+            create_record("exp", data={"control_number": control_number})
+    dumped_record = LiteratureMetadataSchemaV1().dump(record.json).data
+    assert dumped_record["accelerator_experiments"] == expected_experiment
```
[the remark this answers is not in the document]
pazembrz: Record was already there, it's easier to reuse it than creating something new.

PR #117, opened 2019-04-04
ui: add search rank for literature results

harunurhan:
* Adds search rank for literature search results both on literature
search and author's literature, except small (phone) screens.

* Note that it uses native flexbox because antd's grid wraps Cols
without column size (span) and when it is set, it either takes extra
space for the searchRank unnecessarily and not guaranteed to work
properly for all screen sizes & title, rank lenght

* INSPIR-2062

### DIFF
--- a/ui/src/common/components/__tests__/__snapshots__/SearchResults.test.jsx.snap
+++ b/ui/src/common/components/__tests__/__snapshots__/SearchResults.test.jsx.snap
@@ -20,3 +20,5 @@ exports[`SearchResults renders with all props set 1`] = `
   </div>
 </div>
 `;
+
+exports[`SearchResults renders with required props 1`] = `<div />`;
--- a/ui/src/common/utils.js
+++ b/ui/src/common/utils.js
@@ -129,6 +129,10 @@ export function shallowEqual(objA, objB) {
   return true;
 }
 
+export function getSearchRank(index, page, pageSize) {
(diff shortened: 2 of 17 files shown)


ui/src/common/components/__tests__/__snapshots__/SearchResults.test.jsx.snap
```diff
@@ -20,3 +20,5 @@ exports[`SearchResults renders with all props set 1`] = `
   </div>
 </div>
 `;
+
+exports[`SearchResults renders with required props 1`] = `<div />`;
```
drjova: something is missing from this ``snap``, no? Should it return empty ``div``?

ui/src/common/utils.js
```diff
@@ -129,6 +129,10 @@ export function shallowEqual(objA, objB) {
   return true;
 }
 
+export function getSearchRank(index, page, pageSize) {
+  return (page - 1) * pageSize + index + 1;
```
drjova: do we care for cases ``?page=-20``? Will this break?

drjova: discussed IRL

ui/src/common/components/__tests__/__snapshots__/SearchResults.test.jsx.snap
drjova: discussed IRL

PR #118, opened 2019-04-05
ui: remove unique orcid validation for updates

harunurhan:
* Only validates if orcid exist or not, for NEW author submissions

* INSPIR-2304

### DIFF
--- a/ui/src/submissions/authors/containers/AuthorSubmissionPage.jsx
+++ b/ui/src/submissions/authors/containers/AuthorSubmissionPage.jsx
@@ -3,10 +3,14 @@ import PropTypes from 'prop-types';
 import { connect } from 'react-redux';
 import { Map } from 'immutable';
 import { Row, Col } from 'antd';
+import { object } from 'yup';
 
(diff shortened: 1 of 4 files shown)


ui/src/submissions/authors/containers/AuthorSubmissionPage.jsx
```diff
@@ -41,6 +45,7 @@ class AuthorSubmissionPage extends Component {
                 error={error}
                 onSubmit={this.onSubmit}
                 initialFormData={initialFormData}
+                overrideSchema={extraSchemaForNewAuthor}
```
drjova: Am I missing something all this change will always (create & update) will override the schema?

harunurhan: it does not, actually I think `overrideSchema` is not a good name, because what does is `yup.concat`. We could call it `extendSchema` or `extraSchema` sth like that.

drjova: ``extendSchema`` makes more sense 👍

PR #119, opened 2019-04-08
ui: add clear state middleware

harunurhan:
Dispatches CLEAR_STATE action when we want some state to be cleaned
(for now literature and authors).

Implements logic for only for LOCATION_CHANGE, when
location .pathname OR .search changes

Fixes INSPIR-2179


ui/src/middlewares/__tests__/clearStateDispatcher.test.js
```diff
@@ -0,0 +1,111 @@
+import { LOCATION_CHANGE } from 'react-router-redux';
+
+import { CLEAR_STATE } from '../../actions/actionTypes';
+import middleware from '../clearStateDispatcher';
+
+describe('clearStateDispatcher middleware', () => {
+  describe('on LOCATION_CHANGE returns next(original action) and', () => {
```
drjova: two ``describe``s?  :)

harunurhan: there is nothing wrong about that, `describe` can be used to organize/group test cases. Nested `describe` are not new to the codebase.

This is done like this, when we have another case where we dispatch `CLEAR_STATE`, we'll add another `describe('on ANOTHER_THING return...`

drjova:
```suggestion
  describe('on LOCATION_CHANGE returns next(original action)', () => {
```

PR #120, opened 2019-04-08
ui: persist preferred cite format

harunurhan:
* Adds preferredCiteFormat to state.user which is persisted to
localStorage

* INSPIR-2229


ui/src/literature/components/CiteModalAction.jsx
```diff
@@ -14,15 +14,16 @@ const FORMAT_SELECT_OPTIONS = [
   { value: 'vnd+inspire.latex.eu+x-latex', display: 'LaTex (EU)' },
   { value: 'vnd+inspire.latex.us+x-latex', display: 'LaTex (US)' },
 ];
-
-export const DEFAULT_FORMAT = FORMAT_SELECT_OPTIONS[0].value;
+export const FORMAT_SELECT_VALUES = FORMAT_SELECT_OPTIONS.map(
+  option => option.value
+);
```
drjova: shall we move this in  ``constants`` or something? It seems weird to have it here and need it also in the reducers.

harunurhan: I was thinking the same, but then it's relevant here and only used one other place so far. So I decided keep it like that.

PR #123, opened 2019-04-08
records: record create based on pid type

monaawi:
* Check type of record when doing InspireRecord.create / create_or_update and get_record so that we use the proper subclass
* INSPIR-2210

### DIFF
--- a/inspirehep/records/api/base.py
+++ b/inspirehep/records/api/base.py
@@ -34,7 +34,7 @@ from sqlalchemy.dialects.postgresql import JSONB
 from sqlalchemy.orm.exc import NoResultFound
 
 from inspirehep.pidstore.api import PidStoreBase
-from inspirehep.records.errors import MissingSerializerError
+from inspirehep.records.errors import MissingSerializerError, WrongRecordSubclass
 from inspirehep.records.indexer.base import InspireRecordIndexer
 
 logger = logging.getLogger(__name__)
@@ -246,9 +246,9 @@ class InspireRecord(Record):
     @classmethod
     def get_record(cls, id_, with_deleted=False):
         record = super().get_record(str(id_), with_deleted)
-        type_from_schema = PidStoreBase.get_pid_type_from_schema(record["$schema"])
-        if record.pid_type is None or record.pid_type != type_from_schema:
-            return cls.get_subclasses()[type_from_schema].get_record(id_, with_deleted)
+        record_class = cls.get_class_for_record(record)
+        if record_class != cls:
+            record = record_class(record, model=record.model)
         return record
 
     @classmethod
@@ -265,7 +265,20 @@ class InspireRecord(Record):
             yield cls(data.json)
 
     @classmethod
+    def get_class_for_record(cls, data):
+        type_from_schema = PidStoreBase.get_pid_type_from_schema(data["$schema"])
+        record_class = cls.get_subclasses().get(type_from_schema)
+        if record_class is None:
+            raise WrongRecordSubclass(
+                f"Wrong subclass {cls} used for record of type {type_from_schema}"
+            )
+        return record_class
+
+    @classmethod
     def create(cls, data, **kwargs):
+        record_class = cls.get_class_for_record(data)
+        if record_class != cls:
+            return record_class.create(data, **kwargs)
         id_ = uuid.uuid4()
         data = cls.strip_empty_values(data)
         with db.session.begin_nested():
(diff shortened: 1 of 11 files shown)


inspirehep/records/api/base.py
```diff
@@ -246,9 +246,9 @@ def get_subclasses(cls):
     @classmethod
     def get_record(cls, id_, with_deleted=False):
         record = super().get_record(str(id_), with_deleted)
-        type_from_schema = PidStoreBase.get_pid_type_from_schema(record["$schema"])
-        if record.pid_type is None or record.pid_type != type_from_schema:
-            return cls.get_subclasses()[type_from_schema].get_record(id_, with_deleted)
+        subclass = cls.get_class_from_schema(record)
```
monaawi: I don't like this naming but I don't have a better idea, ideas?

inspirehep/records/api/base.py
```diff
@@ -264,8 +264,16 @@ def get_records_by_pids(cls, pids):
         for data in query.yield_per(100):
             yield cls(data.json)
 
+    @classmethod
+    def get_class_from_schema(cls, data):
+        type_from_schema = PidStoreBase.get_pid_type_from_schema(data["$schema"])
+        return cls.get_subclasses().get(type_from_schema, cls)
```
pazembrz: Isn't better to throw an exception if `type_from_schema` is not in subclasses list?

inspirehep/records/api/base.py
```diff
@@ -264,8 +264,16 @@ def get_records_by_pids(cls, pids):
...
+
     @classmethod
     def create(cls, data, **kwargs):
+        subclass = cls.get_class_from_schema(data)
+        if subclass != cls:
```
pazembrz: You should also add tests to check if it will behave correctly when you call Create from correct subclass :)

inspirehep/records/api/base.py
harunurhan: `get_class_for_record` is better since you're passing the whole thing.

tests/integration/records/test_api_literature.py
```diff
@@ -593,3 +593,10 @@ def test_add_and_remove_figs_and_docs_when_files_flag_disabled(
         record["figures"]
 
     assert record["documents"][0]["key"] == expected_doc_filename
+
+
+def test_create_record_from_db_depending_on_its_pid_type(base_app, db):
```
drjova: what about tests for ``create_or_update``?

monaawi: create_or_update is using either the get_record which does the "class resolution" or the create one so by testing those I am testing that it is using the proper class. But if you think it would be better to have it more explicitly testing, then I can add that as well

inspirehep/records/api/base.py
```diff
@@ -264,8 +264,21 @@ def get_records_by_pids(cls, pids):
         for data in query.yield_per(100):
             yield cls(data.json)
 
+    @classmethod
+    def get_class_for_record(cls, data):
+        type_from_schema = PidStoreBase.get_pid_type_from_schema(data["$schema"])
+        record_class = cls.get_subclasses().get(type_from_schema)
+        if record_class is None:
+            raise WrongRecordSubclass(
```
drjova: do we really want to have such restriction? Maybe just log?

monaawi: I think this is a good restriction to have as you shouldn't be using a wrong subclass to create a record of a different type. (Eg JobsRecord.create(literature_record) ) . What would be the alternative? To create a LiteratureRecord anyway and just log it?

inspirehep/records/api/base.py
```diff
@@ -246,9 +246,9 @@ def get_subclasses(cls):
     @classmethod
     def get_record(cls, id_, with_deleted=False):
         record = super().get_record(str(id_), with_deleted)
-        type_from_schema = PidStoreBase.get_pid_type_from_schema(record["$schema"])
-        if record.pid_type is None or record.pid_type != type_from_schema:
-            return cls.get_subclasses()[type_from_schema].get_record(id_, with_deleted)
+        record_class = cls.get_class_for_record(record)
+        if record_class != cls:
+            return record_class.get_record(id_, with_deleted)
```
drjova: Am I missing something, but can we just ``record_class(record)``? do we need a second query here? ☺️

drjova: https://github.com/inveniosoftware/invenio-records/blob/master/invenio_records/api.py#L33

PR #128, opened 2019-04-10
literature: add `/literature/import` endpoint

ammirate:
* ADD: This commit adds the `import_article` endpoint and api function for importing through inspire-next a new article from arXiv or Crossref. This function will be used in the submission form to automatically fill it.

* INSPIR-1769


tests/integration/records/test_views.py
```diff
@@ -614,3 +619,71 @@ def test_literature_facets_collaboration(api_client, db, create_record):
...
+    assert resp.status_code == 400
+
+
+def test_import_article_view_400_because_article_already_exists(
+    api_client, db, create_record
+):
+    arxiv_value = faker.arxiv()
+    data = {"arxiv_eprints": [{"value": arxiv_value}]}
+    data = faker.record("lit", with_control_number=True, data=data)
+    LiteratureRecord.create(data)
```
harunurhan: should use `create_record` which will also mint stuff after #132 is merged.

tests/integration/records/test_api_literature.py
```diff
@@ -593,3 +600,116 @@ def test_add_and_remove_figs_and_docs_when_files_flag_disabled(
...
+            assert m.last_request.headers["content-type"] == "application/json"
+            assert m.last_request.headers["Authorization"] == "Bearer CHANGE_ME"
+
+
+def test_harvest_article_doi(base_app):
+    expected = faker.record("lit")
+    doi = "10.8247/32543408"
+    with requests_mock.mock() as m:
+        m.post(
+            "http://web-next:5000/import/doi/10.8247%2F32543408",
```
harunurhan: I think, it's better to use `INSPIRE_NEXT_URL` rather than hardcoding.

inspirehep/records/api/literature.py
```diff
@@ -143,3 +152,90 @@ def _dump_for_es(self):
...
+        pid_value = normalize_arxiv(identifier)
+
+    elif is_doi(identifier):
+        pid_type = "doi"
+        pid_value = normalize_doi(identifier)
+
+    else:
+        raise UnknownIdentifierError(identifier)
+
+    pid = PersistentIdentifier.query.filter_by(
```
harunurhan: I feel like this will be done many times for other pids, it would be good to have  `def exists(pid_type, pid_value)` in `PersistentIdentifier` class.

inspirehep/records/api/literature.py
```diff
@@ -143,3 +152,90 @@ def _dump_for_es(self):
...
+
+    Raises:
+        UnknownIdentifierError: if `id_type` is neither "arxiv" or "doi".
+    """
+    if id_type not in ("arxiv", "doi"):
+        raise UnknownIdentifierError(identifier)
+
+    inspire_next_url = current_app.config["INSPIRE_NEXT_URL"]
+    endpoint = (
+        current_app.config["INSPIRE_NEXT_IMPORT_ARXIV_ENDPOINT"]
```
harunurhan: Having this in `config`
```
INSPIRE_NEXT_IMPORT_ENDPOINTS_BY_TYPE = {
    'doi': '<full_url>',
    'arxiv': '<full_url>',
}
```
Would simplify this part.

inspirehep/records/api/literature.py
```diff
@@ -143,3 +169,133 @@ def _dump_for_es(self):
...
+    Raises:
+        ImportConnectionError: if the request doesn't succeed.
+        ImportParsingError: if any error occurs during the response parsing.
+
+    """
+    url = ARXIV_URL.replace(PLACEHOLDER, arxiv_id)
+
+    try:
+        resp = requests.get(url=url)
+    except Exception as e:
```
pazembrz: Try to catch less general exception or at least try to log what you cached, as we will lose this information.

inspirehep/records/api/literature.py
```diff
@@ -143,3 +169,133 @@ def _dump_for_es(self):
...
+        raise ImportConnectionError("Cannot contact arXiv")
+
+    if "Malformed identifier" in str(resp.text):
+        # arXiv will reply 200 for a non existing arXiv ID with a message error
+        return {}
+
+    try:
+        parser = ArxivParser(resp.text)
+        return parser.parse()
+    except Exception as e:
```
pazembrz: same here

inspirehep/records/api/literature.py
```diff
@@ -143,3 +169,133 @@ def _dump_for_es(self):
...
+
+    """
+    url = ARXIV_URL.replace(PLACEHOLDER, arxiv_id)
+
+    try:
+        resp = requests.get(url=url)
+    except Exception as e:
+        raise ImportConnectionError("Cannot contact arXiv")
+
+    if resp.status_code >= 400:
```
pazembrz: Do not create two identical exceptions for different errors (at least message should be different)

inspirehep/records/api/literature.py
```diff
@@ -143,3 +169,133 @@ def _dump_for_es(self):
...
+        ImportConnectionError: if the request doesn't succeed.
+        ImportParsingError: if any error occurs during the response parsing.
+
+    """
+    doi = requests.utils.quote(doi, safe="")
+    url = CROSSREF_URL.replace(PLACEHOLDER, doi)
+
+    try:
+        resp = requests.get(url=url)
+    except Exception:
```
pazembrz: ditto

inspirehep/records/api/literature.py
pazembrz: I agree with that, but this shouldn't be added to this task, we should create a new task for this.

harunurhan: I mean it's a few lines function at the end, separate commit would work OK, IMO

inspirehep/records/api/literature.py
```diff
@@ -143,3 +169,141 @@ def _dump_for_es(self):
...
+    importers = {"arxiv": import_arxiv, "doi": import_doi}
+    importer = importers.get(pid_type, UnknownImportIdentifierError)
+    article = importer(pid_value)
+
+    if not article:
+        raise ImportArticleError(f"No article found for {identifier}")
+    return article
+
+
+def import_arxiv(arxiv_id: str) -> dict:
```
harunurhan: types! 🥇 👌

inspirehep/records/api/literature.py
```diff
@@ -143,3 +169,141 @@ def _dump_for_es(self):
...
+        dict: a json object with either the parsed article or the occurred
+        error.
+
+    Raises:
+        ImportConnectionError: if the request doesn't succeed.
+        ImportParsingError: if any error occurs during the response parsing.
+
+    """
+    doi = requests.utils.quote(doi, safe="")
+    url = CROSSREF_URL.replace(PLACEHOLDER, doi)
```
harunurhan: I would go for `format` here.

inspirehep/records/api/literature.py
```diff
@@ -143,3 +169,141 @@ def _dump_for_es(self):
         serialized_data = super()._dump_for_es()
         serialized_data["_ui_display"] = json.dumps(self.get_ui_data())
         return serialized_data
+
+
+def import_article(identifier: str) -> dict:
```
drjova: we shouldn't introduce types in this PR, this is a bigger decision to make an affect the rest of the project as well ☺️

harunurhan: tbh he's gonna put them in doc string otherwise or remove, perhaps it's not bad to have them like that. It's not like the PR is introducing also a static type checker that is running on CI, so this alone does not seem a big decision to me.

pazembrz: Actually I think that it's great that Antonio introduced them. As Harun said, it's not like introducing static checking in CI but it's a good start for at least discussing about them ;)

ammirate:
End-to-end are passing locally but failing on CircleCI with the error
```
File "/inspirehep/modules/workflows/tasks/matching.py", line 418, in _get_wfs_same_source

'acquisition_source.source' -> AttributeError: 'NoneType' object has no attribute 'lower'
```
Looks like the submission that triggers the workflow is missing `'source': 'submission'`.
Any idea why this is not happening locally or how to debug it?
@harunurhan  @drjova

drjova:
Try to update your local deps and images

ammirate:
The E2E issue has been fixed in https://github.com/inspirehep/inspirehep/pull/141.

inspirehep/records/api/literature.py
harunurhan: do you have a strong reason not to just inline url `string` with `format` here?

ammirate: I don't, what's yours? :)

e2e/utils/user.js
```diff
@@ -4,7 +4,10 @@ async function submitLoginForm(page, username, password) {
   await page.type('[data-test-id=email]', username);
   await page.type('[data-test-id=password]', password);
   await page.click('[data-test-id=login]');
-  await page.waitFor(() => !document.querySelector('[data-test-id=login]'));
+  // await page.waitFor(() => !document.querySelector('[data-test-id=login]'));
```
harunurhan: seriously we merge it like this?

inspirehep/records/api/literature.py
harunurhan: * `replace` seems unnecessary for a simple string contact
* it's easy to read when it's next to the place where it is used, since it's only used just here
* we use `format` everywhere else to create urls

PR #129, opened 2019-04-11
Hide stuff api

harunurhan:


inspirehep/accounts/api.py
```diff
@@ -19,3 +19,10 @@ def check_if_is_logged_in(*args, **kwargs):
         return func(*args, **kwargs)
 
     return check_if_is_logged_in
+
+
+def is_superuser_or_cataloger_logged_in():
+    if current_user.is_authenticated:
+        user_roles = [role.name for role in current_user.roles]
+        return any(role in user_roles for role in ("superuser", "cataloger"))
```
michamos: this is a set intersection:
```python
user_roles = {role.name for role in current_user.roles}
return user_roles & {"superuser", "cataloger"}
```

inspirehep/records/marshmallow/authors/base.py
```diff
@@ -11,34 +11,66 @@
...
+
+class AuthorsMetadataRawAdminSchemaV1(AuthorsMetadataRawPublicSchemaV1):
+    class Meta:
+        include = {"$schema": fields.Raw()}
+
+    _private_notes = fields.Raw(dump_only=True)
+    _collections = fields.Raw(dump_only=True)
+    deleted = fields.Raw(dump_only=True)
+    deleted_records = fields.Raw(dump_only=True)
+    legacy_categories = fields.Raw(dump_only=True)
```
michamos: what are `legacy_categories`?

inspirehep/records/marshmallow/authors/base.py
```diff
@@ -11,34 +11,66 @@
...
+    legacy_creation_date = fields.Raw(dump_only=True)
+    legacy_version = fields.Raw(dump_only=True)
```
michamos: why are all these fields (from `deleted` onwards) only available for admins? they are not private.

inspirehep/serializers.py
```diff
@@ -0,0 +1,25 @@
...
+        super().__init__(**kwargs)
+
+    def dump(self, obj, context=None):
+        """Serialize object with schema."""
+        schema = next(
+            schema
+            for condition, schema in self.condition_schema_pairs
+            if condition is None or condition(obj)
+        )
+        return schema(context=context).dump(obj).data
```
michamos: what is the `context` variable here?

inspirehep/accounts/api.py
harunurhan: i did with `set` but looked more clear this way, since it's always 1 or 2 items doesn't matter performance wise.

inspirehep/records/marshmallow/authors/base.py
harunurhan: probably mixed two field names together 😄

inspirehep/records/marshmallow/authors/base.py
harunurhan: they aren't used for now, and didn't wanna expose them since they are legacy related stuff, might go away in the future. Should I include them on the public one?

inspirehep/serializers.py
harunurhan: No idea where and how it is used, I just wanted to override https://github.com/inveniosoftware/invenio-records-rest/blob/c4185c4eac40ccd8c9b532b2ddccce9f91aa63a8/invenio_records_rest/serializers/marshmallow.py#L26

inspirehep/records/marshmallow/authors/base.py
michamos: I would say so, they are not private so no need to hide them.

inspirehep/serializers.py
```diff
@@ -0,0 +1,25 @@
...
+# inspirehep is free software; you can redistribute it and/or modify it under
+# the terms of the MIT License; see LICENSE file for more details.
+
+
+from invenio_records_rest.serializers.json import JSONSerializer
+
+
+class ConditionalMultiSchemaJSONSerializer(JSONSerializer):
+    def __init__(self, condition_schema_pairs, **kwargs):
+        """Initialize record."""
```
drjova: irrelevant

tests/integration/records/test_views.py
```diff
@@ -19,6 +19,7 @@ def test_literature_search_application_json_get(
 ):
     data = {
         "control_number": 666,
+        "document_type": ["article"],
```
drjova: it's automatically added as well as ``_collections``

tests/integration/records/test_views.py
```diff
@@ -669,10 +655,15 @@ def test_institutions_search_json_get(api_client, db, create_record_factory):
 
 
 def test_literature_facets_collaboration(api_client, db, create_record_factory):
-    data = {"collaborations": [{"value": "Alice"}, {"value": "Collab"}]}
-    record_1 = create_record_factory("lit", data=data, with_indexing=True)
-    data = {"collaborations": [{"value": "Alice"}]}
-    record_2 = create_record_factory("lit", data=data, with_indexing=True)
+    data_1 = {
```
drjova: this should be a ``create_record`` as well + you can always skip ``document_type``, ``control_number``, ``_collections```

inspirehep/records/marshmallow/authors/base.py
```diff
@@ -11,34 +11,62 @@
...
     project_membership = fields.Raw(dump_only=True)
+    public_notes = fields.Raw(dump_only=True)
     status = fields.Raw(dump_only=True)
     stub = fields.Raw(dump_only=True)
     urls = fields.Raw(dump_only=True)
+
+
+class AuthorsMetadataRawAdminSchemaV1(AuthorsMetadataRawPublicSchemaV1):
+    class Meta:
+        include = {"$schema": fields.Raw()}
```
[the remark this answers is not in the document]
harunurhan: not used anywhere

tests/integration/records/test_views.py
harunurhan: I don't want to rely on "automatically" edition, when it changes it will break the test, but feature would be already working.

inspirehep/records/marshmallow/authors/base.py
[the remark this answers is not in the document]
drjova: but it's an important piece of information and it's not related only to ``Admins``, no? cc @michamos

tests/integration/records/test_views.py
harunurhan: ditto

tests/helpers/factories/models/user_access_token.py
```diff
@@ -20,10 +20,14 @@ class Meta:
         model = User
 
     @classmethod
-    def _create(cls, model_class, *args, **kwargs):
+    def _create(cls, model_class, role="user", *args, **kwargs):
```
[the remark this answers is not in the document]
harunurhan: let's do it when we need it?

inspirehep/records/marshmallow/authors/base.py
```diff
@@ -11,34 +11,65 @@
...
     advisors = fields.Raw(dump_only=True)
     arxiv_categories = fields.Raw(dump_only=True)
     awards = fields.Raw(dump_only=True)
     birth_date = fields.Raw(dump_only=True)
     control_number = fields.Raw(dump_only=True)
     death_date = fields.Raw(dump_only=True)
-    deleted = fields.Raw(dump_only=True)
-    email_addresses = fields.Raw(dump_only=True)
+    # deleted = fields.Raw(dump_only=True)
+    # deleted_records = fields.Raw(dump_only=True)
```
michamos: these two fields should be added here too

harunurhan: also for literature too right?

michamos: right

PR #131, opened 2019-04-11
alembic: add alembic to inspirehep

pazembrz:
* ADD: Adds alembic to the project with all migrations which will bring all tables for inspirehep and tables for inspire-next because it's not possible to run migrations form inspirehep and inspire-next in the same db, ad both projects are using same modules and tables and this migrations won't work together.

* INSPIR-2213

### DIFF
--- a/inspirehep/alembic/7be4c8b5c5e8_inspirehep_new_migrations.py
+++ b/inspirehep/alembic/7be4c8b5c5e8_inspirehep_new_migrations.py
@@ -1,0 +1,323 @@
+# -*- coding: utf-8 -*-
+#
+# Copyright (C) 2019 CERN.
+#
+# inspirehep is free software; you can redistribute it and/or modify it under
+# the terms of the MIT License; see LICENSE file for more details.
+
+"""Inspirehep initial revision of migrations
+which makes db identical like in inspire-next"""
+
+import sqlalchemy as sa
+from alembic import op
+from sqlalchemy.dialects import postgresql
+
+# revision identifiers, used by Alembic.
+revision = "7be4c8b5c5e8"
+down_revision = None
+branch_labels = ("inspirehep_new",)
+depends_on = "dbdbc1b19cf2"
+
+
+def upgrade():
+    # """Upgrade database."""
+    op.create_table(
+        "workflows_workflow",
+        sa.Column("uuid", postgresql.UUID(), autoincrement=False, nullable=False),
+        sa.Column("name", sa.VARCHAR(length=255), autoincrement=False, nullable=False),
+        sa.Column(
+            "created", postgresql.TIMESTAMP(), autoincrement=False, nullable=False
+        ),
+        sa.Column(
+            "modified", postgresql.TIMESTAMP(), autoincrement=False, nullable=False
+        ),
+        sa.Column("id_user", sa.INTEGER(), autoincrement=False, nullable=False),
+        sa.Column(
+            "extra_data",
+            postgresql.JSON(astext_type=sa.Text()),
+            autoincrement=False,
+            nullable=False,
+        ),
+        sa.Column("status", sa.INTEGER(), autoincrement=False, nullable=False),
+        sa.PrimaryKeyConstraint("uuid", name="pk_workflows_workflow"),
+    )
+    op.create_table(
+        "workflows_object",
+        sa.Column(
+            "id",
+            sa.INTEGER(),
+            server_default=sa.text("nextval('workflows_object_id_seq'::regclass)"),
+            autoincrement=True,
+            nullable=False,
+        ),
+        sa.Column(
+            "data",
+            postgresql.JSON(astext_type=sa.Text()),
+            autoincrement=False,
+            nullable=False,
+        ),
+        sa.Column(
+            "extra_data",
+            postgresql.JSON(astext_type=sa.Text()),
+            autoincrement=False,
+            nullable=False,
+        ),
+        sa.Column("id_workflow", postgresql.UUID(), autoincrement=False, nullable=True),
+        sa.Column("status", sa.INTEGER(), autoincrement=False, nullable=False),
+        sa.Column("id_parent", sa.INTEGER(), autoincrement=False, nullable=True),
+        sa.Column("id_user", sa.INTEGER(), autoincrement=False, nullable=False),
+        sa.Column(
+            "created", postgresql.TIMESTAMP(), autoincrement=False, nullable=False
+        ),
+        sa.Column(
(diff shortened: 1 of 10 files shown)


inspirehep/alembic/7be4c8b5c5e8_inspirehep_new_migrations.py
```diff
@@ -0,0 +1,323 @@
...
+
+# revision identifiers, used by Alembic.
+revision = "7be4c8b5c5e8"
+down_revision = None
+branch_labels = ("inspirehep_new",)
+depends_on = "dbdbc1b19cf2"
+
+
+def upgrade():
+    # """Upgrade database."""
```
drjova: why? ☺️

inspirehep/alembic/7be4c8b5c5e8_inspirehep_new_migrations.py
```diff
@@ -0,0 +1,323 @@
...
+            ondelete="CASCADE",
+        ),
+        sa.PrimaryKeyConstraint(
+            "record_uuid", "source", name="pk_workflows_record_sources"
+        ),
+    )
+
+
+def downgrade():
+    # """Downgrade database."""
```
drjova: ditto

tests/integration/conftest.py
```diff
@@ -25,6 +26,11 @@ def app_config(app_config):
     # add extra global config if you would like to customize the config
     # for a specific test you can chagne create fixture per-directory
     # using ``conftest.py`` or per-file.
+    app_config[
+        "DEBUG"
+    ] = (
+        False
+    )  # Due to flask error it has to be False otherwise Alembic __init__ will fail.
```
drjova: the style here is weird :/ is it from black?

tests/integration-async/records/conftest.py
```diff
@@ -33,7 +34,11 @@
 def app():
     app = invenio_create_app()
     app_config = {}
-    app_config["DEBUG"] = True
+    app_config[
+        "DEBUG"
+    ] = (
+        False
+    )  # Due to flask error it has to be False otherwise Alembic __init__ will fail.
```
drjova: ditto

setup.py
```diff
@@ -59,6 +59,7 @@
         "invenio_jsonschemas.schemas": ["inspire_records_schemas = inspire_schemas"],
         "invenio_search.mappings": ["records = inspirehep.search.mappings"],
         "invenio_celery.tasks": ["invenio_indexer = inspirehep.records.indexer.tasks"],
+        "invenio_db.alembic": ["inspirehep_new = inspirehep:alembic"],
```
drjova:
```suggestion
        "invenio_db.alembic": ["inspirehep = inspirehep:alembic"],
```

drjova:
@pazembrz maybe I'm missing something but it's really confusing. Why we need in the migrations of ``inspirehep`` to have ``workflows``, ``crawlers`` etc? The models are all related to ``inspire-next`` and should be only managed by this application.

tests/integration/conftest.py
```diff
@@ -25,6 +26,11 @@ def app_config(app_config):
     # add extra global config if you would like to customize the config
     # for a specific test you can chagne create fixture per-directory
     # using ``conftest.py`` or per-file.
+    app_config[
+        "DEBUG"
+    ] = (
+        False
```
drjova: Do we have specific explanation why this is failing?

tests/integration/conftest.py
pazembrz: Unfortunately yes. As I  was formatting this completely different.

inspirehep/alembic/7be4c8b5c5e8_inspirehep_new_migrations.py
pazembrz: It's only for allowing to run tests from inspire-next.  (we were talking about it many times ;) )
This migration will never be executed on production or QA, it's for local environment only.

setup.py
pazembrz: It's called `inspirehep` already in inspire-next. They are separate environments but we should use different names to be safe.

tests/integration/conftest.py
pazembrz: No, we don't . It's a known issue, and it was known when we were implementing alembic in inspire-next. Nothing changed from this time.  This may be similar problem to this https://chase-seibert.github.io/blog/2015/06/12/flask-werkzeug-reloader-python-dash-m.html# but I never had time ti investigate it.

setup.py
drjova: it doesn't matter they are different apps in different ``envs`` it's safe to have the same name :)

tests/integration/conftest.py
drjova: oh I get it is because you have the comment after ``False`` should be in a different line i.e.

```
# Due to flask error it has to be False otherwise Alembic __init__ will fail.
app_config["DEBUG"] = False
```

inspirehep/alembic/7be4c8b5c5e8_inspirehep_new_migrations.py
drjova: But still :/ I don't understand why is on``inspirehep``? ``inspirehep`` should know **nothing** about ``inspire-next``, right? Why we cannot just run inspirehep migrations and then ``inspire-next`` migrations  to add the missing tables?

pazembrz: We were talking about this already. You cannot run migrations from inspirehep and inspire-next on the same db, as both projects are using the same modules.

tests/integration/conftest.py
pazembrz: You are right.

setup.py
pazembrz: As you wish, but I  do not like it.

inspirehep/alembic/7be4c8b5c5e8_inspirehep_new_migrations.py
drjova: The following scenario will not work? 

```bash
$ (inspirehep) -> poetry run inspirehep alembic upgrade
```

```shell
$ (inspire-next) -> inspirehep alembic upgrade
```

pazembrz: Of course it won't. All tables for invenio-records, invenio-files will be already in db after first command, and alembic from second one will scream and refuse to migrate.

drjova: what about ignore tables?

PR #132, opened 2019-04-11
search: source based on the Content-Type

drjova:
* BREAKING: Renames ``create_record`` fixture factory to ``create_record_factory``.

* Creates ``create_record`` fixture to create records from the application level.

* Adds elastic search source based on the ``Content-Type``.

### DIFF
--- a/inspirehep/search/factories/search.py
+++ b/inspirehep/search/factories/search.py
@@ -16,13 +16,17 @@ from .filter import inspire_filter_factory
 
 def get_search_with_source(search):
     source = current_app.config.get("SEARCH_SOURCE_INCLUDES")
-    if not source:
+    has_accept_mimetypes = len(request.accept_mimetypes) > 0
+
+    if not source or not has_accept_mimetypes:
         return search
 
     if isinstance(search, LiteratureSearch):
-        source_literature = source.get("literature")
-        if source_literature:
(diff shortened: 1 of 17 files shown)


tests/integration/records/test_cli.py
```diff
@@ -17,7 +17,7 @@
 
 
 @pytest.mark.vcr()
-def test_create_record_with_one_url(api_client, db, script_info):
+def test_create_record_factory_with_one_url(api_client, db, script_info):
```
harunurhan: `find and replace all` went too much

inspirehep/search/factories/search.py
```diff
@@ -16,13 +16,17 @@
...
 
     if isinstance(search, LiteratureSearch):
-        source_literature = source.get("literature")
-        if source_literature:
-            search = search.source(includes=source_literature)
+        source_literature = source.get("literature", {})
+        accept_type = next(request.accept_mimetypes.values())
+        source_literature_mimetype = source_literature.get(accept_type)
+        if source_literature_mimetype:
+            search = search.source(includes=source_literature_mimetype)
```
harunurhan: it's everything if `includes` is `None` right?

drjova: yes, it will never get there with ``None`` though. It'll `` return search``

PR #133, opened 2019-04-12
Display conference info

monaawi:
* Fixed the conference info serializer to retrieve the conference record properly and return correct data
* Added page_start, page_end and acronyms to the serializer
* Added ConferenceInfo component to display conference as Contribution to {acronyms or title}, {start_page - end_page}

### DIFF
--- a/inspirehep/records/marshmallow/literature/common/conference_info_item.py
+++ b/inspirehep/records/marshmallow/literature/common/conference_info_item.py
@@ -4,7 +4,7 @@
 #
 # inspirehep is free software; you can redistribute it and/or modify it under
 # the terms of the MIT License; see LICENSE file for more details.
-
+from invenio_pidstore.errors import PIDDoesNotExistError
 from marshmallow import Schema, fields, pre_dump
 
 from inspirehep.pidstore.api import PidStoreBase
@@ -14,6 +14,9 @@ from inspirehep.records.api import InspireRecord
 class ConferenceInfoItemSchemaV1(Schema):
     titles = fields.Raw()
     control_number = fields.Raw()
+    page_start = fields.Raw()
+    page_end = fields.Raw()
+    acronyms = fields.Raw()
 
     @pre_dump
     def resolve_conference_record_as_root(self, pub_info_item):
@@ -21,15 +24,16 @@ class ConferenceInfoItemSchemaV1(Schema):
         if conference_record is None:
             return {}
 
+        _, recid = PidStoreBase.get_pid_from_record_uri(conference_record.get("$ref"))
         try:
-            _, recid = PidStoreBase.get_pid_type_from_endpoint(
-                conference_record.get("$ref")
+            conference = InspireRecord.get_record_by_pid_value(
+                pid_value=recid, pid_type="con"
             )
-        except TypeError:
+        except PIDDoesNotExistError:
             return {}
-
-        conference = InspireRecord.get_record_by_pid_value(recid, "con")
 
         titles = conference.get("titles")
         if not titles:
             return {}
+        pub_info_item.update(conference)
+        return pub_info_item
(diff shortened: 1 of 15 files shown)


inspirehep/records/marshmallow/literature/common/conference_info_item.py
```diff
@@ -14,22 +14,26 @@
...
+            conference = InspireRecord.get_record_by_pid_value(
+                pid_value=recid, pid_type="con"
             )
-        except TypeError:
+        except PIDDoesNotExistError:
             return {}
 
-        conference = InspireRecord.get_record_by_pid_value(recid, "con")
-
         titles = conference.get("titles")
```
harunurhan: shouldn't this be `acronyms` now, because we need acronyms to be able display.

harunurhan: and need test for this.

ui/src/literature/components/__tests__/ConferenceInfo.test.jsx
```diff
@@ -0,0 +1,85 @@
+import React from 'react';
+import { shallow } from 'enzyme';
+import { fromJS } from 'immutable';
+
+import ConferenceInfo from '../ConferenceInfo';
+
+describe('PublicationInfo', () => {
```
harunurhan: leftover

ui/src/literature/components/ConferenceInfo.jsx
```diff
@@ -0,0 +1,39 @@
+import React, { Component } from 'react';
+import PropTypes from 'prop-types';
+import { Map, List } from 'immutable';
+import InlineList from '../../common/components/InlineList';
+import { getPageDisplay } from '../utils';
+
+class ConferenceInfo extends Component {
+  static renderAcronyms(acronyms) {
```
harunurhan: can you make this non `static` access `acronyms` from props

ui/src/literature/components/__tests__/__snapshots__/LiteratureItem.test.jsx.snap
```diff
@@ -302,6 +308,9 @@ exports[`LiteratureItem renders with all props set, including sub props 1`] = `
       }
       wrapperClassName="di"
     />
+    <ConferenceInfoList
```
harunurhan: ` renders with all props set, including sub props 1` should have minimal `conference_info` too.

ui/src/literature/utils.js
```diff
@@ -0,0 +1,12 @@
+// eslint-disable-next-line import/prefer-default-export
+export function getPageDisplay(pagesInfo) {
```
harunurhan: I would create separate tests for this, since it's really easy to tests, although I get that it's tested through other ones.

ui/src/literature/components/ConferenceInfo.jsx
harunurhan: we have `static` render methods when we can't exact parameter from `props`. For example for functions that we use in `someCollection.map`

inspirehep/records/marshmallow/literature/common/conference_info_item.py
monaawi: Actually, if we don't have acronyms, we fall back to the titles so we ultimately need titles to be able to display.

monaawi:
> Filtering out conferences from real `publication_info` is missing

Actually we already have that in publication_info, we check if journal_title is None and if it is a conference then it doesn't have a journal_title so we are already filtering out conferences from there

harunurhan:
@monaawi it is both `pubinfo_freetext` and `journal_title` are `None`, and I think conferences can have `pubinfo_freetext` so it's better to add a check for `conference_record`

PR #134, opened 2019-04-12
Fix wrong record data and add all collections to record importer cli

monaawi:
* Fixed wrong record data
* Used InspireRecord.create for cli record creation


drjova:
@monaawi could you please update the commit message title? ☺️

PR #136, opened 2019-04-16
ui: display conference_info inline with bulletpoint in search results

monaawi:
* displayed conference_info between publication_info and eprint with bullet points in between
* INSPIR-2069


ui/src/literature/components/LiteratureItem.jsx
```diff
@@ -101,9 +111,13 @@ class LiteratureItem extends Component {
             wrapperClassName="di"
             publicationInfo={publicationInfo}
           />
-          {this.renderBulletIfPublicationInfoAndEprintsNotEmpty()}
+          {this.renderBulletIfPublicationInfoAndConferenceInfoNotEmpty()}
```
harunurhan: Now that this has more than 2 items, can you do something like https://github.com/inspirehep/inspirehep/blob/master/ui/src/literature/components/ReferenceItem/ReferenceItem.jsx#L86 instead.

Btw here we can see limitation of `InlineList` and approach of passing the `data` along with `renderItem` instead of directly markup. But that's a later refactor.

ui/src/literature/components/__tests__/LiteratureItem.test.jsx
```diff
@@ -55,4 +55,81 @@ describe('LiteratureItem', () => {
     );
     expect(wrapper).toMatchSnapshot();
   });
+
+  it('renders bulletpoint if conference_info and publication_info present', () => {
```
harunurhan: you don't care about bullet points anymore, having one test case where you have all props with full data (all fields of literature that are used) and one case with only required ones are enough I think

PR #137, opened 2019-04-16
ui: display public_notes

monaawi:
* Displayed public notes under abstract in an inLineList with middle dot as a separator
* INSPIR-1830

### DIFF
--- a/ui/src/literature/components/PublicNotesList.jsx
+++ b/ui/src/literature/components/PublicNotesList.jsx
@@ -1,0 +1,34 @@
+import React, { Component } from 'react';
+import PropTypes from 'prop-types';
+import { List } from 'immutable';
+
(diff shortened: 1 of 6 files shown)


ui/src/literature/components/PublicNotesList.jsx
```diff
@@ -0,0 +1,30 @@
...
+
+class PublicNotesList extends Component {
+  render() {
+    const { publicNotes } = this.props;
+    return (
+      <InlineList
+        label="Note"
+        items={publicNotes}
+        extractKey={note => note.get('value')}
+        renderItem={note => <span>{note.get('value')}</span>}
```
harunurhan: could you actually make `static getValue` and use it for both `extractKey` and `renderItem`

PR #154, opened 2019-04-30
indexer related records refactoring

harunurhan:


inspirehep/records/api/base.py
```diff
@@ -967,8 +948,8 @@ def _dump_for_es(self):
         Returns:
             dict: Properly serialized and prepared record
         """
-        serialized_data = self.get_serialized_data()
-        return serialized_data
+        enhanced_data = self.get_enhanced_es_data()
```
drjova:
```suggestion
        return self.get_enhanced_es_data()
```

inspirehep/records/api/authors.py
```diff
@@ -5,6 +5,8 @@
 # inspirehep is free software; you can redistribute it and/or modify it under
 # the terms of the MIT License; see LICENSE file for more details.
 
+from inspirehep.records.marshmallow.authors import AuthorsMetadataRawAdminSchemaV1
```
drjova: why not relative as the other imports?

harunurhan: becuase it's not relative on `literature`

PR #167, opened 2019-05-06
Add more ranges to the citation summary endpoint

monaawi:
* Added more ranges to the citation summary endpoint
* INSPIR-2301


tests/integration/records/test_views.py
```diff
@@ -977,3 +1001,124 @@ def test_h_index_with_as_many_papers_as_citations(
...
+                            {
+                                "key": "250.0-500.0",
+                                "from": 250.0,
+                                "to": 500.0,
+                                "doc_count": 0,
+                            },
+                            {"key": "500.0-*", "from": 500.0, "doc_count": 0},
+                        ]
+                    },
+                    "average_citations": {"value": None},
```
drjova: is this correct?

tests/integration/records/test_views.py
```diff
@@ -977,3 +1001,124 @@ def test_h_index_with_as_many_papers_as_citations(
...
+                    "doc_count": 1,
+                    "citations_count": {"value": 8.0},
+                    "citation_buckets": {
+                        "buckets": [
+                            {"key": "0.0-1.0", "from": 0.0, "to": 1.0, "doc_count": 0},
+                            {
+                                "key": "1.0-10.0",
+                                "from": 1.0,
+                                "to": 10.0,
+                                "doc_count": 1,
```
drjova: maybe I'm missing something, you have ``[409, 83, 26]``, none of them should go here right?

tests/integration/records/test_views.py
monaawi: This is the behavior of the elastic search average aggregation. Basically, we are doing an average over 0 documents so they do not return 0 but None. We handle this on the UI though, where the average appears as 0. 
Here is a discussion on the elastic search thing https://github.com/elastic/elasticsearch/issues/31920

tests/integration/records/test_views.py
monaawi: So, I added the doc_type book chapter filter and only the book_chapter_paper has that. The others which have citations [409, 83, 26] are published and citeable papers but they are not book chapter so this is the expected behavior (this is actually what the test is testing - that other filters are applied to the results as well)

PR #172, opened 2019-05-09
submissions: populate $refs if field autocompleted

harunurhan:


ui/src/submissions/common/components/SuggesterField.jsx
```diff
@@ -18,19 +20,38 @@ class SuggesterField extends Component {
...
+      form.setFieldValue(recordFieldPath, null);
+    }
+
+    this.hasChangedBySuggestionSelection = false;
+  }
+
+  onSelect(_, selectedOptionComponent) {
+    const { self } = selectedOptionComponent.props.result._source;
+    const { form, recordFieldPath } = this.props;
+    form.setFieldValue(recordFieldPath, self);
```
drjova: what will happen if I delete an autocompleted value? Example scenario:

I autocomplete conference record and then delete the value, will the ``conference_record`` field stay?

harunurhan: nope, you can see `onChange` where it's set to `null`

tests/unit/submissions/test_literature.py
```diff
@@ -313,3 +322,18 @@ def test_load_authors():
...
+        "parent_book_record": {"$ref": "http://api/literature/123"},
+    }
+
+    expected = {
+        **DEFAULT_LOAD,
+        "publication_info": [{"parent_record": {"$ref": "http://api/literature/123"}}],
+    }
+    result = Literature().load(form).data
+
+    assert result == expected
```
drjova: just to note, ``journals`` and ``conferences`` are missing from tests

PR #177, opened 2019-05-13
Interactive citatation table

harunurhan:


ui/src/common/components/CitationSummaryTable/CitationSummaryTable.jsx
```diff
@@ -0,0 +1,123 @@
...
+    );
+  }
+}
+
+CitationSummaryTable.propTypes = {
+  publishedBucket: PropTypes.instanceOf(Map),
+  citeableBucket: PropTypes.instanceOf(Map),
+  hIndex: PropTypes.instanceOf(Map),
+  renderNumberOfCiteablePapers: PropTypes.func,
+  renderNumberOfPublishedPapers: PropTypes.func,
```
monaawi: Why did you make these as props instead of just putting the NumberOfPublished/CiteableContainer in here? Won't we most probably have the same behavior for the literature search as well? Is there a case where the behavior is different?

ui/src/common/components/CitationSummaryTable/index.js
```diff
@@ -0,0 +1,3 @@
+import CitateSummaryTable from './CitationSummaryTable';
```
monaawi: CitateSummaryTable?

ui/src/common/components/CitationSummaryTable/CitationSummaryTable.jsx
harunurhan: we usually use `render` methods rather passing down components. That was the only reason.

monaawi: I would pass down the component in this case as I think we will need it the same way for literature search and this will add duplicated code in the search and author detail page

harunurhan: well for now it seems to cause duplication of single line functions, but when it gets complicated (and it will), this will be better than passing component that expects certain props..

ui/src/authors/containers/NumberOfPublishedPapersContainer.jsx
```diff
@@ -0,0 +1,18 @@
+import { connect } from 'react-redux';
+
+import {
+  fetchAuthorPublications,
+  fetchAuthorPublicationsFacets,
+} from '../../actions/authors';
+import LinkLikeButton from '../../common/components/LinkLikeButton';
+
+const PUBLISHED_QUERY = { q: 'citeable:true and refereed:true' };
```
monaawi: Is it possible to move PUBLISHED_QUERY and CITEABLE_QUERY so we can share them with the graph as well?

harunurhan: you can move them out in your PR.

ui/src/common/components/CitationSummaryTable/CitationSummaryTable.jsx
```diff
@@ -0,0 +1,123 @@
...
+                      : renderNumberOfCiteablePapers(
+                          citeableBucket.get('doc_count')
+                        )}
+                  </td>
+                  <td>
+                    {!publishedBucket.get('doc_count')
+                      ? '0'
+                      : renderNumberOfPublishedPapers(
+                          publishedBucket.get('doc_count')
+                        )}
```
drjova: maybe move it out??

drjova: as well as the above :)

PR #182, opened 2019-05-15
marshmallow: add default raw marshmallow serializer and apply it to…

pazembrz:
…all types which are missing serializer

* Add: Default raw serializer.
* Add: Use this serializer to all records without serializer.
* Add: Add enchancements for affiliation, experiment, and journal serializers
* Fix: Change test layout. Moved all api tests to separate folder, all indexer tests to separate folder and move to separate files, based by testing type.

**INSPIR-2356**


tests/unit/records/marshmallow/conferences/__init__.py
```diff
@@ -0,0 +1,7 @@
+# -*- coding: utf-8 -*-
+#
+# Copyright (C) 2019 CERN.
+#
+# inspirehep is free software; you can redistribute it and/or modify it under
+# the terms of the MIT License; see LICENSE file for more details.
+# flake8: noqa
```
drjova: left over?

tests/unit/records/marshmallow/data/__init__.py
```diff
@@ -0,0 +1,7 @@
+# -*- coding: utf-8 -*-
+#
+# Copyright (C) 2019 CERN.
+#
+# inspirehep is free software; you can redistribute it and/or modify it under
+# the terms of the MIT License; see LICENSE file for more details.
+# flake8: noqa
```
drjova: ditto

tests/integration/records/views/test_views_journals.py
```diff
@@ -0,0 +1,67 @@
+# -*- coding: utf-8 -*-
+#
+# Copyright (C) 2019 CERN.
+#
+# inspirehep is free software; you can redistribute it and/or modify it under
+# the terms of the MIT License; see LICENSE file for more details.
+
+
+def test_journals_application_json_get(api_client, db, create_record_factory):
+    record = create_record_factory("jou", with_indexing=True)
```
drjova: can we just reuse the create_record now?

tests/integration/records/views/test_views_jobs.py
```diff
@@ -0,0 +1,67 @@
+# -*- coding: utf-8 -*-
+#
+# Copyright (C) 2019 CERN.
+#
+# inspirehep is free software; you can redistribute it and/or modify it under
+# the terms of the MIT License; see LICENSE file for more details.
+
+
+def test_jobs_application_json_get(api_client, db, create_record_factory):
```
drjova: this tests are quite useless, since we're not testing the actual API response for each "Content-Type"

tests/integration/records/views/test_views_data.py
```diff
@@ -0,0 +1,67 @@
...
+    create_record_factory("dat", with_indexing=True)
+
+    expected_status_code = 200
+    response = api_client.get("/data")
+    response_status_code = response.status_code
+
+    assert expected_status_code == response_status_code
+
+
+def test_data_record_search_results(api_client, db, es_clear, create_record):
```
drjova: isn't this the same as above? because we test also the expected? (we can delete`` test_data_search_json_get``) this applies on the others too

tests/integration/records/views/test_views_journals.py
pazembrz: I didn't touch old tests for views except moving them to separate folder. If you wish I can change this tests to use create_record.

tests/integration/records/views/test_views_jobs.py
pazembrz: Still this tests are moved from test_views.py This are not tests created by me. You never told me to also redo this tests, you only requested to move them.

drjova:
@pazembrz I think this one is misising https://github.com/inspirehep/inspire-next/blob/master/inspirehep/modules/records/receivers.py#L214

pazembrz:
Looks that you are right, and it was missed in the task/refinement unfortunately. I will add it. And tests for it.

tests/unit/records/marshmallow/experiments/test_base.py
```diff
@@ -0,0 +1,105 @@
...
+
+def test_experiment_serializer_should_serialize_whole_basic_record():
+    schema = ExperimentsMetadataRawSchemaV1()
+    expected_result = {
+        "$schema": "http://localhost:5000/schemas/records/experiments.json",
+        "_collections": ["Experiments"],
+        "project_type": ["experiment"],
+    }
+
+    experiment = ExperimentsRecord(faker.record("exp"))
```
harunurhan: I would put data here explicitly rather than using `faker`s defaults, when you need to assert directly the data. Because now if we change default experiment record in `faker`, everything will fail unnecessarily.

tests/unit/records/marshmallow/institutions/test_base.py
```diff
@@ -0,0 +1,230 @@
...
+    schema = load_schema("institutions")
+    subschema = schema["properties"]["ICN"]
+
+    data = {
+        "$schema": "http://localhost:5000/schemas/records/institutions.json",
+        "ICN": ["CERN, Geneva"],
+        "legacy_ICN": "CERN",
+    }
+    record = InstitutionsRecord(faker.record("ins", data))
+    assert validate(record["ICN"], subschema) is None
```
harunurhan: schema is already validated in `record_provider` which is used by `faker`.

tests/unit/records/marshmallow/experiments/test_base.py
pazembrz: If I am not mistaken, we are updating data to fake records after generating fake data, so if we add in the future 'fake' experiment data it will still be overwritten by experiment put in `data` parameter so it should not break anything.

pazembrz: But right, here we are not providing any data in this test. I will add empty experiment key in data then.

inspirehep/records/marshmallow/base.py
```diff
@@ -0,0 +1,98 @@
...
+
+    @post_dump(pass_original=True)
+    def include_original_fields(self, object, original_data):
+        for key, value in original_data.items():
+            if key not in object and key not in self.exclude:
+                object[key] = original_data[key]
+        object = self.populate_recid_from_ref(object)
+        return strip_empty_values(object)
+
+    def populate_recid_from_ref(self, record):
```
drjova: this should be part of the other serializers as well on inspire-next is for all records (including ``lit`` and ``aut``) ☺️

pazembrz: Yes, it should, but other serializes were never a part of this task... There should be separate task which would make this also available for literature and authors.

inspirehep/records/marshmallow/base.py
```diff
@@ -0,0 +1,117 @@
...
+    def include_original_fields(self, object, original_data):
+        for key, value in original_data.items():
+            if key not in object and key not in self.exclude:
+                object[key] = original_data[key]
+        for dump_func in self._post_dumps:
+            object = dump_func(object, original_data)
+        return strip_empty_values(object)
+
+
+class PopulateRecidMixin(object):
```
harunurhan: I think both (`IncludeAll` too) should be `schema` or `mixin`if possible for consistency

tests/unit/records/test_base.py
```diff
@@ -330,3 +333,31 @@ def test_get_records_ids_by_pids_function_is_properly_called_with_parameters(PI_
...
+
+    record_data = {
+        "simple_key": {"$ref": "http://x/y/1"},
+        "key_with_record": {"$ref": "http://x/y/2"},
+        "record": {"$ref": "http://x/y/3"},
+        "embedded_list": [{"record": {"$ref": "http://x/y/4"}}],
+        "embedded_record": {"record": {"$ref": "http://x/y/5"}},
+    }
+
+    ConferencesMetadataRawFieldsSchemaV1().populate_recid_from_ref(record_data)
```
harunurhan: why not create a dummy schema class and use directly the mixin on it and test it that way?

tests/unit/records/marshmallow/journals/test_base.py
```diff
@@ -0,0 +1,77 @@
...
+    title_variants_schema = schema["properties"]["title_variants"]
+
+    data = {
+        "$schema": "http://localhost:5000/schemas/records/journals.json",
+        "journal_title": {"title": "The Journal of High Energy Physics (JHEP)"},
+        "short_title": "JHEP",
+        "title_variants": ["JOURNAL OF HIGH ENERGY PHYSICS"],
+    }
+    record = JournalsRecord(faker.record("jou", data))
+    assert validate(record["journal_title"], journal_title_schema) is None
```
harunurhan: we can skip the `validate`

tests/unit/records/marshmallow/experiments/test_base.py
```diff
@@ -0,0 +1,107 @@
...
+        "legacy_name": "foo",
+        "long_name": "foobarbaz",
+        "name_variants": ["bar", "baz"],
+        "collaboration": {"value": "D0"},
+        "accelerator": {"value": "LHC"},
+        "experiment": {"short_name": "SHINE", "value": "NA61"},
+        "institutions": [{"value": "ICN"}],
+    }
+    record = ExperimentsRecord(faker.record("exp", data))
+    assert validate(record["legacy_name"], legacy_name_schema) is None
```
harunurhan: ditto

tests/unit/records/marshmallow/institutions/test_base.py
```diff
@@ -0,0 +1,200 @@
...
+    }
+    record = InstitutionsRecord(faker.record("ins", data))
+
+    schema = InstitutionsMetadataRawFieldsSchemaV1()
+    result = schema.dump(record).data["affiliation_suggest"]
+
+    expected = {
+        "input": [
+            "CERN",
+            u"Centre Européen de Recherches Nucléaires",
```
drjova: why we need the ``u`` here? :)

tests/unit/records/test_base.py
drjova: +1

pazembrz: ?

PR #186, opened 2019-05-16
config: add suggesters for all collections

ammirate:
* Note: only literature and authors are tested

* INSPIR-2364

### DIFF
--- a/inspirehep/config.py
+++ b/inspirehep/config.py
@@ -210,10 +210,28 @@ LITERATURE.update(
             # "application/vnd+inspire.latex.eu+x-latex": f"{INSPIRE_SERIALIZERS}:latex_search_response_eu",
             # "application/vnd+inspire.latex.us+x-latex": f"{INSPIRE_SERIALIZERS}:latex_search_response_us",
         },
-        "list_route": "/literature",
+        "list_route": "/literature/",
         "item_route": '/literature/<pid(lit,record_class="inspirehep.records.api.LiteratureRecord"):pid_value>',
         "create_permission_factory_imp": api_access_permission_check,
         "update_permission_factory_imp": api_access_permission_check,
+        "suggesters": {
+            "abstract_source": {
+                "completion": {
+                    "field": "abstracts.abstract_source_suggest",
+                },
+            },
+            "book_title": {
+                "_source": [
+                    "control_number",
+                    "self",
+                    "titles",
+                    "authors",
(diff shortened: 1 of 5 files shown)


tests/integration/test_config.py
```diff
@@ -0,0 +1,107 @@
...
+from invenio_search import current_search_client as es
+
+
+def _get_suggester_result(response, suggester):
+    return response.json[suggester][0]['options'][0]['_source']
+
+
+def test_literature_suggesters_book_title(api_client, create_record):
+    lit = create_record("lit", data={"titles":[{"title": "Suggested title"}]})
+    es.indices.refresh('records-hep')
```
drjova: do you need to do this? 🤔 it's already on ``create_record``

tests/integration/test_config.py
```diff
@@ -0,0 +1,107 @@
+# -*- coding: utf-8 -*-
```
drjova: I'd move this file to ``tests/integrations/config/views/test_suggesters.py`` WDYT?

tests/integration/test_config.py
```diff
@@ -0,0 +1,107 @@
...
+    assert result_rec_id == expected_rec_id
+
+
+def test_literature_suggesters_abstract_source(api_client, create_record):
+    lit = create_record("lit", data={"abstracts":[{"value": "Fancy abstract", "source": "WSP"}]})
+    es.indices.refresh('records-hep')
+
+    resp = api_client.get('literature/_suggest?abstract_source=ws')
+
+    result_rec_id = _get_suggester_result(resp, 'abstract_source')['control_number']
```
drjova: we should check also the other fields defined, some of the are required in the forms

tests/integration/test_config.py
```diff
@@ -0,0 +1,107 @@
...
+
+    result = resp.json['book_title'][0]['options']
+    expected = []
+
+    assert result == expected
+
+
+@pytest.mark.xfail
+def test_author_suggesters(api_client, create_record):
+    # FIXME: author_suggest field is not populated when indexing authors
```
drjova:
```python
@pytest.mark.xfail(reason="author_suggest field is not populated when indexing authors")
```
:)

tests/integration/test_config.py
ammirate: Actually nope :+1:

PR #191, opened 2019-05-21
citations: Remove non Literature collections from citations

pazembrz:
(they cannot cite)

### DIFF
--- a/inspirehep/records/api/base.py
+++ b/inspirehep/records/api/base.py
@@ -1030,7 +1030,12 @@ class InspireRecord(Record):
         """
         RecordCitations.query.filter_by(citer_id=self.id).delete()
 
-        if self.is_superseded() or self.get("deleted") or self.pid_type not in ["lit"]:
+        if (
+            self.is_superseded()
+            or self.get("deleted")
+            or self.pid_type not in ["lit"]
+            or "Literature" not in self["_collections"]
+        ):
             # Record is not eligible to cite
             return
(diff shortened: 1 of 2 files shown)


inspirehep/records/api/base.py
```diff
@@ -1030,7 +1030,12 @@ def _update_refs_in_citation_table(self, save_every=100):
         """
         RecordCitations.query.filter_by(citer_id=self.id).delete()
 
-        if self.is_superseded() or self.get("deleted") or self.pid_type not in ["lit"]:
+        if (
+            self.is_superseded()
+            or self.get("deleted")
+            or self.pid_type not in ["lit"]
+            or "Literature" not in self["_collections"]
+        ):
```
michamos: is this the right place? what happens if a record goes from being not deleted to deleted? does its citations get removed correctly? (if so the docstring is wrong).

michamos: what about making a `can_cite` property instead of inlining this?

tests/integration/records/test_api_literature.py
```diff
@@ -800,3 +800,27 @@ def test_literature_citation_count_property(base_app, db):
 
     assert record.citation_count == 1
     assert record2.citation_count == 0
+
+
+def test_literature_without_literature_collection_cannot_cite_record_whic_can_be_cited(
```
michamos: `whic`->`which`

PR #192, opened 2019-05-21
mappings: update job mapping

ammirate:
* INSPIR-2419

### DIFF
--- a/inspirehep/search/mappings/v5/records/jobs.json
+++ b/inspirehep/search/mappings/v5/records/jobs.json
@@ -8,19 +8,66 @@
       "numeric_detection": false,
       "properties": {
         "$schema": {
-          "type": "string"
+          "type": "keyword"
         },
         "_collections": {
-          "analyzer": "lowercase_analyzer",
-          "type": "string"
+          "normalizer": "lowercase_analyzer",
+          "type": "keyword"
         },
         "_private_notes": {
           "properties": {
             "source": {
-              "type": "string"
+              "type": "keyword"
             },
             "value": {
-              "type": "string"
+              "type": "text"
+            }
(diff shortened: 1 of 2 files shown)


inspirehep/search/mappings/v5/records/jobs.json
```diff
@@ -8,57 +8,100 @@
...
+              },
+              "type": "object"
+            }
+          },
+          "type": "object"
+        },
+        "acquisition_source": {
+          "properties": {
+            "$ref": {
+              "type": "keyword"
```
michamos: I think you got confused between `$ref` to reference other schema elements and `$ref` inside the fields to reference other records.

inspirehep/search/mappings/v5/records/jobs.json
```diff
@@ -8,57 +8,100 @@
...
+            }
+          },
+          "type": "object"
+        },
+        "arxiv_categories": {
+          "properties": {
+            "$ref": {
+              "type": "keyword"
```
michamos: same

michamos:
don't you also want to bump `inspire-schemas` or will this be done in another PR?

ammirate:
The bump will be done separately, since there's the whole chain of dependencies to update:
* inspire-utils
* inspire-matcher
* inspire-dojson
* inspire-json-merger
and something else probably.

harunurhan:
What is blocking this one? @ammirate 

I need it for the Jobs search... well also `inspire-schemas` bump (@monaawi would need new schema too)

ammirate:
The comment above is listing the dependencies for this. 
I'm merging it, since we can bump all of them now.

PR #193, opened 2019-05-21
search: server-side sort options

harunurhan:
* Sort option now comes with search results and rendered on the UI
dynamically, so that we can avoid hardcoding them on the UI and
creating different SortBy for each collections since sort options
change by collection.

* Makes [web] and [worker] services "restart: always" to avoid crashing
due to syntax errors during development

* Fixes literature record in `data/records` which had duplicate pid
(arxiv)

### DIFF
--- a/inspirehep/config.py
+++ b/inspirehep/config.py
@@ -399,7 +399,9 @@ CONFERENCES.update(
                     "acronyms",
                     "titles",
                     "address",
-                    "opening_date" "cnum" "control_number",
+                    "opening_date",
+                    "cnum",
+                    "control_number",
                     "self",
                 ],
                 "completion": {"field": "conferenceautocomplete"},
@@ -541,13 +543,13 @@ RECORDS_REST_FACETS = {
 RECORDS_REST_SORT_OPTIONS = {
     "records-hep": {
         "mostrecent": {
-            "title": "Most recent",
+            "title": "Most Recent",
             "fields": ["-earliest_date"],
             "default_order": "asc",  # Used for invenio-search-js config
             "order": 1,
         },
         "mostcited": {
-            "title": "Most cited",
+            "title": "Most Cited",
             "fields": ["-citation_count"],
             "default_order": "asc",  # Used for invenio-search-js config
             "order": 2,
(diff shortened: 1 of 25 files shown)


inspirehep/config.py
```diff
@@ -422,14 +399,10 @@
                     "acronyms",
                     "titles",
                     "address",
-                    "opening_date"
-                    "cnum"
-                    "control_number",
+                    "opening_date" "cnum" "control_number",
```
drjova: ``,``

tests/integration/records/serializers/test_json.py
```diff
@@ -314,6 +314,28 @@ def test_literature_json_ui_v1_response_search(api_client, db, create_record):
...
+    response = api_client.get("/literature", headers=headers)
+
+    response_status_code = response.status_code
+    response_data = json.loads(response.data)
+    sort_options = response_data["sort_options"]
+
+    assert expected_status_code == response_status_code
+    assert expected_sort_options == sort_options
+
+
```
drjova: it needs another test when is ``None``

ui/src/common/components/SortBy.jsx
```diff
@@ -2,38 +2,30 @@ import React, { Component } from 'react';
...
     return (
-      <SelectBox
-        onChange={onSortChange}
-        defaultValue={sort}
-        options={SORT_BY_OPTIONS}
-      />
+      sortOptions && (
+        <SelectBox
+          onChange={onSortChange}
+          defaultValue={sort || sortOptions[0].value}
```
drjova: doesn't it make sense to specify which one is the default one in the config?

tests/integration/records/serializers/test_json.py
```diff
@@ -622,6 +644,24 @@ def test_authors_default_json_v1_response_search(
...
+    expected_status_code = 200
+    expected_sort_options = None
+    response = api_client.get("/authors", headers=headers)
+
+    response_status_code = response.status_code
+    response_data = json.loads(response.data)
+    sort_options = response_data["sort_options"]
+
+    assert expected_status_code == response_status_code
+    assert expected_sort_options == sort_options
```
drjova: == ``None`` on the json response? :/

PR #195, opened 2019-05-23
views: Add Annual summary view

pazembrz:
* ADD: AnnualSummaryResource which returns json with summary of all citations for specified author

* INSPIR-2300


michamos:
stupid question: why did you decide to do it through a new endpoint instead of indexing this data in ES for each paper (you need to reindex anyway when the citation count changes)? then you can use an aggregation to compute the sums over all documents in the search query.

pazembrz:
New implementation in new PR https://github.com/inspirehep/inspirehep/pull/205

PR #197, opened 2019-05-23
Citations per year graph

harunurhan:


ui/src/actions/__tests__/citations.test.js
```diff
@@ -101,4 +105,57 @@ describe('citations - async action creator', () => {
     expect(store.getActions()).toEqual(expectedActions);
     done();
   });
+
+  it('creates CITATIONS_BY_YEAR_SUCCESS if succesful [authors]', async done => {
```
drjova:
```suggestion
  it('creates CITATIONS_BY_YEAR_SUCCESS if successful [authors]', async done => {
```

ui/src/actions/__tests__/citations.test.js
```diff
@@ -101,4 +105,57 @@ describe('citations - async action creator', () => {
...
+          },
+        },
+      }),
+    });
+    await store.dispatch(fetchCitationsByYear());
+    expect(store.getActions()).toEqual(expectedActions);
+    done();
+  });
+
+  it('creates CITATIONS_BY_YEAR_ERROR if unsuccesful [literature]', async done => {
```
drjova:
```suggestion
  it('creates CITATIONS_BY_YEAR_ERROR if unsuccessful [literature]', async done => {
```

ui/src/common/utils.js
```diff
@@ -48,6 +49,10 @@ export function getSizeOfArrayOrImmutableList(arrayOrList) {
   return Array.isArray(arrayOrList) ? arrayOrList.length : arrayOrList.size;
 }
 
+export function getFromObjectOrImmutableMap(objectOrMap, key) {
+  return Map.isMap(objectOrMap) ? objectOrMap.get(key) : objectOrMap[key];
```
drjova: maybe an option https://github.com/rtfeldman/seamless-immutable we are adding more and more helpers like this

harunurhan: As far as I can see this is completely different immutable library. Another thing is that we shouldn't add more of these but do a refactor to improve consistency at some point... Like all dumb components should take `JS` props and all state tree should be `immutable`.

harunurhan:
We can merge this and test when the backend is on QA too. There might be some issues how graph would look with dirty or small data (e.g. for only 2 years) but we can do a small patch for this.

PR #198, opened 2019-05-23
mappings: update hep mappings to tokenize $ref

drjova:
* Closes INSPIR-2427


drjova:
I'll do it in a separate there are too many changes in the tests. I'll prefer to test this fast for the migrator

PR #205, opened 2019-05-24
facets: Add citations by year facet

pazembrz:
* ADD: citations-by-year facet which returns json with summary of all citations for specified author

* INSPIR-2300

### DIFF
--- a/inspirehep/alembic/c6570e49b7b2_change_record_citations_to_records_.py
+++ b/inspirehep/alembic/c6570e49b7b2_change_record_citations_to_records_.py
@@ -1,0 +1,44 @@
+# -*- coding: utf-8 -*-
+#
+# Copyright (C) 2019 CERN.
+#
+# inspirehep is free software; you can redistribute it and/or modify it under
+# the terms of the MIT License; see LICENSE file for more details.
+
+"""Change record_citations to records_citations for consistency"""
+
+from alembic import op
+
+# revision identifiers, used by Alembic.
+revision = "c6570e49b7b2"
+down_revision = "5ce9ef759ace"
+branch_labels = ()
+depends_on = None
+
+
+def upgrade():
+    """Upgrade database."""
+    op.rename_table("record_citations", "records_citations")
+    op.execute(
+        "ALTER TABLE records_citations RENAME CONSTRAINT fk_record_citations_cited TO fk_records_citations_cited_id"
+    )
+    op.execute(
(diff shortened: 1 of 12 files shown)


inspirehep/alembic/c6570e49b7b2_change_record_citations_to_records_.py
```diff
@@ -0,0 +1,26 @@
...
+
+
+def upgrade():
+    """Upgrade database."""
+    op.rename_table("record_citations", "records_citations")
+
+
+def downgrade():
+    """Downgrade database."""
+    op.rename_table("records_citations", "record_citations")
```
michamos: also the name of the indexes should be changed to follow the sqlalchemy naming convention:
* `idx_citations_cited` -> `ix_records_citations_cited_id`
* `fk_record_citations_cited` -> `fk_records_citations_cited_id`
* `fk_record_citations_citer` -> `fk_records_citations_citer_id`

inspirehep/records/api/mixins.py
```diff
@@ -29,3 +31,18 @@ def citation_count(self):
             record.
         """
         return self._citation_query().count()
+
+    @property
+    def citations_by_year(self):
+        """Builds dict with cittation summary by year for current record
```
michamos: * typo: `citation`. 
* `Builds dict` is not nice as it doesn't build anything persistent, maybe
```suggestion
        """Return the number of citations received per year for the current record.
```

tests/integration/records/api/test_api_base.py
```diff
@@ -21,6 +22,7 @@
 from invenio_pidstore.models import PersistentIdentifier, RecordIdentifier
 from invenio_records.errors import MissingModelError
 from invenio_records.models import RecordMetadata
+from psycopg2.tz import FixedOffsetTimezone
```
michamos: unused import?

inspirehep/search/facets.py
```diff
@@ -98,3 +98,32 @@ def citation_summary():
...
+
+def citations_by_year():
+    excluded_filters = [
+        "citeable",
+        "refereed",
+        "citation_count",
+        "author_count",
+        "earliest_date",
+        "collaboration",
+    ]
```
michamos: why?

pazembrz: It is the most reasonable setup in my opinion. Do not enable things we don't need. For now this will be only used to retrieve authors by facet names. When this requirement will change this also can be easily changed.

michamos: I have the opposite opinion: don't write *code* you don't need. But it's your call.

PR #208, opened 2019-05-28
global: bump inspire-schemas and inspire-dojson

ammirate:


inspirehep/records/cli.py
```diff
@@ -1,3 +1,4 @@
+# the terms of the MIT License; see LICENSE file for more details.
```
michamos: wrong copy/paste?

ammirate: I'd say so :dart:

PR #210, opened 2019-05-28
marshmallow: Fixed Author BAI generation on literature

pazembrz:


inspirehep/records/marshmallow/literature/base.py
```diff
@@ -245,9 +245,12 @@ def get_inspire_document_type(self, record):
...
         """Prepare record for ``facet_author_name`` field."""
-        authors_with_record = InspireRecord.get_linked_records_from_dict_field(
-            record, "authors.record"
-        )
+        authors_with_record = [
+            aut
+            for aut in InspireRecord.get_linked_records_from_dict_field(
+                record, "authors.record"
+            )
+        ]
```
michamos: why not `list(...)`? maybe also make clear in the called method docstring that it returns an iterator?

PR #224, opened 2019-06-03
orcid: add MIGRATION_REDIS_URL

drjova:

### DIFF
--- a/inspirehep/config.py
+++ b/inspirehep/config.py
@@ -40,7 +40,7 @@ DEBUG = True
 
 # Migration
 # =========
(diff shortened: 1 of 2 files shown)


inspirehep/config.py
```diff
@@ -40,7 +40,7 @@
 
 # Migration
 # =========
-#: Special redis for continuous migration
+#: Special redis for continuous migration and ORCID push
```
michamos:
```suggestion
#: Special redis for continuous migration and ORCID token migration
```

drjova:
@michamos done

PR #226, opened 2019-06-04
records: correctly handle update

michamos:
Invenio uses `Record.commit` to perform the update to a record, whereas
`Record.update` is the simple `dict.update`. In `InspireRecord`, it was
decided that `update` should do the actual update in the CRUD sense.
This was not implemented consistently, however, and this commit fixes
that, by:

* overriding `commit` to do nothing;
* performing validation on `update`;
* fixing `delete` to mark the model as modified;
* fixing the ES serializer to avoid modifying the record;
* calling the `orcid_push` explicitly instead of relying on a receiver
  for the update signal, that doesn't get triggered anymore.

* INSPIR-2459


inspirehep/records/marshmallow/literature/base.py
```diff
@@ -305,11 +305,12 @@ def get_bookautocomplete(self, record):
 
     @pre_dump
     def populate_ui_display(self, data):
-        data["_ui_display"] = deepcopy(data)
+        data["_ui_display"] = copy(data)
```
michamos: This is needed because `_ui_display` is defined as a `Nested` field in the schema, so it has to be there from the start. I think it's better to use a `Method` there that calls the UI serializer on the record explicitly in order to avoid these hacks.

inspirehep/records/api/data.py
```diff
@@ -17,3 +20,15 @@ class DataRecord(InspireRecord, CitationMixin):
     es_serializer = DataMetadataRawFieldsSchemaV1
     pid_type = "dat"
     pidstore_handler = PidStoreData
+
+    @classmethod
+    def create(cls, data, **kwargs):
+        with db.session.begin_nested():
+            record = super().create(data, **kwargs)
+            record._update_refs_in_citation_table()
```
michamos: I don't think this is needed, `data` can only be cited, not cite.

inspirehep/records/marshmallow/literature/base.py
```diff
@@ -223,6 +223,9 @@ class LiteratureESEnhancementV1(
     id_field = fields.Integer(dump_only=True, dump_to="id", attribute="control_number")
     thesis_info = fields.Nested(ThesisInfoSchemaForESV1, dump_only=True)
 
+    def get_ui_display(self, record):
+        return json.dumps(LiteratureMetadataUISchemaV1().dump(dict(record)).data)
```
michamos: is the `dict` correct? don't you need data that's inside attributes of the record, like the citation count?

drjova: nice catch 👍it's a left over

inspirehep/records/api/literature.py
```diff
@@ -44,32 +50,60 @@
...
-class LiteratureRecord(InspireRecord, CitationMixin):
+class LiteratureRecord(FilesMixin, CitationMixin, InspireRecord):
     """Literature Record."""
 
     es_serializer = LiteratureESEnhancementV1
     pid_type = "lit"
     pidstore_handler = PidStoreLiterature
 
+    @property
+    def earliest_date(self):
```
pazembrz: `CitationMixin` is the only code which is using `earliest_date` method, it won't work without it. Is it wise to put this method directly in `LiteratureRecord` then and not in `CitationMixin`?

inspirehep/records/api/literature.py
```diff
@@ -44,32 +50,60 @@
...
+            return date
+        return None
+
     @classmethod
     def create(cls, data, **kwargs):
-        documents = data.pop("documents", None)
-        figures = data.pop("figures", None)
-        record = super().create(data, **kwargs)
-        if documents or figures:
-            record.set_files(documents=documents, figures=figures)
```
pazembrz: Why you are removing this like that? Is it not easier to disable flag in config?

inspirehep/records/api/literature.py
drjova: https://github.com/inspirehep/inspirehep/blob/78f55645b67e2fe39ff37611671c263cd9a79e54/inspirehep/records/marshmallow/literature/base.py#L220

drjova: 1. we're filtering out all other collections https://github.com/inspirehep/inspirehep/blob/e37dde111f8396491d966789e30ce073870841d5/inspirehep/records/api/base.py#L1037
2. ``Data`` doesn't call ``_update_refs_in_citation_table``

inspirehep/records/api/literature.py
drjova: it's completely wrong, especially on update it doesn't take the new ``data`` it takes ``self`` we have to fix this for literature

pazembrz: Thanks for explaining this IRL.

PR #227, opened 2019-06-04
setup: install `invenio-db` with versioning

drjova:
* Adds SQLAlchemy-Continuum.

* INSPIR-2453

### DIFF
--- a/poetry.lock
+++ b/poetry.lock
@@ -586,6 +586,7 @@ werkzeug = "*"
 reference = "84c2633a4af55092650cb6972b126039500b39f9"
 type = "git"
 url = "https://github.com/inspirehep/flask-kvsession.git"
+
 [[package]]
 category = "main"
 description = "Rate limiting for flask applications"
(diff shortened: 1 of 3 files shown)


poetry.lock
```diff
@@ -1628,10 +1632,7 @@ python-versions = "*"
 version = "0.7.1"
 
 [package.dependencies]
-Unidecode = ">=0.4.19"
 dictdiffer = ">=0.6.0"
-editdistance = ">=0.3.1"
-munkres = ">=1.0.7"
```
michamos: are these no longer needed?

pazembrz: As I can see they are still in lock file:
```
[[package]]
category = "main"
description = "munkres algorithm for the Assignment Problem"
name = "munkres"
optional = false
python-versions = "*"
version = "1.0.12"

[[package]]
category = "main"
description = "Fast implementation of the edit distance(Levenshtein distance)"
name = "editdistance"
optional = false
python-versions = "*"
version = "0.5.3"

[[package]]
category = "main"
description = "ASCII transliterations of Unicode text"
name = "unidecode"
optional = false
python-versions = "*"
version = "1.0.23"
```
But `editdistance` is not mentioned in any of the requirements. So it might be removed if it won't be mentioned in `poetry.lock` file

drjova: Honestly I don't understand the update strategy of poetry

PR #230, opened 2019-06-06
rt: init RT

drjova:
* INSPIR-2457


drjova:
Good point actually totally forgot about that, will update it!

PR #234, opened 2019-06-11
Author profile changes

monaawi:
* Added a prop for the ExpandableListToggle label and used show all positions for author positions in the author profile
*  INSPIR-2440
* Add prop to the RangeAggregation for rendering the hovered tooltip
* INSPIR-2438
* Added pointer cursor when hovering citation summary graph bar
* INSPIR-2439

### DIFF
--- a/ui/src/common/components/CitationSummaryGraph/CitationSummaryGraph.jsx
+++ b/ui/src/common/components/CitationSummaryGraph/CitationSummaryGraph.jsx
@@ -179,6 +179,7 @@ class CitationSummaryGraph extends Component {
                       onValueClick={this.onCiteableBarClick}
                       onValueMouseOut={this.onBarMouseOut}
                       data-test-id="citeable-bar-series"
+                      className="series-bar"
                     />
                     <LabelSeries
                       data={citeableSeriesData}
@@ -193,6 +194,7 @@ class CitationSummaryGraph extends Component {
                       onValueClick={this.onPublishedBarClick}
                       onValueMouseOut={this.onBarMouseOut}
                       data-test-id="published-bar-series"
+                      className="series-bar"
                     />
                     <LabelSeries
                       data={publishedSeriesData}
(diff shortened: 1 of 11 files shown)


ui/src/common/components/ExpandListToggle.jsx
```diff
@@ -5,13 +5,13 @@ import SecondaryButton from './SecondaryButton';
 
 class ExpandListToggle extends Component {
   render() {
-    const { size, limit, onToggle, expanded } = this.props;
+    const { size, limit, onToggle, expanded, toggleLabel } = this.props;
```
harunurhan: `expandLabel` would be better since it's only used to expand, otherwise `Hide` is rendered.

Ideally `renderLabel(expanded)` prop would be more flexible.

ui/src/common/components/RangeAggregation/RangeAggregation.jsx
```diff
@@ -249,8 +249,9 @@ class RangeAggregation extends Component {
         <div className="__RangeAggregation__">
           <div className="hovered-info-container">
             {hoveredBar && (
+              // TODO: move the logic for rendering the label for the hovered info so that the component stays generic
               <span className="hovered-info">
-                {hoveredBar.y} found for {hoveredBar.x - HALF_BAR_WIDTH}
+                {hoveredBar.y} papers found for {hoveredBar.x - HALF_BAR_WIDTH}
```
harunurhan: I think it needs to be `<Y> papers in <X>`

ui/src/common/components/CitationSummaryGraph/CitationSummaryGraph.jsx
```diff
@@ -193,6 +194,7 @@ class CitationSummaryGraph extends Component {
                       onValueClick={this.onPublishedBarClick}
                       onValueMouseOut={this.onBarMouseOut}
                       data-test-id="published-bar-series"
+                      style={{ cursor: 'pointer' }}
```
harunurhan: I guess `className` doesn't work in this case?

PR #237, opened 2019-06-12
ui: Align numbers for citation summary graph

monaawi:
* Aligned the numbers for the citation summary graph based on the width of the container INSPIR-2441
* Changed the yAxis domain for the citation by year graph to start from 0
* Changed the yAxis for the citation by year graph to not have non-integer ticks
* Changed peer-review to peer review in the citation summary table INSPIR-2421

### DIFF
--- a/ui/src/common/components/CitationSummaryGraph/CitationSummaryGraph.jsx
+++ b/ui/src/common/components/CitationSummaryGraph/CitationSummaryGraph.jsx
@@ -49,6 +49,8 @@ const typeToColors = {
   [PUBLISHED_BAR_TYPE]: { color: ORANGE, hoveredColor: HOVERED_ORANGE },
 };
 
+export const LABEL_OFFSET_RATIO_TO_GRAPH_WIDTH = 0.016;
+
 class CitationSummaryGraph extends Component {
   constructor() {
     super();
@@ -57,9 +59,21 @@ class CitationSummaryGraph extends Component {
     this.onPublishedBarHover = this.onPublishedBarHover.bind(this);
     this.onCiteableBarClick = this.onCiteableBarClick.bind(this);
     this.onPublishedBarClick = this.onPublishedBarClick.bind(this);
+    this.updateGraphWidth = this.updateGraphWidth.bind(this);
+    this.graphRef = React.createRef();
     this.state = {
       hoveredBar: null,
+      graphWidth: 0,
     };
+  }
+
+  componentDidMount() {
+    this.updateGraphWidth();
+    window.addEventListener('resize', this.updateGraphWidth);
+  }
+
+  componentWillUnmount() {
+    window.removeEventListener('resize', this.updateGraphWidth);
   }
 
   onCiteableBarClick(datapoint) {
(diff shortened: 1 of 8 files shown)


ui/src/common/components/CitationSummaryGraph/CitationSummaryGraph.jsx
```diff
@@ -49,6 +49,8 @@ const typeToColors = {
   [PUBLISHED_BAR_TYPE]: { color: ORANGE, hoveredColor: HOVERED_ORANGE },
 };
 
+const LABEL_OFFSET = 0.016;
```
harunurhan: `LABEL_OFFSET_RATIO_TO_GRAPH_WIDTH`

ui/src/common/components/CitationSummaryGraph/CitationSummaryGraph.jsx
```diff
@@ -57,11 +59,23 @@ class CitationSummaryGraph extends Component {
     this.onPublishedBarHover = this.onPublishedBarHover.bind(this);
     this.onCiteableBarClick = this.onCiteableBarClick.bind(this);
     this.onPublishedBarClick = this.onPublishedBarClick.bind(this);
+    this.updateGraphWidth = this.updateGraphWidth.bind(this);
+    this.graph = React.createRef();
```
harunurhan: `graphRef`

ui/src/common/components/__tests__/CitationSummaryGraph.test.jsx
```diff
@@ -351,4 +368,28 @@ describe('CitationSummaryGraph', () => {
     onCiteableBarUnHover();
     expect(wrapper.state('hoveredBar')).toEqual(null);
   });
+  it('sets graphWidth using ref width and recalculates graphWidth after resize event', () => {
```
harunurhan: how about testing if it is setting `xOffset`?

ui/src/common/components/__tests__/CitationSummaryGraph.test.jsx
```diff
@@ -340,7 +356,8 @@ describe('CitationSummaryGraph', () => {
         loadingCitationSummary={false}
         error={null}
         onSelectBarChange={onSelectBarChange}
-      />
+      />,
+      { disableLifecycleMethods: true }
```
harunurhan: I get that you were having some problems, but this maybe not good in the future when we put more logic in the lifecycle methods. So maybe we could just mock `getBoundingClientRect` for all tests?... like you did in the container, but in `setupTests` so that we don't need to put it or disable lifecycle methods everywhere.

ui/src/common/components/__tests__/CitationSummaryGraph.test.jsx
```diff
@@ -351,4 +368,28 @@ describe('CitationSummaryGraph', () => {
     onCiteableBarUnHover();
     expect(wrapper.state('hoveredBar')).toEqual(null);
   });
+  it('sets graphWidth using ref width and recalculates graphWidth after resize event', () => {
+    const onSelectBarChange = jest.fn();
+    Element.prototype.getBoundingClientRect = jest
```
harunurhan: whenever you're doing this in single test case, you need to restore the original `getBoundingClientRect` back at the end

ui/src/common/components/__tests__/CitationSummaryGraph.test.jsx
monaawi: The issue that I was having was not with mocking getBoundingClientRect, it was with mocking the ref (which doesn't work when using shallow), this is why I added the disable lifecycle methods. I agree that it shouldn't be done this way but mocking the getBoundingClientRect wouldn't solve this issue

PR #238, opened 2019-06-13
ui: add edit button to job detail page

ammirate:
* INSPIR-2446


ammirate:
WIP - missing `canEdit` field from literature and job serializers.

tests/integration/records/serializers/test_json.py
```diff
@@ -300,28 +298,79 @@ def test_literature_json_ui_v1_response(api_client, db, create_record):
...
+    record_titles = record["titles"]
+
+    expected_status_code = 200
+    expected_result_metadata = {
+        "control_number": record_control_number,
+        "document_type": ["article"],
+        "titles": record_titles,
+    }
+    response = api_client.get(
+        "/literature/{}".format(record_control_number), headers=headers
```
monaawi: you could use f-string instead of format here

inspirehep/records/marshmallow/literature/base.py
```diff
@@ -190,17 +192,22 @@ def get_len_or_missing(maybe_none_list):
         return len(maybe_none_list)
 
     @post_dump
-    def strip_empty(self, data):
-        return strip_empty_values(data)
+    def post_dump(self, data):
+        data = strip_empty_values(data)
+        if is_superuser_or_cataloger_logged_in():
+            data["can_edit"] = True
```
monaawi: why not set can_edit False if it is not superuser or cataloger? Also, why does it need to be in the post dump? Can't it be in the LiteratureMetadataUISchemaV1 as a fields.Method and thus will also show up in the ui_display field and you wouldn't need to duplicate logic in get_ui_display? or am i missing something?

tests/integration/records/serializers/test_json.py
```diff
@@ -300,28 +298,79 @@ def test_literature_json_ui_v1_response(api_client, db, create_record):
...
+    response_status_code = response.status_code
+    response_data = json.loads(response.data)
+
+    assert response_data["hits"]['total'] == 1
+
+    expected_data_hits = response_data["hits"]["hits"][0]['metadata']
+
+    assert expected_status_code == response_status_code
+    assert expected_title == expected_data_hits['titles'][0]['title']
+    assert 'can_edit' not in expected_data_hits
```
monaawi: for me it would make more sense to have it as false rather than not there

ui/src/jobs/containers/DetailPage.jsx
```diff
@@ -120,7 +123,10 @@ class DetailPage extends Component {
                   <MoreInfo urls={urls} />
                 </Col>
               </Row>
-              <Row type="flex" justify="end">
```
monaawi: As far as I know, the edit button should also be in the job search page

inspirehep/records/marshmallow/literature/base.py
ammirate: because it does not add any useful info :)

ui/src/jobs/containers/DetailPage.jsx
ammirate: Good catch! It was not in the US, I'll update the User Story in Jira and the PR too :+1:

inspirehep/records/marshmallow/literature/base.py
monaawi: but you would usually expect to check if a flag is true or false not if it is there or not. It doesn't change anything, I agree. But for me, it makes more sense

ui/src/common/containers/EditRecordActionContainer.jsx
```diff
@@ -1,20 +1,15 @@
...
 import ExternalLink from '../components/ExternalLink';
 
 class EditRecordActionContainer extends Component {
   render() {
-    const { userRoles, recordId } = this.props;
-    const isUserCataloger = isCataloger(userRoles);
-    const href = `/workflows/edit_article/${recordId}`;
-    if (isUserCataloger) {
+    const { href, canDisplay } = this.props;
+    if (canDisplay && href) {
```
harunurhan: It is unusual to put to logic as visibility logic that directly (It is OK in cases like, `InlineList` component doesn't render when `items` is an empty array) here in the component. You could put this in the parent {canEdit && <EditRecordAction ... />}`

harunurhan: On the other hand, since this is the `Edit` button. We can pass `pidType`, `pidValue` and push the logic of generating url for editing in this component instead of doing it at the parent level. For example have a map of `{ pidType: editUrlPrefix }` and ```href={`${pidTypeToEditUrlPrefix[pidType]}/${pidValue}`}```

ui/src/jobs/components/JobItem.jsx
```diff
@@ -69,6 +72,11 @@ class JobItem extends Component {
             Posted <DateFromNow date={created} />
           </Col>
         </Row>
+        <Row className="mt3" type="flex" justify="left">
+          <Col>
+            <EditRecordActionContainer href={`${EDIT_JOB}/${recordId}`} canDisplay={ canEdit }/>
```
harunurhan: This is where actions like this one go. (you need to use `rightActions` to put it on the right side)

https://github.com/inspirehep/inspirehep/blob/65486511885301b94cf9b40755410e23654b1fc5/ui/src/literature/components/LiteratureItem.jsx#L46

ui/src/jobs/containers/DetailPage.jsx
```diff
@@ -120,7 +123,10 @@ class DetailPage extends Component {
...
+              <Row className="mt4" type="flex" justify="space-between">
+                <Col>
+                  <EditRecordActionContainer href={`${EDIT_JOB}/${this.recordId}`} canDisplay={canEdit}/>
```
harunurhan: ditto

https://github.com/inspirehep/inspirehep/blob/65486511885301b94cf9b40755410e23654b1fc5/ui/src/literature/containers/DetailPage/DetailPage.jsx#L111

ui/src/common/components/EditRecordAction.jsx
```diff
@@ -0,0 +1,39 @@
...
+import ListItemAction from './ListItemAction';
+import IconText from './IconText';
+import EventTracker from './EventTracker';
+import ExternalLink from './ExternalLink';
+
+import { EDIT_LITERATURE, EDIT_JOB } from '../routes';
+
+
+const pidTypeEditPrefix = {
+  'lit': EDIT_LITERATURE,
```
harunurhan: we only use long versions on the UI, like `literature`.

ui/src/common/components/EditRecordAction.jsx
```diff
@@ -0,0 +1,39 @@
...
+            <IconText text="edit" type="edit" />
+          </ExternalLink>
+        </EventTracker>
+      </ListItemAction>
+    );
+  }
+}
+
+EditRecordAction.propTypes = {
+  pidType: PropTypes.string.isRequired,
```
harunurhan: `PropTypes.oneOf(['literature', 'jobs']).isRequired`

ui/src/common/__tests__/EditRecordAction.test.jsx
```diff
@@ -0,0 +1,12 @@
+import React from 'react';
```
harunurhan: This file should be inside `/components/__tests__`

ui/src/common/__tests__/EditRecordAction.test.jsx
```diff
@@ -0,0 +1,12 @@
...
+import { shallow } from 'enzyme';
+import EditRecordAction from '../components/EditRecordAction';
+
+describe('EditRecordAction', () => {
+  it('renders edit button with pidType and PidValue', () => {
+    const wrapper = shallow(
+      <EditRecordAction pidType='lit' pidValue={1}/>
```
harunurhan: could you add one for the `jobs` too?

ui/src/common/components/EditRecordAction.jsx
```diff
@@ -0,0 +1,39 @@
...
+
+import ListItemAction from './ListItemAction';
+import IconText from './IconText';
+import EventTracker from './EventTracker';
+import ExternalLink from './ExternalLink';
+
+import { EDIT_LITERATURE, EDIT_JOB } from '../routes';
+
+
+const pidTypeEditPrefix = {
```
harunurhan: `pidTypeToEditRoutePrefix`

ui/src/common/components/EditRecordAction.jsx
```diff
@@ -0,0 +1,39 @@
...
+import ListItemAction from './ListItemAction';
+import IconText from './IconText';
+import EventTracker from './EventTracker';
+import ExternalLink from './ExternalLink';
+
+import { EDIT_LITERATURE, EDIT_JOB } from '../routes';
+
+
+const pidTypeToEditRoutePrefix = {
+  'literature': EDIT_LITERATURE,
```
harunurhan: without the quotes.. sorry didn't see that before 😢

PR #239, opened 2019-06-13
ui: add job submission

harunurhan:
* Adds new and update submission pages for jobs

* Adds DateField based on antd DatePicker

* Adds RichTextEditor and RichTextField based on `react-quill`

* Refactors submission actions


ui/src/common/constants.js
```diff
@@ -14,3 +14,7 @@ export const RANK_VALUE_TO_DISPLAY = {
   UNDERGRADUATE: 'Undergrad',
   OTHER: 'other',
 };
+
+export const AUTHORS_PID_TYPE = 'authors';
+export const LITERATURE_PID_TYPE = 'literature';
+export const JOBS_PID_TYPE = 'jobs';
```
drjova: new line

ui/src/submissions/jobs/containers/JobSubmissionPage.jsx
```diff
@@ -0,0 +1,64 @@
...
+  }
+
+  async onSubmit(formData) {
+    const { dispatch } = this.props;
+    await dispatch(submit(JOBS_PID_TYPE, formData));
+  }
+
+  render() {
+    const { error, query } = this.props;
+    const { bai } = query;
```
drjova: do we need ``bai`` for forms?

ui/src/submissions/common/components/DateField.jsx
```diff
@@ -0,0 +1,40 @@
...
+  }
+
+  render() {
+    const { value, ...otherProps } = this.props;
+    return (
+      <DatePicker
+        className="w-100"
+        {...otherProps}
+        data-test-type="date-picker"
+        defaultValue={value && moment(value)}
```
drjova: don't we need a specific format?

ui/src/submissions/common/components/RichTextEditor/RichTextEditor.scss
```diff
@@ -0,0 +1,18 @@
...
+  }
+
+  .ql-container {
+    border: none;
+    min-height: 12em;
+  }
+
+  & + .ant-form-item-children-icon {
+    height: 32px !important;
+    margin-top: -22px !important;
```
drjova: 😿

ui/src/submissions/jobs/components/JobForm.jsx
```diff
@@ -0,0 +1,185 @@
...
+        <Field
+          name="regions"
+          label="* Regions"
+          mode="multiple"
+          options={regionOptions}
+          component={SelectField}
+        />
+        <Field
+          name="field_of_interest"
+          label="* Field of Research"
```
drjova: ``field of interest``

ui/src/submissions/common/components/DateField.jsx
harunurhan: `DatePicker` has default format which is `YYYY-MM-DD`

ui/src/submissions/common/components/RichTextEditor/RichTextEditor.scss
harunurhan: Necessary to fix `error` icon inside `RichTextEditor` which is not an `antd.Input`

ui/src/submissions/jobs/components/JobForm.jsx
harunurhan: Apparently this is the label we want :/

drjova: why we didn't added in the schema from the beginning? anyway :/

michamos: in the schema it's called `arxiv_categories`. @pazembrz insisted we put both `add_arxiv_category` and `add_field_of_interest` to the builder, and now we use a different display name.

harunurhan: Since we can't map form fields to schema fields for `authors` and `literature`, I decided that fields should be consistent with labels at least. So the idea was to do the same thing for job form (even though we could more or less use fields names from schema since it's simple for this case), and label for this field was `Field of Interest` in the requirements which then changed.

michamos:
sorry, but now that I'm thinking about it a bit more, I think `field of interest` is better than `field of research` (despite the latter being present on the search page), as we don't advertize exclusively research positions. So I think we rather need to change the job search.

PR #243, opened 2019-06-14
submissions: Add job create/update views

pazembrz:
*INSPIR-2430


inspirehep/submissions/views.py
```diff
@@ -156,6 +165,149 @@ def start_workflow_for_submission(self, submission_data, control_number=None):
...
+            builder.add_acquisition_source(**acquisition_source)
+        try:
+            builder.validate_record()
+        except ValidationError as e:
+            logger.error(f"Cannot process job submission: {e}")
+            raise RESTDataError(e.args[0])
+        except SchemaError as e:
+            logger.error(f"Schema is broken")
+            abort(500, str(e))
+            data = builder.record
```
monaawi: data should be outside of the except

inspirehep/config.py
```diff
@@ -719,3 +719,10 @@
 RT_PASSWORD = None
 BIBCATALOG_QUEUES = "Test"
 JLAB_ARXIV_CATEGORIES = []
+
+
+# URLS
+UI_SERVER_NAME = "localhost:8081"
+INSPIREHEP_URL = f"https://{UI_SERVER_NAME}"
+JOB_DETAILS = f"{UI_INSPIREHEP_URL}/jobs/<pid_value>"
+JOB_EDIT = f"{UI_INSPIREHEP_URL}/submissions/job/<pid_value>"
```
monaawi: I would use string format instead of replacing <pid_value>

pazembrz: I also thought about this, but we are not using format anymore, am I right?

inspirehep/submissions/views.py
pazembrz: good catch :+1:

inspirehep/submissions/loaders.py
```diff
@@ -7,10 +7,40 @@
...
+    MarshmallowErrors,
+    marshmallow_loader,
+)
+
+from inspirehep.submissions.errors import RESTDataError
 
 from .marshmallow import Author, Job, Literature
 
+
+def inspire_marshmallow_loader(schema_class):
```
drjova: for all the other cases we are using marshmallow_loader, why we need a custom one?

inspirehep/submissions/views.py
```diff
@@ -156,6 +165,151 @@ def start_workflow_for_submission(self, submission_data, control_number=None):
...
+        serialized_record = job_v1.dump(record)
+        return jsonify({"data": serialized_record})
+
+    def post(self):
+        """Adds new job record"""
+        data = job_loader_v1()
+        data["status"] = "pending"
+        data = self.prepare_data(data)
+        record = JobsRecord.create(data)
+        record.commit()
```
drjova: ditto

inspirehep/submissions/views.py
```diff
@@ -156,6 +165,151 @@ def start_workflow_for_submission(self, submission_data, control_number=None):
...
+        db.session.commit()
+        self.create_ticket(record)
+        response = jsonify({"status": "created"})
+        response.status_code = 201
+        response.headers["location"] = url_for(
```
drjova: do you mean ``Link``? http://www.rfc-editor.org/rfc/rfc5988.txt

drjova: but still I don't see any value for this right now, we are not using using it anywhere, and it will be nicer if we envelop the response in something like:

```json
{
   "links": {},
   "data": {}
}
```

inspirehep/rt/templates/rt/new_job.html
```diff
@@ -0,0 +1,18 @@
+Hello,
+
```
drjova: we need also a template for update

inspirehep/submissions/errors.py
```diff
@@ -0,0 +1,12 @@
...
+# Copyright (C) 2019 CERN.
+#
+# inspirehep is free software; you can redistribute it and/or modify it under
+# the terms of the MIT License; see LICENSE file for more details.
+from invenio_rest.errors import RESTValidationError
+
+
+class RESTDataError(RESTValidationError):
+    def get_errors(self):
+        return self.errors
```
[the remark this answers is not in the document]
pazembrz: Yes. Not all of the exceptions we have there have method to_dict().

inspirehep/config.py
```diff
@@ -719,3 +719,10 @@
 RT_PASSWORD = None
 BIBCATALOG_QUEUES = "Test"
 JLAB_ARXIV_CATEGORIES = []
+
+
+# URLS
+UI_SERVER_NAME = "localhost:8081"
+INSPIREHEP_URL = f"https://{UI_SERVER_NAME}"
```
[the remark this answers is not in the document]
pazembrz: Should I use http? This RT is not working locally, and on qa and labs we have https only

inspirehep/submissions/loaders.py
pazembrz: We are not using marshmallow loaders, please check. It's not compatible with our submission views.

inspirehep/rt/templates/rt/new_job.html
pazembrz: done

inspirehep/submissions/views.py
```diff
@@ -156,6 +165,158 @@ def start_workflow_for_submission(self, submission_data, control_number=None):
...
+            abort(404)
+        data = self.prepare_data(data, record)
+        record.update(data)
+        db.session.commit()
+        self.create_ticket(record, "rt/update_job.html")
+        return jsonify({"status": "updated", "pid_value": record["control_number"]})
+
+    def prepare_new_record(self, data):
+        acquisition_source = self.get_acquisition_source()
+        if "$schema" not in data:
```
drjova: https://github.com/inspirehep/inspirehep/blob/master/inspirehep/migrator/utils.py#L48

inspirehep/submissions/views.py
```diff
@@ -156,6 +165,158 @@ def start_workflow_for_submission(self, submission_data, control_number=None):
...
+            data["$schema"] = url_for(
+                "invenio_jsonschemas.get_schema",
+                schema_path="records/jobs.json",
+                _external=True,
+            )
+            data["status"] = "pending"
```
drjova: this should be outside of the if

inspirehep/submissions/views.py
```diff
@@ -156,6 +165,158 @@ def start_workflow_for_submission(self, submission_data, control_number=None):
...
+        except PIDDoesNotExistError:
+            abort(404)
+        data = self.prepare_data(data, record)
+        record.update(data)
+        db.session.commit()
+        self.create_ticket(record, "rt/update_job.html")
+        return jsonify({"status": "updated", "pid_value": record["control_number"]})
+
+    def prepare_new_record(self, data):
+        acquisition_source = self.get_acquisition_source()
```
drjova: why we need ``acquisition_source`` here? :/ and not in ```if "acquisition_source" not in builder.record:```

inspirehep/config.py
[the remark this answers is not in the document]
drjova: but on local will be ``localhost:8081`` why we need UI_SERVER_NAME? can we just use ``SERVER_NAME``?

inspirehep/submissions/views.py
```diff
@@ -156,6 +165,158 @@ def start_workflow_for_submission(self, submission_data, control_number=None):
...
+            "external_job_identifier",
+            "accelerator_experiments",
+            "urls",
+            "contact_details",
+            "reference_letters",
+        ]
+
+        if not is_superuser_or_cataloger_logged_in():
+            old_status = record.get("status", "pending")
+            new_status = data.get("status", old_status)
```
drjova: shall we check if the status has changed?

inspirehep/submissions/views.py
```diff
@@ -156,6 +165,158 @@ def start_workflow_for_submission(self, submission_data, control_number=None):
...
+            if new_status not in self.user_allowed_status_changes[old_status]:
+                raise RESTDataError(
+                    f"Only curator can change status from '{old_status}' to '{new_status}'."
+                )
+        record_data = dict(record)
+        for key in additional_fields:
+            if key not in data and key in record_data:
+                del record_data[key]
```
drjova: no need to do this you already have ``data`` 😕

inspirehep/submissions/views.py
```diff
@@ -156,6 +165,158 @@ def start_workflow_for_submission(self, submission_data, control_number=None):
...
+        else:
+            builder = self.prepare_new_record(data)
+
+        try:
+            builder.validate_record()
+        except ValidationError as e:
+            logger.error(f"Cannot process job submission: {e}")
+            raise RESTDataError(e.args[0])
+        except SchemaError as e:
+            logger.error(f"Schema is broken")
```
drjova: More detailed message here?

inspirehep/submissions/views.py
```diff
@@ -156,6 +165,158 @@ def start_workflow_for_submission(self, submission_data, control_number=None):
...
+            "job_url": current_app.config["JOB_DETAILS"].replace(
+                "<pid_value>", str(control_number)
+            ),
+            "job_url_edit": current_app.config["JOB_EDIT"].replace(
+                "<pid_value>", str(control_number)
+            ),
+            "hep_url": current_app.config["INSPIREHEP_URL"],
+        }
+        try:
+            _ = create_ticket_with_template(
```
drjova: ``create_ticket_with_template`` returns ticket ``id`` if it's ``-1`` then there was something wrong we should log this

inspirehep/submissions/views.py
pazembrz: You are aware that ensure_valid_schema is only checking if current schema is proper or not. It do not adds a schema if it's missing. Actually if I would use it here it would always throw an exception if i do not put at least "jobs.json" in the `$schema` key.

inspirehep/submissions/views.py
pazembrz: This is weird, I already fixed this 2 times, Looks like my IDE or something else is changing this all the time...

inspirehep/submissions/views.py
pazembrz: This is only to creating new record. Acquisition source is never there when you create a new record so adding an if is adding a dead code as this if will always be `True` After creating it, record should have acquisition source and it should never change.

drjova: but you don't need this code until your check in line ``226``,  right? I'd move it there

inspirehep/submissions/views.py
pazembrz: Discussed IRL

inspirehep/submissions/views.py
pazembrz: :+1:

inspirehep/submissions/views.py
pazembrz: :+1:

inspirehep/config.py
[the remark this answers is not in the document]
pazembrz: 1. It has to be decided by someone, as SERVER_NAME do not have protocol. Only way to properly generate this URL is to use url_for but this views are not registered in flask app as this is not a flask view. So protocol here is set to https by random, and it has no matter as locally we do not generate RT tickets at all. 

2.  Problem is that locally SERVER_NAME is unavailable outside docker, so If someone would run `poetry run inspirehep shell` locally it will blow up as 'SERVER_NAME' is not defined. 
Other solution is to set SERVER_NAME in config.  It would be probably better solution.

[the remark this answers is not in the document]
michamos: 1. Is a Flask thing. You can use `PREFERRED_URL_SCHEME` to get the proper scheme.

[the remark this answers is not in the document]
drjova: by default is ``http`` on dev and we override it on prod with ``https``

inspirehep/submissions/views.py
```diff
@@ -156,6 +165,165 @@ def start_workflow_for_submission(self, submission_data, control_number=None):
...
+
+        serialized_record = job_v1.dump(record)
+        return jsonify({"data": serialized_record})
+
+    def post(self):
+        """Adds new job record"""
+        data = job_loader_v1()
+        data = self.prepare_data(data)
+        record = JobsRecord.create(data)
+        record.commit()
```
drjova: left over?

inspirehep/submissions/views.py
```diff
@@ -156,6 +165,164 @@ def start_workflow_for_submission(self, submission_data, control_number=None):
...
+        """Adds new job record"""
+        data = job_loader_v1()
+        data = self.prepare_data(data)
+        record = JobsRecord.create(data)
+        db.session.commit()
+        self.create_ticket(record, "rt/new_job.html")
+        response = jsonify({"status": "created", "pid_value": record["control_number"]})
+        response.status_code = 201
+
+        return response
```
drjova: you can just return ``jsonify({}), 201``, IMHO we don' need the status, we can check the status_code if we need

inspirehep/submissions/views.py
```diff
@@ -156,6 +165,164 @@ def start_workflow_for_submission(self, submission_data, control_number=None):
...
+            record = JobsRecord.get_record_by_pid_value(pid.pid_value)
+            if not self.user_can_edit(record):
+                abort(401)
+        except PIDDoesNotExistError:
+            abort(404)
+        data = self.prepare_data(data, record)
+        record.update(data)
+        db.session.commit()
+        self.create_ticket(record, "rt/update_job.html")
+        return jsonify({"status": "updated", "pid_value": record["control_number"]})
```
drjova: ditto

inspirehep/submissions/views.py
pazembrz: I need pid_value for tests. I told you that. It won't hurt anyone, and it helps testing stuff.
And to be consistent with submit author endpoint where we return workflow_id, this pid_value is required.

inspirehep/submissions/views.py
pazembrz: ditto

inspirehep/submissions/views.py
drjova: I'm not talking about the ``pid_value``, but how to specify the status code :)

``jsonify({........}), 201``

pazembrz: o, ok

pazembrz: Still I would prefer to stay with setting status code explicitly. It's more readable.

drjova: this is the flask way

tests/integration/submissions/test_submissions_views.py
```diff
@@ -276,6 +276,8 @@ def test_update_author(app, api_client, create_user, requests_mock):
         == history.headers["Authorization"]
     )
     assert history.url == f"{app.config['INSPIRE_NEXT_URL']}/workflows/authors"
+    assert "datetime" in post_data["data"]["acquisition_source"]
+    del post_data["data"]["acquisition_source"]["datetime"]
```
drjova: can we just use freezetime?

drjova: why we removed ``freeze_time``?

pazembrz: I didn't removed anything, Rebase merged it in a weird way. Fixed now.

PR #246, opened 2019-06-17
minters: dedup values

drjova:

### DIFF
--- a/inspirehep/pidstore/minters/base.py
+++ b/inspirehep/pidstore/minters/base.py
@@ -5,6 +5,7 @@
 # inspirehep is free software; you can redistribute it and/or modify it under
 # the terms of the MIT License; see LICENSE file for more details.
 
+from inspire_utils.helpers import force_list
 from inspire_utils.record import get_value
 
 from ..errors import MissingSchema
@@ -28,7 +29,10 @@ class Minter:
             raise MissingSchema
 
     def get_pid_values(self):
-        return get_value(self.data, self.pid_value_path, default=None)
+        pid_values = get_value(self.data, self.pid_value_path, default=[])
+        if not isinstance(pid_values, (tuple, list)):
+            pid_values = force_list(pid_values)
+        return set(pid_values)
 
     @property
     def pid_value(self):
@@ -52,14 +56,7 @@ class Minter:
         minter.validate()
         pid_values = minter.get_pid_values()
 
-        if pid_values is None:
-            return
-
-        if isinstance(pid_values, (list, tuple)):
-            for pid_value in pid_values:
-                minter.create(pid_value)
-        else:
-            pid_value = data
+        for pid_value in pid_values:
             minter.create(pid_value)
         return minter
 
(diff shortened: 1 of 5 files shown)


inspirehep/pidstore/minters/base.py
```diff
@@ -28,7 +30,10 @@ def validate(self):
             raise MissingSchema
 
     def get_pid_values(self):
-        return get_value(self.data, self.pid_value_path, default=None)
+        pid_values = get_value(self.data, self.pid_value_path, default=[])
+        if not isinstance(pid_values, (tuple, list)):
+            pid_values = force_list(pid_values)
+        return dedupe_list(pid_values)
```
michamos: maybe simply
```suggestion
        return set(pid_values)
```
as the elements should be hashable, and the order doesn't matter?

tests/integration/pidstore/minters/test_minters_arxiv.py
```diff
@@ -59,6 +59,30 @@ def test_minter_arxiv_eprints_empty(base_app, db, create_record_factory):
     assert expected_pids_len == result_pids_len
 
 
+def test_minter_arxiv_eprints_duplicate(base_app, db, create_record_factory):
+    arxiv_value_1 = faker.arxiv()
+    arxiv_value_2 = faker.arxiv()
```
ammirate: This is not used anywhere

tests/unit/pidstore/test_minters_base.py
```diff
@@ -0,0 +1,80 @@
...
+
+def test_pid_value_with_duplicates():
+    id_ = uuid.uuid4()
+    value_jessica_jones = "Jessica Jones"
+    value_luke_cage = "Luke Cage"
+    data = {
+        "defenders": [
+            {"value": value_jessica_jones},
+            {"value": value_jessica_jones},
+            {"value": "Luke Cage"},
```
ammirate: use `value_luke_cage` instead :)

tests/integration/pidstore/minters/test_minters_arxiv.py
```diff
@@ -59,6 +59,30 @@ def test_minter_arxiv_eprints_empty(base_app, db, create_record_factory):
...
+    data = {"arxiv_eprints": [{"value": arxiv_value_1}, {"value": arxiv_value_1}]}
```
michamos: this can't actually happen as the schema has `uniqueItems: true` for all these lists, so the only way there can be two items with the same value is if there are other `properties` in the `object`s to distinguish them.

tests/integration/pidstore/minters/test_minters_doi.py
```diff
@@ -43,6 +43,29 @@ def test_minter_dois(base_app, db, create_record_factory):
         assert pid.pid_value in epxected_pids_values
 
 
+def test_minter_dois_duplicate(base_app, db, create_record_factory):
+    doi_value_1 = faker.doi()
+    data = {"dois": [{"value": doi_value_1}, {"value": doi_value_1}]}
```
michamos: same

tests/integration/pidstore/minters/test_minters_arxiv.py
drjova: added validation step as well

drjova:
@ammirate what exactly do you mean? select before?

PR #250, opened 2019-06-17
ui: pluralize conditionally

harunurhan:
* Pluralizes `citations` conditionally

* Also uses new `pluralizeUnlessSingular` function for other use cases

* INSPIR-2444

### DIFF
--- a/ui/src/common/utils.js
+++ b/ui/src/common/utils.js
@@ -143,3 +143,8 @@ export function wait(milisec) {
     setTimeout(() => resolve(), milisec);
   });
 }
+
+// appending `s` is all that's needed for the current usages
+export default function pluralizeUnlessSingle(singularWord, count) {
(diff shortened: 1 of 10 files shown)


ui/src/common/utils.js
```diff
@@ -143,3 +143,8 @@ export function wait(milisec) {
     setTimeout(() => resolve(), milisec);
   });
 }
+
+// appending `s` is all that's needed for the current usages
```
monaawi: I would still use something that would work for future cases as well so that we don't come back and change this behavior. Why not use the pluralize mentioned in the US?

harunurhan: Because it's unnecessarily complex (multiple regexp matches etc, + few 100 LOC that you need to ship for no reason) and only useful when you're dealing with dynamic values, not in our case where words are static (hardcoded).

When we have a case that doesn't work we can add map `singularToPlural` and use that instead of that library.

PR #252, opened 2019-06-18
factory: Fix url to backend so url_for will generate proper url

pazembrz:
DO NOT MERGE YET
We can live without it for now, and it's not sure what it will break.
It would be safer to deploy it next sprint

### DIFF
--- a/inspirehep/factory.py
+++ b/inspirehep/factory.py
@@ -12,7 +12,7 @@ from invenio_config import create_config_loader
 
 from . import config
 
-env_prefix = 'INVENIO'
+env_prefix = "INVENIO"
 
 
 def config_loader(app, **kwargs_config):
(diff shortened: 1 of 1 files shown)


inspirehep/factory.py
```diff
@@ -37,7 +41,7 @@ def config_loader(app, **kwargs_config):
 
 create_app = create_app_factory(
     "inspirehep",
-    config_loader=config_loader,
+    config_loader=api_config_loader,
```
michamos: this is the wrong place, as this is the app factory. i was talking about the `create_api` factory above.

pazembrz: When I added it to the first one it wasn't working.

drjova:
we don't need this for now

PR #257, opened 2019-06-19
ui: display update form data fetch errors

harunurhan:
* Displays errors happens while fetchign update data for an update
submission form, and doesn't render empty form fields incase of error.

* Displays loading feedback instead of not rendering while fetching
update form data.

### DIFF
--- a/ui/src/submissions/authors/containers/AuthorUpdateSubmissionPage.jsx
+++ b/ui/src/submissions/authors/containers/AuthorUpdateSubmissionPage.jsx
@@ -2,7 +2,7 @@ import React, { Component } from 'react';
 import PropTypes from 'prop-types';
 import { connect } from 'react-redux';
 import { Map } from 'immutable';
-import { Row, Col } from 'antd';
+import { Row, Col, Alert } from 'antd';
 
 import {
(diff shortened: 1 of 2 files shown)


ui/src/submissions/authors/containers/AuthorUpdateSubmissionPage.jsx
```diff
@@ -50,33 +52,45 @@ class AuthorUpdateSubmissionPage extends Component {
...
-                <AuthorSubmission
-                  error={error}
-                  onSubmit={this.onSubmit}
-                  initialFormData={updateFormData}
-                />
+                {updateFormDataError ? (
+                  <Alert
+                    message={updateFormDataError.get('message')}
+                    type="error"
+                    showIcon
```
drjova: if there is an error we just hide the form and we loose the updates?

harunurhan: `updateFormDataError` happens at the begining while loading the... well `updateFormData`. So it will be loading at the beginning and then error, form will be never visible, therefore nobody can't fill anything.

PR #265, opened 2019-06-20
submissions: skip ticket for job update if curator

harunurhan:
* Does not create ticket when a job updated by a curator


inspirehep/submissions/views.py
```diff
@@ -189,7 +189,10 @@ def post(self):
         data = self.prepare_data(data)
         record = JobsRecord.create(data)
         db.session.commit()
-        self.create_ticket(record, "rt/new_job.html")
+
+        if not is_superuser_or_cataloger_logged_in():
```
michamos: is this affecting new jobs or updates?

harunurhan: ah sorry wrong method, I'll move it to the `put`

PR #270, opened 2019-06-21
travis: fix deploy script for ui

drjova:


drjova:
ui tests are running in ```cd ui && yarn && yarn test;``` ``ui`` dir, travis on deploy is still in that directory we have to change to the checkout dir to run the build script

PR #274, opened 2019-06-24
submissions: make ticket creation async

harunurhan:
* INSPIR-2492

### DIFF
--- a/inspirehep/submissions/views.py
+++ b/inspirehep/submissions/views.py
@@ -27,12 +27,12 @@ from inspirehep.accounts.decorators import login_required_with_roles
 from inspirehep.records.api import AuthorsRecord, JobsRecord
 from inspirehep.records.api.literature import import_doi
 from inspirehep.records.errors import ImportConnectionError, ImportParsingError
-from inspirehep.rt.tickets import create_ticket_with_template
 from inspirehep.submissions.errors import RESTDataError
 
 from .loaders import job_v1 as job_loader_v1
 from .marshmallow import Author, Literature
 from .serializers import author_v1, job_v1  # TODO: use literature_v1 from serializers
+from .tasks import async_create_ticket_with_template
 
 blueprint = Blueprint("inspirehep_submissions", __name__, url_prefix="/submissions")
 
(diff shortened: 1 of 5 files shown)


inspirehep/submissions/views.py
```diff
@@ -165,6 +166,23 @@ def start_workflow_for_submission(self, submission_data, control_number=None):
         abort(503)
 
 
+@shared_task(ingore_result=False, max_retries=5)
+def async_create_ticket_with_template(
```
drjova: I'd move this to ``tasks.py``

inspirehep/submissions/views.py
```diff
@@ -312,22 +330,14 @@ def create_ticket(self, record, rt_template):
...
-        except AttributeError as e:
-            logger.error(f"RT is not initialized properly: {e}")
+        async_create_ticket_with_template(
+            rt_queue,
+            requestor,
+            rt_template,
+            rt_template_context,
+            f"Job {control_number} has been submitted to the Jobs database",
+            control_number,
+        )
```
drjova: what about the test for this call?

harunurhan: there are tests that asserts `create_ticket_with_template` called already, which are passing. However we need to improve them, by asserting the params too which isn't related to making the call `async`.

inspirehep/submissions/views.py
harunurhan: I don't mind but shall we keep this, until we need it somewhere else?

drjova: We always have the tasks in ``tasks.py`` for consistency let's move this one as well.

inspirehep/submissions/views.py
```diff
@@ -165,6 +166,23 @@ def start_workflow_for_submission(self, submission_data, control_number=None):
...
+@shared_task(ingore_result=False, max_retries=5)
+def async_create_ticket_with_template(
+    queue, requestor, template_path, template_context, title, recid=None
+):
+    try:
+        ticket = create_ticket_with_template(
+            queue, requestor, template_path, template_context, title, recid
+        )
+        if ticket == -1:
+            logger.error(
```
michamos: isn't it better to raise an exception here? the celery task will then show as failed, and we will get a nice traceback on sentry. For a view it makes sense not to have a 500, but letting celery tasks blow up is probably OK.

ammirate: I think if we get a job and RT fails to send a ticket, it should fail as loud as possible. Since we receive notifications of failed celery tasks even on our chat, I'd go for that too.

ammirate: @harunurhan :point_up:

PR #276, opened 2019-06-24
Add check for collections literature in the es citations query

monaawi:
* Added a check for _collections Literature in the ES citations query
* Fix the required positional argument error for the batch_index and make it use the request_timeout
* INSPIR-2477

### DIFF
--- a/inspirehep/records/indexer/tasks.py
+++ b/inspirehep/records/indexer/tasks.py
@@ -36,9 +36,9 @@ def get_record(uuid, record_version=None):
 
 
 @shared_task(ignore_result=False, bind=True)
-def batch_index(self, records_uuids, request_timeout):
+def batch_index(self, records_uuids, request_timeout=None):
     logger.info(f"Starting shared task `batch_index for {len(records_uuids)} records")
-    return InspireRecordIndexer().bulk_index(records_uuids)
+    return InspireRecordIndexer().bulk_index(records_uuids, request_timeout)
 
 
 @shared_task(ignore_result=False, bind=True)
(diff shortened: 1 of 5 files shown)


inspirehep/records/indexer/tasks.py
```diff
@@ -36,9 +36,9 @@ def get_record(uuid, record_version=None):
 
 
 @shared_task(ignore_result=False, bind=True)
-def batch_index(self, records_uuids, request_timeout):
+def batch_index(self, records_uuids, request_timeout=None):
     logger.info(f"Starting shared task `batch_index for {len(records_uuids)} records")
-    return InspireRecordIndexer().bulk_index(records_uuids)
+    return InspireRecordIndexer().bulk_index(records_uuids, request_timeout)
```
drjova: ``process_references_for_record`` which was using the ``batch_index`` doesn't have any tests that we have to update?

tests/integration/records/views/test_views_literature.py
```diff
@@ -230,6 +230,58 @@ def test_literature_citations_with_superseded_citing_records(
...
+                "record": {"$ref": "https://link-to-commentor-record/1"},
+                "relation": "commented",
+            },
+            {"record": {"$ref": "https://link-to-any-other-record/2"}},
+        ],
+        "publication_info": [{"year": 2019}],
+    }
+    create_record("lit", data=record_data)
+
+    record_with_invalid_collection_data = {
```
drjova: ``record_with_no_citable_collection``

inspirehep/records/indexer/tasks.py
monaawi: Nope, it doesn't seem to be tested :(

drjova:
@monaawi could you please bump ``inspire-query-parser``?

PR #277, opened 2019-06-24
 migrator: Prepare migrator to properly migrate records

pazembrz:
* allow migrator to run different methods depending of parameters
 * migrator now won't push to orcid
 * migrator now won't index to early
 * migrator now won't recallculate citations too early
 * continous migration will migrate records in proper way, but it will still push to orcid
 * Detach migrator from CLI command, so it will run independently

### DIFF
--- a/inspirehep/migrator/cli.py
+++ b/inspirehep/migrator/cli.py
@@ -19,6 +19,7 @@ from .tasks import (
     migrate_from_mirror,
     migrate_record_from_legacy,
     populate_mirror_from_file,
+    wait_for_all_tasks,
 )
 
 
@@ -52,21 +53,21 @@ def migrate():
     help="Only mirror the records instead of doing a full migration.",
 )
 @click.option(
-    "--wait",
-    "-w",
-    is_flag=True,
-    default=False,
-    help="Wait for migration to complete. This only has an effect if the -m flag is not set.",
-)
-@click.option(
     "-f",
     "--force",
     is_flag=True,
     default=False,
     help="Force the task to run even in debug mode.",
 )
+@click.option(
+    "-w",
+    "--wait",
+    is_flag=True,
+    default=False,
+    help="Wait for migration to complete. This only has an effect if the -m flag is not set.",
+)
 @with_appcontext
-def migrate_file(file_name, mirror_only=False, wait=False, force=False):
+def migrate_file(file_name, mirror_only=False, force=False, wait=False):
     """Migrate the records in the provided file.
 
     The file can be an (optionally-gzipped) XML file containing MARCXML, or a
@@ -77,7 +78,9 @@ def migrate_file(file_name, mirror_only=False, wait=False, force=False):
 
     populate_mirror_from_file(file_name)
     if not mirror_only:
-        migrate_from_mirror(wait_for_results=wait)
+        task = migrate_from_mirror()
+        if wait:
+            wait_for_all_tasks(task)
 
 
 @migrate.command()
@@ -96,23 +99,29 @@ def migrate_file(file_name, mirror_only=False, wait=False, force=False):
     help="Also migrate broken records, which did not migrate correctly in the previous run.",
 )
 @click.option(
-    "--wait", "-w", is_flag=True, default=False, help="Wait for migration to complete."
-)
-@click.option(
     "-f",
     "--force",
     is_flag=True,
     default=False,
     help="Force the task to run even in debug mode.",
 )
+@click.option(
+    "-w",
+    "--wait",
+    is_flag=True,
+    default=False,
+    help="Wait for all subtasks to finish. (No progress bar will be available still)",
+)
 @with_appcontext
-def mirror(also_migrate=None, wait=False, force=False):
+def mirror(also_migrate=None, force=False, wait=False):
     """Migrate records from the mirror.
 
     By default, only records that have not been migrated yet are migrated.
     """
     halt_if_debug_mode(force=force)
-    migrate_from_mirror(also_migrate=also_migrate, wait_for_results=wait)
+    task = migrate_from_mirror(also_migrate=also_migrate, disable_orcid_push=True)
+    if wait:
+        wait_for_all_tasks(task)
 
 
 @migrate.command()
--- a/inspirehep/migrator/tasks.py
+++ b/inspirehep/migrator/tasks.py
@@ -6,17 +6,15 @@
 # the terms of the MIT License; see LICENSE file for more details.
 
 """Manage migration from INSPIRE legacy instance."""
-
-
 import gzip
 import re
 import tarfile
 import zlib
 from contextlib import closing
-from functools import wraps
 
 import requests
-from celery import group, shared_task
+from celery import chord, shared_task
+from celery.result import AsyncResult
 from elasticsearch.helpers import bulk as es_bulk
 from flask import current_app
 from flask_sqlalchemy import models_committed
@@ -28,7 +26,8 @@ from jsonschema import ValidationError
 from redis import StrictRedis
 from redis_lock import Lock
 
-from inspirehep.records.api import InspireRecord
+from inspirehep.orcid.api import push_to_orcid
+from inspirehep.records.api import InspireRecord, LiteratureRecord
 from inspirehep.records.indexer.base import InspireRecordIndexer
 from inspirehep.records.receivers import index_after_commit
 
@@ -36,32 +35,11 @@ from .models import LegacyRecordsMirror
 from .utils import ensure_valid_schema
 
 LOGGER = getStackTraceLogger(__name__)
-
 CHUNK_SIZE = 100
 LARGE_CHUNK_SIZE = 2000
 
+
 split_marc = re.compile("<record.*?>.*?</record>", re.DOTALL)
-
-
-def disable_orcid_push(task_function):
-    """Temporarily disable ORCID push
-    Decorator to temporarily disable ORCID push while a given task is running,
-    and only for that task. Takes care of restoring the previous state in case
-    of errors or when the task is finished. This does not interfere with other
-    tasks, firstly because of ditto, secondly because configuration is only
-    changed within the worker's process (thus doesn't affect parallel tasks).
-    """
-
-    @wraps(task_function)
-    def _task_function(*args, **kwargs):
-        initial_state = current_app.config["FEATURE_FLAG_ENABLE_ORCID_PUSH"]
-        current_app.config["FEATURE_FLAG_ENABLE_ORCID_PUSH"] = False
-        try:
-            task_function(*args, **kwargs)
-        finally:
-            current_app.config["FEATURE_FLAG_ENABLE_ORCID_PUSH"] = initial_state
-
-    return _task_function
 
 
 def chunker(iterable, chunksize=CHUNK_SIZE):
@@ -126,19 +104,18 @@ def migrate_record_from_legacy(recid):
     db.session.commit()
 
 
-def migrate_from_mirror(also_migrate=None, wait_for_results=False):
+def migrate_from_mirror(also_migrate=None, disable_orcid_push=True):
     """Migrate legacy records from the local mirror.
     By default, only the records that have not been migrated yet are migrated.
     Args:
         also_migrate(Optional[string]): if set to ``'broken'``, also broken
             records will be migrated. If set to ``'all'``, all records will be
             migrated.
-        wait_for_results(bool): flag indicating whether the task should wait
-            for the migration to finish (if True) or fire and forget the migration
-            tasks (if False).
+        disable_orcid_push (bool): flag indicating whether the orcid_push
+            should be disabled (if True) or executed at the end of migrations (if False).
     """
-
     query = LegacyRecordsMirror.query.with_entities(LegacyRecordsMirror.recid)
+
     if also_migrate is None:
         query = query.filter(LegacyRecordsMirror.valid.is_(None))
     elif also_migrate == "broken":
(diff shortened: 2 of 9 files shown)


inspirehep/migrator/cli.py
```diff
@@ -77,7 +77,9 @@ def migrate_file(file_name, mirror_only=False, wait=False, force=False):
 
     populate_mirror_from_file(file_name)
     if not mirror_only:
-        migrate_from_mirror(wait_for_results=wait)
+        migration_tasks = migrate_from_mirror()
+        if wait:
+            migration_tasks.wait()
```
michamos: Maybe add an
```
else:
    migration_tasks.forget()
```
otherwise results will accumulate in Redis.

inspirehep/migrator/cli.py
```diff
@@ -95,24 +97,32 @@ def migrate_file(file_name, mirror_only=False, wait=False, force=False):
...
 
     By default, only records that have not been migrated yet are migrated.
     """
     halt_if_debug_mode(force=force)
-    migrate_from_mirror(also_migrate=also_migrate, wait_for_results=wait)
+    migration_tasks = migrate_from_mirror(
+        also_migrate=also_migrate, disable_orcid_push=True
+    )
+    if wait and migration_tasks:
+        migration_tasks.wait()
```
michamos: same

inspirehep/migrator/tasks.py
```diff
@@ -126,54 +106,71 @@ def migrate_record_from_legacy(recid):
...
     By default, only the records that have not been migrated yet are migrated.
     Args:
         also_migrate(Optional[string]): if set to ``'broken'``, also broken
             records will be migrated. If set to ``'all'``, all records will be
             migrated.
-        wait_for_results(bool): flag indicating whether the task should wait
-            for the migration to finish (if True) or fire and forget the migration
-            tasks (if False).
+        disable_orcid_push (bool): flag indicating whether the orcid_push
+            sould be disabled (if True) or executed at the end of migrations (if False).
```
michamos: typo `should`

inspirehep/migrator/tasks.py
```diff
@@ -126,54 +106,71 @@ def migrate_record_from_legacy(recid):
...
-            migrate_recids_from_mirror.delay(chunk)
-
-    if wait_for_results:
-        job = group(tasks)
-        result = job.apply_async()
-        result.join()
-        migrate_recids_from_mirror.ignore_result = True
-        print("All migration tasks have been completed.")
+    recids_chunked = chunker(res.recid for res in query.yield_per(CHUNK_SIZE))
+    tasks = migrate_recids_from_mirror(list(recids_chunked), 0, disable_orcid_push)
```
michamos: I would leave out the `0` here and pass `disable_orcid_push` as a `kwarg`

inspirehep/migrator/tasks.py
```diff
@@ -126,54 +106,71 @@ def migrate_record_from_legacy(recid):
...
+    print("All migration  tasks has been scheduled.")
+    return tasks
+
+
+# Set time_limit for task to 24 hours
+@shared_task(ignore_results=True, queue="migrator")
+def migrate_recids_from_mirror(records_chunks, step_no=0, disable_orcid_push=True):
```
michamos: `recids_chunks` as first argument?

inspirehep/migrator/tasks.py
```diff
@@ -126,54 +106,71 @@ def migrate_record_from_legacy(recid):
...
+    """Function to create group of tasks and adding them to chord."""
```
michamos: ```Task to migrate a chunked list of recids from the mirror.```

inspirehep/migrator/tasks.py
```diff
@@ -126,54 +106,71 @@ def migrate_record_from_legacy(recid):
...
+        recalculate_citations,
+        index_record,
+    ]
+    if not disable_orcid_push:
+        MIGRATION_STEPS.append((run_orcid_push,))
+
+    if step_no >= len(MIGRATION_STEPS):
+        return
+
+    logger.info(f"Running step {step_no+1} from {len(MIGRATION_STEPS)}")
```
michamos:
```suggestion
    logger.info(f"Running migration step {step_no+1}/{len(MIGRATION_STEPS)}")
```

inspirehep/migrator/tasks.py
```diff
@@ -126,54 +106,71 @@ def migrate_record_from_legacy(recid):
...
+    tasks = migrate_recids_from_mirror(list(recids_chunked), 0, disable_orcid_push)
+    print("All migration  tasks has been scheduled.")
+    return tasks
+
+
+# Set time_limit for task to 24 hours
+@shared_task(ignore_results=True, queue="migrator")
+def migrate_recids_from_mirror(records_chunks, step_no=0, disable_orcid_push=True):
+    """Function to create group of tasks and adding them to chord."""
+    MIGRATION_STEPS = [
```
michamos: why all caps, as this is not a global?

inspirehep/migrator/tasks.py
```diff
@@ -126,54 +106,71 @@ def migrate_record_from_legacy(recid):
...
+
+    all_recs = 0
+    for row in records_chunks:
+        all_recs += len(row)
+    if all_recs == 0:
+        logger.warning("There are no records to migrate. Terminating.")
+        return None
+    logger.info(
+        f"Chunking... Creating {len(records_chunks)} tasks with {CHUNK_SIZE} records each"
+    )
```
michamos: leftover? chunking isn't happening inside the task anymore

inspirehep/migrator/tasks.py
```diff
@@ -126,54 +106,71 @@ def migrate_record_from_legacy(recid):
...
+    logger.info(f"Processing {all_recs} records in this step")
```
michamos:
```suggestion
    logger.info(f"Processing {all_recs} records in this migration step")
```

inspirehep/migrator/tasks.py
```diff
@@ -126,54 +106,71 @@ def migrate_record_from_legacy(recid):
...
+    tasks = chord(header)(
+        migrate_recids_from_mirror.s(
+            step_no=step_no + 1, disable_orcid_push=disable_orcid_push
+        )
+    )
+    return tasks
 
 
-def migrate_from_file(source, wait_for_results=False):
+def migrate_from_file(source):
```
michamos: is this still used by anything?

inspirehep/migrator/tasks.py
```diff
@@ -183,8 +180,8 @@ def populate_mirror_from_file(source):
         print(f"Inserted {inserted_records} records into mirror")
 
 
-@shared_task(ignore_result=True)
-def continuous_migration():
+@shared_task(ignore_results=False)
+def continuous_migration(max_timeout=60 * 60, force_wait=False):
```
michamos: why do you think these two parameters are needed? why 1h? I think you can just fire and forget these tasks, no need to wait.

inspirehep/migrator/tasks.py
```diff
@@ -198,33 +195,101 @@ def continuous_migration():
             while r.llen("legacy_records"):
                 raw_record = r.lrange("legacy_records", 0, 0)
                 if raw_record:
-                    migrate_and_insert_record(zlib.decompress(raw_record[0]))
+                    insert_records_for_migration(zlib.decompress(raw_record[0]))
                     db.session.commit()
                 r.lpop("legacy_records")
         finally:
+            migrate_from_mirror.ignore_resuls = False
```
michamos: this doesn't do anything

inspirehep/migrator/tasks.py
```diff
@@ -234,14 +299,25 @@ def insert_into_mirror(raw_records):
...
     """Migrate a record and insert it if valid, or log otherwise."""
     prod_record = LegacyRecordsMirror.from_marcxml(raw_record)
     db.session.merge(prod_record)
-    return migrate_record_from_mirror(prod_record)
+    return migrate_record_from_mirror(
+        prod_record, disable_orcid_push, disable_citation_update
+    )
+
+
+def insert_records_for_migration(raw_record):
```
michamos: `insert_record_into_mirror`? isn't there already a very similar function?

inspirehep/migrator/tasks.py
```diff
@@ -198,33 +195,101 @@ def continuous_migration():
...
+    for recid in recids:
+        try:
+            with db.session.begin_nested():
+                record = migrate_record_from_mirror(
+                    LegacyRecordsMirror.query.get(recid)
                 )
+        except Exception as e:
+            logger.error(f"Cannot process record {recid}: {e}")
+            continue
+        if record and not record.get("deleted", False):
```
michamos: I think this is wrong, as a the next steps need to happen for deleted records too.

inspirehep/migrator/tasks.py
pazembrz: yes, by tests

inspirehep/migrator/tasks.py
pazembrz: If some weird thing will happen, like worker crash etc. this will be waiting forever, and will block migrations if there won't be any timeout.

michamos: why will it block? if you're not waiting, you're releasing the lock. If a worker crashes during a task, it's true that we might get inconsistent data. What about setting `acks_late=True` (see https://docs.celeryproject.org/en/latest/reference/celery.app.task.html#celery.app.task.Task.acks_late) to redo the task in case the worker crashes? all of these tasks should be safe to perform them several times.

inspirehep/migrator/tasks.py
pazembrz: :+1:

inspirehep/migrator/cli.py
pazembrz: :+1:

inspirehep/migrator/cli.py
pazembrz: :+1:

inspirehep/migrator/cli.py
michamos: it's missing the `else:`

inspirehep/migrator/cli.py
michamos: same

inspirehep/migrator/tasks.py
```diff
@@ -198,33 +194,96 @@ def continuous_migration():
...
                 if raw_record:
-                    migrate_and_insert_record(zlib.decompress(raw_record[0]))
-                    db.session.commit()
+                    insert_into_mirror([zlib.decompress(raw_record[0])])
                 r.lpop("legacy_records")
         finally:
+            migration_tasks = migrate_from_mirror(disable_orcid_push=False)
+            if migration_tasks:
+                with allow_join_result():
+                    migration_tasks.wait()
```
michamos: again: why?

inspirehep/migrator/tasks.py
```diff
@@ -28,40 +27,21 @@
...
+from inspirehep.records.api import InspireRecord, LiteratureRecord
 from inspirehep.records.indexer.base import InspireRecordIndexer
 from inspirehep.records.receivers import index_after_commit
 
 from .models import LegacyRecordsMirror
 from .utils import ensure_valid_schema
 
 LOGGER = getStackTraceLogger(__name__)
-
+logger = logging.getLogger(__name__)
```
michamos: we already have a logger right above

inspirehep/migrator/tasks.py
```diff
@@ -126,54 +106,75 @@ def migrate_record_from_legacy(recid):
...
+    header = (step.s(r) for r in recids_chunks)
+    chord_task = chord(header)(
+        migrate_recids_from_mirror.s(
+            step_no=step_no + 1, disable_orcid_push=disable_orcid_push, wait=True
+        )
+    )
+    if wait and chord_task:
+        with allow_join_result():
+            logger.info("Waiting for tasks to finish")
+            chord_task.get()
```
michamos: this is a terrible way to do it, as now you're waiting within the task. Why are you doing this?

pazembrz: I am aware of this but it's not possible to wait the way we talked yesterday. If you want we can discuss.

michamos: yes, sorry, didn't want to be harsh. Thing is there's no point of using chords if we are waiting anyway. Let's discuss it.

inspirehep/migrator/tasks.py
```diff
@@ -183,7 +182,7 @@ def populate_mirror_from_file(source):
         print(f"Inserted {inserted_records} records into mirror")
 
 
-@shared_task(ignore_result=True)
+@shared_task(ignore_results=False)
```
michamos: are we ever waiting for this to finish?

inspirehep/migrator/tasks.py
```diff
@@ -198,33 +197,93 @@ def continuous_migration():
...
+                processed_uuids.append(uuid)
+        except Exception as e:
+            LOGGER.error(f"Cannot recalculate {uuid}: {e}")
+
     db.session.commit()
+    return processed_uuids
+
 
+@shared_task(ignore_results=False, queue="migrator")
+def index_record(uuids):
```
michamos: `index_records`

inspirehep/migrator/tasks.py
```diff
@@ -278,3 +346,11 @@ def migrate_record_from_mirror(prod_record):
         prod_record.valid = True
         db.session.merge(prod_record)
         return record
+
+
+def wait_for_all_tasks(task):
+    print(f"Waiting for {task}.")
```
michamos: Why `print`? maybe better to use a logger if you think you need to log this.

pazembrz: Logger is not logging stuff to stdout when you run cli command.

inspirehep/migrator/tasks.py
```diff
@@ -126,54 +104,75 @@ def migrate_record_from_legacy(recid):
...
+        LOGGER.warning("There are no records to migrate. Terminating.")
+        return None
+    LOGGER.info(f"Processing {all_recs} records in this migration step")
+    step = migration_steps[step_no]
+    header = (step.s(r) for r in recids_chunks)
+    chord_task = chord(header)(
+        migrate_recids_from_mirror.s(
+            step_no=step_no + 1, disable_orcid_push=disable_orcid_push
+        )
+    )
```
ammirate: This triggers a task even if you are at the last step. What about
```
if step_no < len(migration_steps) -1:
    chord_task = chord(header)...
```
?

inspirehep/migrator/tasks.py
```diff
@@ -126,54 +104,75 @@ def migrate_record_from_legacy(recid):
...
-        migrate_recids_from_mirror.ignore_result = True
-        print("All migration tasks have been completed.")
+    """
+    migration_steps = [
+        create_records_from_mirror_recids,
+        recalculate_citations,
+        index_records,
+    ]
+    if not disable_orcid_push:
+        migration_steps.append((run_orcid_push,))
```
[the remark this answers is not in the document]
pazembrz: Right, it's leftover. Before migration_steps was having a tuple with task and it's parameters, but as  I removed parameters from the tasks it's not needed anymore. Good catch! :+1:

inspirehep/migrator/tasks.py
drjova: currently, we are using invenio-logging which means we have to set the level for ``stdout`` like this https://github.com/inveniosoftware/invenio-logging/blob/master/invenio_logging/config.py#L53

inspirehep/migrator/tasks.py
pazembrz: I explained this IRL

inspirehep/migrator/tasks.py
ammirate: but we will merge soon https://github.com/inspirehep/inspirehep/pull/279 and drop invenio-logging. Whu not  `click.echo`?

tests/integration/records/api/test_api_literature.py
```diff
@@ -918,3 +919,87 @@ def test_literature_without_literature_collection_cannot_cite_record_which_can_b
...
+    data2 = faker.record("lit")
+    record1 = InspireRecord.create(data1, disable_citation_update=True)
+    record1.update(data2, disable_citation_update=True)
+    assert citation_recallculate_mock.call_count == 0
+
+
+@mock.patch(
+    "inspirehep.records.api.literature.LiteratureRecord._update_refs_in_citation_table"
+)
+def test_record_update_runs_citation_recallculate_on_default(
```
ammirate: typo, double L in `recallculate`

tests/integration/records/api/test_api_literature.py
```diff
@@ -918,3 +919,87 @@ def test_literature_without_literature_collection_cannot_cite_record_which_can_b
...
+    data2 = faker.record("lit")
+    record1 = InspireRecord.create(data1)
+    record1.update(data2)
+    assert orcid_mock.call_count == 2
+
+
+@mock.patch(
+    "inspirehep.records.api.literature.LiteratureRecord._update_refs_in_citation_table"
+)
+def test_record_update_skips_citation_recallculate_when_passed_parameter_to_skip(
```
ammirate: typo

inspirehep/migrator/tasks.py
ammirate: Then it could be moved in `tests/helpers`, wdyt?

inspirehep/migrator/tasks.py
drjova: why guys we merged it like this? Using print seems small but still IMO doesn't make any sense using it here

PR #279, opened 2019-06-25
logger: initialize inspirehep-logger

drjova:
* Use ``sentry-sdk`` instead of ``raven``.

* Drop ``invenio-logging``.

* INSPIR-2429


inspirehep/logger/ext.py
```diff
@@ -0,0 +1,28 @@
...
+
+
+class InspireLogger:
+    def __init__(self, app=None):
+        if app:
+            self.init_app(app)
+
+    def init_app(self, app):
+        sentry_dsn = app.config.get("SENTRY_DSN")
+        if sentry_dsn is not None:
```
ammirate: `if sentry_dsn:`
:)

drjova:
let's wait until we have labs on k8s

ammirate:
@drjova I think we can merge this now.

inspirehep/logger/config.py
```diff
@@ -0,0 +1,35 @@
...
+dictConfig(
+    {
+        "version": 1,
+        "formatters": {
+            "default": {
+                "format": "[%(asctime)s: %(levelname)s/%(processName)s] %(name)s: %(message)s"
+            }
+        },
+        "handlers": {
+            "stdout_handler": {"class": "logging.StreamHandler", "formatter": "default"}
```
michamos: don't we want to log more on stdout (like WARNING/INFO)?

drjova: updated

drjova: is it what you meant? about the ``level``?

michamos: yeah, but don't you need to set the priority fo sentry too? or is that automatic?

drjova: it's automatic for sentry

PR #282, opened 2019-06-26
ui: bump deps and refactor containers

harunurhan:
* Bumps core dependencies (react, react-dom, react-redux, redux)
* Uses connect-react-router instead of react-router-redux

The reason that this got extremely big is
https://github.com/reduxjs/react-redux/releases/tag/v6.0.0 has a
breaking change (passing `store` directly to the connect components)
which we used to rely on all the containers tests. Therefore all of
them need to change, and change made it impossible to snapshot test
containers. That's why this also refactors containers to separate dummy
component into its own component to be able snaphsot test.

* INSPIR-2498


ui/src/literature/containers/__tests__/CiteModalActionContainer.test.jsx
```diff
@@ -1,15 +1,15 @@
...
-  dispatchToProps,
-} from '../CiteModalActionContainer';
-import * as user from '../../../actions/user';
+import { getStoreWithState, getStore } from '../../../fixtures/store';
+import CiteModalActionContainer from '../CiteModalActionContainer';
+import { setPreferredCiteFormat } from '../../../actions/user';
 import CiteModalAction from '../../components/CiteModalAction';
 
 jest.mock('../../../actions/user');
+setPreferredCiteFormat.mockReturnValue(async () => {});
```
monaawi: maybe you should mockClear before each

ui/src/middlewares/searchDispatcher.js
```diff
@@ -19,32 +19,38 @@ export default function({ getState, dispatch }) {
...
+      const nextLocation = action.payload.location; // use this instead of `getLocationFromRootState(getState())` for testability.
+      const { isFirstRendering } = action.payload;
+      const hasPathnameChanged =
+        nextLocation.pathname !== currentLocation.pathname;
       if (
-        nextLocation.search !== currentLocation.search ||
-        nextLocation.pathname !== currentLocation.pathname
+        isFirstRendering ||
+        hasPathnameChanged ||
+        nextLocation.search !== currentLocation.search
```
monaawi: I know you don't use this anywhere else but it would be nice to put it in hasSearchChanged for readability

ui/src/common/containers/__tests__/CitationSummaryGraphContainer.test.jsx
```diff
@@ -146,8 +161,15 @@ describe('CitationSummaryGraphContainer', () => {
...
+      <Provider store={store}>
+        <CitationSummaryGraphContainer />
+      </Provider>
+    );
+    const onSelectBarChange = wrapper
+      .find(CitationSummaryGraph)
+      .prop('onSelectBarChange');
+    onSelectBarChange(null);
 
     const expectedActions = [
```
monaawi: Maybe here it is better to mock specifically fetchAuthorPublications and fetchAuthorPublicationsFacets and check just that they are called. The actions created are already tested in the authors.test.js. Also, we should unclear the mocks.

harunurhan: There is no mock to clear here, as of now. Not sure what you mean by that. 

Also I think there was a reason (wasn't working) why we used directly `action` rather than mocking the action creators like we do almost everywhere else, that's why I kept it like that. But it may not be the case anymore I will to see if that works.

monaawi: I meant clear the mocks of the fetchAuthorPublications and fetchAuthorPublicationsFacets if we decide to go this way

PR #287, opened 2019-06-27
ui: fix citations by year grap x axis

harunurhan:
* Reduces number of years display on the graph.

Now we set tickValues on the X axis explicitly for all conditions because tickTotal is
unreliable.

* Also makes fetchAuthorError redirectable so that it redirects to
error pages when request fails.

* INSPIR-2443


harunurhan:
@monaawi nope, not yet at least :)

PR #308, opened 2019-07-01
ui: fix citations summary graph label for Safari

harunurhan:
* `dominant-baseline` with `text-after-edge` value works differently on
Safari. That's why we needed to use a different value.

* INSPIR-2525

### DIFF
--- a/ui/src/common/utils.js
+++ b/ui/src/common/utils.js
@@ -145,6 +145,7 @@ export function wait(milisec) {
 }
 
 // appending `s` is all that's needed for the current usages
+// TODO: do not export default (leftover)
 export default function pluralizeUnlessSingle(singularWord, count) {
   return count !== 1 ? `${singularWord}s` : singularWord;
 }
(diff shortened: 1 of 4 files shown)


ui/src/common/utils.js
```diff
@@ -168,3 +169,33 @@ export function pickEvenlyDistributedElements(array, numberOfElements) {
   }
   return someElements;
 }
+
+// adapted from https://developer.mozilla.org/en-US/docs/Web/API/Window/navigator
+// use a library (npmjs.com/package/bowser) if more info from userAgent is needed (like OS, version etc.)
+export function getBrowserName(userAgent) {
```
drjova: can we just return an enum instead of string?

harunurhan: maybe if we used `TypeScript` ;)

drjova: yes let's move them to ``const`` or something

PR #311, opened 2019-07-02
serializers: addition of ``related_records``

drjova:

### DIFF
--- a/tests/integration/records/views/test_views_literature.py
+++ b/tests/integration/records/views/test_views_literature.py
@@ -205,18 +205,12 @@ def test_literature_citations_with_superseded_citing_records(
 
     expected_status_code = 200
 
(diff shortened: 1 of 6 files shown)


tests/integration/records/views/test_views_literature.py
```diff
@@ -205,7 +205,7 @@ def test_literature_citations_with_superseded_citing_records(
 
     expected_status_code = 200
 
-    expected_count = 2
```
michamos: :cry:

drjova: the correct is 1 right?

michamos: yes, it's sad that someone wrote a test for this, but nobody noticed that the test was wrong.

PR #321, opened 2019-07-04
literature: fix earliest_date on ui serializer

harunurhan:
earliest_date was not serialized as `date` on UI response because it
was accessed via `dict.get` while being a regular property on
`LiteratureRecord` class

This also moves date formatting for UI to frontend (where it belongs)

* INSPIR-2539

### DIFF
--- a/inspirehep/records/marshmallow/literature/base.py
+++ b/inspirehep/records/marshmallow/literature/base.py
@@ -9,7 +9,6 @@ import json
 from itertools import chain
 
 from inspire_dojson.utils import strip_empty_values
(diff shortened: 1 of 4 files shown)


inspirehep/records/marshmallow/literature/base.py
```diff
@@ -169,7 +169,7 @@ class Meta:
     thesis_info = fields.Nested(ThesisInfoSchemaV1, dump_only=True)
 
     def get_formatted_date(self, data):
```
drjova: w e don't need this anymore

PR #324, opened 2019-07-05
citations: fix process records references

monaawi:
* Fix get_modified_references so it returns all references for just created records
* Make get_modified_references return list of uuids instead of list of pids
* Fix process_references which was throwing error whenever record didn't have modified references
* INSPIR-2477

### DIFF
--- a/inspirehep/records/api/literature.py
+++ b/inspirehep/records/api/literature.py
@@ -223,7 +223,7 @@ class LiteratureRecord(FilesMixin, CitationMixin, InspireRecord):
         previous version.
 
         Returns:
-            Set[Tuple[str, int]]: pids of references changed from the previous
+            List: uuids of references changed from the previous
             version.
         """
         try:
@@ -235,18 +235,21 @@ class LiteratureRecord(FilesMixin, CitationMixin, InspireRecord):
             "deleted", False
         )
 
-        if changed_deleted_status:
-            return self.get_linked_pids_from_field("references.record")
+        pids_latest = self.get_linked_pids_from_field("references.record")
 
-        ids_latest = set(self.get_linked_pids_from_field("references.record"))
+        if changed_deleted_status:
+            return list(self.get_records_ids_by_pids(pids_latest))
+
         try:
-            ids_oldest = set(
+            pids_oldest = set(
                 self._previous_version.get_linked_pids_from_field("references.record")
             )
         except AttributeError:
-            return []
+            pids_oldest = []
 
-        return set.symmetric_difference(ids_latest, ids_oldest)
+        pids_changed = set.symmetric_difference(set(pids_latest), pids_oldest)
+
+        return list(self.get_records_ids_by_pids(list(pids_changed)))
 
 
 def import_article(identifier):
--- a/tests/integration-async/records/indexer/test_literature.py
+++ b/tests/integration-async/records/indexer/test_literature.py
@@ -5,8 +5,6 @@
 # inspirehep is free software; you can redistribute it and/or modify it under
 # the terms of the MIT License; see LICENSE file for more details.
 
-import time
-
 from flask_sqlalchemy import models_committed
 from helpers.factories.models.user_access_token import AccessTokenFactory
 from helpers.providers.faker import faker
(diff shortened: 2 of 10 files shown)


inspirehep/records/api/literature.py
```diff
@@ -235,18 +235,21 @@ def get_modified_references(self):
...
                 self._previous_version.get_linked_pids_from_field("references.record")
             )
         except AttributeError:
-            return []
+            pids_oldest = []
+
+        pids_changed = set.symmetric_difference(set(pids_latest), pids_oldest)
 
-        return set.symmetric_difference(ids_latest, ids_oldest)
+        return list(self.get_records_ids_by_pids(list(pids_changed)))
```
michamos: is the inner `list` really needed?

tests/integration-async/records/conftest.py
```diff
@@ -127,7 +127,10 @@ def _check(steps={}, timeout=15):
                 _expected_result = step.get("expected_result")
                 _fun = step.get("step")
                 _expected_key = None
-                result = _fun(*_args, **_kwargs)
+                try:
+                    result = _fun(*_args, **_kwargs)
+                except:
+                    pass
```
michamos: what's going on here?

tests/integration-async/records/indexer/test_literature.py
```diff
@@ -95,7 +93,7 @@ def test_index_record_manually(
 ):
     data = faker.record("lit")
     rec = LiteratureRecord.create(data)
-    models_committed.disconnect(index_after_commit)
+
     db.session.commit()
     models_committed.connect(index_after_commit)
```
michamos: if you don't disconnect, you don't need to reconnect the signal. Is that correct though?

inspirehep/records/api/base.py
```diff
@@ -486,3 +486,7 @@ def serialize_for_es(self):
     def get_value(self, field, default=None):
         """Method which makes ``get_value`` more intuitive"""
         return get_value(self, field, default)
+
+    def get_modified_references(self):
+        """Method which is implemented only for LiteratureRecord. For other records, it will return empty list"""
```
harunurhan: Having a condition like `if literature record` on the caller seems more easier to understand to me than this.

tests/integration/records/api/test_api_literature.py
```diff
@@ -1003,3 +1003,35 @@ def test_record_update_runs_citation_recalculate_on_default(
...
+                "$ref": f"http://localhost:5000/api/literature/{cited_record_2['control_number']}"
+            }
+        }
+    ]
+    citing_record.update(citing_data)
+    db.session.commit()
+
+    assert citing_record.get_modified_references() == [cited_record_2.id]
+
+    citing_record.delete()
```
michamos: why don't you need a `db.session.commit()` here?

inspirehep/records/api/literature.py
monaawi: yes, because otherwise I would be giving ```get_records_ids_by_pids``` a set. And ```get_records_ids_by_pids``` is doing the crazy things with the batches and is accessing the elements by index so it needs a list.

tests/integration-async/records/conftest.py
monaawi: This ```retry_until_matched``` wasn't actually retrying when we were checking if a record was in ES because we were getting an error and it was exiting. This is also why we had all those time.sleep in the tests that were actually defeating the purpose of the ```retry_until_matched```

tests/integration-async/records/indexer/test_literature.py
monaawi: I removed it by mistake, good catch

inspirehep/records/api/base.py
monaawi: yes. We had a similar check before but then the method was used somewhere else as well (in the migrator) and the check was missing and it was blowing up. So we decided that it is safer to have it like this so that in case someone forgets to add the check, it wouldn't blow up anymore.

harunurhan: well it was blowing, because it was supposed to blow up. If we keep doing this base class would end up full of irrelevant crap.

The problem here is that we didn't realized that was happening, not that it was "unsafe" to put the check there.

PR #326, opened 2019-07-06
travis: build image from master or prod branch

drjova:


.travis.yml
```diff
@@ -88,11 +88,11 @@ deploy:
     skip_cleanup: true
     script: cd $TRAVIS_BUILD_DIR; ./build-docker-inspirehep-image.sh
     on:
-      branch: master
-      condition: $SUITE = integration
+      all_branches: true
+      condition: $SUITE = integration AND $TRAVIS_BRANCH =~ ^master|prod$
```
michamos: are you sure this shouldn't be
```suggestion
      condition: $SUITE = integration && $TRAVIS_BRANCH =~ ^master|prod$
```
https://docs.travis-ci.com/user/deployment says it's evaluated through `if [[ <condition> ]]` in the shell, but those use `&&`. That's how we do it in hepcrawl: https://github.com/inspirehep/hepcrawl/blob/5b1e0ab16c9bace40450544fa5d5a5fb6fc17ff9/.travis.yml#L77

.travis.yml
```diff
@@ -88,11 +88,11 @@ deploy:
...
   - provider: script
     skip_cleanup: true
     script: cd $TRAVIS_BUILD_DIR; ./build-docker-ui-image.sh
     on:
-      branch: master
-      condition: $SUITE = ui
+      all_branches: true
+      condition: $SUITE = ui AND $TRAVIS_BRANCH =~ ^master|prod$
```
michamos: same

.travis.yml
drjova: as far as I understand from the docs it should work https://docs.travis-ci.com/user/conditions-v1#boolean-operators but I just change it to ``&&``

build-docker-inspirehep-image.sh
```diff
@@ -9,11 +9,13 @@
...
 
 echo "Deploying tag for inspirehep: ${GIT_DESC}"
+echo "Deploying from branch ${TRAVIS_BRANCH}"
+
 curl -X POST "${INSPIREHEP_DEPLOY_URL}" \
     -F token=${INSPIREHEP_DEPLOY_TOKEN} \
     -F ref=master \
     -F "variables[CACHE_DATE]=$(date +%Y-%m-%d:%H:%M:%S)" \
-    -F "variables[BRANCH_NAME]=master" \
+    -F "variables[BRANCH_NAME]=${TRAVIS_BRANCH}" \
```
michamos: This is safer:
```suggestion
    -F "variables[COMMIT_ID]=${TRAVIS_COMMIT}" \
```

PR #329, opened 2019-07-09
ui: fix citations summ table query link bug

harunurhan:
* INSPIR-2447


harunurhan:
@monaawi still a temporary fix, let's wait for the requirements for "interactivitiy" of author page and resetting the table query.

PR #333, opened 2019-07-10
Regular users should see also open option when updating job

monaawi:
* Added open option to status options users see when updating their open job
* INSPIR-2557


harunurhan:
Can you also confirm the order of the status options with @StellaCh ?

StellaCh:
The order should be: pending, open, closed

PR #338, opened 2019-07-11
ui: allow past deadline if job status is closed

harunurhan:
* INSPIR-2558

### DIFF
--- a/ui/src/submissions/jobs/schemas/job.js
+++ b/ui/src/submissions/jobs/schemas/job.js
@@ -71,14 +71,18 @@ const jobSchema = object().shape({
         .label('Contact email'),
     })
   ),
-  deadline_date: date()
-    .required()
-    .test(
-      'is-valid-deadline-date',
-      'Deadline should be within next year',
-      isValidDeadlineDate
-    )
-    .label('Deadline'),
+  deadline_date: date().when('status', {
+    is: 'closed',
+    then: date(),
+    otherwise: date()
+      .required()
+      .test(
+        'is-valid-deadline-date',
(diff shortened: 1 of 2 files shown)


ui/src/submissions/jobs/schemas/job.js
```diff
@@ -71,14 +71,18 @@ const jobSchema = object().shape({
...
-    .required()
-    .test(
-      'is-valid-deadline-date',
-      'Deadline should be within next year',
-      isValidDeadlineDate
-    )
-    .label('Deadline'),
+  deadline_date: date().when('status', {
+    is: 'closed',
+    then: date(),
```
ammirate: We still need to check that the `date` is a valid one (proper format and not just a random string), no?

harunurhan: This checks that it's not a random string but a valid date string. For the format we don't check this at the moment, since `DatePicker` handles it. (we could add it if some users start doing some hacky stuff)

ui/src/submissions/jobs/schemas/job.js
```diff
@@ -71,14 +71,18 @@ const jobSchema = object().shape({
...
+    otherwise: date()
+      .required()
+      .test(
```
ammirate: This ckeck is done only on the front-end side, we should have it on the back-end too

harunurhan: As discussed, we'll do them on another task (with possibly some other validation and refactoring to move validation to loaders from views) as this is supposed to be a simple bug fix.

PR #344, opened 2019-07-15
Rename base schemas

drjova:


inspirehep/records/marshmallow/base.py
```diff
@@ -11,43 +11,62 @@
 from marshmallow.schema import Schema
 
 
-class InspireBaseSchema(RecordSchemaJSONV1):
-    # FIXME: this should be renamed to ``uuid`` depends on ``inspire-next``
-    id_ = fields.String(attribute="pid.object_uuid")
+class InspireEnvelopeSchema(RecordSchemaJSONV1):
```
harunurhan: it may be better if we drop `Inspire` prefix, what do you think?

inspirehep/records/marshmallow/base.py
```diff
@@ -11,43 +11,62 @@
...
+        self.post_dumps.append(self.include_original_fields)
 
     def include_original_fields(self, object_, original_data):
         for key, value in original_data.items():
             if key not in object_ and key not in self.exclude:
                 object_[key] = original_data[key]
         return object_
 
+    @post_dump(pass_original=True)
+    def process_post_dump_in_order(self, object_, original_data):
```
harunurhan: why not all `object_`, `data`?

inspirehep/records/marshmallow/base.py
```diff
@@ -11,58 +11,62 @@
...
+    """ElasticSearch specific extra data."""
 
-class InspireESEnhancementSchema(Schema):
     _created = fields.DateTime(dump_only=True, attribute="created")
     _updated = fields.DateTime(dump_only=True, attribute="updated")
 
 
-class InspireIncludeAllFieldsSchemaMixin:
-    """Include all fields from a record."""
+class InspireBaseSchema(Schema):
```
harunurhan: this sounds like this is a base schema for everything (incl. ES ones), but it doesn't seem so since `class InspireElasticSearchBaseSchema(Schema):`

inspirehep/records/marshmallow/base.py
pazembrz: This might be confusing, as `object_` here is always a dict and `original_data` might be an original object. Maybe they should be called `data` and `original_object`?

inspirehep/records/marshmallow/literature/base.py
```diff
@@ -314,13 +301,19 @@ def get_bookautocomplete(self, record):
...
 
-class LiteratureUISchemaV1(InspireBaseSchema):
-    metadata = fields.Nested(LiteratureMetadataUISchemaV1, dump_only=True)
-
+class LiteratureSearchUISchemaV1(InspireEnvelopeSchema):
+    """Special case for SearchUI.
 
-class LiteratureRawPublicSchemaV1(InspireBaseSchema):
-    metadata = fields.Nested(LiteratureMetadataRawPublicSchemaV1, dump_only=True)
+    We index a stringigy JSON and we have to transform it to JSON again.
```
pazembrz:
```suggestion
    We index a stringified JSON and we have to transform it to JSON again.
```

inspirehep/records/marshmallow/base.py
drjova: ElasticSearchBaseSchema contains fields that are required for ES only i.e. ``_created`` and ``_updated``

harunurhan: Then let's do call this `RecordBaseSchema`, don't have a strong opinion about this though.

drjova: renamed to just ``BaseSchema``

tests/unit/records/marshmallow/test_marshmallow_base.py
```diff
@@ -0,0 +1,277 @@
...
+
+
+def test_base_schema_with_exclude():
+    class TestSchema(BaseSchema):
+        class Meta:
+            exclude = ["description"]
+
+        def __init__(self, *args, **kwargs):
+            super().__init__(*args, **kwargs)
+            self.post_dumps.append(self.another_post_dump)
```
harunurhan: isn't this testing only `exclude`?

harunurhan: also this means we can't use `@post_dump` anymore for the subclasses? Wasn't aware of that :(

PR #347, opened 2019-07-16
serializers: rename id_ to uuid and add _collections to publich schemas

monaawi:
* Renamed id_ to uuid
* INSPIR-2572
* Added _collections to public schemas. This is needed for the record importer by url
* INSPIR-2575

### DIFF
--- a/tests/integration/records/serializers/test_json.py
+++ b/tests/integration/records/serializers/test_json.py
@@ -47,6 +47,7 @@ def test_literature_application_json_without_login(api_client, db, create_record
     record_control_number = record["control_number"]
 
     expected_status_code = 200
+    expected_uuid = str(record.id)
     expected_result = {
         "$schema": "http://localhost:5000/schemas/records/hep.json",
         "document_type": ["article"],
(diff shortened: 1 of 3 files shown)


michamos:
not making `_collections` visible is OK, as a record might also be in a hidden collection the user has no access to, so he shouldn't see those hidden collections mentioned. Why is this needed?

drjova:
```s/publich/publish```

File: tests/integration/records/serializers/test_json.py [hunk 1]
```diff
@@ -287,9 +291,11 @@ def test_literature_json_ui_v1_response(api_client, db, create_record):
     response_status_code = response.status_code
     response_data = json.loads(response.data)
     response_data_metadata = response_data["metadata"]
+    response_data_uuid = response_data["uuid"]
 
     assert expected_status_code == response_status_code
     assert expected_result_metadata == response_data_metadata
+    assert expected_uuid == response_data_uuid
```
drjova: could we also add the ``id``, ``created``, ``updated`` as well?  I think they are missing from the assertions.

monaawi:
> not sure making `_collections` visible is OK, as a record might also be in a hidden collection the user has no access to, so he shouldn't see those hidden collections mentioned. Why is this needed?

As discussed IRL, I removed this part for now and will do it separately without making the _collections public anymore

PR #349, opened 2019-07-17
docker: Fix local env, for local tests and record creation

pazembrz:


docker-compose.core.yml
```diff
@@ -26,6 +26,8 @@ services:
     extends:
       file: docker-compose.services.yml
       service: mq
+    ports:
+      - "5672:5672"
```
drjova: this will collide with ``mq-next``, why we need to specify the port? just the host should be enough, no?

pazembrz: mq-next is not accessible outside docker. We are also not using it externally. 
Without specifying ports, MQ is not accessible outside so tests are hanging, it's not possible to add records, or import them if running outside docker.

PR #353, opened 2019-07-18
migration:fix logic error in continous_migration

pazembrz:
When last bunch of records added to mirror table have errors, `migrate_from_mirror` won't run

### DIFF
--- a/inspirehep/migrator/api.py
+++ b/inspirehep/migrator/api.py
@@ -28,25 +28,27 @@ def continuous_migration():
     if redis_url is None:
         redis_url = current_app.config.get("CACHE_REDIS_URL")
 
-    r = StrictRedis.from_url(redis_url)
-    lock = Lock(r, "continuous_migration", expire=120, auto_renewal=True)
+    redis_cli = StrictRedis.from_url(redis_url)
+    lock = Lock(redis_cli, "continuous_migration", expire=120, auto_renewal=True)
 
     if lock.acquire(blocking=False):
         try:
-            migrated_records = None
-            num_of_records = r.llen("legacy_records")
+            process_migration = False
+            num_of_records = redis_cli.llen("legacy_records")
             logger.info("Starting migration of %d records.", num_of_records)
 
-            while r.llen("legacy_records"):
-                raw_record = r.lrange("legacy_records", 0, 0)
+            while redis_cli.llen("legacy_records"):
+                raw_record = redis_cli.lrange("legacy_records", 0, 0)
                 if raw_record:
                     migrated_records = insert_into_mirror(
                         [zlib.decompress(raw_record[0])]
                     )
                     logger.debug("Migrated %d records.", len(migrated_records))
-                r.lpop("legacy_records")
+                    if migrated_records:
+                        process_migration = True
+                redis_cli.lpop("legacy_records")
         finally:
-            if migrated_records:
+            if process_migration:
                 task = migrate_from_mirror(disable_orcid_push=False)
                 wait_for_all_tasks(task)
             lock.release()
--- a/tests/unit/migrator/test_migrator_api.py
+++ b/tests/unit/migrator/test_migrator_api.py
@@ -1,0 +1,52 @@
+# -*- coding: utf-8 -*-
+#
+# Copyright (C) 2019 CERN.
+#
+# inspirehep is free software; you can redistribute it and/or modify it under
+# the terms of the MIT License; see LICENSE file for more details.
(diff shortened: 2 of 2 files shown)


tests/unit/migrator/test_migrator_api.py
```diff
@@ -0,0 +1,34 @@
...
+    strict_redis_mock,
+    current_app_mock,
+):
+    strict_redis_mock().llen.side_effect = [2, 2, 1, 0]
+    strict_redis_mock().lrange.side_effect = [[True], [True], [False]]
+    lock_mock.acquire.side_effect = True
+    insert_into_mirror_mock.side_effect = [[1, 2, 3], []]
+
+    continuous_migration()
+    migrate_from_mirror_mock.assert_called_once()
```
drjova: :/ to be honest, I don't understand this test. The side effects like ``[2, 2, 1, 0]`` what does it mean? Why it's not ``[2, 1, 0]``? And ``insert_to_mirror`` why returns ```[1,2,3]```? 

Why we don't write a proper integration test and check that both records are there?

tests/unit/migrator/test_migrator_api.py
```diff
@@ -0,0 +1,34 @@
...
+
+
+@patch("inspirehep.migrator.api.current_app")
+@patch("inspirehep.migrator.api.StrictRedis.from_url")
+@patch("inspirehep.migrator.api.Lock")
+@patch("inspirehep.migrator.api.insert_into_mirror")
+@patch("inspirehep.migrator.api.zlib")
+@patch("inspirehep.migrator.api.migrate_from_mirror")
+@patch("inspirehep.migrator.api.wait_for_all_tasks")
+def test_continous_migration_when_last_record_migrated_wont_be_added_to_mirror_properly(
```
drjova:
```suggestion
def test_continuous_migration_to_add_the_last_migrated_record_in_mirror(
```
?

drjova: ``s/continous/continuous``

inspirehep/migrator/api.py
```diff
@@ -28,25 +28,27 @@ def continuous_migration():
     if redis_url is None:
         redis_url = current_app.config.get("CACHE_REDIS_URL")
 
-    r = StrictRedis.from_url(redis_url)
-    lock = Lock(r, "continuous_migration", expire=120, auto_renewal=True)
+    redis = StrictRedis.from_url(redis_url)
```
drjova: 👍 for renaming ``r``. ``redis`` can be easily confused with the ``redis`` module IMHO

tests/unit/migrator/test_migrator_api.py
pazembrz: This is simple logic of continuous_migration function. Its perfect for unit test and not for integration test. Here you do not want to check any data, or interaction between modules. You only want to check what will happen if last response from `insert_into_mirror` is empty list. Side effects are also easy to understand if you look to continuous_migration function.

tests/unit/migrator/test_migrator_api.py
pazembrz: This name means something completely different...

pazembrz: :+1:

inspirehep/migrator/api.py
pazembrz: `redis` -> `redis_cli` for redis client object

tests/unit/migrator/test_migrator_api.py
drjova: Honestly, for me at least are not easy to understand, I have to go and check each individual function and see what it expects otherwise I'm lost. 

Also I think there is an issue with:
```strict_redis_mock().llen.side_effect = [2, 2, 1, 0]```
should be 
strict_redis_mock().llen.side_effect = [2, 1, 0]

pazembrz: No it shouldn't, as there is first call for `redis_cli.llen()` which is providing value for logger.info message (which happens only once). Second and latter calls are for while loop. I can add more comments in this test.

drjova: Thanks for adding the comments. IMO adding commends to explain 5 lines of tests there is a problem. Maybe someone else can commend if she/he understands what is going on

drjova:
Supersede by #357

drjova:
I cannot reproduce the issue, hence I'll not include it in my PR

PR #354, opened 2019-07-18
records: refactor get_records_ids_by_pids

ammirate:
This commit refactors the `get_records_ids_by_pids` function, by inserting the pids in a temporary table and joining it with the PersistentIdentifier table.

Also, it removes the `get_records_metadata_by_pids` function, replacing it directly with `get_records_ids_by_pids` plus a simpler query.

* INSPIR-2540


ammirate:
@drjova @michamos I haven't had time to keep working on this.
Some tests are still failing, they pass running them one by one though.

drjova:
AFAIS there ``db.session.commit`` in the tests, which means it cleans up the temp table, no?

ammirate:
I am closing this PR since we can not figure out what is the issue with it.
The code works properly in the first place, but running it within the same session multiple times still fail, despite the several attempts. The failure happens when trying to insert data in the temporary table 
```
sqlalchemy.exc.ProgrammingError: (raised as a result of Query-invoked autoflush; consider using a session.no_autoflush block if this flush is occurring prematurely)
(psycopg2.errors.UndefinedTable) relation "temporary_pidstore" does not exist
```

Since this is not critical, we can close it and live with it for now.
Ideally, this won't be needed if we were manipulating only UUID, without the need to convert back and forth from `control_number`, but as discussed IRL, this doesn't look very feasible, because might introduce a lots of troubles in case of inconsistencies.

PR #355, opened 2019-07-18
improve jobs filters ux

harunurhan:


drjova:
@harunurhan Maybe I missed it, do we have a screenshot with Jobs filters open?

harunurhan:
@drjova nope we don't open the dropdowns for screenshots (yet)

PR #356, opened 2019-07-19
serializers: add created and updated to the search with ui content type

monaawi:
* added created and updated to the source of the ES query because they were missing from the literature search with ui content-type 
* INSPIR-2582


drjova:
@monaawi AFAIS it's not only for ui content type (for the commit message)

monaawi:
They were missing only for ui content type - for the rest I added the assertions in the tests because they were missing

PR #357, opened 2019-07-19
migrator: addition of tests

drjova:

### DIFF
--- a/tests/integration-async/migrator/test_migrator_api.py
+++ b/tests/integration-async/migrator/test_migrator_api.py
@@ -5,24 +5,14 @@
 # inspirehep is free software; you can redistribute it and/or modify it under
 # the terms of the MIT License; see LICENSE file for more details.
 
-import os
 import time
 import zlib
 
-import pkg_resources
 import pytest
-from flask_sqlalchemy import models_committed
-from helpers.providers.faker import faker
-from invenio_db import db
 from invenio_pidstore.errors import PIDDoesNotExistError
-from invenio_search import current_search_client as es
-from mock import patch
 
 from inspirehep.migrator.api import continuous_migration
-from inspirehep.migrator.models import LegacyRecordsMirror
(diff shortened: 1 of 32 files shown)


tests/integration-async/migrator/test_migrator_api.py
```diff
@@ -0,0 +1,329 @@
...
+
+    assert cache.llen("legacy_records") == 2
+
+    continuous_migration()
+
+    # I don't like timeouts, it's the only way to wait for this chain
+    time.sleep(10)
+
+    record_citer = InspireRecord.get_record_by_pid_value(citer_control_number, "lit")
+    record_citing = InspireRecord.get_record_by_pid_value(citing_control_number, "lit")
```
ammirate: if the first one is the "citer" then the second should be the "cited"

tests/integration-async/migrator/test_migrator_api.py
```diff
@@ -0,0 +1,329 @@
...
+
+    assert record_citing.citation_count == 1
+
+    record_citer_es = InspireSearch.get_record_data_from_es(record_citer)
+    result_citer_control_number = record_citer_es["control_number"]
+
+    assert citer_control_number == result_citer_control_number
+
+    record_citing_es = InspireSearch.get_record_data_from_es(record_citing)
```
ammirate: same

tests/integration-async/migrator/test_migrator_api.py
```diff
@@ -0,0 +1,329 @@
...
+    result_citing_control_number = record_citing_es["control_number"]
+
+    assert citing_control_number == result_citing_control_number
```
ammirate: [we could also check that the migrator indexed properly the record by checking the citation_count of ES, WDYT?

drjova: good point

drjova:
@ammirate updated

PR #358, opened 2019-07-23
bibtex: display authors and editors

michamos:
* Authors and editors were not displayed in the BibTeX output
  because the serializer was trying to read `authors` from the schema output,
  where it was not present. This logic has now been moved to the
  `BibTexCommonSchema`, so it can access the required data.
* INSPIR-2591

### DIFF
--- a/backend/inspirehep/records/marshmallow/literature/bibtex.py
+++ b/backend/inspirehep/records/marshmallow/literature/bibtex.py
@@ -21,6 +21,8 @@ class BibTexCommonSchema(Schema):
     address = fields.Method("get_address")
     archivePrefix = fields.Method("get_archive_prefix")
     author = fields.Method("get_author")
+    authors_with_role_author = fields.Method("get_authors_with_role_author")
+    authors_with_role_editor = fields.Method("get_authors_with_role_editor")
     booktitle = fields.Method("get_book_title")
     collaboration = fields.Method("get_collaboration")
     doc_type = fields.Raw()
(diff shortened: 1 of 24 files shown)


tests/integration/records/serializers/test_bibtex.py
```diff
@@ -29,6 +29,36 @@ def test_bibtex(api_client, db, create_record_factory):
...
+        "document_type": ["conference paper"],
+        "texkeys": ["Smith:2019abc"],
+    }
+    record = create_record_factory(
+        "lit", data=data, with_indexing=True, with_validation=True
+    )
+    record_control_number = record.json["control_number"]
+
+    expected_status_code = 200
+    expected_result = '@inproceedings{Smith:2019abc,\n    author = "Rossi, Maria",\n    editor = "Smith, John",\n    booktitle = "This is a title.",\n    title = "This is a title."\n}\n'
```
michamos: actually `booktitle` is wrong here (it's supposed to come from the linked book record), probably another regression when moving the serializer to the new instance (compare https://github.com/inspirehep/inspire-next/blob/ba9d162b67ede9006892fea4e7b106495b0e445c/inspirehep/modules/records/serializers/fields_export.py#L278-L281 and https://github.com/inspirehep/inspirehep/blob/2f411395ccf666f683da840fa35f03dee91606f3/inspirehep/records/marshmallow/literature/bibtex.py#L246)

drjova:
rebase after https://github.com/inspirehep/inspirehep/pull/389

PR #365, opened 2019-07-30
jobs: return all arxiv categories on facets

harunurhan:
* Sets size explicitly to 166 to include all arxiv_categories for Field
Of Interested facet.

* INSPIR-2585

### DIFF
--- a/inspirehep/records/config.py
+++ b/inspirehep/records/config.py
@@ -447,7 +447,7 @@ RECORDS_REST_FACETS = {
         },
         "aggs": {
             "field_of_interest": {
-                "terms": {"field": "arxiv_categories", "missing": "Other"},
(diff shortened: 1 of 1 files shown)


inspirehep/records/config.py
```diff
@@ -447,7 +447,7 @@
         },
         "aggs": {
             "field_of_interest": {
-                "terms": {"field": "arxiv_categories", "missing": "Other"},
+                "terms": {"field": "arxiv_categories", "missing": "Other", "size": 166},
```
michamos: `500`? number of arXiv categories is not fixed

harunurhan: this is number from schema, so you're saying we are likely to add more categories in the future?

michamos: yes, they regularly add new categories

PR #383, opened 2019-08-05
move backend to its own folder

harunurhan:


backend/Dockerfile
```diff
@@ -22,7 +22,7 @@ ARG POETRY_EXTRA_ARGS=''
 RUN poetry export --without-hashes --format=requirements.txt ${POETRY_EXTRA_ARGS} && \
     pip install -r requirements.txt
 
-COPY setup.py README.md ./
+COPY setup.py ./
```
drjova: I needs ``` README.md `` https://github.com/inspirehep/inspirehep/blob/master/setup.py#L14

harunurhan: why?

drjova: https://github.com/inspirehep/inspirehep/blob/master/setup.py#L14

drjova: if we remove it from Dockerfile we should remove it from the setup.py

harunurhan: your links are for the current master, it's already in the diff.

drjova: my bad, haven't seen the changes in the ``setup.py``

PR #386, opened 2019-08-06
submissions: make lit form accesiable to all users

harunurhan:
* INSPIR-2609


monaawi:
Shouldn't you also update the link for the submission from the header menu?

harunurhan:
@monaawi it's for later

PR #390, opened 2019-08-07
cli: improve orcid CLI logging

ammirate:
This commit changes also the integration tests for `import_legacy_orcid_tokens`, by making the tests style compliant to the rest of the project.

* INSPIR-2594


backend/inspirehep/orcid/cli.py
```diff
@@ -18,6 +25,49 @@ def orcid():
...
+                        },
+                    )
+            else:
+                secho("Cannot link user and token.", fg='yellow')
+
+        except SQLAlchemyError as ex:
+            secho(str(ex), fg='red')
+
+    secho('No more data to process.')
+    db.session.commit()
```
drjova: just a detail, weird spacing :)

backend/tests/integration/orcid/test_orcid_cli_import_legacy_tokens.py
```diff
@@ -191,22 +236,27 @@ def test_import_multiple_orcid_tokens_no_user_exists(
     assert_db_has_n_legacy_tokens(0, SAMPLE_USER_2)
 
     # Migrate
-    import_legacy_orcid_tokens()
+    result = app_cli_runner.invoke(import_legacy_orcid_tokens)
+
+    # both authors are not found
+    output = result.output.split('\n')
+    assert output.count('No row was found for one()') == 2
```
drjova: do we care about the output? just checking the app if the tokens are there and the queue is empty is not enough?

backend/inspirehep/orcid/cli.py
drjova: Does it make sense to remove the logs? We can use them for stats etc. WDYT

backend/tests/integration/orcid/test_orcid_cli_import_legacy_tokens.py
ammirate: I didn't want to change the test logic, that's why I kept them. Some tests are actually focusing on the code throwing exceptions.

backend/inspirehep/orcid/cli.py
ammirate: I think the CLI itself doesn't really provide any interesting statistic compared to the actual pushing tasks, those are important to observe.

ammirate: @drjova could you be more specific about the spacing? :P

drjova: for example spacing between else and except

ammirate: @drjova updated

PR #394, opened 2019-08-09
ui: add pb for inline agg filters upto lg screens

harunurhan:
*INSPIR-2606


drjova:
``s/*INSPIR-2606/* INSPIR-2606``
``s/upto/up to``
``s/pb/padding``?

PR #406, opened 2019-08-13
authors: use athor_list serializer for search ui

harunurhan:


drjova:
``s/authors: use athor_list serializer for search ui/authors: use author_list serializer for search ui``

harunurhan:
redundant cause we have already `authors: `, also over commit message first line char limit.

drjova:
``s/athor_list/author_list`` :)

PR #410, opened 2019-08-14
setup: pin celery <4.3.0

drjova:


pazembrz:
So why we are downgrading celery? It's not better to update msgpack?

ammirate:
@drjova would be nice to have here the reason of the pin, for future references :)

drjova:
we want to upgrade to celery 4.3 which requires 1.1.0 invenio-celery but we' re blocked from invenio 3.2.0 release (https://github.com/inveniosoftware/invenio/commit/7ae53922d3afad980a9ce6b22af3d7b9235aace3).

PR #411, opened 2019-08-14
setup: drop invenio dependency

drjova:

### DIFF
--- a/backend/poetry.lock
+++ b/backend/poetry.lock
@@ -397,11 +397,11 @@ description = "Python Git Library"
 name = "dulwich"
 optional = false
 python-versions = "*"
-version = "0.19.11"
+version = "0.19.12"
 
 [package.dependencies]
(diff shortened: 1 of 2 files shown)


backend/poetry.lock
```diff
@@ -1088,7 +1086,7 @@ jsonschema = ">=2.6.0,<3.0"
 pyyaml = "*"
 rfc3987 = "*"
 six = "*"
-urllib3 = "1.23"
+urllib3 = ">=1.21.1,<1.25.0 || >1.25.0,<1.25.1 || >1.25.1,<1.26"
```
pazembrz: What is that?

pazembrz: Ah, ok, this is lock. Sorry

drjova: inspire-schema requirements I guess..

PR #415, opened 2019-08-15
disambiguation: Save record phonetic blocks to db

monaawi:
* phonetic blocks were not saved to db because we didn't have flag modified on record.model.json so this commit moves adding the phonetic blocks before the super update/create
* INSPIR-2616

### DIFF
--- a/backend/inspirehep/records/api/literature.py
+++ b/backend/inspirehep/records/api/literature.py
@@ -17,6 +17,7 @@ from hepcrawl.parsers.crossref import CrossrefParser
 from idutils import is_doi, normalize_doi
 from inspire_schemas.builders import LiteratureBuilder
 from inspire_schemas.utils import is_arxiv, normalize_arxiv
+from inspire_utils.record import get_value
 from invenio_db import db
 from invenio_pidstore.models import PersistentIdentifier
 from redis import StrictRedis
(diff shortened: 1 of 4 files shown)


backend/inspirehep/records/api/literature.py
```diff
@@ -237,7 +237,8 @@ def get_modified_references(self):
 
         return list(self.get_records_ids_by_pids(list(pids_changed)))
 
-    def update_authors_signature_blocks_and_uuids(self):
+    @staticmethod
+    def update_authors_signature_blocks_and_uuids(data):
```
drjova: why ``static``?

backend/tests/integration-async/records/api/__init__.py
```diff
@@ -0,0 +1,6 @@
+# -*- coding: utf-8 -*-
+#
+# Copyright (C) 2019 CERN.
+#
+# inspirehep is free software; you can redistribute it and/or modify it under
+# the terms of the MIT License; see LICENSE file for more details.
```
drjova: do we really need these files in tests?

monaawi: I was getting ```import mismatch``` error otherwise because test files had the same name

backend/inspirehep/records/api/literature.py
monaawi: Because I don't have access to the record instance where I use it. An idea would also be to take it out of the LiteratureRecord class but I thought it is better to have it there since it is only used for Lit records

backend/inspirehep/records/api/literature.py
```diff
@@ -246,35 +247,45 @@ def update_authors_signature_blocks_and_uuids(self):
...
+        for author in data.get("authors", []):
+            author_signature_block = signature_blocks.get(author["full_name"])
+            if author_signature_block:
+                author["signature_block"] = author_signature_block
+            if "uuid" not in author:
+                author["uuid"] = str(uuid.uuid4())
+
+    def push_authors_phonetic_blocks_to_redis(self):
+        """Sends the phonetic blocks to redis to be consumed by the disambiguation service.
+        """
```
drjova: could you please put  this one line up``"""``

PR #416, opened 2019-08-15
submissions: fetch crossref during arxiv import

harunurhan:


backend/tests/integration/submissions/test_submissions_views.py
```diff
@@ -457,182 +457,6 @@ def test_new_literature_submit_with_workflows_api_error(
     assert response.status_code == 503
 
 
-@freeze_time("2019-06-17")
-def test_new_literature_submit_arxiv_merges_with_crossref(
```
drjova: why are we deleting these tests?

harunurhan: because we don't merge on `submit` now, it's already done on `import`

PR #426, opened 2019-08-28
suggesters: add weights for experiment suggester

harunurhan:
* INSPIR-2406

### DIFF
--- a/backend/tests/unit/records/marshmallow/experiments/test_es.py
+++ b/backend/tests/unit/records/marshmallow/experiments/test_es.py
@@ -45,44 +45,19 @@ def test_experiment_serializer_populates_experiment_suggest():
         "name_variants": ["NAME_V1", "NAME_V2", "NAME_V3"],
     }
 
-    expected_experiment_suggest = {
-        "input": [
-            "ACC",
-            "COLLABORATION",
-            "EXP SHORT NAME",
-            "Experiment value",
-            "INST_VALUE",
-            "LEGACY-NAME",
-            "{Long Name}",
-            "NAME_V1",
-            "NAME_V2",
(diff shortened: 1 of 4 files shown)


michamos:
I don't think this is the right approach, here you're boosting the legacy name because you assume that's what people know, but it's very often not the case.

harunurhan:
@michamos complains came when they were trying to auto-complete with legacy name AFAIK, that's why we decided to try this out temporarily while we we are decide what to do with suggesters in general (only prefix based suggestion problem)

michamos:
OK I see :+1: but this is a symptom of a much more general problem

backend/tests/unit/records/marshmallow/experiments/test_es.py
```diff
@@ -45,44 +45,19 @@ def test_experiment_serializer_populates_experiment_suggest():
...
+        {"input": "ACC", "weight": 1},
+        {"input": "COLLABORATION", "weight": 1},
+        {"input": "EXP SHORT NAME", "weight": 1},
+        {"input": "Experiment value", "weight": 1},
+        {"input": "INST_VALUE", "weight": 1},
+        {"input": "{Long Name}", "weight": 1},
+        {"input": "NAME_V1", "weight": 1},
+        {"input": "NAME_V2", "weight": 1},
+        {"input": "NAME_V3", "weight": 1},
+        {"input": "LEGACY-NAME", "weight": 5},
```
drjova: don't we need to change something on UI part?

harunurhan: as far as I can see from the docs, input value that matched always comes in `options[].text`, weight is just used for matching.

PR #430, opened 2019-09-04
ui: set doc title for each page

harunurhan:
* Also sets meta.description for Jobs Search

* INSPIR-2672


ui/src/submissions/index.jsx
```diff
@@ -53,47 +54,54 @@ const JobUpdateSubmissionSuccessPage$ = Loadable({
...
-          />
-          <Route
-            exact
-            path={SUBMISSION_SUCCESS}
-            component={SubmissionSuccessPage$}
-          />
-        </SafeSwitch>
-      </div>
+      <>
+        <DocumentHead title="Submit" />
```
monaawi: Why not do ```Literature Submit``` , ```Author Submit``` and so on for each submission page container?

harunurhan: as discussed IRL, putting titles for submission pages wasn't a requirement. I just put something for now in order not to display irrelevant titles for some pages. For example `Literature Search` for job submission page, when use moves to job submission after literature search.

PR #433, opened 2019-09-04
migrator: make pid_types blacklistable

harunurhan:
* Now a pid_type for continous migration can be blacklisted

* INSPIR-2666

### DIFF
--- a/backend/inspirehep/migrator/tasks.py
+++ b/backend/inspirehep/migrator/tasks.py
@@ -16,6 +16,7 @@ import requests
 import structlog
 from celery import chord, shared_task
 from celery.result import AsyncResult
+from flask import current_app
 from flask_sqlalchemy import models_committed
 from inspire_dojson import marcxml2record
 from invenio_db import db
@@ -23,6 +24,7 @@ from invenio_pidstore.errors import PIDValueError
 from jsonschema import ValidationError
 
 from inspirehep.orcid.api import push_to_orcid
+from inspirehep.pidstore.api import PidStoreBase
 from inspirehep.records.api import InspireRecord, LiteratureRecord
 from inspirehep.records.indexer.tasks import batch_index
 from inspirehep.records.receivers import index_after_commit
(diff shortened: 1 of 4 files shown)


backend/inspirehep/migrator/tasks.py
```diff
@@ -323,6 +325,11 @@ def migrate_and_insert_record(
     )
 
 
+def does_record_have_blacklisted_pid_type(record):
```
monaawi: maybe rename this to ```has_blacklisted_pid_type```?

harunurhan: I don't like the name neither but I was trying to make it obvious that you should be reading the function to understand what to pass.

otherwise, how do you know what to pass when you read this function? it could recid, record dict, record class, or only pid_type 😄

backend/inspirehep/migrator/tasks.py
```diff
@@ -342,6 +349,9 @@ def migrate_record_from_mirror(
         return None
 
     if "$schema" in json_record:
+        if does_record_have_blacklisted_pid_type(json_record):
+            return
+
         ensure_valid_schema(json_record)
```
michamos: I think you should move it after this line (otherwise you might not have a valid schema path yet)

harunurhan: I was thinking why validate a record if we won't use it.

backend/inspirehep/migrator/tasks.py
monaawi: how about ```is_record_pid_type_blacklisted```?

backend/inspirehep/migrator/tasks.py
michamos: it's a misnomer, `ensure_valid_schema transforms the `$schema` into a full URI instead of `jobs.json`.

harunurhan: ahh it totally is, that's sad :(

PR #437, opened 2019-09-09
subscription: MailChimp replaced for Mailtrain

pazembrz:
INSPIR-2676

### DIFF
--- a/backend/inspirehep/mailing/providers/mailtrain.py
+++ b/backend/inspirehep/mailing/providers/mailtrain.py
@@ -1,0 +1,42 @@
+# -*- coding: utf-8 -*-
+#
+# Copyright (C) 2019 CERN.
+#
+# inspirehep is free software; you can redistribute it and/or modify it under
+# the terms of the MIT License; see LICENSE file for more details.
+
+import requests
+import structlog
+from flask import current_app
+
+LOGGER = structlog.getLogger()
+
+
+def mailtrain_subscribe_user_to_list(list_id, email, first_name, last_name):
+    """Subscribe user to a list.
+
+    Args:
+        list_id (str): the list id .
+        email (str): user email.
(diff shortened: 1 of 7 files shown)


backend/inspirehep/mailing/providers/mailtrain.py
```diff
@@ -0,0 +1,42 @@
...
+def mailtrain_subscribe_user_to_list(list_id, email, first_name, last_name):
+    """Subscribe user to a list.
+
+    Args:
+        list_id (str): the list id .
+        email (str): user email.
+        first_name (str): user first name.
+        last_name (str): user last name.
+    """
+    LOGGER.info("Adding user to the mailing list.", list_id=list_id, email=email)
```
ammirate: Not sure we should explicitly log email addresses.

backend/inspirehep/mailing/providers/mailtrain.py
```diff
@@ -0,0 +1,42 @@
...
+    if response.status_code != 200:
+        LOGGER.error(
+            "Cannot add user to MailTrain! Error: %s. Status_code: %s",
+            response.reason,
+            response.status_code,
+        )
+        raise ValueError(response.reason)
+    else:
+        LOGGER.info(
+            "User successfully added to the mailing list.", list_id=list_id, email=email
```
ammirate: same here

backend/inspirehep/mailing/providers/mailtrain.py
pazembrz: I am also not sure. It was done like that before for previous user submission.

backend/inspirehep/mailing/providers/mailtrain.py
pazembrz: same here

PR #444, opened 2019-09-11
ui: add confirmation after job mails subscription

harunurhan:
* INSPIR-2679


ui/src/jobs/components/SubscribeJobsModalButton.jsx
```diff
@@ -1,10 +1,28 @@
...
 export default class SubscribeJobsModalButton extends Component {
+  static renderConfirmation() {
+    return (
+      <ModalSuccessResult>
+        <Typography.Paragraph>
+          You have successfully subscribed!
+        </Typography.Paragraph>
+        <Typography.Paragraph>
+          You will receive a weekly update on new INSPIRE job listings every
+          Monday. and it should go away automatically after 3 seconds as the
```
monaawi: I don't think this should be in the message

PR #445, opened 2019-09-12
disambiguation: add disambiguation endpoint

monaawi:
* Added disambiguation endpoint which handles the cases with 0 authors and with exactly 1 author
* INSPIR-2674


backend/inspirehep/disambiguation/utils.py
```diff
@@ -0,0 +1,91 @@
...
+    Args:
+        author (dict): The author to be updated.
+        signatures (list): List of signatures data.
+    """
+    name_variants = set()
+    for signature in signatures:
+        name_variants.add(signature["full_name"])
+    name_variants = sorted(list(name_variants), reverse=True, key=len)
+    author["name"]["name_variants"] = name_variants
+    author["name"]["value"] = name_variants[0]
```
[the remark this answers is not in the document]
monaawi: I didn't know that the actual name doesn't need to be in the name variant as well

backend/inspirehep/disambiguation/utils.py
```diff
@@ -0,0 +1,95 @@
...
+    name_variants will be a list of the names from all the signatures.s
+
+    Args:
+        author (dict): The author to be updated.
+        signatures (list): List of signatures data.
+    """
+    name_variants = {signature["full_name"] for signature in signatures}
+    name_variants = sorted(name_variants, reverse=True, key=len)
+    author["name"]["value"] = name_variants.pop(0)
+    if name_variants:
```
monaawi: here I am not sure, should we put an empty list in name_variants? or not add it at all?

michamos: empty list is not schema compliant. You can also use the author builder, it does this kind of things for you.

backend/inspirehep/disambiguation/views.py
```diff
@@ -0,0 +1,42 @@
...
+    data = request.get_json()
+    request_data = DisambiguateSignaturesSchema().load(data)
+    if request_data.errors:
+        LOGGER.info("Validation error.", user=data, errors=request_data.errors)
+        return (
+            jsonify({"message": "Validation Error.", "errors": request_data.errors}),
+            400,
+        )
+    clusters = request_data.data["clusters"]
+    disambiguation_task = disambiguate_signatures.apply_async(
```
michamos: We discussed IRL yesterday putting this behind a feature flag so we can disable it in prod for now.

backend/inspirehep/disambiguation/utils.py
```diff
@@ -0,0 +1,95 @@
...
+        signature_data (list): List containing 2 elements: the publication_id and the signature uuid.
+
+    Returns:
+        dict: The signature data from the publication with the linked author.
+    """
+    record = LiteratureRecord.get_record_by_pid_value(signature_data["publication_id"])
+    signature = next(
+        (
+            author
+            for author in record.get("authors")
```
michamos: As I said in an earlier comment, you can replace everywhere `.get(foo)` by `[foo]` as you know you're getting valid input.

backend/inspirehep/disambiguation/utils.py
```diff
@@ -0,0 +1,95 @@
...
+            if author.get("uuid") == signature_data["signature_uuid"]
+        ),
+        None,
+    )
+    if signature:
```
michamos: minor: I would unnest this a bit with
```
if not signature:
    return None
```

backend/inspirehep/disambiguation/utils.py
```diff
@@ -0,0 +1,95 @@
...
+        if signature.get("curated_relation"):
+            if not LiteratureRecord.get_linked_records_from_dict_field(
```
michamos: you can put these two `if`s into one

PR #446, opened 2019-09-12
ui: improved error handling for cite

miguelgrc:
* INSPIR-2630

### DIFF
--- a/ui/src/common/__tests__/utils.test.js
+++ b/ui/src/common/__tests__/utils.test.js
@@ -327,13 +327,19 @@ describe('utils', () => {
 
     it('convert http network error', () => {
       const error = {
-        response: null,
+        message: 'Network Error',
       };
       const result = httpErrorToActionPayload(error);
       const expected = {
         status: 'network',
       };
       expect(result).toEqual(expected);
+    });
+
+    it('throw other errors', () => {
+      const error = new Error('whatever');
+      const expected = 'whatever';
+      expect(() => httpErrorToActionPayload(error)).toThrow(expected);
     });
   });
 
--- a/ui/src/literature/__tests__/citeArticle.test.js
+++ b/ui/src/literature/__tests__/citeArticle.test.js
@@ -21,14 +21,22 @@ describe('citeArticle', () => {
     done();
   });
 
-  it('sends request with Accept header based on format and returns text', async done => {
+  it('returns a status code that is not 2xx without data', async done => {
     const citeUrl = '/literature/12345';
-    const format = 'x-another';
+    const format = 'x-test';
     mockHttp
       .onGet(citeUrl, null, { Accept: 'application/x-test' })
-      .replyOnce(200, 'Test');
-    const content = await citeArticle(format, 12345);
-    expect(content).toEqual('Not supported yet :(');
+      .replyOnce(500);
+    await expect(citeArticle(format, 12345)).rejects.toThrow(new Error('Request failed with status code 500'));
+    done();
+  });
+
+  it('returns a network error', async done => {
+    const citeUrl = '/literature/12345';
+    const format = 'x-test';
+    mockHttp
+      .onGet(citeUrl, null, { Accept: 'application/x-test' }).networkError();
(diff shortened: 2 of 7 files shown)


ui/src/common/__tests__/utils.test.js
```diff
@@ -327,14 +327,21 @@ describe('utils', () => {
...
       };
       const result = httpErrorToActionPayload(error);
       const expected = {
         status: 'network',
       };
       expect(result).toEqual(expected);
     });
+
+    it('throw other errors', () => {
+      const error = 'whatever';
```
harunurhan: new Error('whatever');

ui/src/common/__tests__/utils.test.js
```diff
@@ -327,14 +327,21 @@ describe('utils', () => {
...
+      const result = httpErrorToActionPayload;
```
harunurhan: we don't need this line, you can directly call `httpErrorToActionPayload` on the next line.

ui/src/literature/__tests__/citeArticle.test.js
```diff
@@ -21,14 +21,34 @@ describe('citeArticle', () => {
...
     mockHttp
       .onGet(citeUrl, null, { Accept: 'application/x-test' })
-      .replyOnce(200, 'Test');
-    const content = await citeArticle(format, 12345);
-    expect(content).toEqual('Not supported yet :(');
+      .replyOnce(500);
+    let error;
+    try { 
+      await citeArticle(format, 12345);
+    } catch (e) {
```
harunurhan: why not  `toThrow` instead of catching the error explicitly in the tests.

ui/src/literature/components/CiteModalAction.jsx
```diff
@@ -28,6 +28,7 @@ class CiteModalAction extends Component {
 
     this.state = {
       modalVisible: false,
+      errorMessage: '',
```
harunurhan: `null` for default.

ui/src/literature/components/__tests__/CiteModalAction.test.jsx
```diff
@@ -84,6 +84,22 @@ describe('CiteModalAction', () => {
...
+        recordId={12345}
+        initialCiteFormat={initialCiteFormat}
+        onCiteFormatChange={jest.fn()}
+      />
+    );
+    const errorMessage = 'There is an error';
+    wrapper.setState({ errorMessage });
+    wrapper.update();
+    const alert = wrapper.find(Alert);
+    expect(alert.prop('description')).toEqual(errorMessage);
```
harunurhan: just match snapshot

ui/src/literature/components/CiteModalAction.jsx
```diff
@@ -92,6 +100,17 @@ class CiteModalAction extends Component {
           onCancel={this.onModalCancel}
         >
           <div>
+            <Row>
```
harunurhan: wrapper `Row` is redundant.

ui/src/literature/__tests__/citeArticle.test.js
miguelgrc: I was having some problems with this. The error was uncatched so the execution of the test stopped before performing the check. It's a bit tricky because this https://jestjs.io/docs/en/expect#tothrowerror doesn't work like that with async/await. Anyway, I found a better and shorter solution:
`await expect(citeArticle(format, 12345)).rejects.toThrow(new Error('Request failed with status code 500'));`
which I'll push now.

ui/src/literature/components/CiteModalAction.jsx
```diff
@@ -65,16 +66,23 @@ class CiteModalAction extends Component {
     let citeContent = this.citeContentCacheByFormat[format];
     if (!citeContent) {
       const { recordId } = this.props;
-      citeContent = await citeArticle(format, recordId);
-      this.citeContentCacheByFormat[format] = citeContent;
+      try {
+        citeContent = await citeArticle(format, recordId);
+        this.citeContentCacheByFormat[format] = citeContent;
+        this.setState({ errorMessage: '' });
```
harunurhan: also `null` here.

PR #452, opened 2019-09-17
submissions: better error message when paper exists

miguelgrc:
* INSPIR-2668

### DIFF
--- a/backend/inspirehep/records/api/literature.py
+++ b/backend/inspirehep/records/api/literature.py
@@ -22,6 +22,7 @@ from inspire_utils.record import get_value
 from invenio_db import db
 from invenio_pidstore.models import PersistentIdentifier
 from redis import StrictRedis
+from sqlalchemy.orm import aliased
 
 from inspirehep.orcid.api import push_to_orcid
 from inspirehep.pidstore.api import PidStoreLiterature
@@ -301,7 +302,7 @@ class LiteratureRecord(FilesMixin, CitationMixin, InspireRecord):
 def import_article(identifier):
     """Import a new article from arXiv or Crossref based on the identifier.
 
-    This function attempts to parse  and normalize the identifier as a valid
+    This function attempts to parse and normalize the identifier as a valid
     arXiv id or DOI. If the identifier is valid and there is no record in
     Inspire matching the ID, it queries the arXiv/CrossRef APIs and parses
     the record to make it inspire compliant.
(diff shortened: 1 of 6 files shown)


backend/inspirehep/records/api/literature.py
```diff
@@ -335,9 +335,14 @@ def import_article(identifier):
     ).one_or_none()
 
     if pid:
+        pid_for_recid = PersistentIdentifier.query.filter_by(
+            pid_provider="recid", object_uuid=pid.object_uuid
+        ).one_or_none()
```
michamos: * you need to add `object_type='rec'` to be able to use an index (see https://github.com/inveniosoftware/invenio-pidstore/blob/master/invenio_pidstore/models.py#L98).
* the proper way to do this would be with a self-join on the pidstore in order to do one fewer query. You need to alias the table for that: https://docs.sqlalchemy.org/en/13/orm/tutorial.html#using-aliases (we can discuss this).

backend/inspirehep/records/errors.py
```diff
@@ -10,7 +10,9 @@
 
 
 class ExistingArticleError(Exception):
-    pass
+    def __init__(self, message, recid):
+        self.message = message
+        self.recid = recid
```
michamos: you shouldn't have to do this (see further).

backend/inspirehep/records/views.py
```diff
@@ -55,7 +55,7 @@ def import_article_view(identifier):
         return jsonify({"data": literature_v1.dump(article)})
 
     except ExistingArticleError as e:
-        return jsonify(message=str(e)), 409
+        return jsonify(message=str(e.message), recid=str(e.recid)), 409
```
michamos:
```suggestion
        message, recid = e.args
        return jsonify(message=str(message), recid=str(recid)), 409
```

ui/src/submissions/literature/components/DataImporter.jsx
```diff
@@ -25,6 +26,16 @@ class DataImporter extends Component {
     onImportClick(this.importValue);
   }
 
+  // eslint-disable-next-line class-methods-use-this
+  alertMessage(error) {
```
harunurhan: `renderAlertMessage` and read `error` from `this.props` inside this method.

ui/src/submissions/literature/components/DataImporter.jsx
```diff
@@ -25,6 +26,16 @@ class DataImporter extends Component {
...
+    return <React.Fragment>
```
harunurhan: `<> </>` is also fragment, so better to use this.

ui/src/submissions/literature/components/DataImporter.jsx
```diff
@@ -25,6 +26,16 @@ class DataImporter extends Component {
...
+      {error.get('message', DEFAULT_ERROR_MESSAGE)}
+      {error.get('recid') &&
+        <React.Fragment>{". "} <Link to={`/literature/${error.recid}`}>See the record</Link></React.Fragment>
```
harunurhan: ditto

ui/src/submissions/literature/components/__tests__/__snapshots__/DataImporter.test.jsx.snap
```diff
@@ -20,7 +20,105 @@ exports[`DataImporter renders with error with a message 1`] = `
...
+        closable={true}
+        message={
+          <React.Fragment>
+            error
+            <React.Fragment>
+              . 
+               
+              <Link
+                replace={false}
+                to="/literature/undefined"
```
harunurhan: something seems be wrong, because you're accessing `error.recid` instead of `error.get('recid')`

ui/src/submissions/literature/components/DataImporter.jsx
```diff
@@ -25,6 +26,16 @@ class DataImporter extends Component {
     onImportClick(this.importValue);
   }
 
+  renderAlertMessage() {
+    const { error } = this.props;
+    return <>
+      {error.get('message', DEFAULT_ERROR_MESSAGE)}
+      {error.get('recid') &&
```
harunurhan: can you do `const recordId = error.get('recid')` not repeat the `.get` 2 times.

backend/tests/integration/records/views/test_views_article.py
```diff
@@ -39,14 +39,17 @@ def test_import_article_view_409_because_article_already_exists(
...
     resp = api_client.get(f"/literature/import/arXiv:{arxiv_value}")
 
-    expected_msg = f"Article arXiv:{arxiv_value} already in Inspire"
+    expected_msg = f"The article arXiv:{arxiv_value} already exists in Inspire"
+    expected_recid = str(record["control_number"])
     result_msg = json.loads(resp.data)["message"]
+    result_recid = json.loads(resp.data)["recid"]
 
     assert expected_msg in result_msg
+    assert expected_recid in result_recid
```
michamos: why not `==` instead of `in` for both?

PR #453, opened 2019-09-18
jobs: add rss feed for weekly jobs

harunurhan:
* INSPIR-2689

### DIFF
--- a/backend/inspirehep/mailing/views.py
+++ b/backend/inspirehep/mailing/views.py
@@ -5,8 +5,12 @@
 # inspirehep is free software; you can redistribute it and/or modify it under
 # the terms of the MIT License; see LICENSE file for more details.
 
+from datetime import datetime
+
 import structlog
-from flask import Blueprint, jsonify, request
+from flask import Blueprint, current_app, jsonify, request
+from redis import StrictRedis
+from werkzeug.contrib.atom import AtomFeed
 
 from inspirehep.mailing.api.jobs import subscribe_to_jobs_weekly_list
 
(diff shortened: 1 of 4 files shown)


backend/tests/integration/mailing/test_views.py
```diff
@@ -76,3 +77,23 @@ def test_view_subscribe_to_list_with_missing_data(api_client, db, es_clear):
...
+
+
+def test_get_weekly_jobs_rss(app, api_client, redis, shared_datadir):
+    with patch.dict(app.config, {"WEEKLY_JOBS_EMAIL_REDIS_KEY": "weekly_jobs_email"}):
+        entry = {
+            "id": 12345,
+            "title": "New HEP positions opened last week",
+            "timestamp": 1568789887.583032,
+            "html_content": "HEP Jobs",
+        }
```
michamos: this is not what gets put into redis: https://github.com/inspirehep/inspirehep/pull/451/files#diff-b0bc1d0b37b02da3a85dae03c090d755R66

backend/inspirehep/mailing/views.py
```diff
@@ -38,3 +42,36 @@ def subscribe_jobs_weekly():
...
+        weekly_jobs_email_key, ["title", "html_content", "timestamp"]
+    )
+
+    title = raw_title.decode("UTF-8")
+    content = raw_content.decode("UTF-8")
+    timestamp = float(raw_timestamp)
+    date = datetime.fromtimestamp(timestamp)
+
+    feed = AtomFeed(
+        title="Weekly HEP jobs - INSPIRE", feed_url=request.url, url=request.url_root
```
harunurhan: feed `title` and feed entry `author` are hardcoded, still not sure if we should make them config vars, since I think they will not appear anywhere in the user email (TBC), therefore unlikely to be changed later.

michamos:
Once #452 is merged, I think you should add an integration test that checks the whole thing.

backend/tests/integration/mailing/test_views.py
harunurhan: yep that's leftover format

harunurhan:
@michamos I was thinking about this, something that should start from CLI and checks the rss.

michamos:
I think you need https://docs.python.org/3/library/datetime.html#datetime.datetime.utcfromtimestamp

PR #454, opened 2019-09-18
mail: replace sendgrid with flask_mail

ammirate:
* INSPIR-2682


backend/inspirehep/mailing/providers/flask_mail.py
```diff
@@ -0,0 +1,20 @@
...
+from flask import current_app
+from flask_mail import Message
+
+
+def send_email(sender, recipient, subject, body, cc=None):
+    msg = Message(
+        subject,
+        sender=sender,
+        recipients=[recipient],
+        html=body,
```
michamos: I think you want to add `body=strip_all_html_tags(body)` (with a suitably defined function) to have a plain text version.

ammirate: Good suggestion

backend/inspirehep/mailing/errors.py
```diff
@@ -8,7 +8,3 @@
 
 class MailingError(Exception):
```
pazembrz: It's not used anywhere now. Please remove it.

backend/pyproject.toml
```diff
@@ -69,7 +69,8 @@ colorama = "^0.4.1"
 structlog-sentry = "^1.1"
 mailchimp3 = "^3.0"
 humanize = "^0.5.1"
-sendgrid = "^6.0"
+flask-mail = "^0.9.1"
+flask_mail = "^0.9.1"
```
pazembrz: ?

backend/inspirehep/mailing/config.py
```diff
@@ -19,11 +20,12 @@
 MAILTRAIN_URL = "https://lists.labs.inspirehep.net"
 MAILTRAIN_API_TOKEN = ""
 MAILTRAIN_JOBS_WEEKLY_LIST_ID = ""
-SENDGRID_API_KEY = "CHANGE_ME"
-"""SendGrid api authentication key"""
 
 JOBS_DEADLINE_PASSED_SENDER_EMAIL = "jobs@inspirehep.net"
 """Email address for notifying expired jobs"""
 
+MAIL_SERVER = os.environ.get("MAIL_SERVER")
```
michamos: why? you can already set `INVENIO_MAIL_SERVER` in the env without this line.

backend/inspirehep/mailing/errors.py
ammirate: yeah, this code is keeping changing and I am constantly rebasing.

backend/inspirehep/mailing/config.py
ammirate: Because we wanted to be explicit: set INVENIO_MAIL_SERVER var for flask-mail which is not an invenio package, because it is automatically loaded and set in our app, ecc ecc. Too much magic IMO.
But if you guys prefer it, I am okay with it.

michamos: then I would do
```suggestion
MAIL_SERVER = "localhost"
```
if the goal is to make its existence visible

PR #456, opened 2019-09-19
mailing: change from atom to rss feed

harunurhan:
For weekly jobs feed.

### DIFF
--- a/backend/inspirehep/mailing/views.py
+++ b/backend/inspirehep/mailing/views.py
@@ -7,10 +7,11 @@
 
 from datetime import datetime
 
+import pytz
 import structlog
-from flask import Blueprint, current_app, jsonify, request
+from feedgen.feed import FeedGenerator
+from flask import Blueprint, Response, current_app, jsonify, request
 from redis import StrictRedis
-from werkzeug.contrib.atom import AtomFeed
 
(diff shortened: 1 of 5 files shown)


backend/inspirehep/mailing/views.py
```diff
@@ -56,20 +57,20 @@ def get_weekly_jobs_rss():
...
-        title=title,
-        content=content,
-        content_type="html",
-        author="inspirehep.net",
-        updated=date,
-        published=date,
-    )
-
-    return feed.get_response()
+    date = datetime.utcfromtimestamp(timestamp).replace(tzinfo=pytz.UTC)
```
pazembrz: why? You have already UTC there.

harunurhan: nop it doesn't have `tzinfo` and `feedgen` is complaining because of that.

pazembrz: O. You are right. https://docs.python.org/3.6/library/datetime.html#datetime.datetime.utcfromtimestamp This is weird. 
Cannot you use datetime.fromtimestamp(timestamp, timezone.utc) then?

harunurhan: yep, that's same result and seems cleaner.

PR #459, opened 2019-09-19
BibTeX/LaTeX typo

miguelgrc:
* INSPIR-2636


miguelgrc:
> Can you make sure you're running `prettier` on pre-commit, change of format seems weird and not sure if yours or the existing one is correct.

Yeah, I installed it before commiting this, that's why it changed the format of those lines that belong to a previous PR.

PR #467, opened 2019-09-24
records: minor fixes in ``hard_delete``

drjova:

### DIFF
--- a/backend/inspirehep/records/api/base.py
+++ b/backend/inspirehep/records/api/base.py
@@ -11,6 +11,7 @@ import uuid
 from datetime import datetime
 
 import structlog
+from elasticsearch import NotFoundError
 from flask_celeryext.app import current_celery_app
 from inspire_dojson.utils import strip_empty_values
 from inspire_schemas.api import validate as schema_validate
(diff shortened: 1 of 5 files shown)


backend/inspirehep/records/api/base.py
```diff
@@ -335,9 +335,11 @@ def hard_delete(self):
                 PersistentIdentifier.object_uuid == self.id
             ).all()
             for pid in pids:
-                RecordIdentifier.query.filter_by(recid=pid.pid_value).delete()
+                if pid.pid_provider == "recid":
+                    RecordIdentifier.query.filter_by(recid=pid.pid_value).delete()
```
michamos: don't you also need to delete the other PIDs?

pazembrz: `pid` is not only recid but also doi or arXiv. If you will try to remove them from RecordIdentifier it will blow up as there are only recids.
pid from PersistentIdentifier table should be deleted after this check. (and originally was).

drjova: Actually it doesn't blow up :)

michamos: OK, I understand. Was confused between `RecordIdentifier` and `PersistentIdentifier`. I agree it should be there.

PR #469, opened 2019-09-25
serializers: Get referenced records from ES

pazembrz:
* INSPIR-2713

### DIFF
--- a/backend/inspirehep/records/api/literature.py
+++ b/backend/inspirehep/records/api/literature.py
@@ -4,9 +4,8 @@
 #
 # inspirehep is free software; you can redistribute it and/or modify it under
 # the terms of the MIT License; see LICENSE file for more details.
-
-
 import datetime
+import json
 import uuid
 
 import requests
@@ -39,6 +38,7 @@ from inspirehep.records.utils import (
     get_authors_phonetic_blocks,
     get_literature_earliest_date,
 )
+from inspirehep.search.api import LiteratureSearch
 
 from .base import InspireRecord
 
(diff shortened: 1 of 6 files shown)


backend/inspirehep/records/api/literature.py
```diff
@@ -89,6 +89,45 @@ def create(
...
+            >>> data = {
+                'references': [
+                    {
+                        'record': {
+                            '$ref': 'http://localhost/literature/1'
+                        }
+                    }
+                ]
+            }
+            >>>  records = LiteratureRecord.get_linked_records_from_field(
```
ammirate: I love examples like this, it is missing the function output though ;)

backend/inspirehep/records/api/literature.py
```diff
@@ -89,6 +89,45 @@ def create(
...
+        """
+        pids = cls._get_linked_pids_from_field(data, path)
+        if pids:
+            results = LiteratureSearch.get_records_by_pids(pids)
+            for result in results:
+                try:
+                    rec_data = json.loads(result._ui_display)
+                except AttributeError:
+                    LOGGER.exception(
+                        "Record do not have _ui_display field!", record=result.meta.id
```
ammirate: ops, `doesn't`.
Also, I think we have a lot of logs with recid reather than records uuid

pazembrz: UUID here is the only thing which will be there always. recid might be missing from there if the record won't be indexed properly (and it's not indexed properly if _ui_display field is missing).

backend/inspirehep/records/api/literature.py
```diff
@@ -89,6 +89,53 @@ def create(
...
+                    )
+                    continue
+                result._d_.update(rec_data)
+                yield InspireRecord.get_class_for_record(result._d_)(result._d_)
```
michamos: * Isn't this always a `LiteratureRecord` given you're doing a `LiteratureSearch`?
* I know the semantics of this in invenio are not very clear (you can have a `Record` that's not backed by a model instance) but why should this `LiteratureRecord`?

backend/inspirehep/records/api/literature.py
```diff
@@ -89,6 +89,53 @@ def create(
...
+        """
+        pids = cls._get_linked_pids_from_field(data, path)
+        if pids:
+            results = LiteratureSearch.get_records_by_pids(pids)
+            for result in results:
+                try:
+                    rec_data = json.loads(result._ui_display)
+                except AttributeError:
+                    LOGGER.exception(
+                        "Record do not have _ui_display field!", record=result.meta.id
```
michamos:
```suggestion
                        "Record does not have _ui_display field!", record=result.meta.id
```

backend/inspirehep/records/api/literature.py
```diff
@@ -89,6 +89,53 @@ def create(
...
+                    )
+                    continue
+                result._d_.update(rec_data)
```
michamos: what is this?

backend/inspirehep/records/api/literature.py
```diff
@@ -89,6 +89,53 @@ def create(
...
+                    'collaborations': [(...)],
+                }
+            ]
+        """
+        pids = cls._get_linked_pids_from_field(data, path)
+        if pids:
+            results = LiteratureSearch.get_records_by_pids(pids)
+            for result in results:
+                try:
+                    rec_data = json.loads(result._ui_display)
```
michamos: how do you know the data you're interested in is actually present in `_ui_display`? and what's the point of using that if you're deserializing it anyway? wouldn't it make more sense to have a configurable `_source`?

backend/inspirehep/records/api/literature.py
pazembrz: `_d_` is a dictionary of the record from ES, I'm updating it with data loaded from `_ui_display` json.

backend/inspirehep/records/api/literature.py
pazembrz: We don't have everything in `_ui_display`, we are missing  `arxiv_eprints`, `collaborations`, `dois` and `misc` field from there, this is why we are merging deserialized `_ui_display` with this missing fields in line 135. I preffer to use `_ui_data` as this is only place where we have limited number of authors there. Big number of authors was the main reason why loading of references was so slow before.

backend/inspirehep/search/api.py
```diff
@@ -145,6 +145,22 @@ def citations(record, page=1, size=10):
         )
         return citations_search.execute().hits
 
+    @staticmethod
+    def get_records_by_pids(pids, source=None, size=1000):
```
michamos: what if there are more than 1000 refs?

pazembrz: :+1:

PR #474, opened 2019-09-26
global: bump hepcrawl

miguelgrc:
* INSPIR-2668


drjova:
I think is something is wrong, the hepcrawl is not updated :/

michamos:
Commit message is wrong too

PR #475, opened 2019-09-27
serializers: Get referenced records from ES

pazembrz:
Fix bug when reference from rec is not in db but have collaborations attached.
* INSPIR-2713

### DIFF
--- a/backend/inspirehep/records/marshmallow/literature/common/reference_item.py
+++ b/backend/inspirehep/records/marshmallow/literature/common/reference_item.py
@@ -10,7 +10,6 @@ from inspire_utils.helpers import force_list
 from inspire_utils.record import get_value
 from marshmallow import Schema, fields, missing, post_dump, pre_dump
 
-from inspirehep.records.api import InspireRecord
 from inspirehep.records.marshmallow.fields import (
     ListWithLimit,
     NestedWithoutEmptyObjects,
@@ -82,9 +81,9 @@ class ReferenceItemSchemaV1(Schema):
 
     def get_resolved_references_by_control_number(self, data):
         data = force_list(data)
-        resolved_records = InspireRecord.get_linked_records_from_dict_field(
-            data, "record"
-        )
+        from inspirehep.records.api.literature import LiteratureRecord
+
+        resolved_records = LiteratureRecord.get_es_linked_references(data)
 
         return {record["control_number"]: record.dumps() for record in resolved_records}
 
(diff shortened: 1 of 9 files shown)


backend/inspirehep/records/marshmallow/literature/common/reference_item.py
```diff
@@ -82,9 +81,9 @@ def get_reference_record_id(self, data):
 
     def get_resolved_references_by_control_number(self, data):
         data = force_list(data)
-        resolved_records = InspireRecord.get_linked_records_from_dict_field(
-            data, "record"
-        )
+        from inspirehep.records.api.literature import LiteratureRecord
```
drjova: Why we need to import it in the function?

backend/tests/integration/search/test_api.py
```diff
@@ -0,0 +1,28 @@
...
+    result = LiteratureSearch().get_records_by_pids([("lit", record1_control_number)])
+    assert len(result) == 1
+    assert (
+        json.loads(result[0]._ui_display)["control_number"] == record1["control_number"]
+    )
+
+    result = LiteratureSearch().get_records_by_pids(
+        [("lit", record1_control_number), ("lit", record2_control_number)]
+    )
+    assert len(result) == 2
```
drjova: Isn't it better to check the actual control number in the date instead of len results?

backend/tests/integration/search/test_api.py
```diff
@@ -0,0 +1,28 @@
...
+# Copyright (C) 2019 CERN.
+#
+# inspirehep is free software; you can redistribute it and/or modify it under
+# the terms of the MIT License; see LICENSE file for more details.
+import json
+
+from inspirehep.search.api import LiteratureSearch
+
+
+def test_literature_get_records_by_pids_returns_correct_record(
```
drjova:
```suggestion
def test_literature_get_records_by_pids_returns_correct_record(base_app,
```

backend/inspirehep/records/marshmallow/literature/common/reference_item.py
pazembrz: Circular Imports :/

backend/tests/integration/search/test_api.py
pazembrz: :+1:

backend/inspirehep/records/marshmallow/literature/common/reference_item.py
drjova: ```from inspirehep.records.api import LiteratureRecord``` also?

pazembrz: yes

PR #477, opened 2019-09-30
ui: fix commas in linebreaks for authors

miguelgrc:
* INSPIR-2590


ui/src/common/components/InlineList/InlineList.scss
```diff
@@ -16,10 +16,10 @@
   }
 
   $separator-padding: 0.25em;
-  @mixin separte-items-with($seperator, $padding-left: 0) {
-    // For every not empty `li`, add separator before all the not empty `li` that come after it
-    & > li:not(:empty) ~ li:not(:empty):before {
-      content: $seperator;
+  @mixin separte-items-with($separator, $padding-left: 0) {
+    // Add separator after all the non-empty `li` except the last one
```
monaawi: I don't think this works. What if we have a non-empty li then an empty li? It shouldn't add a separator between them since the next one is empty but it will add it in your case

ui/src/jobs/components/JobItem.jsx
```diff
@@ -48,8 +48,8 @@ class JobItem extends Component {
               wrapperClassName="di"
               separateItemsClassName="separate-items-with-middledot"
             >
-              <InstitutionsList institutions={institutions} />
-              <RegionsList regions={regions} />
+              {institutions && <InstitutionsList institutions={institutions} />}
+              {regions && <RegionsList regions={regions} />}
```
monaawi: you can remove the check for regions here as regions is required by the schema

ui/src/jobs/containers/DetailPageContainer.jsx
```diff
@@ -99,8 +99,10 @@ class DetailPage extends Component {
               <Row className="mt1">
                 <Col>
                   <InlineUL separateItemsClassName="separate-items-with-middledot">
-                    <InstitutionsList institutions={institutions} />
-                    <RegionsList regions={regions} />
+                    {institutions && (
+                      <InstitutionsList institutions={institutions} />
+                    )}
+                    {regions && <RegionsList regions={regions} />}
```
monaawi: here as well

miguelgrc:
I know, but the alternative is, for every child in InlineUL, to traverse its own children to see if they're empty, which I think is even worse.

PR #480, opened 2019-10-01
mailing: use right args for send_email

ammirate:

### DIFF
--- a/backend/tests/integration/mailing/test_api_jobs.py
+++ b/backend/tests/integration/mailing/test_api_jobs.py
@@ -126,3 +126,5 @@ def test_send_email_to_contact_details_without_putting_it_in_cc(
     mock_call = mock_send_email.mock_calls[0][2]
     assert mock_call["recipient"] == expected_recipient
     assert mock_call["cc"] == [expected_cc]
+    assert mock_call["body"]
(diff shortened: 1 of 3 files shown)


backend/tests/integration/mailing/test_api_jobs.py
```diff
@@ -126,3 +126,5 @@ def test_send_email_to_contact_details_without_putting_it_in_cc(
     mock_call = mock_send_email.mock_calls[0][2]
     assert mock_call["recipient"] == expected_recipient
     assert mock_call["cc"] == [expected_cc]
+    assert mock_call["body"]
+    assert mock_call["subject"] == "Expired deadline for your INSPIRE job: Tester"
```
michamos: pass `autospec=True` to avoid this issue? see https://docs.python.org/3/library/unittest.mock.html#unittest.mock.patch

ammirate: it is a nice addition, but it won't check the actual content, so this is still needed :+1:

michamos: yeah, comment was about avoiding the issue of misnaming arguments, not about this specific line.

ammirate: Sure, totally agree.

PR #482, opened 2019-10-01
app: add displatcher middleware without host name checking

drjova:
* INSPIR-2740

### DIFF
--- a/backend/inspirehep/config.py
+++ b/backend/inspirehep/config.py
@@ -133,7 +133,7 @@ SQLALCHEMY_DATABASE_URI = (
 # JSONSchemas
 # ===========
 #: Hostname used in URLs for local JSONSchemas.
-JSONSCHEMAS_HOST = "localhost:5000"
+JSONSCHEMAS_HOST = "localhost:8000"
 
 # Flask configuration
 # ===================
--- a/backend/inspirehep/factory.py
+++ b/backend/inspirehep/factory.py
@@ -7,8 +7,9 @@
 
 from invenio_app.factory import app_class, instance_path
 from invenio_base.app import create_app_factory
-from invenio_base.wsgi import create_wsgi_factory, wsgi_proxyfix
+from invenio_base.wsgi import wsgi_proxyfix
 from invenio_config import create_config_loader
+from werkzeug.middleware.dispatcher import DispatcherMiddleware
 
 from . import config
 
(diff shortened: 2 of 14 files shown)


backend/inspirehep/factory.py
```diff
@@ -22,6 +23,29 @@ def config_loader(app, **kwargs_config):
...
 
+class DispatcherMiddlewareWithoutHostNameChecking(DispatcherMiddleware):
+    """WSGI middleware without checking hostname.
+
+    It will override the request ``HOST`` to always match
+    the ``SERVER_NAME``.
+    """
+
+    def __call__(self, environ, start_response):
+        if "SERVER_NAME" in environ:
```
michamos: this is always present according to the spec: https://www.python.org/dev/peps/pep-0333/#environ-variables

backend/inspirehep/factory.py
```diff
@@ -22,6 +23,29 @@ def config_loader(app, **kwargs_config):
...
+            environ["HTTP_HOST"] = environ["SERVER_NAME"]
+        return super().__call__(environ, start_response)
+
+
+def inspirehep_create_wsgi_factory(mounts_factories):
```
michamos: is the `inspirehep_` prefix really needed?

backend/inspirehep/factory.py
michamos: on the other hand this is passed by the WSGI server, so it's probably not the same thing as `app.config['SERVER_NAME']`

backend/inspirehep/factory.py
drjova: just not to make it clear is not the default invenio one

backend/inspirehep/config.py
```diff
@@ -133,7 +133,7 @@
 # JSONSchemas
 # ===========
 #: Hostname used in URLs for local JSONSchemas.
-JSONSCHEMAS_HOST = "localhost:5000"
+JSONSCHEMAS_HOST = "localhost:8000"
```
michamos: Is this needed?

drjova: just for consinstency with the ``SERVER_NAME``

backend/inspirehep/factory.py
```diff
@@ -22,6 +23,37 @@ def config_loader(app, **kwargs_config):
...
+
+    def __init__(self, app, mounts=None, server_name=None):
+        self.app = app
+        self.mounts = mounts or {}
+        self.server_name = server_name
+        super().__init__(app, mounts)
+
+    def __call__(self, environ, start_response):
+        if self.server_name:
+            environ["HTTP_HOST"] = self.server_name
```
michamos: so taking it from the `environ['SERVER_NAME']` is not OK in the end?

drjova: I was wrong, actually if ```HTTP_HOST``` is present (https://github.com/pallets/werkzeug/blob/b81aa0f18c7336a0421b569d8a5dab8d5c0d7517/src/werkzeug/wsgi.py#L161 and https://github.com/pallets/werkzeug/blob/master/src/werkzeug/routing.py#L1510) is been checked against the wsgi_app ``SERVER_NAME`` not the ``environ['SERVER_NAME']`` because the app ``SERVER_NAME`` is not ``None`` https://github.com/pallets/werkzeug/blob/master/src/werkzeug/routing.py#L1513-L1516

drjova: I can saw you IRL

PR #483, opened 2019-10-02
serializers: use control_number instead of BAI on facet_author_name

monaawi:
* INSPIR-2752

### DIFF
--- a/backend/inspirehep/records/marshmallow/utils.py
+++ b/backend/inspirehep/records/marshmallow/utils.py
@@ -6,7 +6,7 @@
 # the terms of the MIT License; see LICENSE file for more details.
 
 from inspire_utils.name import ParsedName
-from inspire_utils.record import get_value, get_values_for_schema
+from inspire_utils.record import get_value
 
 
 def get_display_name_for_author_name(author_name):
@@ -15,14 +15,10 @@ def get_display_name_for_author_name(author_name):
 
 
 def get_facet_author_name_for_author(author):
-    author_ids = author.get("ids", [])
-    author_bai = get_values_for_schema(author_ids, "INSPIRE BAI")
-    bai = author_bai[0] if author_bai else "BAI"
+    author_control_number = author["control_number"]
     author_preferred_name = get_value(author, "name.preferred_name")
 
     if author_preferred_name:
-        return "{}_{}".format(bai, author_preferred_name)
+        return f"{author_control_number}_{author_preferred_name}"
 
-    return "{}_{}".format(
-        bai, get_display_name_for_author_name(get_value(author, "name.value"))
-    )
+    return f"{author_control_number}_{get_display_name_for_author_name(get_value(author, 'name.value'))}"
--- a/backend/tests/integration-async/disambiguation/test_disambiguation_tasks.py
+++ b/backend/tests/integration-async/disambiguation/test_disambiguation_tasks.py
@@ -1,0 +1,70 @@
+# -*- coding: utf-8 -*-
+#
+# Copyright (C) 2019 CERN.
+#
+# inspirehep is free software; you can redistribute it and/or modify it under
+# the terms of the MIT License; see LICENSE file for more details.
+
+import json
+
+from helpers.providers.faker import faker
+from invenio_db import db
+from invenio_pidstore.models import PersistentIdentifier
+from invenio_search import current_search_client as es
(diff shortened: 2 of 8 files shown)


backend/inspirehep/records/marshmallow/utils.py
```diff
@@ -15,14 +15,15 @@ def get_display_name_for_author_name(author_name):
 
 
 def get_facet_author_name_for_author(author):
-    author_ids = author.get("ids", [])
-    author_bai = get_values_for_schema(author_ids, "INSPIRE BAI")
-    bai = author_bai[0] if author_bai else "BAI"
+    control_number = (
+        author["control_number"] if author.get("control_number") else "NOREC"
```
michamos: There should always be a control number here. Maybe log an error if that's not the case and don't put it in the facet?

backend/inspirehep/records/marshmallow/utils.py
```diff
@@ -15,14 +15,15 @@ def get_display_name_for_author_name(author_name):
...
+    )
     author_preferred_name = get_value(author, "name.preferred_name")
 
     if author_preferred_name:
-        return "{}_{}".format(bai, author_preferred_name)
+        return "{}_{}".format(control_number, author_preferred_name)
```
michamos: Use f-strings instead of `format` (wherever it makes sense):
```suggestion
        return f"{control_number}_{author_preferred_name}"
```

backend/tests/integration/disambiguation/test_tasks.py
```diff
@@ -84,6 +85,43 @@ def test_disambiguate_signatures_cluster_with_0_authors(
     assert expected_ref == record["authors"][0]["record"]["$ref"]
 
 
+@freeze_time("2019-02-15")
```
michamos: I don't see any assertions on dates. Why is this needed?

backend/tests/integration/disambiguation/test_tasks.py
```diff
@@ -84,6 +85,43 @@ def test_disambiguate_signatures_cluster_with_0_authors(
...
+def test_disambiguate_signatures_cluster_creates_author_with_facet_author_name(
```
michamos: how is this different from the async test you also added?

backend/inspirehep/records/marshmallow/utils.py
monaawi: but in the schema the control number isn't required, this is why I added this. Do you think this check should be skipped?
What do you mean by don't put it in the facet? Put NOREC instead? I am already doing this.

backend/tests/integration/disambiguation/test_tasks.py
monaawi: In the async one I'm checking the facet_author_name on the literature record once an author is linked by disambiguation (I moved it to async so I don't trigger the indexing myself, can also be moved to the integration tests and I would trigger the indexing, wdyt?), I guess the naming is misleading. In the integration one, I'm checking the facet_author_name on a newly created author (from the author's detail page)

backend/inspirehep/records/marshmallow/utils.py
monaawi: Actually, I don't think that this check is necessary anymore - I think I was being paranoid, I'll remove it. This method is either called in the literature serializer (where we are already retrieving linked authors by id so it has to be there) or from the author details serializer.

michamos: We had some records in the past without a control number. It shouldn't be happening, so I think it's OK to blow up.

backend/tests/integration-async/disambiguation/test_disambiguation_tasks.py
```diff
@@ -0,0 +1,70 @@
...
+from invenio_db import db
+from invenio_pidstore.models import PersistentIdentifier
+from invenio_search import current_search_client as es
+
+from inspirehep.disambiguation.tasks import disambiguate_signatures
+from inspirehep.records.api import AuthorsRecord
+from inspirehep.records.api.literature import LiteratureRecord
+
+
+def test_author_created_by_disambiguation_has_author_facet_name(
```
michamos: then maybe
```suggestion
def test_signature_linked_by_disambiguation_has_correct_facet_author_name(
```

PR #484, opened 2019-10-03
migrator: change log to contain the number of records

drjova:

### DIFF
--- a/backend/inspirehep/migrator/api.py
+++ b/backend/inspirehep/migrator/api.py
@@ -39,7 +39,7 @@ def continuous_migration():
         try:
             migrated_records = None
             num_of_records = r.llen("legacy_records")
-            LOGGER.info(f"Starting migration of {num_of_records} records.")
+            LOGGER.info("Starting migration of records.", records_total=num_of_records)
 
             while r.llen("legacy_records"):
                 raw_record = r.lrange("legacy_records", 0, 0)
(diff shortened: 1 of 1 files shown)


backend/inspirehep/migrator/api.py
```diff
@@ -39,15 +39,22 @@ def continuous_migration():
         try:
             migrated_records = None
             num_of_records = r.llen("legacy_records")
-            LOGGER.info(f"Starting migration of {num_of_records} records.")
+            LOGGER.info(
+                f"Starting migration of {num_of_records} records.",
```
michamos: Isn't the point of structured logging to have the vars outside of the message, which serves only for the type of message? so
```suggestion
                "Starting migration of records.",
```

backend/inspirehep/migrator/api.py
```diff
@@ -39,15 +39,22 @@ def continuous_migration():
...
             while r.llen("legacy_records"):
                 raw_record = r.lrange("legacy_records", 0, 0)
                 if raw_record:
                     migrated_records = insert_into_mirror(
                         [zlib.decompress(raw_record[0])]
                     )
-                    LOGGER.debug(f"Migrated {len(migrated_records)} records.")
+                    migrated_records_len = len(migrated_records)
+                    LOGGER.debug(
+                        f"Migrated {migrated_records_len} records.",
```
michamos:
```suggestion
                        "Finished migrating records.",
```

backend/inspirehep/migrator/api.py
```diff
@@ -39,15 +39,22 @@ def continuous_migration():
...
+                        records_migrated=migrated_records_len,
```
michamos: maybe use same key name for both messages?

PR #492, opened 2019-10-07
facets: rename peer reviewed to published

miguelgrc:
* INSPIR-2363

Note: server.js has been modified since the previous default values are not being read at the time of generating the recordings for the UI tests. The url and http schema have to be set via environmental variables.

### DIFF
--- a/backend/inspirehep/records/config.py
+++ b/backend/inspirehep/records/config.py
@@ -409,7 +409,17 @@ HEP_COMMON_AGGS = {
     },
     "doc_type": {
         "terms": {"field": "facet_inspire_doc_type", "size": 20},
-        "meta": {"title": "Document Type", "order": 6, "type": "checkbox"},
+        "meta": {
+            "title": "Document Type",
+            "order": 6,
+            "type": "checkbox",
+            "bucket_help": {
+                "published": {
+                    "text": "Published papers are believed to have undergone rigorous peer review.",
+                    "link": "https://inspirehep.net/info/faq/general#published",
+                }
+            },
+        },
     },
     "author_count": {
         "range": {
@@ -514,7 +524,7 @@ RECORDS_REST_SORT_OPTIONS = {
         },
     },
     "records-jobs": {
-        "mostrecent": {"title": "Most Recent", "fields": ["-_created"], "order": 1},
+        "mostrecent": {"title": "Most Recent", "fields": ["-_created"], "order": 1}
     },
 }
 
--- a/backend/tests/integration/records/views/test_views_literature.py
+++ b/backend/tests/integration/records/views/test_views_literature.py
@@ -420,6 +420,29 @@ def test_literature_facets_arxiv(api_client, db, es_clear, create_record):
         assert expected_data_hits_source == source["_source"]
 
 
+def test_literature_facets_doc_type_has_bucket_help(
+    api_client, db, es_clear, create_record
+):
+    response = api_client.get("/literature/facets")
+    response_data = json.loads(response.data)
+    response_status_code = response.status_code
+    response_data_facet_bucket_help = (
+        response_data.get("aggregations").get("doc_type").get("meta").get("bucket_help")
+    )
+
+    expected_status_code = 200
+    expected_text = (
+        "Published papers are believed to have undergone rigorous peer review."
+    )
+    expected_link = "https://inspirehep.net/info/faq/general#published"
+
+    assert expected_status_code == response_status_code
+    assert "published" in response_data_facet_bucket_help
+    assert expected_text == response_data_facet_bucket_help["published"]["text"]
+    assert expected_link == response_data_facet_bucket_help["published"]["link"]
+    assert len(response_data["hits"]["hits"]) == 0
+
+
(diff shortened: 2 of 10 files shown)


pazembrz:
Python looks correct, but someone have to check JS part.

backend/inspirehep/records/config.py
```diff
@@ -514,7 +524,7 @@
         },
     },
     "records-jobs": {
-        "mostrecent": {"title": "Most Recent", "fields": ["-_created"], "order": 1},
+        "mostrecent": {"title": "Most Recent", "fields": ["-_created"], "order": 1}
```
drjova: why? :D

backend/tests/integration/records/views/test_views_literature.py
```diff
@@ -420,6 +420,29 @@ def test_literature_facets_arxiv(api_client, db, es_clear, create_record):
...
+    response = api_client.get("/literature/facets")
+    response_data = json.loads(response.data)
+    response_status_code = response.status_code
+    response_data_facet_bucket_help = (
+        response_data.get("aggregations").get("doc_type").get("meta").get("bucket_help")
+    )
+
+    expected_status_code = 200
+
+    print(response_data_facet_bucket_help)
```
drjova: left over ☺️

backend/tests/integration/records/views/test_views_literature.py
```diff
@@ -420,6 +420,29 @@ def test_literature_facets_arxiv(api_client, db, es_clear, create_record):
...
+
+    assert expected_status_code == response_status_code
+    assert "published" in response_data_facet_bucket_help
+    assert all(
+        key in response_data_facet_bucket_help.get("published")
+        for key in ["text", "link"]
+    )
```
drjova: I'd prefer to have something more explicit for testing for example:

```python
expected_text == response_data_facet_bucket_help['published']['text']
```

backend/inspirehep/records/config.py
```diff
@@ -409,7 +409,17 @@
...
     "doc_type": {
         "terms": {"field": "facet_inspire_doc_type", "size": 20},
-        "meta": {"title": "Document Type", "order": 6, "type": "checkbox"},
+        "meta": {
+            "title": "Document Type",
+            "order": 6,
+            "type": "checkbox",
+            "bucket_help": {
+                "published": {
+                    "text": "Published papers are believed to have undergone rigorous peer review",
```
drjova:
```suggestion
                    "text": "Published papers are believed to have undergone rigorous peer review.",
```

ui/src/common/components/CheckboxAggregation.jsx
```diff
@@ -69,6 +71,44 @@ class CheckboxAggregation extends Component {
...
+          help={
+            <>
+              {text}
+              {separator}
+              {link}
+            </>
+          }
+        />
+      </>
+    );
```
drjova: we can reduce this code with something like:

```js
const bucketText = bucketKey.get('text');
const bucketLink = bucketKey.get('link');

return (
<HelpIconTooltip
          help={
              <>
                {bucketText} { bucketLink && <ExternalLink href={bucketLink}>Learn More</ExternalLink>
              </>
          } 
/>
)
```

backend/inspirehep/records/config.py
miguelgrc: It's because of Black, its criteria for commas is weird... it puts it here and not above in line 420

ui/src/common/components/CheckboxAggregation.jsx
```diff
@@ -69,6 +71,32 @@ class CheckboxAggregation extends Component {
     });
   }
 
+  // eslint-disable-next-line class-methods-use-this
+  generateTooltip(bucketHelp, bucketKey) {
+    const bucketHelpKey = bucketHelp.get(bucketKey);
+
```
drjova:
```js
if (!bucketHelpKey) {
   return null;
}

```

This way you exit in the beggining.

ui/src/common/components/CheckboxAggregation.jsx
```diff
@@ -88,9 +116,9 @@ class CheckboxAggregation extends Component {
 
   renderBucket(bucket) {
     const { selectionMap } = this.state;
-    const { splitDisplayName } = this.props;
-
+    const { splitDisplayName, bucketHelp } = this.props;
     const bucketKey = bucket.get('key');
```
drjova: const bucketHelpKey = bucketHelp.get(bucketKey); and pass this directly to ``generateTooltip``

ui/src/common/components/CheckboxAggregation.jsx
```diff
@@ -69,6 +71,32 @@ class CheckboxAggregation extends Component {
     });
   }
 
+  // eslint-disable-next-line class-methods-use-this
+  generateTooltip(bucketHelp, bucketKey) {
```
drjova:
```suggestion
  renderBucketHelpTooltip(bucketHelp, bucketKey) {
```

PR #494, opened 2019-10-07
indexer: Retry indexer after db connection issue

pazembrz:
* Adds an auto retry when one of sqlalchemy connection issue Exceptions are thrown
* fixes unit.records.indexer.base tests as they were never executed

INSPIR-2686

### DIFF
--- a/backend/inspirehep/records/indexer/tasks.py
+++ b/backend/inspirehep/records/indexer/tasks.py
@@ -8,6 +8,13 @@
 import structlog
 from celery import shared_task
 from elasticsearch import NotFoundError
+from sqlalchemy.exc import (
+    DisconnectionError,
+    OperationalError,
+    ResourceClosedError,
+    TimeoutError,
+    UnboundExecutionError,
+)
 from sqlalchemy.orm.exc import NoResultFound, StaleDataError
 
 from inspirehep.records.api import LiteratureRecord
@@ -15,6 +22,17 @@ from inspirehep.records.indexer.base import InspireRecordIndexer
 from inspirehep.records.indexer.utils import get_record
 
 LOGGER = structlog.getLogger()
+
+
+CELERY_INDEX_RECORD_RETRY_ON_EXCEPTIONS = (
+    NoResultFound,
+    StaleDataError,
+    DisconnectionError,
+    TimeoutError,
+    UnboundExecutionError,
+    ResourceClosedError,
+    OperationalError,
+)
 
 
 @shared_task(ignore_result=False, bind=True)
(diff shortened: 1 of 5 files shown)


backend/inspirehep/records/indexer/tasks.py
```diff
@@ -81,6 +89,23 @@ def index_record(self, uuid, record_version=None, force_delete=None):
...
+    ) as e:
+        backoff = 2 ** (self.request.retries + 1)
+        if self.max_retries < self.request.retries + 1:
+            LOGGER.exception(
+                "DB Connection error. Scheduled retry for record",
+                uuid=uuid,
+                record_version=record_version,
+                force_delete=force_delete,
+            )
+        raise self.retry(countdown=backoff, exc=e)
```
drjova: why not https://docs.celeryproject.org/en/v4.2.2/userguide/tasks.html#automatic-retry-for-known-exceptions ? :)

pazembrz: Task requirements ;) But actually it's good idea, we should also do this for StaleDataError. I will create task for this.

backend/inspirehep/records/indexer/tasks.py
```diff
@@ -81,6 +89,23 @@ def index_record(self, uuid, record_version=None, force_delete=None):
...
             )
         raise self.retry(countdown=backoff, exc=e)
+    except (
+        DisconnectionError,
+        TimeoutError,
+        InvalidatePoolError,
+        OperationalError,
+        UnboundExecutionError,
+        ResourceClosedError,
+    ) as e:
```
michamos: isn't there a root exception the other inherit from and that you could catch?

pazembrz: Only one which is common for all of them is `SQLAlchemyError`
For `DisconnectionError` and `InvalidatePoolError` I could use only `DisconnectionError` as `InvalidatePoolError` is a subclass `DisconnectionError`. But for other exceptions it's not possible :/

backend/inspirehep/records/indexer/tasks.py
```diff
@@ -17,6 +24,17 @@
 LOGGER = structlog.getLogger()
 
 
+retry_on = (
```
drjova:
```suggestion
CELERY_INDEX_RECORD_RETRY_ON_EXCEPTIONS = (
```

backend/inspirehep/records/indexer/tasks.py
```diff
@@ -66,23 +90,8 @@ def index_record(self, uuid, record_version=None, force_delete=None):
...
-            LOGGER.debug(
-                "Record not yet at version on DB - Too many retries",
-                uuid=str(uuid),
-                version=record_version,
-                attempts=self.max_retries,
-            )
-        raise self.retry(countdown=backoff, exc=e)
-
     LOGGER.debug("Indexing record", uuid=str(uuid), version=record_version)
+    record = get_record(uuid, record_version)
```
drjova: We can add a ``TODO`` to move this function in ``InspireRecord``

backend/tests/unit/records/indexer/test_index_record.py
```diff
@@ -0,0 +1,101 @@
+# -*- coding: utf-8 -*-
+#
+# Copyright (C) 2019 CERN.
+#
+# inspirehep is free software; you can redistribute it and/or modify it under
+# the terms of the MIT License; see LICENSE file for more details.
+
+
```
drjova: One empty space between license and imports :P

backend/inspirehep/records/indexer/tasks.py
pazembrz: :+1:

PR #497, opened 2019-10-09
cite: cite multiple papers

monaawi:
* Enabled latex and bibtex serializers
* Add Cite All functionality
* INSPIR-2663


ui/src/literature/components/CiteAllAction.jsx
```diff
@@ -0,0 +1,82 @@
...
+    } catch (error) {
+      this.setState({ loading: false });
+    }
+  }
+
+  renderDropdownTitle(disabled) {
+    const { loading } = this.state;
+    return (
+      <Tooltip
+        title={disabled ? 'Only up to 500 results can be exported.' : null}
```
drjova:
```suggestion
        title={disabled ? `Only up to ${MAX_NUMBER_OF_CITED_RECORDS} results can be exported.` : null}
```

drjova:
@monaawi LGTM 👍 I'd add a screenshot test for the disabled case as well. WDYT?

monaawi:
@drjova There is already a test for disabled in CiteAllAction tests. There was also a case in ui tests in the search where we had more than 500 results and we could see it disabled. But I made it available only for superusers so it is not visible there for now.

PR #498, opened 2019-10-10
literature: Add fulltext_links

pazembrz:
* INSPIR-2737

### DIFF
--- a/backend/inspirehep/records/marshmallow/literature/ui.py
+++ b/backend/inspirehep/records/marshmallow/literature/ui.py
@@ -7,6 +7,7 @@
 
 import json
 
+from flask import current_app
 from inspire_utils.date import format_date
 from inspire_utils.record import get_value
 from marshmallow import fields, missing
@@ -83,6 +84,7 @@ class LiteratureDetailSchema(LiteraturePublicSchema):
     external_system_identifiers = fields.Nested(
         ExternalSystemIdentifierSchemaV1, dump_only=True, many=True
     )
+    fulltext_links = fields.Method("get_fulltext_links", dump_only=True)
     isbns = fields.List(fields.Nested(IsbnSchemaV1, dump_only=True))
     number_of_authors = fields.Method("get_number_of_authors")
     number_of_references = fields.Method("get_number_of_references")
@@ -96,6 +98,49 @@ class LiteratureDetailSchema(LiteraturePublicSchema):
         if earliest_date is None:
             return missing
         return format_date(earliest_date)
+
+    def get_fulltext_links(self, data):
+        field_data = []
+        fields_to_include = {
+            "arxiv_eprints": self.get_arxiv_fulltext_link,
+            "external_system_identifiers": self.get_kek_fulltext_link,
+            "documents": self.get_internal_fulltext_link,
+        }
+        for field, process_method in fields_to_include.items():
+            for item in data.get(field, []):
+                field_data.append(process_method(item))
+        return field_data
+
+    def get_kek_fulltext_link(self, data):
+        description = "KEK scanned document"
+        kek_id = data.get("value")
+        if kek_id and data.get("schema", "") == "KEKSCAN":
+            return {
+                "description": description,
+                "value": ExternalSystemIdentifierSchemaV1.get_link_for_kekscan_schema(
+                    kek_id
+                ),
+            }
+        return missing
+
+    def get_arxiv_fulltext_link(self, data):
+        description = "arXiv"
+        arxiv_id = data.get("value")
+        if arxiv_id:
+            return {
+                "description": description,
+                "value": "https://arxiv.org/pdf/%s" % arxiv_id,
+            }
+        return missing
+
+    def get_internal_fulltext_link(self, data):
+        if not current_app.config.get("FEATURE_FLAG_ENABLE_FILES"):
+            return missing
+        description = data.get("description") or "fulltext"
+        url = data.get("url")
+        if url and not data.get("hidden", False) and data.get("fulltext"):
+            return {"description": description, "value": url}
+        return missing
 
     def get_number_of_authors(self, data):
         authors = data.get("authors")
--- a/backend/tests/integration/records/indexer/test_indexer_literature.py
+++ b/backend/tests/integration/records/indexer/test_indexer_literature.py
@@ -7,12 +7,8 @@
 
 import json
 
-import pytest
 from invenio_search import current_search_client as es
 from mock import patch
-from sqlalchemy.exc import DisconnectionError
-
-from inspirehep.records.indexer.tasks import index_record
 
 
 def test_index_literature_record(es_clear, db, datadir, create_record):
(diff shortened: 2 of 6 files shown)


backend/inspirehep/records/marshmallow/base.py
```diff
@@ -89,3 +89,21 @@ class MetadataSchema(EnvelopeSchema):
         metadata = fields.Nested(schema, dump_only=True)
 
     return MetadataSchema
+
+
+class RequiredFieldsSchema(Schema):
```
drjova: IMHO this is too complex, each just a check on the specific serializer instead on ``post_dump`` or even better https://github.com/inspirehep/inspirehep/pull/498/files#diff-19d3d162ce2a08a6469fc1ae8afae85fR112

backend/inspirehep/records/marshmallow/literature/common/fulltext_links.py
```diff
@@ -0,0 +1,70 @@
...
+)
+
+
+class FulltextLinkBase(RequiredFieldsSchema):
+    value = fields.Method("get_fulltext_url", dump_only=True, required=True)
+    description = fields.Method(
+        "get_fulltext_description", dump_only=True, required=True
+    )
+
+    DESCRIPTION = None
```
drjova: Maybe?
```suggestion
    default_description = None
```

backend/inspirehep/records/marshmallow/literature/common/fulltext_links.py
```diff
@@ -0,0 +1,70 @@
...
+            if entry["schema"].lower() == "kekscan":
+                return ExternalSystemIdentifierSchemaV1.get_link_for_kekscan_schema(
+                    value
+                )
+        return missing
+
+
+class ArxivFulltextLink(FulltextLinkBase):
+    DESCRIPTION = "arXiv"
+    ARXIV_FULLTEXT_URL = "https://arxiv.org/pdf/{arxiv_eprint_id}"
```
drjova:
```suggestion
    arxiv_fulltext_url = "https://arxiv.org/pdf/{arxiv_eprint_id}"
```

backend/inspirehep/records/marshmallow/literature/common/fulltext_links.py
```diff
@@ -0,0 +1,70 @@
...
+        value = entry.get("value")
+        if value:
+            return self.ARXIV_FULLTEXT_URL.format(arxiv_eprint_id=value)
+        return missing
+
+
+class InternalFulltextLink(FulltextLinkBase):
+    DESCRIPTION = "fulltext"
+
+    def get_fulltext_url(self, entry):
```
drjova:
```python
if current_app.config.get("FEATURE_FLAG_ENABLE_FILES"):
    return missing
```

backend/inspirehep/records/marshmallow/literature/common/fulltext_links.py
```diff
@@ -0,0 +1,70 @@
...
+        raise NotImplementedError
+
+    def get_fulltext_description(self, entry):
+        return self.DESCRIPTION
+
+
+class KekFulltextLink(FulltextLinkBase):
+    DESCRIPTION = ("KEK scanned document",)
```
drjova:
```
"KEK scanned document"
```

backend/inspirehep/records/marshmallow/literature/ui.py
```diff
@@ -97,6 +103,18 @@ def get_formatted_earliest_date(self, data):
             return missing
         return format_date(earliest_date)
 
+    def get_fulltext_links(self, data):
+        field_data = []
+        fields_to_include = {
+            "arxiv_eprints": ArxivFulltextLink(),
+            "external_system_identifiers": KekFulltextLink(),
+            "documents": InternalFulltextLink(),
+        }
```
drjova: IMHO we shouldn't use schema in schema, why we just create three methods that handle this? i.e.

```python
def get_arxiv_fulltext_link():
   ....

def get_kek_fulltext_link():
   ....

def get_internal_fulltext_link():
   ....
```

ammirate: @drjova what's the advantage of this?

drjova: we're not calling ``dump`` schema in schema and it's just what we need

pazembrz: IMHO your sollution will be even more complex. Here we have full info about the data source (so it if it arxiv, kek or internal url) so we do not kave to check almost anything. Inside Serializer we don't, so this will require more checks like: if keys are called `description` and `value` or are they called `schema` and `value`. Additionally recognition of schema KEK vs arXiv will require more logic inside this simple serializer. So there will be quire complex logic for simple thing.
Also Moving this amount of logic so deep into serializers is really bad idea.

drjova: it's the wrong way to use marshmallow, if you want to put the logic in a ``schema`` you can use ``Nested``, calling ``schema`` in ``schema`` it's wrong. Which means ``pre_dump`` you analyzing the date and you let marshmallow to handle the dumping.

pazembrz: But wait. You want to serialize it without serializer? Just pure dict in here? Then why we use serializes at all? Is this really what we want?

backend/tests/integration/records/indexer/test_indexer_literature.py
```diff
@@ -72,3 +68,49 @@ def test_indexer_deletes_record_from_es(es_clear, db, datadir, create_record):
     hits_total = response["hits"]["total"]
 
     assert hits_total == expected_total
+
+
+@patch("inspirehep.records.api.literature.LiteratureRecord.add_files")
+def test_indexer_creates_proper_fulltext_links_in_ui_display(
+    add_file_mock, base_app, es_clear, db, create_record
```
drjova: We have ``enable_files`` AFAIR, we can use that :)

pazembrz: :+1:

backend/tests/integration/records/indexer/test_indexer_literature.py
```diff
@@ -72,3 +68,49 @@ def test_indexer_deletes_record_from_es(es_clear, db, datadir, create_record):
     hits_total = response["hits"]["total"]
 
     assert hits_total == expected_total
+
+
+@patch("inspirehep.records.api.literature.LiteratureRecord.add_files")
+def test_indexer_creates_proper_fulltext_links_in_ui_display(
```
harunurhan: we might need a test case when all the source fields are empty, and we don't return `fulltext_links` at all.

also tests for the detail view would be good to

harunurhan:
commit message is misleading, as this doesn't only `_ui_display`, but also detail pages. `literature: ` prefix would be more appropriate.

backend/tests/integration/records/indexer/test_indexer_literature.py
pazembrz: It only sets flag in config and looks like `enable_files` is a little broken, it's not restoring previous value in config  when test ends.

backend/tests/integration/records/indexer/test_indexer_literature.py
pazembrz: :+1:

PR #501, opened 2019-10-14
ui: display all fulltext links

harunurhan:
* INSPIR-2638


ui/src/literature/components/FullTextLinksAction.jsx
```diff
@@ -0,0 +1,68 @@
...
+class FullTextLinksAction extends Component {
+  static renderIcon() {
+    return <IconText text="pdf" type="download" />;
+  }
+
+  renderDropdown() {
+    const { fullTextLinks } = this.props;
+    const hrefDisplayPairs = fullTextLinks.map(fullTextLinkToHrefDisplayPair);
+    return (
+      <DropdownMenu title={<Button>{FullTextLinksAction.renderIcon()}</Button>}>
```
drjova: can we just put the logic on ``renderIcon`` and rename it to something like ``renderIconButton``?

ui/src/literature/components/FullTextLinksAction.jsx
```diff
@@ -0,0 +1,68 @@
...
+        {hrefDisplayPairs.map(pair => (
+          <Menu.Item key={pair.get('href')}>
+            <EventTracker eventId="PdfDownload">
```
drjova: ``fullTextLink``?

ui/src/literature/components/FullTextLinksAction.jsx
```diff
@@ -0,0 +1,68 @@
...
+      </EventTracker>
+    );
+  }
+
+  render() {
+    const { fullTextLinks } = this.props;
+    return (
+      <ListItemAction>
+        {fullTextLinks.size > 1 ? this.renderDropdown() : this.renderOne()}
+      </ListItemAction>
```
drjova: do we want to dinstinguish between 1 and more? IMO we can always show the dropdown?

ui/src/literature/components/FullTextLinksAction.jsx
harunurhan: ?

ui/src/literature/components/FullTextLinksAction.jsx
harunurhan: nope, because it's not always button.. when it's single we need `ExternalLink` that wraps the icon.

ui/src/literature/components/FullTextLinksAction.jsx
harunurhan: this has been done like this for a while now :(, for 3-4 other components.

ui/src/literature/components/FullTextLinksAction.jsx
drjova: for the eventId do we always have pdfs?

harunurhan: yeah for now, we don't care I guess.

PR #503, opened 2019-10-14
ui: change rich text editor's text block tag.

harunurhan:
Quill's default tag after the line break is `p` which has overriden
style inside quill editor, no bottom margin 1em which is a web default
and exist everywhere including job detail page. Since we don't want to
break a web standard by removing margin for `p` inside job description,
this PR changes `P` tag to `DIV` instead, which has consistent
behaviour everywhere.

Note that adding margin back to P tag for Quill instead of changing whole
tag, has a limitation. It doesn't allow linebreaks without spaces.

* INSPIR-2799

### DIFF
--- a/ui/src/submissions/common/components/RichTextEditor/RichTextEditor.jsx
+++ b/ui/src/submissions/common/components/RichTextEditor/RichTextEditor.jsx
@@ -1,9 +1,14 @@
 import React, { Component } from 'react';
-import QuillEditor from 'react-quill';
+import QuillEditor, { Quill } from 'react-quill';
 
 import 'react-quill/dist/quill.snow.css';
 import './RichTextEditor.scss';
 import EditorToolbar from './EditorToolbar';
+
+// change default text default (`P`)
+const Block = Quill.import('blots/block');
+Block.tagName = 'DIV';
+Quill.register(Block, true);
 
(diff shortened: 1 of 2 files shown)


ui/src/submissions/common/components/RichTextEditor/RichTextEditor.jsx
```diff
@@ -1,10 +1,15 @@
...
-import QuillEditor from 'react-quill';
+import QuillEditor, { Quill } from 'react-quill';
 
 import 'react-quill/dist/quill.snow.css';
 import './RichTextEditor.scss';
 import EditorToolbar from './EditorToolbar';
 
+// change default text default (`P`)
+const Block = Quill.import('blots/block');
+Block.tagName = 'DIV';
```
michamos: that's not so nice as tt's really a paragraph semantically :disappointed: can we strip empty paragraphs instead? or change the style in Quill?

harunurhan: It's already in the commit message, If we change the styles for Quill, then we lose ability to do linebreaks without blank space, as it would create `p` whenever you hit enter, and each `p` would have a `margin-bottom: 1em`

PR #505, opened 2019-10-15
setup: add xrootd dependencies as an extra

drjova:
* poetry install -E "xrootd"

Signed-off-by: Harris Tzovanakis <me@drjova.com>

PR #506, opened 2019-10-15
Add the xrootd dependency and use the poetry base image

benjamin-bergia:


drjova:
Supersede by https://github.com/inspirehep/inspirehep/pull/505

PR #507, opened 2019-10-15
Citations by year for literature

harunurhan:


ui/src/actions/__tests__/citations.test.js
```diff
@@ -63,24 +62,15 @@ describe('citations - async action creator', () => {
...
-          query: {
-            author: ['J.M.Maxson.1_Jared'],
-          },
-        },
-      }),
-    });
-    await store.dispatch(fetchCitationSummary());
+    const store = getStore();
+    await store.dispatch(
+      fetchCitationSummary({ author: ['J.M.Maxson.1_Jared'] })
```
drjova: correct my if I'm wrong but we changed the BAI, shall we update these tests (since we're touching them) to reflect the changes? (I know it's not important but just as a guide)

PR #509, opened 2019-10-16
submissions: add preprint_date to form data

harunurhan:
* INSPIR-2806


backend/inspirehep/submissions/marshmallow/literature.py
```diff
@@ -163,6 +165,7 @@ def build_literature(self, data):
         literature.add_document_type(data["document_type"])
 
         literature.add_arxiv_eprint(data.get("arxiv_id"), data.get("arxiv_categories"))
+        literature.add_preprint_date(data.get("preprint_date"))
```
drjova: add a test for the ``None`` case

harunurhan: all other unit tests expect `test_load_preprint_date`,  are already testing when this is `None`.

PR #511, opened 2019-10-16
ui: curators will see 100 results on author detail page

pazembrz:
*INSPIR-2767


ui/src/actions/authors.js
```diff
@@ -51,10 +53,13 @@ export function fetchAuthor(recordId) {
   };
 }
 
-function getAuthorPublicationsQuery(authorsState, newQuery) {
+function getAuthorPublicationsQuery(authorsState, newQuery, user) {
+  const userIsCataloger = isCataloger(Set(user.getIn(['data', 'roles'])));
+  const userIsLoggedIn = user.get('loggedIn');
```
harunurhan: no need to check this, since `userIsCataloger` only when `loggedIn` is `true`.

ui/src/actions/authors.js
```diff
@@ -51,10 +53,13 @@ export function fetchAuthor(recordId) {
...
   return {
     ...authorsState.getIn(['publications', 'query']).toJS(),
     ...newQuery,
+    ...(userIsLoggedIn && userIsCataloger ? { size: 100 } : {}),
```
harunurhan: `size: userIsCataloger ? 100 : undefined` is more direct, if you don't have any reason to do this.

ui/src/actions/__tests__/authors.test.js
```diff
@@ -96,6 +96,78 @@ describe('AUTHOR - async action creators', () => {
...
+            roles: ['cataloger'],
+          },
+        }),
+      });
+      await store.dispatch(fetchAuthorPublications({ page: 1, q: 'test' }));
+      expect(store.getActions()).toEqual(expectedActions);
+      done();
+    });
+
+    it('and queries publications with proper size when user is not cataloger', async done => {
```
harunurhan: this isn't needed.

ui/src/actions/authors.js
harunurhan: And move this just before the `newQuery`

ui/src/actions/__tests__/authors.test.js
```diff
@@ -96,6 +96,78 @@ describe('AUTHOR - async action creators', () => {
       done();
     });
 
+    it('and queries publications with proper size when user is cataloger', async done => {
```
harunurhan: `but uses size=100 when user is cataloger and creates AUTHOR_PUBLICATIONS_SUCCESS if successful`

ui/ui-tests/utils/constants.js
```diff
@@ -15,6 +15,7 @@ const routes = {
     literatureDetail1472986: `${appBaseUrl}/literature/1472986`,
     authorSearch: `${appBaseUrl}/authors`,
     authorDetail983328: `${appBaseUrl}/authors/983328`,
```
harunurhan: you can remove this if not used.

ui/src/actions/__tests__/authors.test.js
pazembrz: We don't want to be sure that normal users get 10 results and not 100?

ui/src/actions/authors.js
pazembrz: If i do this, then size is completely removed from queries for non-catalogers

ui/src/actions/__tests__/authors.test.js
harunurhan: the existing test is already testing this, no?

ui/src/actions/authors.js
harunurhan: ah right, that's why it should be `100 : 10` instead of `100 : undefined`, but anyhow keeping like this already not bad.

ui/src/actions/__tests__/authors.test.js
pazembrz: No, we're not. Every other test in authors.test.js is using `size=5` as specified in test payload. None of them are using default `10`

harunurhan: default doesn't matter, it's not concern of the the action here, but the reducer. So as long as we're testing the size is not overwritten with `100` for non catalogers, that's OK.

pazembrz: :+1:

PR #514, opened 2019-10-17
facets: affiliation and different author names facets for catalogers

miguelgrc:
* INSPIR-2762


backend/inspirehep/serializers.py
```diff
@@ -82,4 +86,18 @@ def serialize_search(self, pid_fetcher, search_result, **kwargs):
             facets. This is used with
             ``inspirehep.search.factories.search.search_factory_only_with_aggs``.
         """
+
+        aggregations = search_result.get("aggregations")
+        new_aggs = deepcopy(aggregations)
```
michamos: do you really need to **deep**copy here?

backend/inspirehep/serializers.py
```diff
@@ -82,4 +86,18 @@ def serialize_search(self, pid_fetcher, search_result, **kwargs):
...
+
+        for agg_key, agg_value in aggregations.items():
```
michamos: this blows up if there are no aggregations (which you assume might happen given you use `get` on L90).

backend/inspirehep/serializers.py
```diff
@@ -82,4 +86,18 @@ def serialize_search(self, pid_fetcher, search_result, **kwargs):
...
+            if "nested" in agg_value:
+                nested = agg_value.get("nested")
```
michamos: you know that there is a `nested` key here so you don't have to use `get` but can use `[`...`]`

backend/inspirehep/serializers.py
```diff
@@ -82,4 +86,18 @@ def serialize_search(self, pid_fetcher, search_result, **kwargs):
...
+                for key in nested:
+                    if key != "doc_count":
+                        new_aggs[key] = nested.get(key)
+                new_aggs.pop(agg_key)
+
+        search_result["aggregations"] = new_aggs
```
michamos: FYI, I would write this more like follows. First define a helper function:
```python
def unnest_aggregation(agg):
    """Flatten the aggregation dict in case there are nested aggregations.
    
    Warning:
         This mutates the argument.
    """
    if "nested" not in agg:
        return agg
    del agg["nested"]["doc_count"]
    return agg["nested"]
```
(I think it's fine to remove the `doc_count` by mutating the dict here as we won't need the original later, but better be explicit about it).

Then you can do here
```python
aggregations = search_result.get("aggregations", {})
search_result["aggregations"] = {name: unnest_aggregation(agg) for (name, agg) in aggregations.items()}
```
which I believe is more readable and has the benefit of iterating only once over the aggregations instead of twice when using `copy` (or much worse with `deepcopy`).

backend/inspirehep/search/facets.py
```diff
@@ -30,8 +30,26 @@ def inner(values):
     return inner
 
 
+def must_match_all_filter_nested(field):
```
michamos:
```suggestion
def must_match_all_filter_nested(nested_path, match_field):
```

backend/inspirehep/serializers.py
michamos: I'm stupid, this doesn't work because you need to unnest twice. Anyway, the general point still stands: it's better to build the resulting `dict` as you go rather than copy the original and mutate it.

backend/inspirehep/serializers.py
```diff
@@ -82,4 +86,25 @@ def serialize_search(self, pid_fetcher, search_result, **kwargs):
...
             ``inspirehep.search.factories.search.search_factory_only_with_aggs``.
         """
+
+        search_result["aggregations"] = self._unnest_aggregations(
+            search_result.get("aggregations", {})
+        )
+
         return json.dumps(search_result)
+
+    def _unnest_aggregations(self, aggregations):
```
michamos: should be decorated with `@staticmethod`, as you don't use `self` in the body of the function.

backend/inspirehep/search/facets.py
```diff
@@ -30,8 +30,24 @@ def inner(values):
...
+def must_match_all_filter_nested(nested_path, match_field):
+    """Bool filter containing a list of must matches for nested queries."""
+
+    def inner(values):
+        if not values:
+            return
+        return Q(
+            "bool",
+            must=Q(
+                "nested", path=nested_path, query=Q("terms", **{match_field: values})
```
michamos: is this going to work if there are several `values`, i.e. if multiple checkboxes are ticket in the facet? from the difference with `must_match_all_filter`, I strongly suspect it won't.

backend/inspirehep/search/facets.py
```diff
@@ -30,8 +30,22 @@ def inner(values):
     return inner
 
 
+def must_match_all_filter_nested(nested_path, match_field):
+    """Bool filter containing a list of must matches for nested queries."""
+
+    def inner(values):
+        filters = []
+        for value in values:
+            filters.append(Q("match", **{match_field: value}))
```
michamos: The common pattern
```python
out_list = []
for element in in_list:
    if p(element):
        out_list.append(f(element))
```
can be expressed more clearly and concisely as a list comprehension
```python
out_list = [f(element) for element in in_list if p(element)]
```

backend/inspirehep/serializers.py
```diff
@@ -7,11 +7,14 @@
 
 import json
 
+import structlog
 from flask import current_app
 from invenio_records_rest.serializers.json import (
     JSONSerializer as InvenioJSONSerializer,
 )
 
+LOGGER = structlog.getLogger()
```
michamos: leftover?

backend/tests/integration/records/views/test_views_authors.py
```diff
@@ -7,25 +7,69 @@
...
+        f"/literature/facets?facet_name=hep-author-publication&author_recid=9999"
+    )
 
-    response = api_client.get("/literature/facets?facet_name=hep-author-publication")
     response_data = json.loads(response.data)
     response_status_code = response.status_code
     response_data_facet_keys = list(response_data.get("aggregations").keys())
 
+    print("KEYS -----------")
+    print(response_data_facet_keys)
```
michamos: leftover?

backend/tests/integration/records/views/test_views_authors.py
```diff
@@ -7,25 +7,69 @@
...
         "earliest_date",
         "collaboration",
+        "subject",
+        "arxiv_categories",
+        "self_affiliations",
+        "self_author_names",
     ]
     expected_facet_keys.sort()
     response_data_facet_keys.sort()
+    print(response_data_facet_keys)
```
michamos: leftover?

PR #515, opened 2019-10-17
cite: cite multiple papers with latex and bibtex display fields

monaawi:
* INSPIR-2663
* Add latex and bibtex display fields and use them for the latex/bibtex search
* Add author to the cite query in author profiles
* Add the sort to the cite query


backend/inspirehep/records/marshmallow/literature/bibtex.py
```diff
@@ -277,7 +278,12 @@ def hyphenate_if_possible(no_hyphens):
 
     @pre_dump
     def filter_data(self, data):
-        control_number = str(data.get("control_number"))
-        data["doc_type"] = BibTexCommonSchema.get_bibtex_document_type(data)
-        data["texkey"] = get_value(data, "texkeys[0]", default=control_number)
-        return data
+        processed_data = deepcopy(data)
```
michamos: do you really need the **deep**copy here?

backend/inspirehep/records/marshmallow/literature/es.py
```diff
@@ -44,6 +47,25 @@ class LiteratureElasticSearchSchema(ElasticSearchBaseSchema, LiteratureRawSchema
     def get_ui_display(self, record):
         return json.dumps(LiteratureDetailSchema().dump(record).data)
 
+    def get_latex_us_display(self, record):
+        from inspirehep.records.serializers.latex import latex_US
```
michamos: :cry: I guess you get circular import otherwise?

backend/inspirehep/records/marshmallow/literature/latex.py
```diff
@@ -41,7 +42,8 @@ def get_author_names(self, data):
         return [name.replace(". ", ".~") for name in author_names]
 
     def get_publication_info(self, data):
-        publication_info = BibTexCommonSchema.get_best_publication_info(data)
+        processed_data = deepcopy(data)
+        publication_info = BibTexCommonSchema.get_best_publication_info(processed_data)
```
michamos: so `BibTexCommonSchema.get_best_publication_info` mutates its argument? That's very misleading. I think you should fix that if it's the case rather than do a `deepcopy` here.

backend/tests/integration/records/indexer/test_indexer_literature.py
```diff
@@ -28,6 +28,9 @@ def test_index_literature_record(es_clear, db, datadir, create_record):
 
     result = response["hits"]["hits"][0]["_source"]
     result_ui_display = json.loads(result.pop("_ui_display"))
+    result.pop("_latex_us_display")
+    result.pop("_latex_eu_display")
+    result.pop("_bibtex_display")
```
michamos: no assertions here?

backend/inspirehep/records/marshmallow/literature/es.py
monaawi: :( yes

backend/inspirehep/records/marshmallow/literature/latex.py
monaawi: no, ```get_best_publication_info``` is not mutating its data. The data is mutated afterwards adding ```doc_type``` and ```texkey```. I should actually maybe copy the publication_info not the entire data

backend/inspirehep/records/marshmallow/literature/bibtex.py
monaawi: I'm actually not sure how to do this. So the data is being mutated here as well with adding ```texkey``` and ```doc_type``` which are being used by the other methods. Another idea would be to ditch the pre_dump and have a method to get those 2 fields whenever we need them without adding them to the data?

michamos: I think it's an ordering issue: other methods need the `doc_type` to decide how to serialize other fields, but there's no way to enforce the order in which they get evaluated. The `texkey` though looks like it could be moved out to a regular field. In any case, `copy` is sufficient, you don't need to `deepcopy` because you're only modifying the `data` itself (by adding a key), not nested fields.

backend/inspirehep/records/marshmallow/literature/es.py
michamos: no way to rearrange things to avoid that?

monaawi: Couldn't find a way, things in here are way too twisted together

backend/inspirehep/records/serializers/bibtex.py
```diff
@@ -136,8 +136,10 @@ def serialize(self, pid, record, links_factory=None):
     def serialize_search(
         self, pid_fetcher, search_result, links=None, item_links_factory=None
     ):
-        records = [hit["_source"] for hit in search_result["hits"]["hits"]]
-        return self.create_bibliography(records)
+        records = [
+            hit["_source"]["_bibtex_display"] for hit in search_result["hits"]["hits"]
+        ]
+        return "\n".join(records)
```
michamos: maybe add an extra newline here, like for LaTeX?

backend/inspirehep/records/marshmallow/literature/latex.py
```diff
@@ -42,23 +42,27 @@ def get_author_names(self, data):
 
     def get_publication_info(self, data):
         publication_info = BibTexCommonSchema.get_best_publication_info(data)
```
michamos: it would have been simpler to add a `.copy()` here rather than renaming everything.

backend/inspirehep/records/serializers/bibtex.py
monaawi: I checked how it was before and we had only one newline for bibtex, should I change it to 2?

michamos: Yes, I think it would be better to separate the entries by a blank line.

backend/inspirehep/records/marshmallow/literature/bibtex.py
```diff
@@ -6,6 +6,7 @@
 # the terms of the MIT License; see LICENSE file for more details.
 
 import re
+from copy import deepcopy
```
michamos: unused import

monaawi:
cool, I asked Harun to have a look at the frontend part. Yes, I already added tests for those with the first implementation

backend/inspirehep/records/serializers/bibtex.py
monaawi: Actually, I just tested again locally and the bibtex entry already has a newline at the end so we need to add only one here. I'll switch it back

ui/src/literature/components/__tests__/CiteAllAction.test.jsx
```diff
@@ -33,12 +36,32 @@ describe('CiteAllAction', () => {
   it('calls OnCiteClick when option clicked', () => {
```
harunurhan: this test is redundant, but I guess we can keep it for now, until the other is fixed

ui/src/literature/components/__tests__/CiteAllAction.test.jsx
```diff
@@ -33,12 +36,32 @@ describe('CiteAllAction', () => {
...
     );
     wrapper.find(DropdownMenu).prop('onClick')(CITE_FORMAT_VALUES[1]);
-    expect(onCiteClickSpy).toBeCalledWith(CITE_FORMAT_VALUES[1]);
+    expect(CiteAllAction.prototype.onCiteClick).toBeCalledWith(
+      CITE_FORMAT_VALUES[1]
+    );
     CiteAllAction.prototype.onCiteClick = originalOnCiteClick;
   });
+
+  // openCallCount stays 0 and toBeCalledWith doesn't work
```
harunurhan: could you add 
```js
// remove `calls OnCiteClick when option clicked` case, after this is enabled.
```

ui/src/literature/containers/CiteAllActionContainer.jsx
```diff
@@ -1,11 +1,14 @@
...
-import omit from 'lodash.omit';
 
+import omit from 'lodash.omit';
 import CiteAllAction from '../components/CiteAllAction';
 
 const stateToProps = state => ({
   numberOfResults: state.search.get('total'),
-  query: omit(state.router.location.query, ['sort', 'page']),
+  query: {
+    sort: state.search.getIn(['scope', 'query', 'sort']),
```
harunurhan: maybe it would be better to add something like:
```js
// set default `sort` incase there is no `sort`, in the location query
```

harunurhan: btw reading from the scope is OK, but since this is only for literature, alternatively you can directly read it from `searchScopes`.

ui/src/literature/containers/__tests__/CiteAllActionContainer.test.jsx
```diff
@@ -8,18 +8,41 @@ import CiteAllActionContainer from '../CiteAllActionContainer';
...
+        location: { query: { q: 'query', page: 1, size: 1 } },
+      },
+    });
+    const wrapper = mount(
+      <Provider store={store}>
+        <CiteAllActionContainer />
+      </Provider>
+    );
+    expect(wrapper.find(CiteAllAction)).toHaveProp({
+      query: { q: 'query', sort: 'mostrecent' },
```
harunurhan: maybe it's better import it in the test too, instead of hardcoding.

ui/src/literature/containers/CiteAllActionContainer.jsx
```diff
@@ -1,11 +1,16 @@
...
+import omit from 'lodash.omit';
 import CiteAllAction from '../components/CiteAllAction';
+import { searchScopes } from '../../reducers/search';
 
 const stateToProps = state => ({
   numberOfResults: state.search.get('total'),
-  query: omit(state.router.location.query, ['sort', 'page']),
+  query: {
+    // set default `sort` in case there is no `sort` in the location query
+    sort: searchScopes.getIn(['literature', 'query', 'sort']),
```
harunurhan: it's a bit weird to not to use `state` in `stateToProps`, perhaps this could go inside the component itself. But i'd say optional since it's not a problem for `EmbeddedSearch` yet.

monaawi: we are using state as well, we are combining this with the state

harunurhan: I meant that this particular logic having default `sort`, doesn't need the app state at all.

PR #517, opened 2019-10-20
records: simpler files handling

drjova:


backend/inspirehep/pidstore/api/base.py
```diff
@@ -41,7 +41,10 @@ def delete(cls, object_uuid, data):
         pid_base = cls(object_uuid, data)
         pid_base.delete_external_pids()
         for minter in cls.minters:
-            minter.delete(object_uuid, data)
+            try:
+                minter.delete(object_uuid, data)
+            except PIDDoesNotExistError:
+                LOGGER.warning("Cannot delete pid", object_uuid=object_uuid)
```
pazembrz: This message is misleading. Isn't it?

drjova: any suggestions? :)

backend/inspirehep/records/api/literature.py
```diff
@@ -185,7 +179,7 @@ def add_files(self):
             builder.add_document(**document)
             keys.append(document["key"])
             if "documents" in builder.record:
```
pazembrz: This should be outside for loop otherwise you update it every loop which is unnecessary.

drjova: oups missed that

backend/inspirehep/records/api/mixins.py
```diff
@@ -114,242 +112,119 @@ def update_refs_in_citation_table(self, save_every=100):
...
+            raise MissingModelError()
+        bucket = self.get_bucket()
+        return self.files_iter_cls(self, bucket=bucket, file_cls=self.file_cls)
+
+    def add_file(self, url, original_url=None, key=None, filename=None, **kwargs):
+        if self.local_url(url):
+            bucket, key = self.find_bucket_and_key_from_local_url(url)
+            key = self.add_local_file(key, bucket)
+        else:
+            key = self.add_file_from_url(url)
```
pazembrz: You are not handling situation when local url will be broken and you have to re-download file form original_url (this is why we decided to store original url) - requirement from files refactoring

backend/inspirehep/records/api/mixins.py
```diff
@@ -114,242 +112,122 @@ def update_refs_in_citation_table(self, save_every=100):
...
 
-        If everything is ok then adds it to files in current record.
-        If file with same hash already found in db, tries to use this one
-        instead of creating duplicate (uses `ObjectVersion.copy()` method)
+        self.files.flush()
+        return key_hashed
 
-        Args:
-            url (str): Local or remote url/filepath
+    def add_file_from_url(self, url):
```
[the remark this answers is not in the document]
drjova: If I understand correctly but it's handled and there is a test for this https://github.com/inspirehep/inspirehep/blob/c731ebab84a2af86173d2d510ed6c7e51b6652d9/backend/tests/integration/records/api/test_api_mixins.py#L64

backend/inspirehep/records/api/mixins.py
drjova: if there is local url then we copy from the storage, what's the point to download it from the original url ? 😌

drjova:
@pazembrz could you please be more specific on what functionality I'm missing? And why it looks weird and incomplete? ☺️ 

About the doc strings, I believe in most cases function names and the code is clear enough what is doing and there is no need to explain it

backend/inspirehep/records/api/mixins.py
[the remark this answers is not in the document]
pazembrz: No, this test checks local file only so actually we know the key as it's in the local url already.
I'm writing about situation when you have only url (like during harvest) and actually file is already in other record. Now we will add it again in another bucket, and requirement was to just to use copy method which actually won't copy data but it will increase reference count for it.

backend/inspirehep/records/api/mixins.py
pazembrz: I do not remember now but It was one of the requirements on the original files refactor.

pazembrz: I see 2 use-caces here. 1. File updated on original url, 2. file is broken on local storage.

drjova:
@pazembrz updated

drjova:
@pazembrz ping

backend/tests/integration-async/records/api/test_literature.py
```diff
@@ -38,3 +40,319 @@ def test_authors_signature_blocks_and_uuids_added_after_create_and_update(
...
+        record_control_number = record["control_number"]
+        record_bucket = record.bucket
+        db.session.commit()
+
+        record_from_db = LiteratureRecord.get_record_by_pid_value(record_control_number)
+
+        assert record_from_db["_files"]
+        assert len(record_from_db["_files"]) == 2
+
+        for document in record_from_db["documents"]:
```
ammirate: We can avoid this, since there's only one item in the list :)

backend/tests/integration-async/records/api/test_literature.py
```diff
@@ -38,3 +40,319 @@ def test_authors_signature_blocks_and_uuids_added_after_create_and_update(
...
+            assert "fulltext" in document
+            assert expected_document_filename == document["filename"]
+            assert expected_document_original_url == document["original_url"]
+            assert f'/api/files/{record_bucket}/{document["key"]}' == document["url"]
+
+            response = app.test_client().get(document["url"])
+            assert 200 == response.status_code
+
+        for figure in record_from_db["figures"]:
```
ammirate: Same

backend/tests/integration-async/records/api/test_literature.py
```diff
@@ -38,3 +40,319 @@ def test_authors_signature_blocks_and_uuids_added_after_create_and_update(
...
+
+        record_from_db = LiteratureRecord.get_record_by_pid_value(record_control_number)
+
+        expected_figure_filename = "WZ_fig4.png"
+        expected_figure_original_url = data_updated["figures"][0]["url"]
+
+        assert record_from_db["_files"]
+        assert len(record_from_db["_files"]) == 2
+
+        for document in record_from_db["documents"]:
```
ammirate: Same

backend/tests/integration-async/records/api/test_literature.py
```diff
@@ -38,3 +40,319 @@ def test_authors_signature_blocks_and_uuids_added_after_create_and_update(
...
+            assert "fulltext" in document
+            assert expected_document_filename == document["filename"]
+            assert expected_document_original_url == document["original_url"]
+            assert f'/api/files/{record_bucket}/{document["key"]}' == document["url"]
+
+            response = app.test_client().get(document["url"])
+            assert 200 == response.status_code
+
+        for figure in record_from_db["figures"]:
```
ammirate: Same

backend/tests/integration-async/records/api/test_literature.py
```diff
@@ -38,3 +40,319 @@ def test_authors_signature_blocks_and_uuids_added_after_create_and_update(
...
+        record_from_db.update(dict(record_from_db))
+
+        db.session.commit()
+        record_from_db = LiteratureRecord.get_record_by_pid_value(record_control_number)
+
+        assert record_from_db["_files"]
+        assert len(record_from_db["_files"]) == 1
+        assert "figures" not in record_from_db
+
+        for document in record_from_db["documents"]:
```
ammirate: Same

backend/tests/integration-async/records/api/test_literature.py
```diff
@@ -38,3 +40,319 @@ def test_authors_signature_blocks_and_uuids_added_after_create_and_update(
...
+        record_from_db.update(dict(record_from_db))
+
+        db.session.commit()
+        record_from_db = LiteratureRecord.get_record_by_pid_value(record_control_number)
+
+        assert record_from_db["_files"]
+        assert len(record_from_db["_files"]) == 1
+        assert "documents" not in record_from_db
+
+        for figure in record_from_db["figures"]:
```
ammirate: Here too

PR #519, opened 2019-10-22
global: add release tags for sentry

harunurhan:
* INSPIR-2802


backend/inspirehep/logger/ext.py
```diff
@@ -40,6 +42,7 @@ def init_sentry(self, app):
             dsn=sentry_dsn,
             integrations=[FlaskIntegration(), CeleryIntegration()],
             send_default_pii=send_default_pii,
+            release=os.environ.get("VERSION"),
```
michamos:
```suggestion
            release=curren_app.config.get("VERSION"),
```
instead?

harunurhan: we won't put it in the config.

PR #520, opened 2019-10-22
cite: handle exception for the bibtex and latex display fields

monaawi:
* Handle exception for the bibtex and latex display fields so the indexing doesn't fail
* INSPIR-2663

### DIFF
--- a/backend/inspirehep/records/marshmallow/literature/es.py
+++ b/backend/inspirehep/records/marshmallow/literature/es.py
@@ -8,6 +8,7 @@
 import json
 from itertools import chain
 
+import structlog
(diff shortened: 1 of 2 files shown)


backend/inspirehep/records/marshmallow/literature/es.py
```diff
@@ -26,6 +27,8 @@
 from .base import LiteratureRawSchema
 from .ui import LiteratureDetailSchema
 
+LOGGER = structlog.getLogger(__name__)
```
drjova:
```suggestion
LOGGER = structlog.getLogger()
```

PR #522, opened 2019-10-22
files: link bucket with record if record is created without bucket

drjova:


backend/inspirehep/records/api/mixins.py
```diff
@@ -114,11 +115,21 @@ def update_refs_in_citation_table(self, save_every=100):
 
 class FilesMixin:
     @classmethod
```
pazembrz: do you need this create_bucket as class-method? You are not calling it without initialized object.
Check how it was done before:
https://github.com/pazembrz/inspirehep/blob/15f90a513006a709f3fe1758460ec6b244993420/inspirehep/records/api/base.py#L389
https://github.com/pazembrz/inspirehep/blob/15f90a513006a709f3fe1758460ec6b244993420/inspirehep/records/api/base.py#L342

You could do it the same way.

drjova: https://github.com/inveniosoftware/invenio-records-files/blob/master/invenio_records_files/api.py#L325 just for reference :)

pazembrz: This is still so bad. Would be great if you overwrite create method and properly use create_bucket in there...

drjova: Honestly, I prefer to open a PR in invenio

pazembrz: but then you should also modify this again.

drjova: indeed and probably remove the ``create_bucket_and_link_to_record``

pazembrz: But still you could do this like that: https://github.com/pazembrz/inspirehep/blob/15f90a513006a709f3fe1758460ec6b244993420/inspirehep/records/api/base.py#L342-L408 which is already tested in old files code.

drjova: isn't it exactly what I'm doing here https://github.com/inspirehep/inspirehep/pull/522/files#diff-0e2e76aebfb3829e4197cc77b180fda8R227 and https://github.com/inveniosoftware/invenio-records-files/blob/master/invenio_records_files/api.py#L370-L375?

drjova: and the code that you send me is missing things like: ``self._bucket = bucket`` and ``self.dump_bucket(self, bucket)``

pazembrz: as this was changed in invenio ;) Actually you could use old code by only adding this things ;)

PR #523, opened 2019-10-23
Add a version env variable to the images

benjamin-bergia:


harunurhan:
let's merge this anyway, so that at least we have `version` associated with issues on sentry.

PR #526, opened 2019-10-23
references: Add pagination to references

monaawi:
* Added pagination to the references request
* INSPIR-2765


pazembrz:
Backend part looks ok.

backend/inspirehep/records/views.py
```diff
@@ -48,6 +51,28 @@ def get(self, pid, record):
...
+    @pass_record
+    def get(self, pid, record):
+        page = request.values.get("page", 1, type=int)
+        size = request.values.get("size", 25, type=int)
+
+        if page < 1 or size < 1:
+            abort(400)
+
+        references = record.get("references", [])
+        selected_references = references[(page - 1) * size : page * size]
```
drjova: if we allow any page and size this might cause ``500`` we should catch this and display with a migniful message or even better restrict the values

monaawi:
@drjova @harunurhan updated

ui/src/literature/components/ReferenceList.jsx
```diff
@@ -1,61 +1,77 @@
...
-
-ReferenceList.defaultProps = {
-  references: null,
-  error: null,
-  loading: false,
+  total: PropTypes.number.isRequired,
+  references: PropTypes.instanceOf(List).isRequired,
+  error: ErrorPropType,
+  loading: PropTypes.bool.isRequired,
+  onPageDisplay: PropTypes.func.isRequired,
```
harunurhan: `onQueryChange` would be better

monaawi: right, missed that

backend/inspirehep/records/views.py
```diff
@@ -48,6 +55,31 @@ def get(self, pid, record):
...
+    @pass_record
+    def get(self, pid, record):
+        page = request.values.get("page", 1, type=int)
+        size = request.values.get("size", 25, type=int)
+
+        if page < 1 or size < 1:
+            abort(400)
+
+        if size > current_app.config["MAX_API_RESULTS"]:
+            raise MaxResultWindowRESTError()
```
ammirate: I am wondering if we actually want to throw an error to the web page or overwrite `size` with `current_app.config["MAX_API_RESULTS"]`, WDYT? 
Same above.

monaawi: we want to throw an error to the web page because this is what we (invenio actually) are doing for other search endpoints. So we want to keep the behavior consistent for all

monaawi: Also, I think the API user should be informed that he isn't getting all the results he asked for

ammirate: Eventually, would be the same for the check `< 1`.

monaawi: I don't think I follow. What do you mean? Ofc it is easy to add a check if it exceeds the ```MAX_API_RESULTS``` and overwrite it but this isn't what we want since this is inconsistent with all our other search endpoints

ammirate: That sounds reasonable

PR #528, opened 2019-10-24
Fix small literature display

harunurhan:


ui/src/common/utils.js
```diff
@@ -189,3 +189,7 @@ export function getRecordIdFromRef($ref) {
   const parts = $ref.split('/');
   return parts[parts.length - 1];
 }
+
+export function requireOneOf(markup, dependencies) {
+  return dependencies.some(Boolean) ? markup : null;
```
drjova: Can we just do it as a component? Like ``ChildrenOrNone``?

ui/public/config.js
```diff
@@ -3,6 +3,7 @@
     REACT_APP_PIWIK_URL: null,
     REACT_APP_PIWIK_SITE_ID: null,
     REACT_APP_SENTRY_DSN: null,
+    FIGURES_FEATURE_FLAG: true,
```
drjova:
```suggestion
    FEATURE_FLAG_ENABLE_FIGURES: true,
```
wdyt?

PR #531, opened 2019-10-28
records: fix missing ``_bucket`` on update

drjova:
* While migrating, records don't have ``bucket`` hence the
  ``_bucket`` field which is required from invenio is missing
  and never been added.

* Sentry HEP-6JC


backend/inspirehep/records/api/literature.py
```diff
@@ -157,18 +157,28 @@ def update(
...
         if "deleted" in data and data["deleted"]:
+            LOGGER.info("Record is deleted", uuid=self.id)
             return data
 
         documents = data.pop("documents", [])
         figures = data.pop("figures", [])
 
+        self.pop("documents", None)
+        self.pop("figures", None)
+        self.delete_all_files()
```
pazembrz: What's the benefit of removing all files?

backend/inspirehep/records/api/literature.py
```diff
@@ -191,12 +201,10 @@ def add_files(self, data):
         if "figures" in builder.record:
             data["figures"] = builder.record["figures"]
 
-        if self.files:
-            for key in list(self.files.keys):
-                if key not in keys:
-                    del self.files[key]
         if "_files" in self:
             data["_files"] = copy(self["_files"])
+        if "_bucket" in self:
```
pazembrz: Isn't just easier here to do:
```suggestion
        data["_bucket"] = str("self.bucket")
```
It will also create bucket if it's missing.

backend/inspirehep/records/api/mixins.py
```diff
@@ -137,9 +147,13 @@ def files(self):
         bucket = self.get_bucket()
         return self.files_iter_cls(self, bucket=bucket, file_cls=self.file_cls)
 
+    def delete_all_files(self):
+        for key in list(self.files.keys):
+            del self.files[key]
```
pazembrz: I'm testing this locally now. When you do this, file is marked as removed and when you are trying to add it, it's pulled from local file object. Which is removed already. 
You should modify `get_file_object` method to query only for non deleted files.

backend/inspirehep/records/api/literature.py
drjova: my bad, that's wrong

backend/inspirehep/records/api/literature.py
drjova: I want to be sure I'm using the same bucket as the one used for adding the files hence it has updated the ``self``

drjova:
@pazembrz updated

backend/inspirehep/records/api/literature.py
```diff
@@ -157,18 +157,27 @@ def update(
...
         builder = LiteratureBuilder(record=data)
 
         for document in documents:
+            if document.get("hidden", False):
+                builder.add_document(**document)
+                continue
+
             file_data = self.add_file(document=True, **document)
             document.update(file_data)
             if "hidden" not in document:
```
pazembrz: Now you do not need this `if` here right?

backend/inspirehep/records/api/literature.py
pazembrz: Then if `_bucket` is not in self we should throw some exception here. right?

backend/inspirehep/records/api/mixins.py
pazembrz: I still believe that you should fix `get_file_object` method to query only for non deleted files.

drjova: I completely removed the method :/

backend/inspirehep/records/api/mixins.py
```diff
@@ -175,7 +197,7 @@ def add_local_file(self, key, bucket=None):
         file_object = self.get_file_object(key, bucket)
```
pazembrz: When file_object will be None don't you want to re-download file form original_url if it's possible?

pazembrz: You should also add test for re-adding deleted file (we talked about it)

backend/inspirehep/records/api/literature.py
drjova: it will throw a ``KeyError``, but this case I don't see how it can happen

backend/inspirehep/records/api/literature.py
drjova: removed

PR #542, opened 2019-10-31
ui: add common conference components

harunurhan:
* Also moves some components that will be use for conferences to common folder

INSPIR-2855


ui/src/common/components/__tests__/JournalInfo.test.jsx
```diff
@@ -0,0 +1,26 @@
...
+describe('JournalInfo', () => {
+  it('renders with only journal_title', () => {
+    const info = fromJS({
+      journal_title: 'Test Journal',
+    });
+    const wrapper = shallow(<JournalInfo info={info} />);
+    expect(wrapper).toMatchSnapshot();
+  });
+
+  it('renders with journal_title and all others fields', () => {
```
monaawi: other*

ui/src/conferences/components/ConferenceWebsitesAction.jsx
```diff
@@ -0,0 +1,46 @@
...
+    </Menu.Item>
+  );
+}
+
+function renderWebsiteAction(website, title) {
+  return <ExternalLink href={website.get('value')}>{title}</ExternalLink>;
+}
+
+const ACTION_TITLE = (
+  <Tooltip title="Conference website">
```
monaawi: in the mockup it has the link icon and the text ```websites``` next to it and not a tooltip. A tooltip would make it more consistent with what is in authors, I agree. But should we ask Stella too just to make sure?

ui/src/conferences/components/ProceedingsAction.jsx
```diff
@@ -0,0 +1,49 @@
...
+import ExternalLink from '../../common/components/ExternalLink';
+import ActionsDropdownOrAction from '../../common/components/ActionsDropdownOrAction';
+import IconText from '../../common/components/IconText';
+import JournalInfo from '../../common/components/JournalInfo';
+
+function renderProceedingsDropdownAction(proceeding, index) {
+  const href = proceeding.getIn(['record', '$ref']);
+  const publicationInfo = proceeding.getIn(['publication_info', 0], Map());
+  return (
+    <Menu.Item key={href}>
```
monaawi: what about the arXiv part? I thought that if there is an arXiv then we need to display that, otherwise display the publication info

ui/src/conferences/components/ProceedingsAction.jsx
```diff
@@ -0,0 +1,49 @@
...
+      <ExternalLink href={href}>
+        {publicationInfo.has('journal_title') ? (
+          <JournalInfo info={publicationInfo} />
+        ) : (
+          <span>Proceedings {index + 1}</span>
```
monaawi: should this be Proceeding or Proceedings?

ui/src/conferences/components/ConferenceWebsitesAction.jsx
StellaCh: We should have the text 'website' next to the icon to be consistent with all the literature search. We don't have tooltips in literature, but having it here is fine! We should consider adding tooltips in all buttons, as it's a way to help users.

ui/src/conferences/components/ProceedingsAction.jsx
StellaCh: After discussion, we decided to remove that part. there's an updated mockup in the issue

ui/src/conferences/components/ConferenceWebsitesAction.jsx
monaawi: so you would need to add the text 'website' as well

ui/src/conferences/components/ProceedingsAction.jsx
harunurhan: apparently `proceedingS`

ui/src/conferences/components/ConferenceWebsitesAction.jsx
harunurhan: I do disagree with idea of putting tooltips in general... The most important reason is that they don't work on mobile. Especially that, we are not trying more intuitive and obvious options for UI, because well, we have tooltips for that. It's alright for the Desktop users but then mobile UX suffers, since there is no explanation for the vague feature/functionality on mobile.

harunurhan: On another note, we have some sort of consistency with the actions... they have tooltip if they are only an icon, they don't when they have also text next the icon.

harunurhan:
@monaawi for conference date you're right, it was missing from the refinement too unfortunately.

PR #544, opened 2019-11-01
cite: Enable cite all for all users

monaawi:
* INSPIR-2884


monaawi:
Merge and deploy this only after remapping and reindexing prod

PR #545, opened 2019-11-01
citations: refactor CitationList pagination

monaawi:
* INSPIR-2889

### DIFF
--- a/ui/src/actions/citations.js
+++ b/ui/src/actions/citations.js
@@ -12,9 +12,10 @@ import {
 } from './actionTypes';
 import { httpErrorToActionPayload } from '../common/utils';
 
-function fetchingCitations() {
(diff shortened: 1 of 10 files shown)


ui/src/actions/citations.js
```diff
@@ -52,12 +53,17 @@ function fetchCitationsSummaryError(error) {
   };
 }
 
-export function fetchCitations(pidType, recordId, paginationOptions) {
+export function fetchCitations(pidType, recordId, newQuery = {}) {
```
harunurhan: since you're at it, could you also remove this `pidType` since it's only `literature`

monaawi: oh ok, I thought it was like that because we foresee there will be something other than literature later

PR #546, opened 2019-11-04
files: Quick fixes for files

pazembrz:
* Remove signal for Timestamp.before_update event for all invenio_files_rest.model models
* Add index for files_object table on key and is_head columns

### DIFF
--- a/backend/inspirehep/records/api/mixins.py
+++ b/backend/inspirehep/records/api/mixins.py
@@ -6,7 +6,12 @@ import requests
 import structlog
 from flask import current_app
 from invenio_db import db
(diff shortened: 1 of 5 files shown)


backend/inspirehep/records/api/mixins.py
```diff
@@ -16,6 +21,8 @@
 
 LOGGER = structlog.getLogger()
 
+db.event.remove(Timestamp, "before_update", timestamp_before_update)
```
michamos: add a test maybe?

pazembrz: Done

PR #547, opened 2019-11-06
mailing: don't use null CC from contact_details

ammirate:
* INSPIR-2886


backend/inspirehep/mailing/api/jobs.py
```diff
@@ -69,7 +69,7 @@ def send_job_deadline_reminder(job):
     cc_addresses = [
         cd.get("email")
         for cd in job.get("contact_details")
-        if cd.get("email") != recipient
+        if cd.get("email") and cd.get("email") != recipient
     ]
```
michamos:
``` 
   cc_addresses = [
        cd["email"]
        for cd in job.get("contact_details")
        if "email" in cd and cd["email"] != recipient
    ]

PR #548, opened 2019-11-07
Search UI Refactor

harunurhan:


ui/src/middlewares/searchDispatcher.js
```diff
@@ -1,68 +1,96 @@
...
-          (hasPathnameChanged || isFirstRendering)
-        ) {
-          dispatch(fetchSearchAggregationsForCurrentLocation(false));
-        }
+        dispatch(fetchSearchAggregationsForCurrentQuery(namespace));
       }
       return result;
     }
+
+    if (action.type === SEARCH_BASE_QUERIES_UPDATE) {
```
monaawi: what use cases do we have for the ```SEARCH_BASE_QUERIES_UPDATE```? Is it only used for displaying 100 results for catalogers and adding the author filter?

ui/src/actions/search.js
```diff
@@ -1,149 +1,179 @@
...
 import { UI_SERIALIZER_REQUEST_OPTIONS } from '../common/http';
 import { httpErrorToActionPayload } from '../common/utils';
+import { FETCH_MODE_NEVER, FETCH_MODE_INITIAL } from '../reducers/search';
 
-export function changeSearchScope(scope) {
-  return {
-    type: CHANGE_SEARCH_SCOPE,
-    payload: scope,
-  };
+function getQueryForCurrentState(namespace, state) {
```
monaawi: This is defined also later but for current state, maybe could be extracted somewhere and we pass the current state in the other case?

ui/src/actions/search.js
```diff
@@ -1,149 +1,179 @@
...
-    search,
-  } = state;
-  const baseQueryForCurrentSearchScope = search
-    .getIn(['scope', 'query'])
-    .toJS();
-  return stringify(
-    { ...baseQueryForCurrentSearchScope, ...location.query },
-    { indices: false }
-  );
+function getPathnameForNamespace(namespace, state) {
```
monaawi: ditto

ui/src/actions/search.js
```diff
@@ -1,149 +1,179 @@
...
+
+    const searchQuery =
+      aggregationsFetchMode === FETCH_MODE_INITIAL
+        ? {}
+        : getQueryForCurrentState(namespace, state);
+    const baseAggregationsQuery = state.search
+      .getIn(['namespaces', namespace, 'baseAggregationsQuery'])
+      .toJS();
+    const aggregationsQuery = {
+      ...searchQuery,
```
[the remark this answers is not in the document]
harunurhan: when it comes to `baseAggregationsQuery` we want to impose everything, to be sure. Perhaps we could call it differently `overrideAggregationsQuery`

ui/src/middlewares/searchDispatcher.js
```diff
@@ -1,68 +1,96 @@
...
-      const hasSearchChanged = nextLocation.search !== currentLocation.search;
-      if (isFirstRendering || hasPathnameChanged || hasSearchChanged) {
-        dispatch(searchForCurrentLocation());
+      const nextState = getState();
+      const nextQuery = getQueryFromState(namespace, nextState);
+      const baseQuery = getBaseQueryFromState(namespace, nextState);
+
+      const hasQueryChanged = prevQuery !== nextQuery;
+      // TODO: better name (this is basically to trigger search when initial location sth like /literature)
+      const isInitialQueryUpdateEmpty = nextQuery === baseQuery;
```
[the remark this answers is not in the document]
harunurhan: To dispatch search when you hit any search page without any query in the url  like `/literature`.

But just realized that I already broke while fixing another thing, I'll fix this first and later we can have a look.

ui/src/middlewares/searchDispatcher.js
harunurhan: it's for embedded searches only, so far we have only one.

ui/src/actions/search.js
harunurhan: it's only `getIn(['namespaces', namespace, 'query'])` basically, that's I just didn't wanna create common functions everywhere, and I created for places just to make the main functions that use this, a bit more readable.

But I am considering adding helpers for paths like `['namespaces', namespace, 'query']`

ui/src/authors/containers/__tests__/AuthorPublicationsContainer.test.jsx
```diff
@@ -5,86 +5,41 @@ import { Provider } from 'react-redux';
...
-import {
-  AUTHOR_PUBLICATIONS_REQUEST,
-  AUTHOR_PUBLICATIONS_FACETS_REQUEST,
-} from '../../../actions/actionTypes';
+import LiteratureSearchContainer from '../../../literature/containers/LiteratureSearchContainer';
+import { AUTHOR_PUBLICATIONS_NS } from '../../../reducers/search';
 
 describe('AuthorPublicationsContainer', () => {
-  it('passes all props to embedded search', () => {
+  it('passes all props to LiteratureSarchContainer', () => {
```
monaawi: LiteratureSearch*

ui/src/literature/components/CiteAllAction.jsx
```diff
@@ -24,21 +22,18 @@ class CiteAllAction extends Component {
   async onCiteClick({ key }) {
     const { query } = this.props;
     const citeQuery = {
-      // set default `sort` in case there is no `sort` in the query
-      sort: searchScopes.getIn(['literature', 'query', 'sort']),
-      ...omit(query, ['size', 'page']),
+      ...query,
```
monaawi: this is wrong, it should omit the page

ui/src/literature/containers/SearchPage.jsx
```diff
@@ -0,0 +1,22 @@
+import React from 'react';
+import { Row, Col } from 'antd';
+
+import LiteratureSearchContainer from './LiteratureSearchContainer';
+import { LITERATURE_NS } from '../../reducers/search';
+import DocumentHead from '../../common/components/DocumentHead';
+
+// TODO: move it out from containers
```
monaawi: Maybe better to move it out now?

ui/src/middlewares/searchDispatcher.js
```diff
@@ -1,68 +1,109 @@
...
+      const hasQueryQParamChanged = query.q && query.q !== stateQuery.get('q');
+      if (hasQueryQParamChanged) {
+        dispatch(newSearch(namespace));
       }
-      // TODO: change the shallowEqual once we make router immutable. This works for now because we generate the query from scratch everytime
-      const hasQueryChangedExceptSortOrPage = !shallowEqual(
-        omit(currentLocation.query, ['sort', 'page']),
-        omit(nextLocation.query, ['sort', 'page'])
+
+      // get prevQuery after newSearch because it might update the g
```
monaawi: q?

ui/src/middlewares/searchDispatcher.js
```diff
@@ -1,68 +1,109 @@
...
-      const hasSearchChanged = nextLocation.search !== currentLocation.search;
-      if (isFirstRendering || hasPathnameChanged || hasSearchChanged) {
-        dispatch(searchForCurrentLocation());
+    if (action.type === SEARCH_QUERY_UPDATE) {
+      const { namespace, query } = action.payload;
+
+      const stateQuery = getQueryFromState(namespace, getState());
+      const hasQueryQParamChanged = query.q && query.q !== stateQuery.get('q');
+      if (hasQueryQParamChanged) {
+        dispatch(newSearch(namespace));
```
monaawi: newSearch naming was a bit confusing for me, maybe refreshSearch is better? or resetSearch?

ui/src/reducers/search.js
```diff
@@ -5,104 +5,203 @@ import {
...
+export const FETCH_MODE_ALWAYS = 'always';
+export const FETCH_MODE_INITIAL = 'only-initial-without-query';
+
+export const initialState = fromJS({
+  searchBoxNamespace: LITERATURE_NS,
+  namespaces: {
+    [LITERATURE_NS]: {
+      ...initialNamespaceState,
+      pathname: LITERATURE,
+      embedded: false,
```
monaawi: maybe embedded could go as false by default?

ui/src/reducers/search.js
```diff
@@ -5,104 +5,203 @@ import {
...
+      baseAggregationsQuery: {
+        facet_name: 'hep-author-publication',
+      },
+    },
+  },
 });
+// TODO: maybe move all static things into config so that we can easily add new collection
+// from single, less complicated file.
+
+// TODO: reuse embedded=true in the namepsace to generate this dynamically
```
monaawi: I think it would be good if we had this. Otherwise, you could easily forget to change this

ui/src/reducers/search.js
```diff
@@ -5,104 +5,203 @@ import {
...
+export const SEARCHABLE_COLLECTION_PATHNAMES = [LITERATURE, JOBS, AUTHORS];
 
-function getSearchScopeForLocationChangeAction(action) {
+export const SEARCH_BOX_NAMESPACES = [LITERATURE_NS, AUTHORS_NS, JOBS_NS];
```
monaawi: This can also be generated dynamically

ui/src/reducers/search.js
```diff
@@ -5,104 +5,203 @@ import {
...
     case SEARCH_AGGREGATIONS_SUCCESS:
-      if (state.get('initialAggregations').isEmpty()) {
+      if (
+        state.getIn(['namespaces', namespace, 'initialAggregations']).isEmpty()
+      ) {
         // eslint-disable-next-line no-param-reassign
-        state = state.set(
-          'initialAggregations',
-          fromJS(action.payload.aggregations)
+        state = state.setIn(
```
monaawi: I don't get it, why are initialAggregations needed?

monaawi:
I also found another bug: when you type something in the search box and then search, then select a facet. Then remove what you had typed in the search box and search again, the facets remain there

monaawi:
Something is off with the date facet as well. Again, if you search for something, then remove it from the searchbox and search again, the date facet doesn't get reset. It's working on labs so I don't think it is the same bug as the one in the sprint

monaawi:
Maybe an improvement: in the authorPublications, we redo the call for the facets if the page/sort change. Maybe this could be omitted there as well

monaawi:
Another bug: search for something, then go to the logo (you would expect the search to be refreshed), You get the same search that you had. Also, when you press search, it doesn't search anymore because the query didn't change

ui/src/authors/containers/DetailPageContainer.jsx
```diff
@@ -36,193 +30,158 @@ import AuthorWebsitesAction from '../components/AuthorWebsitesAction';
...
 
-    const query = publicationsQuery.toJS();
-    dispatch(fetchCitationSummary(query));
-    dispatch(fetchCitationsByYear(query));
+function DetailPage({ record, loading, dispatch, match, publicationsQuery }) {
+  useEffect(
+    () => {
+      dispatch(fetchAuthor(match.params.id));
+      dispatch(newSearch(AUTHOR_PUBLICATIONS_NS));
+      // window.scrollTo(0, 0);
```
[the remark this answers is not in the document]
harunurhan: actually not, it should be uncommented.

ui/src/authors/containers/DetailPageContainer.jsx
```diff
@@ -36,193 +30,158 @@ import AuthorWebsitesAction from '../components/AuthorWebsitesAction';
...
+    },
+    [dispatch, match.params.id]
+  );
+
+  const authorFacetName = publicationsQuery.getIn(['author', 0]);
```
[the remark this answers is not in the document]
harunurhan: this is only to check if author facet name is pushed to the publications query already, so we don't care about the order.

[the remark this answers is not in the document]
harunurhan: could be even `publicationsQuery.has('author')`

ui/src/reducers/search.js
```diff
@@ -5,104 +5,203 @@ import {
...
   loading: false,
   total: 0,
-  scope: searchScopes.get('literature'),
   error: null,
   sortOptions: null,
   aggregations: {},
   initialAggregations: {},
   loadingAggregations: false,
   aggregationsError: null,
+  baseQuery: initialBaseQuery,
```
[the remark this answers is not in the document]
harunurhan: it's needed in multiple places that's why it's like that :/

ui/src/authors/containers/DetailPageContainer.jsx
```diff
@@ -36,193 +30,158 @@ import AuthorWebsitesAction from '../components/AuthorWebsitesAction';
...
+  const experiments = metadata.get('project_membership');
+
+  const twitter = metadata.get('twitter');
+  const linkedin = metadata.get('linkedin');
+  const urls = metadata.get('urls');
+  const orcid = metadata.get('orcid');
+  const emails = metadata.get('email_addresses');
+  return (
+    <>
+      {name && <DocumentHead title={getAuthorDisplayName(name)} />}
```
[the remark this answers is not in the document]
harunurhan: it was needed because I was trying to fix another issue too, but ended up not fixing it, so this isn't needed anymore.

ui/src/literature/components/CiteAllAction.jsx
harunurhan: I don't get why it would be wrong?, `page=1` is already a default on the backend.

ui/src/literature/containers/SearchPage.jsx
harunurhan: nope

ui/src/middlewares/searchDispatcher.js
harunurhan: I agree it was like that before, if there is time I'll rename it.

ui/src/reducers/search.js
harunurhan: I want it to be explicitly state for every namespace.

ui/src/reducers/search.js
harunurhan: it existed before. for now it's only need for literature date facet.

ui/src/literature/components/CiteAllAction.jsx
monaawi: ah, missed the fact that you added ```page=1``` after it. Thought if you were on page 2 and pressed on ```citeAll`` it won't give results, my bad

PR #549, opened 2019-11-07
serializers: add serializers for conferences

pazembrz:
Little cleanup in the serializers code

* INSPIR-2850

### DIFF
--- a/backend/inspirehep/records/api/conferences.py
+++ b/backend/inspirehep/records/api/conferences.py
@@ -4,8 +4,12 @@
 #
 # inspirehep is free software; you can redistribute it and/or modify it under
 # the terms of the MIT License; see LICENSE file for more details.
+
 from inspirehep.records.marshmallow.conferences import ConferencesElasticSearchSchema
-from inspirehep.records.models import ConferenceLiterature
+from inspirehep.records.models import (
+    ConferenceLiterature,
+    ConferenceToLiteratureRelationshipType,
+)
 
 from ...pidstore.api import PidStoreConferences
 from .base import InspireRecord
@@ -28,3 +32,25 @@ class ConferencesRecord(InspireRecord):
     def hard_delete(self):
         self.delete_relations_with_literature()
         super().hard_delete()
+
+    @property
+    def linked_papers_query(self):
+        return ConferenceLiterature.query.filter_by(conference_uuid=self.id)
+
+    @property
+    def number_of_contributions(self):
+        return self.linked_papers_query.filter_by(
+            relationship_type=ConferenceToLiteratureRelationshipType.conference_paper
+        ).count()
+
+    @property
+    def proceedings(self):
+        proceeding_records = self.linked_papers_query.filter_by(
+            relationship_type=ConferenceToLiteratureRelationshipType.proceedings
+        ).all()
+        proceedings = []
+        for record in proceeding_records:
+            # assuming there is only one publication_info element in record
+            data = record.conference_document.json
+            proceedings.append(data)
+        return proceedings
--- a/backend/inspirehep/records/api/mixins.py
+++ b/backend/inspirehep/records/api/mixins.py
@@ -14,7 +14,7 @@ from invenio_files_rest.models import (
 )
 from invenio_records.errors import MissingModelError
 from invenio_records_files.models import RecordsBuckets
-from sqlalchemy import func
+from sqlalchemy import func, or_
 
 from inspirehep.records.errors import DownloadFileError
 from inspirehep.records.models import (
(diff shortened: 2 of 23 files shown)


backend/inspirehep/records/indexer/tasks.py
```diff
@@ -52,8 +52,9 @@ def batch_index(self, records_uuids, request_timeout=None):
 
 
 def process_references_for_record(record):
-    """Tries to find differences in record references and forces to reindex
-    records which reference changed to update their citation statistics.
+    """Tries to find differences in record references
```
ammirate: Docstrings have a specific format: first line is a short description, then empty line and then a more verbose description on the purpose of the function :)

backend/inspirehep/records/api/conferences.py
```diff
@@ -28,3 +32,25 @@ def delete(self):
     def hard_delete(self):
         self.delete_relations_with_literature()
         super().hard_delete()
+
+    @property
+    def papers_query(self):
```
ammirate: `linked_papers_query` maybe? :)

backend/inspirehep/records/api/mixins.py
```diff
@@ -69,6 +69,15 @@ def citations_by_year(self):
         db_query = db_query.group_by("year").order_by("year")
         return [{"year": r.year.year, "count": r.sum} for r in db_query.all() if r.year]
 
+    def hard_delete(self):
+        with db.session.begin_nested():
+            LOGGER.warning("Hard Deleting citations")
+            # Removing citations from RecordCitations table
+            RecordCitations.query.filter_by(citer_id=self.id).delete()
+            # Removing references to this record from RecordCitations table
+            RecordCitations.query.filter_by(cited_id=self.id).delete()
```
ammirate: Here we could do something like `RecordCitations.query.filter_by(or_(citer_id=self.id, cited_id=self.id))`, WDYT?

backend/inspirehep/records/api/mixins.py
```diff
@@ -381,7 +390,50 @@ def update(self, data, disable_relations_update=False, *args, **kwargs):
...
+    def get_updated_referenced_conference(self):
+        try:
+            prev_version = self._previous_version
+        except AttributeError:
+            prev_version = {}
+
+        changed_deleted_status = self.get("deleted", False) ^ prev_version.get(
+            "deleted", False
+        )
+
```
ammirate: I would check `changed_deleted_status` already here and if `True` loads `pids_latest` and returns, without doing all the other operations

backend/inspirehep/records/api/mixins.py
```diff
@@ -381,7 +390,50 @@ def update(self, data, disable_relations_update=False, *args, **kwargs):
...
+        doc_type_previous = prev_version.get("document_type", [])
+        doc_type_latest = self.get("document_type", [])
+        doc_type_diff = set.symmetric_difference(
+            set(doc_type_previous), doc_type_latest
```
ammirate: Aren't they both supposed to be `set`?

backend/inspirehep/records/api/mixins.py
```diff
@@ -381,7 +390,50 @@ def update(self, data, disable_relations_update=False, *args, **kwargs):
             self.update_conference_paper_and_proccedings()
         else:
             LOGGER.info(
-                "Record conference papaers and proccedings update disabled",
+                "Record conference papers and proceedings update disabled",
                 recid=self.get("control_number"),
                 uuid=str(self.id),
             )
+
+    def get_updated_referenced_conference(self):
```
ammirate: This name is not very intuitive... I would put something like `get_newest_linked_conferences_uuid`. Adding a docstring would also help a lot IMO

backend/tests/integration/records/api/test_api_conferences.py
```diff
@@ -206,3 +205,73 @@ def test_hard_delete_conference_clears_entries_in_conference_literature_table(
...
+
+    expected_contributions_number = 1
+    assert expected_contributions_number == conference.number_of_contributions
+
+    rec2 = create_record("lit", rec_data)
+
+    expected_contributions_number = 2
+    assert expected_contributions_number == conference.number_of_contributions
+
+    rec1.hard_delete()
```
ammirate: Nice that we test this, but simple delete is a more common use case :)

backend/tests/integration/records/api/test_api_conferences.py
```diff
@@ -206,3 +205,73 @@ def test_hard_delete_conference_clears_entries_in_conference_literature_table(
...
+    rec2 = create_record("lit", rec_data)
+
+    expected_proceedings_count = 2
+    expected_proceedings = [dict(rec1), dict(rec2)]
+    assert expected_proceedings_count == len(conference.proceedings)
+    assert expected_proceedings == sorted(
+        conference.proceedings, key=itemgetter("control_number")
+    )
+
+    rec1.hard_delete()
```
ammirate: Same here

backend/tests/unit/records/marshmallow/literature/common/test_publication_info_item.py
```diff
@@ -40,6 +42,7 @@ def test_returns_non_empty_fields_if_pubinfo_freetext_present():
     assert expected == json.loads(result)
 
 
+@pytest.mark.xfail(reason="It's not a correct behaviour anymore.")
```
harunurhan: why not remove this?

pazembrz: As I do not have any requirement or written proof that this is actually not correct.  ;)

ammirate: as discussed IRL, this is the correct behavior.

PR #550, opened 2019-11-07
ui: conferences detail page

miguelgrc:
* INSPIR-2856


ui/ui-tests/__api_recordings__/ConferenceDetail_2344640018/recording.har
```diff
@@ -0,0 +1,113 @@
+{
```
harunurhan: this is redundant at this point because tests are `xfail`

ui/src/conferences/containers/DetailPageContainer.jsx
```diff
@@ -1,11 +1,147 @@
...
+              <Col>
+                <ConferenceDates
+                  openingDate={openingDate}
+                  closingDate={closingDate}
+                />
+                {location && (
+                  <>
+                    {'. '}
+                    {location.getIn(['cities', 0])}
+                    {', '}
```
harunurhan: how about when it doesn't have a `city` or `country`?

harunurhan: maybe it makes sense to move this another component `ConferenceAddress` and implement all this logic there and test separately.

miguelgrc: I think all the records must have `city` and `country`, right? At least according to the required fields in the submission form task.
Also, now we know that `cnum` is required, so it's not really necessary to check that in line 91 I guess.

ui/src/conferences/components/ConferenceLocation.jsx
```diff
@@ -0,0 +1,22 @@
...
+import PropTypes from 'prop-types';
+import { Map } from 'immutable';
+
+function ConferenceLocation({ location }) {
+  const city = location.getIn(['cities', 0]);
+  const country = location.get('country');
+  const renderSeparator = city && country;
+  return (
+    <span>
+      {city && city}
```
harunurhan: just `city` is enough

ui/src/conferences/components/ConferenceLocation.jsx
```diff
@@ -0,0 +1,22 @@
...
+      {renderSeparator && ', '}
+      {country && country}
```
harunurhan: ditto

PR #552, opened 2019-11-08
migrator: don't download file if available locally

michamos:
* INSPIR-2901


michamos:
I plan on running something like this to warm the cache:

pazembrz:
2 things to fix:
* ./inspirehep/migrator/utils.py:90:1: W293 blank line contains whitespace
* remove old recordings and rerun test locally to record new ones.

PR #554, opened 2019-11-12
bibtex: Fix error when serializing broken data

pazembrz:
INSPIR-2843

### DIFF
--- a/backend/inspirehep/records/serializers/bibtex.py
+++ b/backend/inspirehep/records/serializers/bibtex.py
@@ -4,13 +4,15 @@
 #
 # inspirehep is free software; you can redistribute it and/or modify it under
 # the terms of the MIT License; see LICENSE file for more details.
-
+import structlog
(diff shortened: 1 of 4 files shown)


backend/inspirehep/records/serializers/bibtex.py
```diff
@@ -120,18 +122,20 @@ def create_bibliography_entry(self, record):
...
+        bib_data = BibliographyData(data)
         writer = BibtexWriter()
         return writer.to_string(bib_data)
 
     def serialize(self, pid, record, links_factory=None):
-        return self.create_bibliography([record])
+        try:
+            return self.create_bibliography(record)
+        except Exception as e:
+            LOGGER.exception("Bibtex serialization error", error=e)
```
ammirate: I would add the recid as well

pazembrz: :+1:

PR #557, opened 2019-11-13
Improve logging

michamos:
* adds more context when logging in Celery tasks (task ID and task name) and Flask requests (request path)
* overrides the default Celery logger so it doesn't prepend useless stuff and we get straight JSON
* changes the labels used on the prometheus exporter to have metrics grouped by `url_rule` rather than `path` (i.e. `/literature/<pid_value>/` instead of `/literature/4512`).


backend/inspirehep/logger/ext.py
```diff
@@ -17,6 +18,22 @@
...
+def remove_request_logging(*args, **kwargs):
+    structlog.threadlocal.clear_threadlocal()
+
+
+def url_rule(request):
+    # Pretty-printing adapted from werkzeug.routing.Route.__repr__
+    return "".join(
+        f"<{data}>" if is_dynamic else data
+        for (is_dynamic, data) in request.url_rule._trace
+    ).lstrip("|")
```
michamos: this uses the private `_trace` property, otherwise you get ugly stuff like the value in https://github.com/inspirehep/inspirehep/blob/611756c422e5c906d87ab0abd19b0fa521aa3d19/backend/inspirehep/records/config.py#L94

backend/inspirehep/logger/config.py
```diff
@@ -55,3 +57,19 @@
...
+# Celery logging
+# ==============
+@task_prerun.connect()
+def log_task_context(sender, task_id, task, *args, **kwargs):
+    structlog.threadlocal.bind_threadlocal(task_id=task_id, task=task.name)
+
+
+@task_postrun.connect()
+def remove_task_context_logging(*args, **kwargs):
+    structlog.threadlocal.clear_threadlocal()
```
michamos: @benjamin-bergia we could also export Celery metrics from workers directly by plugging into these signals

PR #558, opened 2019-11-14
pidstore: make doi case insensitive

pazembrz:
* This commit requires to remove dois from pidstore and records migration. See: [INSPIR-2925](https://its.cern.ch/jira/browse/INSPIR-2925)
* INSPIR-2822

### DIFF
--- a/backend/inspirehep/pidstore/minters/doi.py
+++ b/backend/inspirehep/pidstore/minters/doi.py
@@ -12,3 +12,10 @@ from .base import Minter
 class DoiMinter(Minter):
     pid_value_path = "dois.value"
     pid_type = "doi"
+
+    def create(self, pid_value):
+        return super().create(pid_value.lower())
+
+    def get_pid_values(self):
+        pid_values = {str(pid_value).lower() for pid_value in super().get_pid_values()}
+        return pid_values
--- a/backend/setup.py
+++ b/backend/setup.py
@@ -74,6 +74,10 @@ setup(
             "inspirehep_records = inspirehep.records:InspireRecords",
             "inspirehep_rt = inspirehep.rt:InspireRtExt",
         ],
+        "invenio_base.api_converters": [
+            "doipath = inspirehep.pidstore.converters:DOIPathConverter",
+            "arxivpath = inspirehep.pidstore.converters:ArXivPathConverter",
+        ],
         "invenio_jsonschemas.schemas": ["inspirehep_records_schemas = inspire_schemas"],
         "invenio_search.mappings": ["records = inspirehep.search.mappings"],
         "invenio_db.models": [
--- a/backend/tests/integration/pidstore/minters/test_minters_doi.py
+++ b/backend/tests/integration/pidstore/minters/test_minters_doi.py
@@ -108,3 +108,41 @@ def test_mitner_dois_missing_schema(base_app, db, es, create_record_factory):
 
     with pytest.raises(MissingSchema):
         DoiMinter.mint(record_id, record_data)
+
+
+def test_new_doi_converted_to_lowercase(base_app, db, create_record):
+    expected_doi = "10.1109/tpel.2019.2900393"
+    doi_value = "10.1109/TPEL.2019.2900393"
+    data = {"dois": [{"value": doi_value}]}
+    create_record("lit", data=data)
+    doi = PersistentIdentifier.query.filter_by(pid_type="doi").first()
+    assert expected_doi == doi.pid_value
+
+
+def test_adding_same_doi_different_case_raises_pid_already_exists(
+    base_app, db, create_record
+):
+    doi_value_1 = "10.1109/tpel.2019.2900393"
+    doi_value_2 = "10.1109/TPEL.2019.2900393"
+    data_1 = {"dois": [{"value": doi_value_1}]}
+    data_2 = {"dois": [{"value": doi_value_2}]}
+    create_record("lit", data=data_1)
+    with pytest.raises(PIDAlreadyExists):
+        create_record("lit", data=data_2)
+
+
+def test_adding_record_with_duplicated_dois_different_case(base_app, db, create_record):
+    doi_value_1 = "10.1109/tpel.2019.2900393"
+    doi_value_2 = "10.1109/TPEL.2019.2900393"
+
+    expected_doi = "10.1109/tpel.2019.2900393"
+    expected_dois_count = 1
+
+    expected_dois_in_record = [{"value": doi_value_1}, {"value": doi_value_2}]
+    data = {"dois": [{"value": doi_value_1}, {"value": doi_value_2}]}
+    record = create_record("lit", data=data)
+    pids = PersistentIdentifier.query.filter_by(pid_type="doi").all()
+
+    assert record["dois"] == expected_dois_in_record
+    assert expected_dois_count == len(pids)
+    assert expected_doi == pids[0].pid_value
--- a/backend/tests/integration/pidstore/test_pidstore_api_base.py
+++ b/backend/tests/integration/pidstore/test_pidstore_api_base.py
@@ -21,3 +21,47 @@ def test_get_config_for_schema(appctx):
     pids_to_endpoints = PidStoreBase._get_config_pid_types_to_schema()
 
     assert pids_to_endpoints is not None
+
+
+def test_doi_resolve_ignores_case(app, api_client, create_record):
+    expected_response_code = 200
+    lowercase_doi = "10.1109/tpel.2019.2900393"
+    uppercase_doi = "10.1109/TPEL.2019.2900393"
+    data = {"dois": [{"value": uppercase_doi}]}
+    record = create_record("lit", data)
+    expected_control_number = record["control_number"]
+    response = api_client.get(f"api/doi/{uppercase_doi}")
+    assert expected_response_code == response.status_code
+    assert expected_control_number == response.json["metadata"]["control_number"]
+
+    response = api_client.get(f"api/doi/{lowercase_doi}")
+    assert expected_response_code == response.status_code
+    assert expected_control_number == response.json["metadata"]["control_number"]
+
+
+def test_arxiv_path_converter_works_for_all_arxiv_pid_values(
+    app, api_client, create_record
+):
+    data = {
+        "arxiv_eprints": [
+            {"value": "1607.06746", "categories": ["hep-th"]},
+            {"categories": ["hep-ph"], "value": "hep-ph/9709356"},
+        ]
+    }
+    record = create_record("lit", data)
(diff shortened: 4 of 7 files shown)


backend/setup.py
```diff
@@ -74,6 +74,10 @@
             "inspirehep_records = inspirehep.records:InspireRecords",
             "inspirehep_rt = inspirehep.rt:InspireRtExt",
         ],
+        "invenio_base.api_converters": [
+            "pid = inspirehep.pidstore.converters:DOIConverter",
+            "pidpath = inspirehep.pidstore.converters:DOIPathConverter",
```
michamos: won't this kick-in for all PIDs, not only DOIs?

pazembrz: From what i can understand we cannot assign more than one converter and path converter to pid/pidpath. So yes, it will affect all pids we use, but fortunately only doi's have letters.

backend/setup.py
```diff
@@ -74,6 +74,12 @@
             "inspirehep_records = inspirehep.records:InspireRecords",
             "inspirehep_rt = inspirehep.rt:InspireRtExt",
         ],
+        "invenio_base.api_converters": [
+            "doi = inspirehep.pidstore.converters:DOIConverter",
+            "doipath = inspirehep.pidstore.converters:DOIPathConverter",
+            "arxiv = invenio_records_rest.utils:PIDConverter",
+            "arxivpath = inspirehep.pidstore.converters:ArXivPathConverter",
```
michamos: Do you need to declare `doi` and `arxiv` here, given you're not using them (only the path version)?

backend/tests/integration-async/alembic/test_alembic_data_migrations.py
```diff
@@ -0,0 +1,78 @@
...
+    db.session.execute(
+        """
+        INSERT INTO public.pidstore_pid(
+        created, updated, id, pid_type, pid_value,
+        pid_provider, status, object_type, object_uuid
+        ) VALUES ('{0}','{1}', {2}, '{3}', '{4}','{5}', '{6}', '{7}', '{8}');
+        """.format(
+            *data
+        )
+    )
```
michamos: this is what `sqlalchemy.text` is for: https://docs.sqlalchemy.org/en/13/core/tutorial.html#using-text-fragments-inside-bigger-statements.

backend/tests/integration/pidstore/test_pidstore_api_base.py
```diff
@@ -21,3 +21,47 @@ def test_get_config_for_schema(appctx):
     pids_to_endpoints = PidStoreBase._get_config_pid_types_to_schema()
 
     assert pids_to_endpoints is not None
+
+
+def test_doi_resolve_ignores_case(app, api_client, create_record):
+    expected_response_code = 200
+    lowercase_doi = "10.1109/tpel.2019.2900393"
+    big_letters_doi = "10.1109/TPEL.2019.2900393"
```
michamos:
```suggestion
    uppercase_doi = "10.1109/TPEL.2019.2900393"
```

backend/tests/integration/pidstore/test_pidstore_api_base.py
```diff
@@ -21,3 +21,47 @@ def test_get_config_for_schema(appctx):
...
+    data = {"dois": [{"value": big_letters_doi}]}
+    record = create_record("lit", data)
+    expected_control_number = record["control_number"]
+    response = api_client.get(f"api/doi/{big_letters_doi}")
```
michamos:
```suggestion
    response = api_client.get(f"api/doi/{uppercase_doi}")
```

backend/inspirehep/alembic/d7965f995d30_convert_all_dois_to_lowercase.py
```diff
@@ -0,0 +1,29 @@
...
+revision = "d7965f995d30"
+down_revision = "f563233434cd"
+branch_labels = ()
+depends_on = None
+
+
+def upgrade():
+    """Upgrade database."""
+    op.execute(
+        "UPDATE pidstore_pid SET pid_value=LOWER(pid_value) WHERE pid_type='doi'"
```
michamos: this might blow up, in case we already have a different row with the same DOI.

backend/tests/integration/pidstore/test_pidstore_api_base.py
```diff
@@ -21,3 +21,47 @@ def test_get_config_for_schema(appctx):
     pids_to_endpoints = PidStoreBase._get_config_pid_types_to_schema()
 
     assert pids_to_endpoints is not None
+
+
+def test_doi_resolve_ignores_case(app, api_client, create_record):
+    expected_response_code = 200
+    lowercase_doi = "10.1109/tpel.2019.2900393"
+    big_letters_doi = "10.1109/TPEL.2019.2900393"
+    data = {"dois": [{"value": big_letters_doi}]}
```
michamos:
```suggestion
    data = {"dois": [{"value": uppercase_doi}]}
```

backend/tests/integration/pidstore/minters/test_minters_doi.py
```diff
@@ -108,3 +108,12 @@ def test_mitner_dois_missing_schema(base_app, db, es, create_record_factory):
...
         DoiMinter.mint(record_id, record_data)
+
+
+def test_new_doi_converted_to_lowercase(base_app, db, create_record):
+    expected_doi = "10.1109/tpel.2019.2900393"
+    doi_value = "10.1109/TPEL.2019.2900393"
+    data = {"dois": [{"value": doi_value}]}
+    create_record("lit", data=data)
+    doi = PersistentIdentifier.query.filter_by(pid_type="doi").first()
+    assert expected_doi == doi.pid_value
```
michamos: can you also add a test where you create a record with the same DOI twice in the record, but each time with different case? this might happen, and I'm not sure it's handled correctly.

backend/tests/integration/pidstore/minters/test_minters_doi.py
```diff
@@ -108,3 +108,24 @@ def test_mitner_dois_missing_schema(base_app, db, es, create_record_factory):
...
+def test_adding_same_doi_different_case_raises_pid_already_exists(
+    base_app, db, create_record
+):
+    doi_value_1 = "10.1109/tpel.2019.2900393"
+    doi_value_2 = "10.1109/TPEL.2019.2900393"
+    data_1 = {"dois": [{"value": doi_value_1}]}
+    data_2 = {"dois": [{"value": doi_value_2}]}
+    create_record("lit", data=data_1)
+    with pytest.raises(PIDAlreadyExists):
+        create_record("lit", data=data_2)
```
michamos: So I guessed right, this is a bug, it shouldn't raise but create the PID only once and ignore the second one.

backend/inspirehep/alembic/d7965f995d30_convert_all_dois_to_lowercase.py
```diff
@@ -0,0 +1,37 @@
...
+        """
+        UPDATE pidstore_pid SET pid_value=LOWER(pid_value)
+        WHERE pid_type='doi' AND NOT EXISTS(
+          SELECT 1
+          FROM pidstore_pid pid
+          WHERE pid.pid_type=pidstore_pid.pid_type AND
+          pid.pid_value=LOWER(pidstore_pid.pid_value)
+        )
+        """
+    )
```
michamos: not sure this will work properly, we can discuss it.

backend/tests/integration/pidstore/minters/test_minters_doi.py
pazembrz: Why? It's illogical to not raise here if you are trying to create second time the same pid (because in this case it's the same pid as it should be now case insensitive)

pazembrz: :+1:

backend/inspirehep/pidstore/minters/doi.py
```diff
@@ -12,3 +12,10 @@
 class DoiMinter(Minter):
     pid_value_path = "dois.value"
     pid_type = "doi"
+
+    def create(self, pid_value):
+        super().create(pid_value.lower())
```
michamos:
```suggestion
        return super().create(pid_value.lower())
```

backend/inspirehep/pidstore/minters/doi.py
```diff
@@ -12,3 +12,10 @@
...
+
+    def get_pid_values(self):
+        pid_values = [str(pid_value).lower() for pid_value in super().get_pid_values()]
+        return set(pid_values)
```
michamos: You can do a set comprehension directly:
```python
return {str(pid_value).lower() for pid_value in super.().get_pid_values()}
```

backend/tests/integration/pidstore/minters/test_minters_doi.py
```diff
@@ -108,3 +108,33 @@ def test_mitner_dois_missing_schema(base_app, db, es, create_record_factory):
...
+
+
+def test_adding_record_with_duplicated_dois_differen_case(base_app, db, create_record):
```
michamos:
```suggestion
def test_adding_record_with_duplicated_dois_different_case(base_app, db, create_record):
```

backend/tests/integration/pidstore/minters/test_minters_doi.py
```diff
@@ -108,3 +108,33 @@ def test_mitner_dois_missing_schema(base_app, db, es, create_record_factory):
...
+    doi_value_1 = "10.1109/tpel.2019.2900393"
+    doi_value_2 = "10.1109/TPEL.2019.2900393"
+    expected_dois = [{"value": doi_value_1}, {"value": doi_value_2}]
+    data = {"dois": [{"value": doi_value_1}, {"value": doi_value_2}]}
+    record = create_record("lit", data=data)
+    assert record["dois"] == expected_dois
```
michamos: maybe also assert that the DOI is in the pidstore?

backend/inspirehep/pidstore/minters/doi.py
pazembrz: :+1:

backend/inspirehep/pidstore/minters/doi.py
pazembrz: :+1:

backend/tests/integration/pidstore/minters/test_minters_doi.py
pazembrz: :+1:

backend/tests/integration/pidstore/minters/test_minters_doi.py
pazembrz: :+1:

PR #561, opened 2019-11-14
migrator: job record migration doesn't raise anymore

ammirate:
* INSPIR-2896

### DIFF
--- a/backend/inspirehep/migrator/tasks.py
+++ b/backend/inspirehep/migrator/tasks.py
@@ -19,10 +19,18 @@ from celery.result import AsyncResult
 from flask import current_app
 from flask_sqlalchemy import models_committed
 from inspire_dojson import marcxml2record
+from inspire_dojson.errors import NotSupportedError
 from invenio_db import db
 from invenio_pidstore.errors import PIDValueError
 from jsonschema import ValidationError
 
+from inspirehep.migrator.models import LegacyRecordsMirror
+from inspirehep.migrator.utils import (
+    cache_afs_file_locations,
+    ensure_valid_schema,
+    remove_cached_afs_file_locations,
+    replace_afs_file_locations_with_local,
+)
 from inspirehep.orcid.api import push_to_orcid
 from inspirehep.pidstore.api import PidStoreBase
 from inspirehep.records.api import InspireRecord, LiteratureRecord
@@ -30,14 +38,6 @@ from inspirehep.records.errors import DownloadFileError
 from inspirehep.records.indexer.tasks import batch_index
 from inspirehep.records.receivers import index_after_commit
 from inspirehep.records.tasks import update_records_relations
-
-from .models import LegacyRecordsMirror
-from .utils import (
-    cache_afs_file_locations,
-    ensure_valid_schema,
-    remove_cached_afs_file_locations,
-    replace_afs_file_locations_with_local,
-)
 
 LOGGER = structlog.getLogger()
 CHUNK_SIZE = 100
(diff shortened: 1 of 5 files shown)


backend/inspirehep/migrator/tasks.py
```diff
@@ -344,11 +343,16 @@ def migrate_record_from_mirror(
     logger = LOGGER.bind(recid=prod_record.recid)
     try:
         json_record = marcxml2record(prod_record.marcxml)
+    except NotImplementedError as exc:
```
jacquerie: This works because the `NotSupportedError` exception introduced in https://github.com/inspirehep/inspire-dojson/pull/261 inherits from `NotImplementedError`, but it doesn't follow the spirit of the exception translation pattern that you probably had in mind.

In fact, I'd argue that exception translation is an antipattern in a language without checked exceptions like Python. In Java, the compiler will check for you that any user of a function returning type `Foo` and throwing checked exception `Bar` does either catch the exception or throw it again. In Python, exceptions are not part of the signature (even when using any of the available typing mechanisms!), so you don't gain additional safety, but you have to write more boilerplate.

ammirate: That sounds very interesting but in this case I don't see any issue in applying polymorphism principles, it's one of the basics of OOP. 
Indeed, it works because `NotSupportedError` inherits from `NotImplementedError`, that's exactly the sole purpose of that class.

ammirate: On the other hand, being explicit and specific is another basic principle of Python. I don't think catching `NotImplementedError` will ever catch errors that are not `NotSupportedError`, but the devil is in the details so I guess explicit is better than implicit.

jacquerie: That's not the point I'm trying to make: all I'm saying is that the exception translation pattern works in Java because you have checked exceptions, but is less useful in Python because it doesn't have them.

Another example is https://github.com/inspirehep/inspirehep/blob/4d73b2073a9c2de64ef344e481d7357ba11a7c79/backend/inspirehep/records/views.py#L85-L103
If someone modifies `import_article` so that it raises a new `FooException` in certain cases, no type checker is going to complain about it. One possible solution would be to use a union type as the return value, or to just ignore the problem and hope for the best. I'm claiming that exception translation is an antipattern in Python because it achieves the worst of both worlds: you end up having to write a bunch of boilerplate types, but you don't get any additional type safety out of it.

backend/inspirehep/migrator/tasks.py
```diff
@@ -344,11 +344,16 @@ def migrate_record_from_mirror(
     logger = LOGGER.bind(recid=prod_record.recid)
     try:
         json_record = marcxml2record(prod_record.marcxml)
+    except NotSupportedError as exc:
+        logger.warning(str(exc), recid=prod_record.recid)
+        prod_record.valid = True
+        db.session.merge(prod_record)
+        return
```
michamos: `return None` as per PEP8 (the statement is not used only for control-flow, this is a function that actually returns a value, so all branches should return a value).

backend/inspirehep/migrator/tasks.py
```diff
@@ -344,11 +344,16 @@ def migrate_record_from_mirror(
     logger = LOGGER.bind(recid=prod_record.recid)
     try:
         json_record = marcxml2record(prod_record.marcxml)
+    except NotSupportedError as exc:
+        logger.warning(str(exc), recid=prod_record.recid)
```
michamos: Do we have guidelines about logging priority somewhere? to me this is an `info` rather than a `warning`.

ammirate: To mee too. Apparently, during our refinement people agreed that this is a warning, thus I am sticking to our requirements.

ammirate: If you are also for `info` than we can reevaluate it.

backend/inspirehep/migrator/tasks.py
```diff
@@ -344,11 +344,16 @@ def migrate_record_from_mirror(
...
+        prod_record.valid = True
+        db.session.merge(prod_record)
+        return
     except Exception as exc:
         logger.exception("Error converting from marcxml")
         prod_record.error = exc
         db.session.merge(prod_record)
-        return None
+        return
```
michamos: same

PR #562, opened 2019-11-14
ui: add message when 0 results in lit search

miguelgrc:
* INSPIR-2838

Also deleted a wild semicolon and replaced a few &nbsp for consistency


ui/src/common/components/EmptyOrChildren.jsx
```diff
@@ -28,8 +31,10 @@ EmptyOrChildren.propTypes = {
     PropTypes.object,
     PropTypes.instanceOf(List),
     PropTypes.instanceOf(Map),
+    PropTypes.number,
   ]),
   description: PropTypes.string,
+  extraInfo: PropTypes.node,
```
harunurhan: why not just use `description`?

ui/src/literature/containers/LiteratureSearchContainer.jsx
```diff
@@ -56,43 +58,53 @@ function LiteratureSearch({
...
-                renderItem={renderLiteratureItem}
-              />
-              <PaginationContainer namespace={namespace} />
-            </Col>
-          </Row>
+          <EmptyOrChildren
+            data={numberOfResults}
+            description="0 results"
+            extraInfo={
+              <div className="i">
```
harunurhan: why `italic`? 

also make sure you use semantic html tags, for this instead of css classes.

https://developer.mozilla.org/en-US/docs/Web/HTML/Element/em#Usage_notes

ui/src/common/components/EmptyOrChildren.jsx
miguelgrc: `description` is a `<p>` in ant's `<Empty>`, so you cannot put something like 
```html
<div>0 results</div>
<div>Oops! Your might want to check out our 
  <ExternalLink ... >search tips</ExternalLink>
</div>
```
inside it

ui/src/common/components/EmptyOrChildren.jsx
```diff
@@ -6,16 +6,19 @@ import { Empty } from 'antd';
 function isEmptyCollection(data) {
   return (
     data == null ||
-    Object.keys(data).length === 0 || // object/array
+    data <= 0 || // number
```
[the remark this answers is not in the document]
miguelgrc: Do you think it's better to do that? I'm not using `results` in `<LiteratureSearchContainer>` for anything else.

ui/src/common/components/__tests__/EmptyOrChildren.test.jsx
```diff
@@ -94,4 +94,16 @@ describe('EmptyOrChildren', () => {
     );
     expect(wrapper).toMatchSnapshot();
   });
+
+  it('does not render children and shows title and description', () => {
+    const data = null;
+    const title = 'title';
+    const description = 'description';
```
harunurhan: put `<div>description</div>` to be more clear, incase somebody reads the test cases as a source of documentation.

ui/src/common/components/__tests__/__snapshots__/LabelWithHelp.test.jsx.snap
```diff
@@ -3,7 +3,7 @@
 exports[`LabelWithHelp renders label with help 1`] = `
 <span>
   Label
-   
+
```
harunurhan: what is this?

ui/src/common/components/LabelWithHelp.jsx
```diff
@@ -7,8 +7,7 @@ class LabelWithHelp extends Component {
     const { help, label } = this.props;
     return (
       <span>
-        {label}&nbsp;
```
harunurhan: why are you changing those?

ui/src/literature/containers/LiteratureSearchContainer.jsx
```diff
@@ -2,6 +2,7 @@ import React, { useCallback, useEffect } from 'react';
 import PropTypes from 'prop-types';
 import { Row, Col } from 'antd';
 import { connect } from 'react-redux';
+import Immutable from 'immutable';
```
harunurhan: `import { List } from`

ui/src/common/components/LabelWithHelp.jsx
miguelgrc: We were using `{' '}` in some places and `&nbsp;` in others. I haven't found if there's a best practice for whitespaces in JSX, but we should stick to one of them.
(In this particular case we don't even need the explicit whitespace since both expressions can fit in the same line)

ui/src/common/components/__tests__/__snapshots__/LabelWithHelp.test.jsx.snap
miguelgrc: The previous one is an `&nbsp;` and the new one is a normal whitespace.

PR #563, opened 2019-11-14
Literature Submission bugs

harunurhan:


ui/src/submissions/literature/components/ArticleForm.jsx
```diff
@@ -65,7 +66,7 @@ class ArticleForm extends Component {
             />
             <Field name="volume" label="Volume" component={TextField} />
             <Field name="issue" label="Issue" component={TextField} />
-            <Field name="year" label="Year" component={TextField} />
+            <Field name="year" label="Year" component={NumberField} />
```
drjova: do we have a tests for these value types?

ui/src/submissions/literature/components/LiteratureSubmission.jsx
```diff
@@ -37,60 +43,75 @@ const FORMS_BY_DOC_TYPE = {
...
-    this.onFormikSubmit = this.onFormikSubmit.bind(this);
-  }
+  const [isInitilValid, setInitialValid] = useState(false);
+  useEffect(
+    () => {
+      const validate = async () => {
+        setInitialValid(await schema.isValid(initialValues));
+      };
+      validate();
+    },
```
drjova: mann it's too complicated to read :(

harunurhan: yeah, I agree, `async` calls in effect hook is pretty weird. I'll create a custom hook for this, to simplify here.

ui/src/submissions/literature/components/ArticleForm.jsx
harunurhan: tested via `e2e`, test is probably not filling the year field.

harunurhan: it's visible in screenshot though, that year is a number field.

drjova: can we add the year field?

harunurhan:
@drjova addressed the comments

PR #565, opened 2019-11-19
conftest: remove commit to not generate more sql statements at test ends

pazembrz:
* Don't destroy db, just clean tables.


backend/inspirehep/records/api/mixins.py
```diff
@@ -373,6 +374,16 @@ def create_conferences_relations(self, document_type):
                         ),
                     )
                 )
+        if len(conference_literature_relations_waiting_for_commit) > 0:
+            db.session.bulk_save_objects(
+                conference_literature_relations_waiting_for_commit
+            )
+        LOGGER.info(
+            "Conferecnce documents references set",
```
monaawi: should this be inside the if? Also, maybe something like ```Added conference-literature relations```

pazembrz: :+1:

PR #566, opened 2019-11-19
conferences: auth update view via session

harunurhan:
* PUT is only allowed if request has session of a cataloger.

* Also fixes login_required_with_roles to return `403` when user is
logged in but do not have access for the resource.

INSPIR-2854

### DIFF
--- a/backend/tests/integration/records/views/test_views_conferences.py
+++ b/backend/tests/integration/records/views/test_views_conferences.py
@@ -4,6 +4,8 @@
 #
 # inspirehep is free software; you can redistribute it and/or modify it under
 # the terms of the MIT License; see LICENSE file for more details.
+import json
+
 from invenio_accounts.testutils import login_user_via_session
 
 from inspirehep.accounts.roles import Roles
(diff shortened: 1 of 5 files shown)


backend/tests/integration/records/views/test_views_conferences.py
```diff
@@ -115,3 +106,56 @@ def test_conferences_contribution_filters(api_client, db, es, create_record):
...
+
+
+def test_conferences_application_json_put_without_auth(
+    api_client, db, es_clear, create_record
+):
+    record = create_record("con")
+    record_control_number = record["control_number"]
+
+    expected_status_code = 401
+    response = api_client.put("/conferences/{}".format(record_control_number))
```
monaawi: use f-strings instead of format

backend/tests/integration/records/views/test_views_conferences.py
```diff
@@ -115,3 +106,56 @@ def test_conferences_contribution_filters(api_client, db, es, create_record):
...
+    api_client, db, es_clear, create_user, create_record
+):
+    user = create_user()
+    login_user_via_session(api_client, email=user.email)
+
+    record = create_record("con")
+    record_control_number = record["control_number"]
+
+    expected_status_code = 403
+    response = api_client.put("/conferences/{}".format(record_control_number))
```
monaawi: ditto

PR #569, opened 2019-11-19
migrator: Trying to fix few errors on migration

pazembrz:
* INSPIR-2923

### DIFF
--- a/backend/inspirehep/migrator/tasks.py
+++ b/backend/inspirehep/migrator/tasks.py
@@ -229,15 +229,19 @@ def create_records_from_mirror_recids(recids):
                 record = migrate_record_from_mirror(
                     LegacyRecordsMirror.query.get(recid)
                 )
+            if record:
+                processed_records.add(str(record.id))
+            else:
+                LOGGER.warning("Record is empty", recid=recid)
         except Exception:
             LOGGER.exception("Cannot migrate record", recid=recid)
             continue
-
-        if record:
-            processed_records.add(str(record.id))
-        else:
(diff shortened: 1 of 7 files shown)


backend/inspirehep/migrator/tasks.py
```diff
@@ -200,6 +201,20 @@ def migrate_recids_from_mirror(
...
+    task_id=None, exception=None, args=None, kwargs=None, traceback=None, einfo=None
+):
+    LOGGER.error(
+        "One of the tasks in chord just failed.",
+        task_id=task_id,
+        exception=exception,
+        task_args=args,
+        task_kwargs=kwargs,
+        traceback=traceback,
+    )
```
michamos: This hooks into the signal far all tasks, so the message should be more generic, like `Celery task failed`, and it should be moved to the `logger.config` module with the other celery logging hooks.

backend/inspirehep/migrator/tasks.py
```diff
@@ -200,6 +201,20 @@ def migrate_recids_from_mirror(
     return str(chord_task.id)
 
 
+@task_failure.connect
+def log_error(
+    task_id=None, exception=None, args=None, kwargs=None, traceback=None, einfo=None
+):
```
michamos: you should add `*args, **kwargs` in case celery decides to add extra args in the future.

backend/inspirehep/migrator/tasks.py
michamos: better to set `exc_info=(type(exception), exception, traceback)` so it gets interpreted correctly.

PR #570, opened 2019-11-19
conferences: search

monaawi:
* Added search  conferences
* INSPIR-2857

### DIFF
--- a/ui/src/common/components/AggregationFilters.jsx
+++ b/ui/src/common/components/AggregationFilters.jsx
@@ -23,19 +23,23 @@ class AggregationFilters extends Component {
       inline,
       displayWhenNoResults,
       initialAggregations,
+      staticAggregations,
     } = this.props;
     const rowClassName = classnames('bg-white', {
       pa3: !inline,
       pv3: inline,
     });
+    const aggregationsToDisplay = aggregations
+      .entrySeq()
+      .filter(([, aggregation]) => aggregation.get('buckets').size > 0)
+      .sort(AggregationFilters.compareAggregationEntries);
     return (
       aggregations &&
       (numberOfResults > 0 || displayWhenNoResults) && (
         <Row className={rowClassName} type="flex" justify="space-between">
-          {aggregations
+          {staticAggregations
             .entrySeq()
-            .filter(([, aggregation]) => aggregation.get('buckets').size > 0)
-            .sort(AggregationFilters.compareAggregationEntries)
+            .concat(aggregationsToDisplay)
             .map(([aggregationKey, aggregation]) => (
               <Col
                 className={classnames({ 'md-pb3': inline })}
@@ -76,6 +80,7 @@ AggregationFilters.propTypes = {
   inline: PropTypes.bool,
   onAggregationChange: PropTypes.func.isRequired,
   aggregations: PropTypes.instanceOf(Immutable.Map).isRequired,
+  staticAggregations: PropTypes.instanceOf(Immutable.Map),
   initialAggregations: PropTypes.instanceOf(Immutable.Map).isRequired,
   query: PropTypes.objectOf(PropTypes.any).isRequired,
   numberOfResults: PropTypes.number.isRequired,
@@ -85,6 +90,7 @@ AggregationFilters.propTypes = {
 AggregationFilters.defaultProps = {
   inline: false,
   displayWhenNoResults: false,
+  staticAggregations: Immutable.Map(),
 };
 
 export default AggregationFilters;
--- a/ui/src/common/components/DateRangeAggregation.jsx
+++ b/ui/src/common/components/DateRangeAggregation.jsx
@@ -1,0 +1,109 @@
+import React, { useCallback, useMemo } from 'react';
+import PropTypes from 'prop-types';
+import moment from 'moment';
+import { DatePicker, Row } from 'antd';
+import AggregationBox from './AggregationBox';
+import {
+  DATE_RANGE_FORMAT,
+  RANGE_AGGREGATION_SELECTION_SEPARATOR,
+} from '../constants';
+
+function DateRangeAggregation({ name, onChange, selections }) {
+  const [startDate = '', endDate = ''] = useMemo(
+    () => selections.split(RANGE_AGGREGATION_SELECTION_SEPARATOR),
+    [selections]
+  );
+
+  const startDateMoment = useMemo(
+    () => (startDate ? moment.utc(startDate, DATE_RANGE_FORMAT) : null),
+    [startDate]
+  );
+
+  const endDateMoment = useMemo(
+    () => (endDate ? moment.utc(endDate, DATE_RANGE_FORMAT) : null),
+    [endDate]
+  );
+
+  const isLaterThanEndDate = useCallback(
+    dateAsMoment => {
+      if (!dateAsMoment || !endDateMoment) {
+        return false;
+      }
+      return dateAsMoment.isSameOrAfter(endDateMoment);
+    },
+    [endDateMoment]
+  );
+
+  const isEarlierThanStartDate = useCallback(
+    dateAsMoment => {
+      if (!dateAsMoment || !startDateMoment) {
+        return false;
+      }
+      return dateAsMoment.isSameOrBefore(startDateMoment);
+    },
+    [startDateMoment]
+  );
+
+  const onStartDateChange = useCallback(
+    (date, dateString) => {
+      const willHaveAnySelectedDate = dateString || endDate;
+      return onChange(
+        willHaveAnySelectedDate
+          ? `${dateString}${RANGE_AGGREGATION_SELECTION_SEPARATOR}${endDate}`
+          : undefined
+      );
+    },
+    [endDate, onChange]
+  );
+
+  const onEndDateChange = useCallback(
+    (date, dateString) => {
+      const willHaveAnySelectedDate = dateString || startDate;
+      return onChange(
+        willHaveAnySelectedDate
+          ? `${startDate}${RANGE_AGGREGATION_SELECTION_SEPARATOR}${dateString}`
+          : undefined
+      );
+    },
+    [startDate, onChange]
+  );
+
+  return (
+    <AggregationBox name={name}>
+      <Row className="mb1">From:</Row>
+      <Row className="mb2">
+        <DatePicker
+          data-test-id="start-date-picker"
+          format={DATE_RANGE_FORMAT}
+          className="w-100"
+          onChange={onStartDateChange}
+          value={startDateMoment}
+          disabledDate={isLaterThanEndDate}
+        />
+      </Row>
+      <Row className="mb1">To:</Row>
+      <Row>
+        <DatePicker
+          data-test-id="end-date-picker"
+          format={DATE_RANGE_FORMAT}
+          onChange={onEndDateChange}
+          className="w-100"
+          value={endDateMoment}
+          disabledDate={isEarlierThanStartDate}
+        />
+      </Row>
+    </AggregationBox>
+  );
+}
+
+DateRangeAggregation.propTypes = {
+  onChange: PropTypes.func.isRequired,
+  name: PropTypes.string.isRequired,
+  selections: PropTypes.string,
+};
+
+DateRangeAggregation.defaultProps = {
+  selections: '',
+};
+
+export default DateRangeAggregation;
--- a/ui/src/conferences/components/ConferenceItem.jsx
+++ b/ui/src/conferences/components/ConferenceItem.jsx
@@ -1,0 +1,96 @@
+import React, { Component } from 'react';
+import { Link } from 'react-router-dom';
+import PropTypes from 'prop-types';
+import { Map } from 'immutable';
+
+import { Row, Col } from 'antd';
+import EditRecordAction from '../../common/components/EditRecordAction';
+import ResultItem from '../../common/components/ResultItem';
+import { CONFERENCES } from '../../common/routes';
+import ConferenceTitle from './ConferenceTitle';
+import ConferenceDates from './ConferenceDates';
+import ConferenceLocation from './ConferenceLocation';
+import InspireCategoryList from './InspireCategoryList';
+import ConferenceWebsitesAction from './ConferenceWebsitesAction';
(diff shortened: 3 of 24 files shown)


ui/src/conferences/components/ConferenceWebsitesAction.jsx
```diff
@@ -22,11 +23,7 @@ function renderWebsiteAction(website, title) {
   return <ExternalLink href={website.get('value')}>{title}</ExternalLink>;
 }
 
-const ACTION_TITLE = (
-  <Tooltip title="Conference website">
```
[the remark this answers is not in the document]
monaawi: yes, from our discussion with Stella on the common components PR, I understood that we want websites text in front and that we would drop the tooltip

ui/src/conferences/components/ConferenceItem.jsx
```diff
@@ -0,0 +1,96 @@
...
+              openingDate={openingDate}
+              closingDate={closingDate}
+            />
+            {location && (
+              <>
+                {'. '}
+                <ConferenceLocation location={location} />
+              </>
+            )}
+            {cnum && ` (${cnum})`}
```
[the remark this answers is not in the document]
monaawi: It isn't in the requirements to display it only for catalogers :/ . It is also like this in the ```DetailPage```

ui/src/common/components/AggregationFilters.jsx
```diff
@@ -85,6 +88,7 @@ AggregationFilters.propTypes = {
 AggregationFilters.defaultProps = {
   inline: false,
   displayWhenNoResults: false,
+  staticAggregations: fromJS({}),
```
harunurhan: `Map()`

ui/src/common/components/DateRangeAggregation.jsx
```diff
@@ -0,0 +1,119 @@
...
+import moment from 'moment';
+import { DatePicker, Row } from 'antd';
+import AggregationBox from './AggregationBox';
+import {
+  DATE_RANGE_FORMAT,
+  RANGE_AGGREGATION_SELECTION_SEPARATOR,
+} from '../constants';
+
+function DateRangeAggregation({ name, onChange, selections }) {
+  const splitSelections = useMemo(
```
harunurhan: `const [startDate, endDate]`

ui/src/common/components/DateRangeAggregation.jsx
```diff
@@ -0,0 +1,119 @@
...
+    () => selections && selections.split(RANGE_AGGREGATION_SELECTION_SEPARATOR),
+    [selections]
+  );
+
+  const startDate = useMemo(
```
harunurhan: reduntant

harunurhan: also redundant use of `useMemo`. It should be used when you do "heavy" calculation or reguler one that returns always a different reference (which then causes rerendering if you pass it as prop to any child).

ui/src/common/components/DateRangeAggregation.jsx
```diff
@@ -0,0 +1,119 @@
...
+    () => (splitSelections && splitSelections[0]) || null,
+    [splitSelections]
+  );
+
+  const endDate = useMemo(
```
harunurhan: ditto

ui/src/common/components/DateRangeAggregation.jsx
```diff
@@ -0,0 +1,119 @@
...
+    () => startDate && moment.utc(startDate, DATE_RANGE_FORMAT),
+    [startDate]
+  );
+
+  const endDateMoment = useMemo(
+    () => endDate && moment.utc(endDate, DATE_RANGE_FORMAT),
+    [endDate]
+  );
+
+  const disabledStartDate = useCallback(
```
harunurhan: `isLaterThanEndDate`

ui/src/common/components/DateRangeAggregation.jsx
```diff
@@ -0,0 +1,119 @@
...
+    givenStartDate => {
+      if (!givenStartDate || !endDateMoment) {
+        return false;
+      }
+      return givenStartDate.valueOf() > endDateMoment.valueOf();
+    },
+    [endDateMoment]
+  );
+
+  const disabledEndDate = useCallback(
```
harunurhan: `isEarlierThanStartDate`

ui/src/common/components/DateRangeAggregation.jsx
```diff
@@ -0,0 +1,119 @@
...
+        return false;
+      }
+      return givenEndDate.valueOf() <= startDateMoment.valueOf();
+    },
+    [startDateMoment]
+  );
+
+  const onStartDateChange = useCallback(
+    (date, dateString) => {
+      const query = [dateString, endDate].join(
```
harunurhan: how about  ?
```js
`${startDate}${SEP...}${endDate}
```

ui/src/common/components/DateRangeAggregation.jsx
```diff
@@ -0,0 +1,119 @@
...
+      return onChange(
+        query !== RANGE_AGGREGATION_SELECTION_SEPARATOR ? query : undefined
+      );
+    },
+    [endDate, onChange]
+  );
+
+  const onEndDateChange = useCallback(
+    (date, dateString) => {
+      const query = [startDate, dateString].join(
```
harunurhan: ditto

ui/src/common/components/DateRangeAggregation.jsx
```diff
@@ -0,0 +1,119 @@
...
+    [startDateMoment]
+  );
+
+  const onStartDateChange = useCallback(
+    (date, dateString) => {
+      const query = [dateString, endDate].join(
+        RANGE_AGGREGATION_SELECTION_SEPARATOR
+      );
+      return onChange(
+        query !== RANGE_AGGREGATION_SELECTION_SEPARATOR ? query : undefined
```
harunurhan: this is a bit confusion, I would move out `const hasAnySelectedDate = startDate || endDate` and, here 
```js
const query  = hasAnyDateSelected ? `${startDate}${SEP...}${endDate}` : undefined
```

ui/src/common/components/DateRangeAggregation.jsx
```diff
@@ -0,0 +1,119 @@
...
+    [endDate, onChange]
+  );
+
+  const onEndDateChange = useCallback(
+    (date, dateString) => {
+      const query = [startDate, dateString].join(
+        RANGE_AGGREGATION_SELECTION_SEPARATOR
+      );
+      return onChange(
+        query !== RANGE_AGGREGATION_SELECTION_SEPARATOR ? query : undefined
```
harunurhan: ditto

ui/src/conferences/containers/SearchPageContainer.jsx
```diff
@@ -1,11 +1,111 @@
...
+
+function renderConferenceItem(result) {
+  return <ConferenceItem metadata={result.get('metadata')} />;
+}
+
+const STATIC_AGGREGATIONS = fromJS({
+  start_date: {
+    meta: {
+      title: 'Start Date',
+      order: 1,
```
harunurhan: I think having `1` here and starting `2` from the backend feels awkward if you read the backend code and if somebody was to add a new facet before `subject`, it's likely that they would just do `order: 1`. So maybe we can put this `0` or `-1` and start at `1` on the otherside. Or at least a comment there?

monaawi: or another idea would be to put the staticAggregation always first and not add order for them at all, wdyt? I did it like that initially because I thought maybe in the future, the staticAggregation would even be between 2 aggregations, but you are right, it is confusing for the backend and there is no need to have it like this yet.

harunurhan: I was thinking to make them always first, but you know how it goes.. we will probably need some of them last or sth like that in the future.

ui/src/common/components/DateRangeAggregation.jsx
```diff
@@ -0,0 +1,119 @@
...
+    () => endDate && moment.utc(endDate, DATE_RANGE_FORMAT),
+    [endDate]
+  );
+
+  const disabledStartDate = useCallback(
+    givenStartDate => {
+      if (!givenStartDate || !endDateMoment) {
+        return false;
+      }
+      return givenStartDate.valueOf() > endDateMoment.valueOf();
```
harunurhan: you can use `moment` functions like https://momentjs.com/docs/#/query/is-same-or-before/

ui/src/conferences/containers/SearchPageContainer.jsx
monaawi: I still say let's go always first for now. And if needed, we will change it later. Adding ```-1``` and things like that isn't too nice

ui/src/common/components/AggregationFilters.jsx
```diff
@@ -85,6 +93,7 @@ AggregationFilters.propTypes = {
 AggregationFilters.defaultProps = {
   inline: false,
   displayWhenNoResults: false,
+  staticAggregations: Map(),
```
harunurhan: `Immutable.Map()`, since `Immutable` is already imported

ui/src/common/components/DateRangeAggregation.jsx
```diff
@@ -0,0 +1,110 @@
...
+import AggregationBox from './AggregationBox';
+import {
+  DATE_RANGE_FORMAT,
+  RANGE_AGGREGATION_SELECTION_SEPARATOR,
+} from '../constants';
+
+function DateRangeAggregation({ name, onChange, selections }) {
+  const [startDate = '', endDate = ''] = useMemo(
+    () =>
+      selections ? selections.split(RANGE_AGGREGATION_SELECTION_SEPARATOR) : [],
```
harunurhan: it may be better to default selections `selections = ''`, instead of this check.

ui/src/common/components/DateRangeAggregation.jsx
```diff
@@ -0,0 +1,110 @@
...
+    [selections]
+  );
+
+  const startDateMoment = useMemo(
+    () => (startDate && moment.utc(startDate, DATE_RANGE_FORMAT)) || null,
```
harunurhan: `startDate ? moment.utc(startDate, DATE_RANGE_FORMAT)) : null` would be more readable (to me at least)

ui/src/common/components/DateRangeAggregation.jsx
```diff
@@ -0,0 +1,110 @@
...
+import moment from 'moment';
+import { DatePicker, Row } from 'antd';
+import AggregationBox from './AggregationBox';
+import {
+  DATE_RANGE_FORMAT,
+  RANGE_AGGREGATION_SELECTION_SEPARATOR,
+} from '../constants';
+
+function DateRangeAggregation({ name, onChange, selections }) {
+  const [startDate = '', endDate = ''] = useMemo(
```
harunurhan: why do we need defaults?

ui/src/common/components/DateRangeAggregation.jsx
```diff
@@ -0,0 +1,110 @@
...
+    [startDate]
+  );
+
+  const endDateMoment = useMemo(
+    () => (endDate && moment.utc(endDate, DATE_RANGE_FORMAT)) || null,
+    [endDate]
+  );
+
+  const isLaterThanEndDate = useCallback(
+    givenStartDate => {
```
harunurhan: this can be called `dateAsMoment`, `dateMoment` to indicated that it is a `moment` object, and since it's a generic callback doesn't necessarily need to take start date.

ui/src/common/components/DateRangeAggregation.jsx
```diff
@@ -0,0 +1,110 @@
...
+          ? `${dateString}${RANGE_AGGREGATION_SELECTION_SEPARATOR}${endDate}`
+          : undefined
+      );
+    },
+    [endDate, onChange]
+  );
+
+  const onEndDateChange = useCallback(
+    (date, dateString) => {
+      const hasAnySelectedDate = dateString || startDate;
```
harunurhan: `willHaveAnySelectedDate` maybe even more appropriate

ui/src/reducers/search.js
```diff
@@ -102,6 +108,24 @@ export const initialState = fromJS({
...
+      embedded: false,
+      baseQuery: {
+        ...initialBaseQuery,
+        start_date: `${moment
+          .utc()
+          .format(DATE_RANGE_FORMAT)}${RANGE_AGGREGATION_SELECTION_SEPARATOR}`,
+      },
+      query: {
+        ...initialBaseQuery,
+        start_date: `${moment
```
harunurhan: maybe you can pull this out `const today = moment.utc()...`

ui/src/common/components/AggregationFilters.jsx
```diff
@@ -32,10 +33,16 @@ class AggregationFilters extends Component {
       aggregations &&
       (numberOfResults > 0 || displayWhenNoResults) && (
         <Row className={rowClassName} type="flex" justify="space-between">
-          {aggregations
+          {staticAggregations
             .entrySeq()
-            .filter(([, aggregation]) => aggregation.get('buckets').size > 0)
-            .sort(AggregationFilters.compareAggregationEntries)
+            .concat(
+              aggregations
```
harunurhan: this nested calls are not so elegant you can do,

```
aggregations
 .entrySeq... 
 .filter...
 .sort...
 .splice(0, 0, ...staticAggregations.entrySeq())
```

`splice` have `O(N)` complexity, but since we have upto 5 items in this list, we don't care that much I think.

Otherwise I would throw the `order: 0 or -1`, instead of having these nested calls here.

monaawi: I can't use ```splice``` here because we```entrySeq``` gives a sequence and it doesn't have ```splice``` nor ```unshift```. But I still think this is better than having ```0``` or ```-1```. It's just a ```concat```

ui/src/common/components/DateRangeAggregation.jsx
monaawi: I added them to avoid checks for undefined in onStart/EndDateChange

PR #572, opened 2019-11-20
ui: display authors native names

monaawi:
* INSPIR-2927


ui/src/authors/components/AuthorResultItem.jsx
```diff
@@ -21,11 +21,12 @@ class AuthorResultItem extends Component {
     );
     const arxivCategories = metadata.get('arxiv_categories');
     const experiments = metadata.get('project_membership');
+    const nativeName = metadata.getIn(['name', 'native_names', 0]);
```
michamos: what if there's more than one native name?

StellaCh: we only allow 1 native name in the author form. Are there cases with more?

monaawi: the schema has a list of native_names so it allows for more than one to be there. But do we really want to display all native names in the title like that?

michamos: maybe not in the title, but in the detailed record we probably should.

monaawi: We checked and found only 6 authors out of 125k with more than one native name. After discussion with Stella, we will display only first native name

ui/src/authors/components/AuthorName.jsx
```diff
@@ -5,15 +5,21 @@ import { getAuthorDisplayName } from '../utils';
...
+        {displayName}
+        {nativeName && <span> ({nativeName})</span>}
+      </span>
+    );
   }
 }
 
 AuthorName.propTypes = {
   name: PropTypes.instanceOf(Map).isRequired,
+  nativeName: PropTypes.string,
```
harunurhan: `nativeName` is already inside `name` prop, so this isn't needed.

monaawi: right, missed this

ui/src/authors/components/__tests__/AuthorName.test.jsx
```diff
@@ -11,6 +11,15 @@ describe('AuthorName', () => {
     expect(wrapper).toMatchSnapshot();
   });
 
+  it('renders with native_name', () => {
+    const name = fromJS({
+      preferred_name: 'Harun Urhan',
+      native_names: ['赵新丽'],
```
harunurhan: I think my native name is misspelled little bit 😄

PR #575, opened 2019-11-20
tests: don't remove all tables, just remove content

pazembrz:
* don't remove indexes from ES only remove data
 * There is ES error on version 5, when you try to remove records using delete_by_query and record have external visioning and it's current version is 0 it's failing. So in this case only way is to remove index. invenio 1.1.1 allows to remove all indexes only. After upgrading to invenio 1.2 we will be able to remove only specified index

### DIFF
--- a/backend/tests/conftest.py
+++ b/backend/tests/conftest.py
@@ -32,3 +32,13 @@ def vcr_config():
         ),
         "record_mode": "once",
     }
+
+
+@pytest.fixture(scope="module")
+def es(appctx):
+    """Setup all registered Elasticsearch indices."""
+    from invenio_search import current_search, current_search_client
+
+    list(current_search.create(ignore=[400]))
+    list(current_search.put_templates(ignore=[400]))
+    yield current_search_client
--- a/backend/tests/helpers/cleanups.py
+++ b/backend/tests/helpers/cleanups.py
@@ -1,0 +1,46 @@
+# -*- coding: utf-8 -*-
+#
+# Copyright (C) 2019 CERN.
+#
+# inspirehep is free software; you can redistribute it and/or modify it under
+# the terms of the MIT License; see LICENSE file for more details.
+
+import elasticsearch
+from pytest_invenio.fixtures import _es_create_indexes, _es_delete_indexes
+from sqlalchemy_utils import create_database, database_exists
+
+
+def es_cleanup(es):
+    """Removes all data from es indexes
+
+    There is ES error on version 5, when you try to remove records
+    using delete_by_query and record have internal visioning and
+    it's current version is 0 it's failing. So in this case only way is to
+    remove index but invenio 1.1.1 allows to remove all indexes only
+    After upgrading to invenio 1.2 we will be able
+    to remove only specified index
+    """
+    from invenio_search import current_search, current_search_client
+
+    es.indices.refresh()
+    try:
+        for index in es.indices.stats()["indices"].keys():
+            es.delete_by_query(index, "{}")
+    except elasticsearch.exceptions.RequestError:
+        _es_delete_indexes(current_search)
+        _es_create_indexes(current_search, current_search_client)
+    es.indices.refresh()
+
+
+def db_cleanup(db_):
+    """Truncate tables."""
+    if not database_exists(str(db_.engine.url)):
+        create_database(str(db_.engine.url))
+    db_.session.remove()
+    db_.create_all()
+    all_tables = db_.metadata.tables
+    for table_name, table_object in all_tables.items():
+        db_.session.execute(f"ALTER TABLE {table_name} DISABLE TRIGGER ALL;")
+        db_.session.execute(table_object.delete())
+        db_.session.execute(f"ALTER TABLE {table_name} ENABLE TRIGGER ALL;")
+    db_.session.commit()
--- a/backend/tests/integration-async/conftest.py
+++ b/backend/tests/integration-async/conftest.py
@@ -4,7 +4,6 @@
 #
 # inspirehep is free software; you can redistribute it and/or modify it under
 # the terms of the MIT License; see LICENSE file for more details.
-
 import time
 from copy import deepcopy
 from datetime import datetime, timedelta
@@ -14,6 +13,7 @@ import pytest
 import structlog
 from click.testing import CliRunner
 from flask.cli import ScriptInfo
+from helpers.cleanups import db_cleanup, es_cleanup
 from helpers.providers.faker import faker
 from inspire_utils.record import get_value
 from invenio_db import db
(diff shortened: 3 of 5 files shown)


michamos:
I remember looking into this and getting convinced that this should be slower than deleting the index. Have you compared performance?

backend/tests/integration/conftest.py
```diff
@@ -76,8 +77,30 @@ def create_app():
...
+@pytest.fixture(scope="module")
+def inspire_database(appctx):
+    """Setup database.
+
+    Scope: module
+
+    Normally, tests should use the function-scoped :py:data:`db` fixture
+    instead. This fixture takes care of creating the database/tables and
+    removing the tables once tests are done.
+    """
```
drjova: this is a left over?

backend/tests/integration/conftest.py
```diff
@@ -76,8 +77,30 @@ def create_app():
...
+
+    clear_db(db_)
+    yield db_
+    db_.session.remove()
+
+
 @pytest.fixture(scope="function")
-def db_(database):
+def db_alembic(database):
+    yield database
```
drjova: why not just use ```database`` fixture in alembic?

backend/tests/integration-async/conftest.py
```diff
@@ -57,24 +58,12 @@ def clear_environment(app):
...
-        for table_name, table_object in all_tables.items():
-            db_.session.execute(f"ALTER TABLE {table_name} DISABLE TRIGGER ALL;")
-            db_.session.execute(table_object.delete())
-            db_.session.execute(f"ALTER TABLE {table_name} ENABLE TRIGGER ALL;")
-        db.session.commit()
-        _es = app.extensions["invenio-search"]
-        list(_es.delete(ignore=[404]))
-        list(_es.create(ignore=[400]))
+        clear_db(db_)
+        clear_es(es)
```
drjova: I'd make them fixtures instead of passing objects around

backend/tests/helpers/cleanups.py
```diff
@@ -0,0 +1,46 @@
...
+    it's current version is 0 it's failing. So in this case only way is to
+    remove index but invenio 1.1.1 allows to remove all indexes only
+    After upgrading to invenio 1.2 we will be able
+    to remove only specified index
+    """
+    from invenio_search import current_search, current_search_client
+
+    es.indices.refresh()
+    try:
+        for indice in es.indices.stats()["indices"].keys():
```
drjova:
```suggestion
        for index in es.indices.stats()["indices"].keys():
```

backend/tests/helpers/cleanups.py
```diff
@@ -0,0 +1,46 @@
...
+            es.delete_by_query(indice, "{}")
+    except elasticsearch.exceptions.RequestError:
+        _es_delete_indexes(current_search)
+        _es_create_indexes(current_search, current_search_client)
+    es.indices.refresh()
+
+
+def clear_db(db_):
+    """Remove data from tables not tables itself."""
```
drjova:
```suggestion
    """Truncate tables."""
```

backend/tests/helpers/cleanups.py
```diff
@@ -0,0 +1,46 @@
+# -*- coding: utf-8 -*-
```
drjova: the name is ```cleanups``` and the functions are for ``clear`` isn't it a bit confusing? :/

backend/tests/helpers/cleanups.py
```diff
@@ -0,0 +1,46 @@
...
+        create_database(str(db_.engine.url))
+    db_.create_all()
+    db_.session.rollback()
+    db_.session.remove()
+    all_tables = db_.metadata.tables
+    for table_name, table_object in all_tables.items():
+        db_.session.execute(f"ALTER TABLE {table_name} DISABLE TRIGGER ALL;")
+        db_.session.execute(table_object.delete())
+        db_.session.execute(f"ALTER TABLE {table_name} ENABLE TRIGGER ALL;")
+    db_.session.commit()
```
drjova: Also if we do something like this do we need to rollback? (for integration)

drjova:
I'd change the commit message:

```
tests: ....

```

backend/tests/integration/conftest.py
pazembrz: :+1:

backend/tests/integration/conftest.py
pazembrz: yup :/

backend/tests/integration-async/conftest.py
pazembrz: But then It would not work too well with async tests, as fixtures there are different :/
It's similar to invenio's `_es_delete/create_indexes` functions ;)

pazembrz:
> I remember looking into this and getting convinced that this should be slower than deleting the index. Have you compared performance?

Yes, and difference is quite big. Before changes integration tests on travis (without build) took around 830 seconds after changes they take around 530 seconds.

backend/tests/helpers/cleanups.py
```diff
@@ -0,0 +1,46 @@
...
+    to remove only specified index
+    """
+    from invenio_search import current_search, current_search_client
+
+    es.indices.refresh()
+    try:
+        for index in es.indices.stats()["indices"].keys():
+            es.delete_by_query(index, "{}")
+    except elasticsearch.exceptions.RequestError:
+        print("================================ES CLEANUP FAILED=====================")
```
drjova: 🙄

backend/tests/integration/conftest.py
```diff
@@ -76,6 +77,17 @@ def create_app():
     return inspire_create_app
 
 
+@pytest.fixture(scope="module")
+def database(appctx):
+    """Setup database.
```
drjova:
```suggestion
    """Setup database."""
```

backend/tests/integration/alembic/test_alembic.py
```diff
@@ -9,172 +9,189 @@
...
+        "conference_literature", database
+    )
+    assert "enum_conference_to_literature_relationship_type" in _get_custom_enums(
+        database
     )
-    assert "enum_conference_to_literature_relationship_type" in _get_custom_enums(db)
 
 
-def _get_indexes(tablename, db):
+def _get_indexes(tablename, db_alembic):
```
drjova: db_alembic?

backend/tests/helpers/cleanups.py
pazembrz: ouch :P

backend/tests/integration/alembic/test_alembic.py
pazembrz: there is no db_alembic anymore, I found a way to run it on normal fixture like other tests.

drjova: you are using ``db_alembic`` :P

pazembrz: no I'm not ;)

backend/tests/conftest.py
```diff
@@ -32,3 +32,18 @@ def vcr_config():
         ),
         "record_mode": "once",
     }
+
+
+@pytest.fixture(scope="module")
+def es(appctx):
+    """Setup and teardown all registered Elasticsearch indices.
+
+    Scope: module
```
drjova: redundant :)

backend/tests/helpers/cleanups.py
```diff
@@ -0,0 +1,45 @@
+# -*- coding: utf-8 -*-
+#
+# Copyright (C) 2019 CERN.
+#
+# inspirehep is free software; you can redistribute it and/or modify it under
+# the terms of the MIT License; see LICENSE file for more details.
+import elasticsearch
```
drjova: new line between docs and imports

pazembrz: This fixture is required to overwrite invenio es fixture which also deletes indexes.

backend/tests/conftest.py
pazembrz: This fixture is required to overwrite invenio es fixture which also deletes indexes.

backend/tests/helpers/cleanups.py
pazembrz: :+1:

PR #577, opened 2019-11-21
ui: rework InlineList and InlineUL

miguelgrc:
* INSPIR-2891

Also removed the .bulleted-list CSS class and replaced usages with InlineUL


ui/src/common/components/PublicNotesList.jsx
```diff
@@ -3,6 +3,7 @@ import PropTypes from 'prop-types';
 import { List } from 'immutable';
 
 import InlineList from './InlineList';
+import { SEPARATOR_MIDDLEDOT } from './InlineList/constants';
```
harunurhan: those imports are not nice, let's change the `index.js` of `InlineList` to export stuff directly.

```js

export default InlineList;

export { default as InlineUL } from './InlineUL'
export { SEPARATOR_AND, ... } from './constants'
```

PR #579, opened 2019-11-21
migrator: add cli command to run specified migration step

pazembrz:


drjova:
@pazembrz why you always add an uppercase in the commit message :P you're the uppercase master now

backend/inspirehep/migrator/tasks.py
```diff
@@ -109,6 +110,37 @@ def migrate_record_from_legacy(recid):
...
+            str(res.object_uuid) for res in query.yield_per(CHUNK_SIZE)
+        )
+    task = migrate_recids_from_mirror(
+        list(recids_chunked),
+        disable_orcid_push=disable_orcid_push,
+        disable_references_processing=disable_references_processing,
+        step_no=step_no,
+        one_step=True,
+    )
+    print("All migration tasks has been scheduled.")
```
drjova: ditto

backend/inspirehep/migrator/tasks.py
```diff
@@ -109,6 +110,37 @@ def migrate_record_from_legacy(recid):
...
+    """Allows to easily run step by step migration only for valid records """
+    if step_no == 0:
+        query = LegacyRecordsMirror.query.with_entities(LegacyRecordsMirror.recid)
+        query = query.filter(LegacyRecordsMirror.valid.is_(True))
+    elif 0 < step_no < 3:
+        query = PersistentIdentifier.query.with_entities(
+            PersistentIdentifier.object_uuid
+        )
+    else:
+        print("Wrong step number!")
```
drjova: can we use ``echo``? also more info will be better like, the step etc

backend/inspirehep/migrator/cli.py
```diff
@@ -156,3 +157,41 @@ def continuously():
...
+    default=False,
+    help="Wait for all subtasks to finish. (No progress bar will be available still)",
+)
+@click.option(
+    "-s",
+    "--step",
+    default=0,
+    type=int,
+    help="""
+    1 - Migrate from mirror table
```
drjova: this will not have a new line on help I think

backend/inspirehep/migrator/cli.py
```diff
@@ -156,3 +157,41 @@ def continuously():
...
+    is_flag=True,
+    default=False,
+    help="Force the task to run even in debug mode.",
+)
+@click.option(
+    "-w",
+    "--wait",
+    is_flag=True,
+    default=False,
+    help="Wait for all subtasks to finish. (No progress bar will be available still)",
```
drjova:
```suggestion
    help="Wait for all subtasks to finish without progress bar.",
```

backend/inspirehep/migrator/tasks.py
```diff
@@ -109,6 +110,37 @@ def migrate_record_from_legacy(recid):
     db.session.commit()
 
 
+def migrate_from_mirror_run_step(
+    disable_orcid_push=True, disable_references_processing=True, step_no=1
+):
+    """Allows to easily run step by step migration only for valid records """
+    if step_no == 0:
+        query = LegacyRecordsMirror.query.with_entities(LegacyRecordsMirror.recid)
+        query = query.filter(LegacyRecordsMirror.valid.is_(True))
```
drjova: do we need to assignments for this?

backend/inspirehep/migrator/tasks.py
```diff
@@ -190,16 +223,28 @@ def migrate_recids_from_mirror(
     logger.info(f"Processing {all_recs} records in this migration step")
     step = migration_steps[step_no]
     header = (step.s(r) for r in recids_chunks)
-    chord_task = chord(header)(
-        migrate_recids_from_mirror.s(
+    if one_step:
+        callback = stop_after_one_step.s()
```
drjova: shouldn't this be 

```python
 callback = migrate_recids_from_mirror.s(
            step_no=step_no ,
            disable_orcid_push=disable_orcid_push,
            disable_orcid_push=disable_orcid_push,
            disable_references_processing=disable_references_processing,
            disable_references_processing=disable_references_processing,
        )
```

backend/inspirehep/migrator/tasks.py
pazembrz: Actually for all tasks in migrator we are using print.

drjova: amazing :D Let's change that bit by bit

pazembrz: And I don't see why i should print step number to this message. It will be the same number provided by user few seconds before displaying this message)

backend/inspirehep/migrator/cli.py
pazembrz: This is how  it looks like:
```
[pazembrz@pawel-pc backend]$ poetry run inspirehep migrate mirror_step --help
Usage: inspirehep migrate mirror_step [OPTIONS]

  Migrate records step by step

  (Re)Migrates only records marked as valid

Options:
  -f, --force         Force the task to run even in debug mode.
  -w, --wait          Wait for all subtasks to finish. (No progress bar will
                      be available still)
  -s, --step INTEGER  1 - Migrate from mirror table
                      2 - Recalculate citations
                      proceedings etc.
                      3 - Reindex
  --help              Show this message and exit.
```

backend/inspirehep/migrator/cli.py
pazembrz: It means that there is no progress bar with out without this option. Your help message suggests that without this option there is progress bar.

backend/inspirehep/migrator/tasks.py
pazembrz: It's more readable than one long line.

backend/inspirehep/migrator/tasks.py
pazembrz: No, why?

backend/inspirehep/migrator/tasks.py
pazembrz: I changed all prints to echo in this file

backend/inspirehep/migrator/tasks.py
drjova: it's not IMO, the following looks better :)

``` python
LegacyRecordsMirror \
    .query.with_entities(LegacyRecordsMirror.recid) \
    .filter(LegacyRecordsMirror.valid.is_(True))
```

or 

```python
query = (
    LegacyRecordsMirror
    .query.with_entities(LegacyRecordsMirror.recid)
    .filter(LegacyRecordsMirror.valid.is_(True))
)
```

backend/inspirehep/migrator/cli.py
drjova: I'm missing something :/ it's still not clear for me, this option will display a progressbar or not?

backend/inspirehep/migrator/tasks.py
pazembrz: Really? I don't see it but i can change it if it's that important for you ;)

backend/inspirehep/migrator/tasks.py
drjova: my bad sorry I didn't follow the code correclty

backend/inspirehep/migrator/tasks.py
pazembrz: Black is forcing:
```
query = LegacyRecordsMirror.query.with_entities(
            LegacyRecordsMirror.recid
        ).filter(LegacyRecordsMirror.valid.is_(True))
```
And this looks terrible IMO

drjova: yes please 👍 either option will work

backend/inspirehep/migrator/cli.py
pazembrz: This option has nothing to do with progressbar, This comment is there only because old migration had progressbar. Comment is copy-paste of same option from `mirror` command.

drjova: let's remove then ``` (No progress bar will be available still)`` we don't need it right?

backend/inspirehep/migrator/tasks.py
pazembrz: As i wrote Black is forcing 
```
query = LegacyRecordsMirror.query.with_entities(
            LegacyRecordsMirror.recid
        ).filter(LegacyRecordsMirror.valid.is_(True))
```
for both options. Is is really better?

drjova: from assigning twice the same ``var`` it's better IMO :)

pazembrz: If you say so. For me this what black is doing is less readable.

drjova: we're using black exactly for that reason, not to have different ways to style things

backend/inspirehep/migrator/cli.py
pazembrz: :+1:

backend/inspirehep/migrator/tasks.py
pazembrz: done

backend/inspirehep/migrator/tasks.py
```diff
@@ -109,6 +111,38 @@ def migrate_record_from_legacy(recid):
...
+        )
+    else:
+        echo("Wrong step number!")
+        return
+    if step_no == 0:
+        recids_chunked = chunker(str(res.recid) for res in query.yield_per(CHUNK_SIZE))
+    else:
+        recids_chunked = chunker(
+            str(res.object_uuid) for res in query.yield_per(CHUNK_SIZE)
+        )
```
drjova: Just last thing, can we just reduce the ``if``s to the following?

```python

    if step_no == 0:
        query = LegacyRecordsMirror.query.with_entities(
            LegacyRecordsMirror.recid
        ).filter(LegacyRecordsMirror.valid.is_(True))
        recids_chunked = chunker(str(res.recid) for res in query.yield_per(CHUNK_SIZE))
    elif 0 < step_no < 3:
        query = PersistentIdentifier.query.with_entities(
            PersistentIdentifier.object_uuid
        )
        recids_chunked = chunker(
            str(res.object_uuid) for res in query.yield_per(CHUNK_SIZE)
        )
    else:
        echo("Wrong step number!")
        return
 
    task = migrate_recids_from_mirror(...)
```

pazembrz: :+1:

PR #580, opened 2019-11-22
tests: ressult not assigned error fix

pazembrz:
* Fix for old error which was making async tests to fail randomly


backend/tests/integration-async/conftest.py
```diff
@@ -127,7 +127,7 @@ def _check(steps={}, timeout=15):
                 try:
                     result = _fun(*_args, **_kwargs)
                 except:
-                    pass
+                    break
```
drjova: we should replace this fixture in the future :/

pazembrz: I agree

PR #591, opened 2019-11-27
setup: add web_pdb dev dependency

miguelgrc:
This will make it much easier for us to debug inside a docker container.

To use, ``import web_pdb`` and invoke ``web_pdb.set_trace(port=7777)`` where desired. Then go to ``localhost:7777`` to access the debugger.


michamos:
don't we install dev packages also inside prod containers?

pazembrz:
I didn't noticed this before, but yes, this is true. Why we are doing this?

harunurhan:
why don't we do `--no-dev` while building the image?

pazembrz:
I don't know. Probably we should.

harunurhan:
so we can maybe take this in, because it is useful. And add `--no-dev` to `Dockerfile`s to reduce the size of the images

@drjova @benjamin-bergia

drjova:
We can just create a US for adding ``--no-dev``

PR #592, opened 2019-11-27
conferences: add submission form

harunurhan:
INSPIR-2858


ui/src/submissions/conferences/schemas/constants.js
```diff
@@ -0,0 +1,248 @@
+import getValue from '../../common/getValue';
+
+export const countryOptions = [
```
ksachs: don't tell me you hard-coded a country list.
There must be a list CountryCode - Country somewhere.

harunurhan: frontend doesn't have to be aware of country code, we need display only things there.

harunurhan: I would be interested to hear how else we could display country select box?

ksachs: there must be a country list from another pull-down/autocomplete/whatever in this project.
Worst case: loop over that list to fill whatever you need here.

I just don't want to maintain several lists in case another former soviet union country pops up or they rename Macedonia.

ksachs: https://github.com/inspirehep/inspire-dojson/blob/702d2733fc6f1288ef4fb54687442fe332acb6a0/inspire_dojson/utils/geo.py

harunurhan: - No we don't have another "pull-down/autocomplete/whatever" country list on the UI.
- Loop over what?
- Well, fortunately it's not often that a country pops up or renamed.
- Effort to share code between `dojson` and `inspirehep/ui` is way too much.

backend/inspirehep/search/facets.py
```diff
@@ -28,6 +28,30 @@ def inner(values):
     return inner
 
 
+def date_range_contains_conferences_filter():
```
drjova: Are you sure will not break without the ``field``?

backend/inspirehep/search/facets.py
```diff
@@ -28,6 +28,30 @@ def inner(values):
...
+    def inner(values):
+        opening_date, closing_date = values[0].split("--")
```
drjova: don't like that tbh, what will happen with ``contains=`` (empty)?

backend/tests/integration/records/views/test_views_conferences.py
```diff
@@ -218,3 +218,37 @@ def test_conferences_filters(api_client, db, es, create_record):
...
+    create_record("con", data=conference_during_april_third_week)
+    create_record("con", data=conference_during_whole_april)
+    create_record("con", data=conference_during_january)
+    create_record("con", data=conference_during_june)
+
+    from_april_5_to_15 = "2019-04-05--2019-04-15"
+    response = api_client.get(f"/conferences?contains={from_april_5_to_15}")
+    response_data = response.json
+    assert response_data["hits"]["total"] == 3
+    # TODO: assert right conferences found
```
drjova: (I'd add a test without a value 🆙 )

ui/src/submissions/conferences/components/ConferenceSubmission.jsx
```diff
@@ -0,0 +1,58 @@
...
+import cleanupFormData from '../../common/cleanupFormData';
+import { convertAllImmutablePropsToJS } from '../../../common/immutableToJS';
+import useIsMounted from '../../../common/hooks/useIsMounted';
+
+const DEFAULT_FORM_DATA = conferenceSchema.cast();
+
+function ConferenceSubmission({ onSubmit, error = null }) {
+  const isMounted = useIsMounted();
+
+  const onFormikSubmit = useCallback(
```
drjova:
```suggestion
  const onSubmit = useCallback(
```

backend/inspirehep/search/facets.py
harunurhan: It's for a single use case for now, that's why I didn't care to handle all the cases. It should blow up pretty obviously, so if it happens we will see and handle the case.

ui/src/submissions/conferences/components/ConferenceSubmission.jsx
harunurhan: onSubmit is already a prop :/

PR #593, opened 2019-11-27
submissions: Change the submission forms to dispatch logout for 401

monaawi:
* INSPIR-2830

### DIFF
--- a/ui/src/middlewares/__tests__/logoutUser.test.jsx
+++ b/ui/src/middlewares/__tests__/logoutUser.test.jsx
@@ -1,0 +1,50 @@
+import middleware from '../logoutUserOn401';
+import { userLogout } from '../../actions/user';
+
+jest.mock('../../actions/user');
+
+describe('LogoutUser middleware', () => {
+  let mirrorNext;
+  let dispatch;
+  let mockDispatch;
+
+  beforeEach(() => {
+    mirrorNext = jest.fn(value => value);
+    mockDispatch = jest.fn();
+    dispatch = middleware({ dispatch: mockDispatch })(mirrorNext);
+  });
+
+  afterEach(() => {
+    userLogout.mockClear();
+  });
+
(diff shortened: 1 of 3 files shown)


ui/src/middlewares/__tests__/logoutUser.test.jsx
```diff
@@ -0,0 +1,40 @@
...
+  });
+
+  afterEach(() => {
+    userLogout.mockClear();
+  });
+
+  it('dispatches logoutUser for SUBMIT_ERROR 401 and returns result of next(action)', () => {
+    const action = {
+      type: 'SUBMIT_ERROR',
+      payload: { status: 401 },
```
harunurhan: we might need another case for `SUBMIT_ERROR` where status is not `401`

ui/src/middlewares/__tests__/logoutUser.test.jsx
```diff
@@ -0,0 +1,40 @@
...
+    };
+    const result = dispatch(action);
+    expect(result).toBe(action);
+    expect(userLogout).toHaveBeenCalled();
+  });
+
+  it('only returns result of next(action) when not a SUBMIT_ERROR 401', () => {
+    const action = {
+      type: 'SOME_ERROR',
+      payload: { status: 500 },
```
harunurhan: `status: 401` would be better, since it's not sure if that test is failing because it's not `SUBMIT_ERROR` or it is not `401`

ui/src/store.js
```diff
@@ -20,6 +20,7 @@ import clearStateDispatcher from './middlewares/clearStateDispatcher';
 import redirectToErrorPageMiddleware from './middlewares/redirectToErrorPage';
 import actionTrackerMiddleware from './middlewares/actionTracker';
 import syncLocationWithSearch from './middlewares/syncLocationWithSearch';
+import logoutUser from './middlewares/logoutUser';
```
harunurhan: sth like `logoutUserOn401` would be better

PR #595, opened 2019-11-28
minter: add CNUM minter and provider

ammirate:
* INSPIR-2853


backend/inspirehep/pidstore/providers/cnum.py
```diff
@@ -0,0 +1,60 @@
...
+from datetime import datetime
+
+import structlog
+from invenio_pidstore.models import PersistentIdentifier, PIDStatus
+from invenio_pidstore.providers.base import BaseProvider
+
+LOGGER = structlog.getLogger()
+
+
+class CNUMIdentifier(object):
```
michamos: why a class? this looks like a function to me.

backend/inspirehep/pidstore/minters/cnum.py
```diff
@@ -0,0 +1,37 @@
...
+
+    pid_value_path = "cnum"
+    pid_type = "cnum"
+    provider = InspireCNUMProvider
+
+    @classmethod
+    def mint(cls, object_uuid, data):
+        minter = cls(object_uuid, data)
+        minter.validate()
+        if "cnum" not in data:
```
michamos: what about existing CNUMs coming from legacy in the migration? are you registering those?

backend/inspirehep/pidstore/minters/cnum.py
```diff
@@ -0,0 +1,37 @@
...
+            record_id_provider = minter.create(data)
```
michamos:
```suggestion
            cnum_provider = minter.create(data)
```

backend/inspirehep/pidstore/providers/cnum.py
ammirate: Because we need to be consistent. For the control number minter we have a class who does the job and having a class there and a function here makes the code more complicated than having a function instead of a class. And BTW, functions are classes too so I don't see any real improvement :)

backend/inspirehep/pidstore/providers/cnum.py
```diff
@@ -0,0 +1,60 @@
...
+            str: a conference CNUM identifier in the form CYY-MM-DD[.X]
+        """
+        opening_date = data.get("opening_date")
+        if not opening_date:
+            return
+
+        adjusted_date = datetime.strptime(opening_date, "%Y-%m-%d").strftime("%y-%m-%d")
+        cnum = f"C{adjusted_date}"
+        cnums_count = PersistentIdentifier.query.filter(
+            PersistentIdentifier.pid_value.like(f"{cnum}%")
```
michamos: I think you also want to set the `pid_type` here to
1. ensure you don't match something that looks like a CNUM but is not
2. use the index

backend/inspirehep/pidstore/minters/cnum.py
ammirate: Nope, this is only for minting new conference records. The migration is not part of this task

backend/inspirehep/pidstore/providers/cnum.py
michamos: Sorry, but a class with one method is a function, not a class. If you want to simplify code, change it to a function there. Functions are objects, true, but not classes.

backend/inspirehep/pidstore/providers/cnum.py
ammirate: yeah why not :+1:

backend/inspirehep/pidstore/minters/cnum.py
michamos: is it part of any task? if not I believe it should be part of this one, otherwise you'll register existing CNUMs for conferences.

ammirate: Agree on migrating conferences to avoid overwriting CNUMs, not sure is part of this though, since it won't run unless we create conferences record through submissions. I'll create another subtask for this User Story, WDYT?

backend/inspirehep/pidstore/providers/cnum.py
ammirate: Nothing enforces a class to have more than one method :) to make everyone happy I moved that method in the `InspireCNUMProvider` class and removed the other one. I think it fits properly there.

backend/inspirehep/pidstore/minters/cnum.py
ammirate: After discussing IRL, we decided to have some checks here. Some edge cases won't be handled, as describer in the `CNUMMinter` documentation.

ammirate:
@michamos updated

backend/inspirehep/pidstore/minters/cnum.py
```diff
@@ -0,0 +1,109 @@
...
+        changed_deleted_status = rec.get("deleted", False) ^ prev_version.get(
+            "deleted", False
+        )
+        if changed_deleted_status:  # record is un-deleted
+            pid = PersistentIdentifier.query\
+                .filter(PersistentIdentifier.pid_type == "cnum")\
+                .filter(PersistentIdentifier.object_uuid == rec.id)\
+                .one_or_none()
+            if pid:
+                pid.status = PIDStatus.REGISTERED  # un-delete pid
```
michamos: this whole thing with versions seems overly complex. I would do the following:
1. Get the cnum pointing to this record from the pidstore
2. Is it the `pid_value` compatible with the metadata? if not -> exception
3. If the record is not marked as `deleted` but the `pid` is deleted, mark it as `registered`.

ammirate: Totally agree

backend/inspirehep/pidstore/minters/cnum.py
```diff
@@ -0,0 +1,101 @@
...
+                status=PIDStatus.REGISTERED,
+            )
+        return minter
+
+    @classmethod
+    def update(cls, object_uuid, data):
+        """Handle the CNUM indentifier in record updates .
+
+        This method performs several checks:
+        * if it is the first update for this record, don't do anything;
```
michamos: docstring out of date

backend/inspirehep/pidstore/minters/cnum.py
```diff
@@ -0,0 +1,101 @@
...
+        Args:
+            object_uuid (str): the record uuid
+            data (dict): the record's metadata
+
+        Returns:
+            None
+        """
+        from inspirehep.records.api import ConferencesRecord
+
+        rec = ConferencesRecord.get_record_by_pid_value(data['control_number'])
```
michamos: is this needed?

backend/inspirehep/pidstore/minters/cnum.py
```diff
@@ -0,0 +1,101 @@
...
+        if cnum_pid:
+            if cnum_pid.pid_value != rec["cnum"]:
+                raise CNUMChanged(f"CNUM identifier changed from the prev version")
+
+            if data.get("deleted"):
+                cnum_pid.status = PIDStatus.REGISTERED  # un-delete pid
+
+    @classmethod
+    def delete(cls, object_uuid, data):
+        """Removes the CNUM pid for the given record from the DB.
```
michamos:
```suggestion
        """Mark the CNUM pid for the given record as deleted.
```

backend/inspirehep/records/api/base.py
```diff
@@ -383,7 +383,7 @@ def index(self, force_delete=None, delay=True):
 
     @property
     def _previous_version(self):
-        """Allows to easily acces previous version of the record"""
+        """Allows to easily access previous version of the record"""
```
michamos:
```suggestion
        """Return the previous version of the record"""
```

backend/tests/helpers/providers/record_provider.py
```diff
@@ -20,6 +20,9 @@ class RecordProvider(BaseProvider):
     def control_number(self):
         return fake.random_number(digits=8, fix_len=True)
 
+    def cnum(self):
+        return f"C-{fake.date()}"
```
michamos: This doesn't look like a CNUM

PR #597, opened 2019-11-28
facets: new option for single author in number of authors facet

miguelgrc:
* INSPIR-2930

### DIFF
--- a/backend/inspirehep/search/facets.py
+++ b/backend/inspirehep/search/facets.py
@@ -13,10 +13,16 @@ from invenio_records_rest.facets import range_filter
 
 
 def range_author_count_filter(field):
-    """Range filter for returning record only with 1 <= authors <= 10."""
+    """Range filter for returning records within the corresponding range(s)."""
+
+    range_for_option = {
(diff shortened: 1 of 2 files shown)


backend/inspirehep/search/facets.py
```diff
@@ -5,15 +5,21 @@
...
+
+    range_for_option = {
+        "Single author": {"gte": 1, "lte": 1},
+        "10 authors or less": {"gte": 1, "lte": 10},
+    }
 
     def inner(values):
-        return Range(**{field: {"gte": 1, "lte": 10}})
+        ranges = [Q("range", **{field: range_for_option[value]}) for value in values]
+        return Q("bool", must=ranges)
```
michamos: I know you just copied what others have been doing, but it's better to use a `filter` rather than a `must` for aggregations (not only here, all aggregations), see https://www.elastic.co/guide/en/elasticsearch/reference/5.6/query-filter-context.html for information about the difference.

drjova:
@miguelgrc could you please rebase?

PR #601, opened 2019-12-02
submission: make submit button always enabled

monaawi:
* INSPIR-2935

### DIFF
--- a/e2e/tests/submissions/author.test.js
+++ b/e2e/tests/submissions/author.test.js
@@ -1,6 +1,6 @@
 const { ResponseInterceptor } = require('../../utils/interceptors');
 const { login } = require('../../utils/user');
-const { FormSubmitter, SUBMIT_BUTTON_SELECTOR } = require('../../utils/form');
+const { FormSubmitter } = require('../../utils/form');
 const routes = require('../../utils/routes');
 
 describe('author submissions', () => {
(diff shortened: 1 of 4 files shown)


ui/src/submissions/common/components/SubmitButton.jsx
```diff
@@ -1,22 +1,29 @@
...
-        loading={isSubmitting || isValidating}
-        disabled={!isValid}
-      >
-        Submit
-      </Button>
-    );
-  }
+  useEffect(
+    () => {
+      if (previousIsSubmitting && !isSubmitting && !isValid) {
```
harunurhan: Perhaps extract this line, to a meaningful variable so that it's more readable:

Example:
`const hasTriedToSubmitInvalidForm = previousIsSubmitting && !isSubmitting && !isValid;`

e2e/tests/submissions/author.test.js
```diff
@@ -100,12 +100,6 @@ describe('author submissions', () => {
       linkEl => linkEl.href
     );
 
-    const isSubmitButtonDisabled = await page.$eval(
-      SUBMIT_BUTTON_SELECTOR,
-      submitButton => submitButton.disabled
-    );
-
     expect(authorUpdateLink).toEqual(`${routes.AUTHOR_SUBMISSION}/999108`);
-    expect(isSubmitButtonDisabled).toBe(true);
```
harunurhan: would be better if you click the button first, then check if it's disabled, rather than removing this completely.

monaawi: but the button never becomes disabled now. It is always enabled but if there is an error, then it scrolls up

PR #604, opened 2019-12-02
ui: hide citation summary graph and table when 0 publications

miguelgrc:
* INSPIR-2949

Also added the type to the author facet since it was missing and was causing warnings in the UI.


ui/src/common/components/CitationSummaryGraph/CitationSummaryGraph.jsx
```diff
@@ -265,6 +275,7 @@ CitationSummaryGraph.propTypes = {
     type: PropTypes.string.isRequired,
     xValue: PropTypes.string.isRequired,
   }),
+  publications: PropTypes.array,
```
harunurhan: pull this logic out to the detail page instead of passing here.

miguelgrc:
No, we decided to display the box with the `Empty`, as for the publications box and the categories' contributions

ui/src/authors/containers/DetailPageContainer.jsx
```diff
@@ -163,14 +170,21 @@ function DetailPage({ record, loading, dispatch, match, publicationsQuery }) {
...
           <ContentBox subTitle="Citation Summary">
-            <Row gutter={{ xs: 0, lg: 32 }}>
-              <Col xs={24} md={24} lg={7}>
-                <CitationsByYearGraphContainer />
-              </Col>
-              <Col xs={24} md={24} lg={17}>
-                <CitationSummaryGraphContainer />
-              </Col>
-            </Row>
+            {publications && publications.size === 0 ? (
```
harunurhan: I don't like that we have this thing in some places and we can't use `EmptyOrChildren`, maybe we could change `EmptyOrChildren`  to handle this case? what do you think?

ui/src/common/containers/CitationSummaryTableContainer.jsx
```diff
@@ -18,6 +19,11 @@ const stateToProps = state => ({
   ]),
   hIndex: state.citations.getIn(['citationSummary', 'h-index', 'value']),
   error: state.citations.get('errorCitationSummary'),
+  publications: state.search.getIn([
```
harunurhan: you can pull this out too.

ui/src/authors/containers/DetailPageContainer.jsx
miguelgrc: We could simply make `EmptyOrChildren` accept a boolean instead of directly an object/collection. We would have to remove the `isEmptyCollection` function (the component would be just like `condition ? content : Empty`) and change all the usages of the component to pass the check that we want to perform.

PR #607, opened 2019-12-04
Upcoming Conferences & start_date filter

harunurhan:


ui/src/reducers/search.js
```diff
@@ -177,14 +174,29 @@ const searchReducer = (state = initialState, action) => {
     case CHANGE_SEARCH_BOX_NAMESPACE:
       return state.set('searchBoxNamespace', searchBoxNamespace);
     case NEW_SEARCH_REQUEST:
+      const persisteQueryParams = state.getIn([
```
drjova:
```suggestion
      const persistQueryParams = state.getIn([
```

PR #623, opened 2019-12-12
ui: fix form fields display on small screens and/or with long labels

miguelgrc:
* INSPIR-2936


ui/src/submissions/common/__tests__/__snapshots__/withFormItem.test.jsx.snap
```diff
@@ -11,16 +11,38 @@ exports[`withFormItem passes all props correctly to Form.Item and wrapped Input
       Error
     </span>
   }
-  label="Test Label"
+  label={
+    <Tooltip
+      arrowPointAtCenter={false}
+      autoAdjustOverflow={true}
+      mouseEnterDelay={0.1}
```
drjova: do we need these values? Can we just use the default ones?

drjova:
Do we have tests for the ellipisis?

drjova:
if not could you please attache some screenshots?

ui/src/submissions/common/__tests__/__snapshots__/withFormItem.test.jsx.snap
miguelgrc: They're default props from Ant's Tooltip, I'm not setting those values.

drjova: if they are default why we set delay etc :/

miguelgrc: We don't. This is the snapshot and `0.1` is the default `mouseEnterDelay` set by Ant (https://ant.design/components/tooltip/#Common-API)

drjova: so we don't have to explicity add it, by default it'll have this value :)

miguelgrc:
We don't have tests for this (we would need to create several UI tests with different viewport sizes and they would take long to run). Here are a few screenshots with the ellipsis and with the new form layout for mobile/small viewports
Ellipsis (I added tooltips)
Mobile
I'm not 100% convinced with the ellipsis thing, but it's much better than what we have, specially the mobile layout.

ui/src/submissions/common/components/FieldInfoAlert.jsx
```diff
@@ -1,14 +1,19 @@
...
 class FieldInfoAlert extends Component {
   render() {
     const { description } = this.props;
     return (
       <Row className="mb1">
-        <Col span={WRAPPER_COL.span} offset={LABEL_COL.span}>
+        <Col
+          {...{
+            sm: { span: 24 },
+            md: { span: 24 - LABEL_COL.md.span, offset: LABEL_COL.md.span },
```
drjova: can we use ``WRAPPER_COL`` for span?

ui/src/submissions/common/withFormItem.jsx
```diff
@@ -21,10 +24,9 @@ export default function withFormItem(FormInputComponent) {
     getWrapperColForOnlyChild() {
       const { label, labelCol } = this.props;
       if (label && labelCol) {
-        const span = 24 - labelCol.span;
-        return { span };
+        return WRAPPER_COL;
       }
-      return { span: 24 };
+      return { sm: { span: 24 }, md: { span: 24 } };
```
drjova: redundant IMHO

ui/src/submissions/common/__tests__/__snapshots__/withFormItem.test.jsx.snap
miguelgrc: We're not adding anything, note that in `withFormItem.jsx` we're only setting the title of the Tooltip. When running the jest tests the other values are added and if we remove them the tests fail 🙁

ui/src/App.scss
```diff
@@ -181,3 +181,13 @@ body {
 .ant-modal-wrap {
   padding-top: $header-min-height;
 }
+
+.ant-form-item-label {
+  text-overflow: ellipsis;
+  z-index: 1;
+  padding-right: 8px !important;
```
drjova: the screenshots are with these values?

miguelgrc: Yes, without this the padding is different when the ellipsis is used vs when the label fits (since Ant adds some padding after the colon).

PR #624, opened 2019-12-12
celery: don't fail acks_late tasks on worker crash

michamos:
* Tasks having `acks_late` set to true automatically retry when they are
  interrupted due to a worker crash. However, this doesn't happen by
  default when the worker is killed, unless `reject_on_worker_lost` is
  also set to true. To ensure this is always the case, this commit uses
  the Celery task annotations machinery to automatically annotate tasks
  having the former option with the latter.

Note that this won't solve the issue when a task in the head of the chord fails until https://github.com/celery/celery/pull/5700 gets released.

### DIFF
--- a/backend/inspirehep/config.py
+++ b/backend/inspirehep/config.py
@@ -124,6 +124,16 @@ CELERY_BEAT_SCHEDULE = {
     # },
 }
 
+
+class Annotator:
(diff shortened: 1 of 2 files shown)


drjova:
no my bad, this is on ``reject_on_worker_lost``

PR #628, opened 2019-12-13
tests: fix random async test failure

pazembrz:
Tries to catch ES exception about wrong records when deleting index content.


michamos:
Still looks flaky: https://travis-ci.org/inspirehep/inspirehep/builds/625591474?utm_medium=notification&utm_source=github_status

PR #629, opened 2019-12-13
deps: update celery to v4.4.0

michamos:
Let's try this :crossed_fingers:


backend/poetry.lock
```diff
@@ -48,6 +48,9 @@ optional = false
 python-versions = "*"
```
michamos: large diff in this file due to using poetry 1.0.0 which changed a bit the format of the lock file.

PR #637, opened 2020-01-06
submissions: send rt ticket when user submits conference

pazembrz:
* INSPIR-3019


backend/inspirehep/submissions/views.py
```diff
@@ -125,13 +125,43 @@ def post(self):
...
+        if not is_superuser_or_cataloger_logged_in():
+            self.create_ticket(record, "rt/new_conference.html")
+        return (
+            jsonify(
+                {"pid_value": record["control_number"], "cnum": record.get("cnum")}
+            ),
+            201,
+        )
+
+    def create_ticket(self, record, rt_template):
```
drjova: this function is very similar to https://github.com/inspirehep/inspirehep/blob/1ae762f48e218debf3ab96062cdd2b3c828f91e6/backend/inspirehep/submissions/views.py#L293 can we move it to BaseSubmissionsResource?

PR #640, opened 2020-01-07
indexer: reindex authors papers when his name changes

pazembrz:
* INSPIR-2990

### DIFF
--- a/backend/inspirehep/records/api/authors.py
+++ b/backend/inspirehep/records/api/authors.py
@@ -6,6 +6,7 @@
 # the terms of the MIT License; see LICENSE file for more details.
 
 from inspirehep.records.marshmallow.authors import AuthorsElasticSearchSchema
+from inspirehep.search.api import AuthorsSearch
 
 from ...pidstore.api import PidStoreAuthors
 from .base import InspireRecord
@@ -17,3 +18,8 @@ class AuthorsRecord(InspireRecord):
     es_serializer = AuthorsElasticSearchSchema
     pid_type = "aut"
     pidstore_handler = PidStoreAuthors
+
+    def get_papers_uuids(self):
+        all_papers = AuthorsSearch.get_author_papers(self, source="_id")
+        papers_ids = [paper.meta["id"] for paper in all_papers]
+        return list(set(papers_ids))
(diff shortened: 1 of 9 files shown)


backend/inspirehep/records/indexer/tasks.py
```diff
@@ -76,6 +76,39 @@ def process_references_for_record(record):
     LOGGER.info("No references changed", uuid=str(record.id))
 
 
+def process_author_papers_if_author_changed_name(record):
+    """Checks if author has changed his name and reindexes all his papers if he did
+
+    Checks `preffered_name` and `value` fields.
```
drjova: this is not true :P

backend/inspirehep/records/api/authors.py
```diff
@@ -17,3 +18,8 @@ class AuthorsRecord(InspireRecord):
     es_serializer = AuthorsElasticSearchSchema
     pid_type = "aut"
     pidstore_handler = PidStoreAuthors
+
+    def get_all_papers_ids(self):
```
drjova:
```suggestion
    def get_papers_uuids(self):
``` 
?

backend/tests/integration-async/records/indexer/test_authors.py
```diff
@@ -125,3 +125,75 @@ def test_record_created_through_api_is_indexed(
...
+        {
+            "step": es.search,
+            "args": ["records-hep"],
+            "expected_result": {
+                "expected_key": "hits.hits[0]._source.facet_author_name[0]",
+                "expected_result": f"{author['control_number']}_{author['name']['value']}",
+            },
+        },
+    ]
+    retry_until_matched(steps)
```
drjova: I'd try to avoid using the ``retry_until_matched`` tbh, can we just timeout instead? I know in the past we were avoiding it but still is better ☺️

pazembrz: Can I ask why?

drjova: because the logic of ``retry_until_matched`` it's wrong, if it doesn't match it goes through **all** the steps again and again. That's why we marked it as deprecated.

pazembrz: ok

PR #644, opened 2020-01-09
records: do not allow to update record with data without control_number

pazembrz:


backend/tests/integration/records/views/test_views_conferences.py
```diff
@@ -146,13 +146,16 @@ def test_conferences_application_json_put_with_cataloger_logged_in(
     login_user_via_session(api_client, email=cataloger.email)
     record = create_record("con")
     record_control_number = record["control_number"]
+    import ipdb
 
+    ipdb.set_trace()
```
drjova: oups

pazembrz: This is why I called this commit WIP ;)

PR #647, opened 2020-01-13
conferences: send confirmation email at user submission

monaawi:
* INSPIR-2878

### DIFF
--- a/backend/tests/integration/mailing/test_api_conferences.py
+++ b/backend/tests/integration/mailing/test_api_conferences.py
@@ -1,0 +1,24 @@
+import pytest
+from flask import current_app, render_template
+from mock import patch
+
+from inspirehep.mailing.api.conferences import send_conference_confirmation_email
+
+
+@patch("inspirehep.mailing.api.conferences.send_email")
+def test_send_conferences_confirmation_email(mock_send_email, base_app, db, es_clear):
+    expected_recipient = "johndoe@gmail.com"
+    conference = {"control_number": "1234"}
+    expected_subject = "Your conference(1234) has been successfully submitted!"
+    expected_content = render_template(
+        "mailing/conferences/submission_confirmation/base.html",
+        conference=conference,
+        host=current_app.config["SERVER_NAME"],
+    )
+    send_conference_confirmation_email(expected_recipient, conference)
+    mock_send_email.assert_called_once_with(
+        sender=current_app.config["CONFERENCES_CONFIRMATION_EMAIL_ADDRESS"],
+        recipient=expected_recipient,
+        subject=expected_subject,
+        body=expected_content,
+    )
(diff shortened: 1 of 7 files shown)


backend/tests/integration/mailing/test_api_conferences.py
```diff
@@ -0,0 +1,19 @@
...
+
+
+@patch("inspirehep.mailing.api.conferences.send_email")
+def test_send_conferences_confirmation_email(mock_send_email, base_app, db, es_clear):
+    expected_recipient = "johndoe@gmail.com"
+    conference = {"control_number": "1234"}
+    send_conference_confirmation_email(expected_recipient, conference)
+    mock_send_email.assert_called_once()
+
+    mock_call = mock_send_email.mock_calls[0][2]
```
harunurhan: use `assert_called_once_with` instead

backend/tests/integration/submissions/test_submissions_views.py
```diff
@@ -1003,3 +1003,57 @@ def test_rt_ticket_when_superuser_submits_conference(
...
+    form_data = deepcopy(CONFERENCE_FORM_DATA)
+    response = api_client.post(
+        "/submissions/conferences",
+        content_type="application/json",
+        data=json.dumps({"data": form_data}),
+    )
+
+    assert response.status_code == 201
+    mock_send_email.assert_called_once()
+    mock_call = mock_send_email.mock_calls[0][2]
```
harunurhan: also `assert_called_once_with`

backend/tests/integration/submissions/test_submissions_views.py
```diff
@@ -1003,3 +1003,57 @@ def test_rt_ticket_when_superuser_submits_conference(
...
+        "/submissions/conferences",
+        content_type="application/json",
+        data=json.dumps({"data": form_data}),
+    )
+
+    assert response.status_code == 201
+    mock_send_email.assert_not_called()
+
+
+@patch("inspirehep.mailing.api.conferences.send_email")
```
harunurhan: it is usually a good practice to mock directly the function you are calling in what you're testing, for especially unit tests, and I think this kind of view integration tests are in a way, unit tests. They are just in `integration` suite because we need `http` client.

So in this case it would be `send_conference_confirmation_email`

backend/tests/integration/submissions/test_submissions_views.py
```diff
@@ -1003,3 +1003,57 @@ def test_rt_ticket_when_superuser_submits_conference(
...
+
+    assert response.status_code == 201
+    mock_send_confirmation_email.assert_not_called()
+
+
+@patch("inspirehep.submissions.views.send_conference_confirmation_email")
+def test_confirmation_email_not_sent_when_user_is_cataloger(
+    mock_send_confirmation_email, app, api_client, create_user
+):
+    user = create_user(role=Roles.cataloger.value)
```
harunurhan: I think `create_user(role="cataloger")` should work the same, but not really important.

PR #649, opened 2020-01-14
celery: add ignore_results in tasks we don't have to wait

drjova:


backend/inspirehep/submissions/tasks.py
```diff
@@ -11,7 +11,7 @@
 from inspirehep.rt.tickets import CreateTicketException, create_ticket_with_template
 
 
-@shared_task(ignore_result=False, max_retries=5)
+@shared_task(ignore_result=True, max_retries=5)
```
harunurhan: I thought ticket creation was important (something to wait for) for some things like job submissions. For example, if there is no ticket nobody would check and approve it [pending -> open].

michamos:
This is a workaround for https://github.com/celery/celery/issues/5919. Celery currently can't recover from a Redis connection issue in a task producer, so we avoid using the result backend in tasks spawned in the web workers to avoid having to restart them when Redis goes down.

backend/inspirehep/submissions/tasks.py
drjova: we will get an exception on Sentry, which is currently the case anyhow no? The user is just getting an error with no information, right?

drjova: ping @harunurhan

PR #652, opened 2020-01-14
jobs: close_expired_jobs with --notify

harunurhan:
* Replaces existing `close_before`.

* This will be used as single command
from cron job that automatically closes expired jobs and notifies users


backend/inspirehep/records/cli.py
```diff
@@ -219,7 +218,12 @@ def close_before(deadline_before_date):
     for job in expired_jobs:
         record = JobsRecord(job.json, model=job)
         record["status"] = "closed"
-        record.update(dict(record))
+        record_dict = dict(record)
+        record.update(record_dict)
+
+        if notify:
+            send_job_deadline_reminder(record_dict)
```
drjova: I'd send the reminder when the record update has been commited to the db

harunurhan: it was like this before actually.. I'll revert.

PR #661, opened 2020-01-20
accounts: check if user is logged in on App mount

harunurhan:
* INSPIR-3056


ui/src/actions/user.js
```diff
@@ -23,6 +24,18 @@ function userLoginError(error) {
   };
 }
 
+function userLogutSuccess() {
```
drjova:
```suggestion
function userLogoutSuccess() {
```

PR #662, opened 2020-01-20
setup: lock on elasticsearch 7.1.x

pazembrz:


backend/pyproject.toml
```diff
@@ -11,6 +11,7 @@ celery = {git = "https://github.com/michamos/celery.git", branch = "fix-redis-co
 # Click in version 7 have broken progressbar in python 3
 # waiting for this PR to be merged https://github.com/pallets/click/pull/1163
 click = "<7.0.0"
+elasticsearch = "<=7.2.0"
```
drjova: isn't it dangerous? shall we add >=7.1.0,<7.2.0

pazembrz: elasticsearch main version is managed by invenio-search.
`invenio-search = {version = "~=1.2.3,<1.3.0", extras = [ "elasticsearch7",]}`

drjova: I'd atleast add ``<`` instead of ``<=``

pazembrz: It doesn't matter, actually I could even add <7.5.1 or <=7.5.0 as they didn't released anything between 7.1.0 and 7.5.1 https://pypi.org/project/elasticsearch/#history

drjova: IMHO I'd pin it properly to the version we want, maybe the will release 7.2 and will break

pazembrz: As you wish.

drjova:
@pazembrz the lock file is not updated :/

PR #663, opened 2020-01-21
pid: return deleted pids if authorized

harunurhan:
Now superuser and catalogers can display records that are marked
`deleted: true`.

* INSPIR-3068

### DIFF
--- a/backend/inspirehep/pidstore/converters.py
+++ b/backend/inspirehep/pidstore/converters.py
@@ -4,11 +4,36 @@
 #
 # inspirehep is free software; you can redistribute it and/or modify it under
 # the terms of the MIT License; see LICENSE file for more details.
+
+from functools import partial
+
 import structlog
-from invenio_records_rest.utils import PIDConverter
+from invenio_records.api import Record
+from invenio_records_rest.utils import PIDConverter, obj_or_import_string
 from werkzeug.routing import PathConverter
 
+from .resolvers import InspireResolver
+
 LOGGER = structlog.getLogger()
+
+
+class InspirePIDConverter(PIDConverter):
+    """Just to use our own Resolver class"""
+
+    def __init__(
+        self, url_map, pid_type, getter=None, record_class=None, object_type="rec"
+    ):
+        super(PIDConverter, self).__init__(url_map)
+        getter = obj_or_import_string(
+            getter,
+            default=partial(
+                obj_or_import_string(record_class, default=Record).get_record,
+                with_deleted=True,
+            ),
+        )
+        self.resolver = InspireResolver(
+            pid_type=pid_type, object_type=object_type, getter=getter
+        )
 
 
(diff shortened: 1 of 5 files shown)


backend/inspirehep/pidstore/converters.py
```diff
@@ -4,13 +4,38 @@
...
+        getter = obj_or_import_string(
+            getter,
+            default=partial(
+                obj_or_import_string(record_class, default=Record).get_record,
+                with_deleted=True,
+            ),
+        )
+        self.resolver = InspireResolver(
+            pid_type=pid_type, object_type=object_type, getter=getter
+        )
```
drjova: whey not something like?
```python
       super().__init__(url_map, pid_type, getter, record_class, object_type)
        self.resolver = InspireResolver(
            pid_type=pid_type, object_type=object_type, getter=getter
        )
```

backend/inspirehep/pidstore/converters.py
```diff
@@ -4,13 +4,38 @@
...
+from invenio_records_rest.utils import PIDConverter, obj_or_import_string
 from werkzeug.routing import PathConverter
 
+from .resolvers import InspireResolver
+
 LOGGER = structlog.getLogger()
 
 
+class InspirePIDConverter(PIDConverter):
+    """Just to use our own Resolver class"""
```
drjova: unnecessary ;)

backend/inspirehep/pidstore/resolvers.py
```diff
@@ -0,0 +1,46 @@
...
+from sqlalchemy.orm.exc import NoResultFound
+
+from inspirehep.accounts.api import is_superuser_or_cataloger_logged_in
+
+
+class InspireResolver(Resolver):
+    def resolve(self, pid_value):
+        """Just override single condition for deleted pids
+        Now if user is cataloger or superuser, PIDDeletedError is not raised
+        """
```
drjova:
```suggestion
        """Override single condition for deleted pids.
        Now if user is cataloger or superuser, PIDDeletedError is not raised.
        """
```

backend/inspirehep/pidstore/converters.py
harunurhan: I think it it, otherwise it's not easy to see why we override PIDConverter. Unless we name it `PidConverterWithCustomResolver` or sth, I'd keep this.

backend/inspirehep/pidstore/converters.py
harunurhan: because of the line  above: `getter = obj_or_import_string(...`

michamos:
I'm not sure we actually want this :/

harunurhan:
@michamos , you and @StellaCh should decide then.

harunurhan:
Agreed IRL to keep it as it

PR #665, opened 2020-01-21
files: integrate boto to inspirehep

monaawi:
* INSPIR-3049

### DIFF
--- a/backend/inspirehep/files/api/__init__.py
+++ b/backend/inspirehep/files/api/__init__.py
@@ -1,0 +1,9 @@
+# -*- coding: utf-8 -*-
+#
+# Copyright (C) 2019 CERN.
+#
+# inspirehep is free software; you can redistribute it and/or modify it under
+# the terms of the MIT License; see LICENSE file for more details.
+# flake8: noqa
+
+from ..proxies import current_s3_instance  # noqa: F401
--- a/backend/inspirehep/files/api/s3.py
+++ b/backend/inspirehep/files/api/s3.py
@@ -1,0 +1,133 @@
+# -*- coding: utf-8 -*-
+#
+# Copyright (C) 2019 CERN.
+#
+# inspirehep is free software; you can redistribute it and/or modify it under
+# the terms of the MIT License; see LICENSE file for more details.
+
+import structlog
+from botocore.exceptions import ClientError
+from flask import current_app
+
+LOGGER = structlog.getLogger()
+
+
+class S3:
+    def __init__(self, client, resource):
+        self.client = client
+        self.resource = resource
+
+    @staticmethod
+    def get_bucket(key):
+        """Return the bucket for the given file key.
+
+        :param key: the file key
+        :return: bucket: The corresponding bucket.
+        """
+        bucket = f"{current_app.config.get('S3_BUCKET_PREFIX')}{key[0]}"
+        return bucket
+
+    def upload_file(self, data, key, filename, mimetype, acl):
+        """Upload a file in s3 bucket with the given metadata
+
+        :param data: the data of the file.
+        :param key: the key of the file.
+        :param filename: the filename.
+        :param mimetype: the mimetype of the file.
+        :param acl: the access control list for the file.
+        :return: dict
+        """
+        try:
+            response = self.client.upload_fileobj(
+                data,
+                self.get_bucket(key),
+                key,
+                ExtraArgs={
+                    "ContentType": mimetype,
+                    "ACL": acl,
+                    "ContentDisposition": self.get_content_disposition(filename),
+                },
+            )
+            return response
+        except ClientError as e:
+            LOGGER.warning(exc=e, key=key)
+            raise
+
+    def delete_file(self, key):
+        """Deletes the given file from S3.
+
+        :param key: the key of the file.
+        :return: dict
+        """
+        try:
+            response = self.client.delete_object(Bucket=self.get_bucket(key), Key=key)
+            return response
+        except ClientError as e:
+            LOGGER.warning(exc=e, key=key)
+            raise
+
+    @staticmethod
+    def get_file_url(key):
+        """Returns the S3 link for the file.
+
+        :param key: the key of the file.
+        :return: string: the s3 link for the file
+        """
+        return f"{current_app.config.get('S3_HOSTNAME')}/{S3.get_bucket(key)}/{key}"
+
+    @staticmethod
+    def get_content_disposition(filename):
+        return f'attachment; filename="{filename}"'
+
+    def replace_file_metadata(self, key, filename, mimetype, acl):
+        """Updates the metadata of the given file.
+
+        :param key: the file key.
+        :param filename: the new filename.
+        :param mimetype: the new mimetype.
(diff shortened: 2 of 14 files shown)


backend/inspirehep/files/ext.py
```diff
@@ -0,0 +1,92 @@
...
+# inspirehep is free software; you can redistribute it and/or modify it under
+# the terms of the MIT License; see LICENSE file for more details.
+
+import boto3
+from botocore.exceptions import UnknownServiceError
+from flask import _app_ctx_stack as stack
+from flask import current_app
+
+
+class InspireS3(object):
```
harunurhan: why do we just call it `S3Extention` ?

backend/inspirehep/files/ext.py
```diff
@@ -0,0 +1,92 @@
...
+        if hasattr(ctx, "boto3_cns"):
+            for c in ctx.boto3_cns:
+                con = ctx.boto3_cns[c]
+                if hasattr(con, "close") and callable(con.close):
+                    ctx.boto3_cns[c].close()
+
+    @property
+    def resources(self):
+        c = self.connections
+        return {k: v for k, v in c.items() if hasattr(c[k].meta, "client")}
```
harunurhan: not so readable, maybe a better naming with variables would help.

backend/inspirehep/files/api/s3.py
```diff
@@ -0,0 +1,130 @@
...
+        """
+        try:
+            self.resource.Object(self.get_bucket(key), key).load()
+        except ClientError as e:
+            if e.response["Error"]["Code"] == "404":
+                return False
+            else:
+                LOGGER.warning(exc=e, key=key)
+                raise
+        else:
```
harunurhan: could  you return in `try` block instead?

backend/inspirehep/files/api/s3.py
```diff
@@ -0,0 +1,130 @@
...
+        :param acl: the new access control list for the file.
+        :return: dict
+        """
+        try:
+            response = self.client.copy_object(
+                ACL=acl,
+                Bucket=self.get_bucket(key),
+                Key=key,
+                CopySource={"Bucket": self.get_bucket(key), "Key": key},
+                ContentDisposition=f'attachment; filename="{filename}"',
```
harunurhan: we can move this to a function and share with `upload_file`

harunurhan:
It looks like we have to `s3 = S3()` whenever we need to use `s3`. AFAIK, flask app extensions are supposed to be used for this as well, so instead of keeping only boto s3 client there, we should keep instance of our s3 client that wraps boto one.

backend/tests/integration/files/test_api_s3.py
```diff
@@ -0,0 +1,117 @@
...
+    filename = "file.txt"
+    mimetype = "text/*"
+    acl = "public-read"
+    expected_extraargs = {
+        "ContentType": mimetype,
+        "ACL": acl,
+        "ContentDisposition": f'attachment; filename="{filename}"',
+    }
+    s3 = S3()
+    with patch.object(s3, "client") as mock_client:
```
michamos: have you looked into using https://github.com/spulec/moto instead of mocking the client manually?

backend/inspirehep/files/ext.py
monaawi: Because all the other extensions are called InspireSomething, so I think we should keep it consistent

backend/inspirehep/files/proxies.py
```diff
@@ -0,0 +1,11 @@
...
+#
+# Copyright (C) 2019 CERN.
+#
+# inspirehep is free software; you can redistribute it and/or modify it under
+# the terms of the MIT License; see LICENSE file for more details.
+
+from flask import current_app
+from werkzeug.local import LocalProxy
+
+s3_instance = LocalProxy(lambda: current_app.extensions["inspirehep-s3"].s3_instance)
```
michamos:
```suggestion
current_s3_instance = LocalProxy(lambda: current_app.extensions["inspirehep-s3"].s3_instance)
```
to keep the Flask naming convention?

backend/inspirehep/files/api/__init__.py
```diff
@@ -0,0 +1,9 @@
+# -*- coding: utf-8 -*-
+#
+# Copyright (C) 2019 CERN.
+#
+# inspirehep is free software; you can redistribute it and/or modify it under
+# the terms of the MIT License; see LICENSE file for more details.
+# flake8: noqa
+
+from .s3 import S3  # noqa: F401
```
michamos: re-export the proxy instance here?

backend/tests/integration/files/test_api_s3.py
```diff
@@ -0,0 +1,114 @@
...
+
+from inspirehep.files.api import S3
+from inspirehep.files.proxies import s3_instance
+
+BUCKET = "inspire-files-b"
+KEY = "b50c2ea2d26571e0c5a3411e320586289fd715c2"
+
+
+def test_upload_file(base_app, appctx, s3, create_s3_bucket):
+    s3 = S3(s3[0], s3[1])
```
michamos: use `s3_instance` in all tests instead?

monaawi: I couldn't find a way to do that and pass it the mocked client and resource. It is easier this way

backend/inspirehep/files/api/s3.py
```diff
@@ -0,0 +1,133 @@
...
+        """
+        try:
+            response = self.client.delete_object(Bucket=self.get_bucket(key), Key=key)
+            return response
+        except ClientError as e:
+            LOGGER.warning(exc=e, key=key)
+            raise
+
+    @staticmethod
+    def get_file_link(key):
```
michamos:
```suggestion
    def get_file_url(key):
```

backend/inspirehep/files/ext.py
```diff
@@ -0,0 +1,103 @@
...
+import boto3
+from botocore.exceptions import UnknownServiceError
+from flask import _app_ctx_stack as stack
+from flask import current_app
+
+from inspirehep.files.api.s3 import S3
+
+
+class InspireS3(object):
+    """Stores boto3 conectors inside Flask's application context
```
michamos:
```suggestion
    """Store boto3 connectors inside Flask's application context
```

backend/tests/integration/files/test_api_s3.py
```diff
@@ -0,0 +1,112 @@
...
+    assert result["ContentDisposition"] == f'attachment; filename="{filename}"'
+    assert result["ContentType"] == mimetype
+    assert result["Metadata"] == {}
+
+
+def test_get_file_metadata(base_app, appctx, s3, create_s3_bucket, create_s3_file):
+    expected_metadata = {"foo": "bar"}
+    create_s3_bucket(BUCKET)
+    create_s3_file(BUCKET, KEY, "this is my data", expected_metadata)
+    metadata = current_s3_instance.get_file_metadata(KEY)["Metadata"]
```
michamos: maybe `get_file_metadata` should return the `Metadata` key only?

PR #668, opened 2020-01-22
config: Add index prefix

pazembrz:
INSPIR-3007


backend/inspirehep/serializers.py
```diff
@@ -67,10 +68,12 @@ def serialize_search(
...
 
     def _get_sort_options(self):
-        sort_options = current_app.config["RECORDS_REST_SORT_OPTIONS"].get(
-            self.index_name
-        )
-
+        if self.index_name:
+            alias_name = build_alias_name(self.index_name, app=current_app)
+        else:
+            alias_name = None
```
drjova:
```suggestion
            alias_name = None
            if self.index_name:
                alias_name = build_alias_name(self.index_name, app=current_app)
```

backend/inspirehep/config.py
```diff
@@ -202,6 +202,7 @@ def annotate(self, task):
 INDEXER_DEFAULT_DOC_TYPE = "_doc"
 INDEXER_BULK_REQUEST_TIMEOUT = 900
 INDEXER_REPLACE_REFS = False
+SEARCH_INDEX_PREFIX = "inspire-"
```
drjova: this will screw up prod in the next deployment

backend/inspirehep/search/api.py
```diff
@@ -43,7 +51,7 @@ def get_source(self, uuid, **kwargs):
         :type uuid: UUID
         :returns: dict
         """
-        return es.get_source(index=self.Meta.index, id=uuid, **kwargs)
+        return es.get_source(index=self.alias, id=uuid, **kwargs)
```
drjova: this means that we'll search in all records?

pazembrz: Discussed IRL

backend/inspirehep/search/api.py
```diff
@@ -24,6 +24,14 @@
 class SearchMixin(object):
     """Mixin that adds helper functions to ElasticSearch DSL classes."""
 
+    @property
+    def base_index(self):
```
drjova: ``original_index`` or just ``index_name``?

pazembrz: Invenio already uses _original_index and _index_name. I didn't want to create property too existing ones. I think that base_index is quite understandable - base like without_prefix one.

backend/inspirehep/records/indexer/base.py
```diff
@@ -62,6 +62,8 @@ def _process_bulk_record_for_index(
         index_from_record, doc_type_from_record = self.record_to_index(record)
         if not index:
             index = index_from_record
+
+        index, doc_type = self._prepare_index(index, doc_type or doc_type_from_record)
```
drjova: why we need the _prepare_index? :/

pazembrz: Because _prepare_index is building index name with prefix etc: https://github.com/inveniosoftware/invenio-indexer/blob/8d35a88b79c9008e2ff5ecbff0c74ec915e428e4/invenio_indexer/api.py#L318-L320
I tried not to replace main logic from Invenio-search to be as much compatible as possible.

PR #673, opened 2020-01-22
accounts: addition of invenio-oauthclient rest

drjova:


ui/src/user/components/SignUpPage.jsx
```diff
@@ -0,0 +1,57 @@
+import React, { useState } from 'react';
+import { Row, Card, Input, Button } from 'antd';
+import { Field, Form, Formik } from 'formik';
+import DocumentHead from '../../common/components/DocumentHead';
+
+function renderFormInput({ field, form, ...props }) {
+  return <Input {...field} {...props} />;
+}
+
+function SignUpPage({ onSignUpClick }) {
```
harunurhan: you can have a look at `SubscribeJobsForm` as an example which is better than `LocalLogin` one (using `TextField` and validation) since it's user facing one.

ui/yarn.lock
```diff
@@ -8677,6 +8683,14 @@ react-app-rewire-sass-rule@^2.1.1:
     node-sass "^4.9.0"
     sass-loader "^7.0.3"
 
+react-app-rewired@1.6.2:
```
harunurhan: ?

ui/src/user/containers/__tests__/SignUpPageContainer.test.jsx
```diff
@@ -0,0 +1,28 @@
...
+      <Provider store={store}>
+        <SignUpPageContainer />
+      </Provider>
+    );
+    const onSignUpClick = wrapper.find(SignUpPage).prop('onSignUpClick');
+    const dataToPass = {
+      email: 'jessica@jones.com',
+    };
+    onSignUpClick(dataToPass);
+    expect(userSignUp).toHaveBeenCalledWith(dataToPass);
```
harunurhan: Lately we started dropping this approach to test if container is dispatching the right action or not. Because this will pass even if you do `userSignUp(data)` without `dispatch(userSignUp(data))`.

Instead you can add `USER_SIGN_UP_REQUEST` (would be useful only for debugging and testing), and assert this here via `store.getActions`. I know that that goes further by testing the action creator's logic (wrong), but I think it's still better than this one.

Actually bonus point, you can set `isSigningUp` in the state and manage it there instead of setting `loading: true` in the form `onSubmit` and never changing it back to `false`. Which would cause problems it `userSignUp` ends up throwing an error, user would stay in the `SignUpPage` look at the `loading` button, without any feedback :/

backend/inspirehep/accounts/handlers.py
```diff
@@ -0,0 +1,70 @@
...
+#
+# inspirehep is free software; you can redistribute it and/or modify it under
+# the terms of the MIT License; see LICENSE file for more details.
+
+from flask import current_app, jsonify, render_template
+from flask_login import current_user
+from invenio_db import db
+from invenio_oauthclient.utils import oauth_link_external_id
+
+RESPONSE_SIGNUP = "Successfully signed up."
```
harunurhan: So these hardcoded in `invenio`? :(

backend/inspirehep/accounts/views.py
```diff
@@ -5,25 +5,45 @@
...
+def sign_up_required():
+    payload = {"user_needs_sign_up": True}
+    return render_template("accounts/postmessage.html", payload=payload)
+
+
+@blueprint.route("/signup", methods=["POST"])
+def sign_up_user():
+    try:
+        current_oauthclient.signup_handlers["orcid"]["view"]()
+    except IntegrityError:
```
harunurhan: should we make it clear that `IntegrityError` is raised when the email already exists? Somebody could read the code and ask "Why" 😄

backend/tests/integration/accounts/test_views.py
```diff
@@ -19,3 +25,84 @@ def test_me_returns_user_data_if_logged_in(api_client, create_user):
...
+
+@mock.patch("flask_login.utils._get_user")
+def test_sign_up_user_success(mock_current_user, api_client, create_user):
+    """It's mocking current user because invenio handlers need a lot of things to
+    setup in order to make it properly work and we don't want to test this functinality."""
+
+    def return_true():
+        return True
+
+    current_oauthclient.signup_handlers["orcid"] = {"view": return_true}
```
harunurhan: maybe it's better to also "teardown" this in order not to affect other test cases, I see that right now the other tests are overriding this, but it might be confusing when we start adding new tests.

ui/src/user/components/__tests__/SignUpPage.test.jsx
```diff
@@ -0,0 +1,30 @@
...
+
+import SignUpPage from '../SignUpPage';
+
+jest.mock('../../../actions/user');
+
+describe('SignUpPage', () => {
+  it('renders page', () => {
+    const wrapper = shallow(
+      <SignUpPage loading={false} onSubmit={jest.fn()} />
+    );
```
harunurhan: needs `dive()` to assert form fields, as you can see in the snapshot they are not there.

ui/src/user/components/__tests__/SignUpPage.test.jsx
```diff
@@ -0,0 +1,36 @@
...
+    expect(wrapper).toMatchSnapshot();
+    expect(formWrapper.dive()).toMatchSnapshot();
+  });
+
+  it('renders page with loading', () => {
+    const wrapper = shallow(<SignUpPage loading onSubmit={jest.fn()} />);
+    const formWrapper = wrapper.find(Formik);
+
+    expect(wrapper).toMatchSnapshot();
+    expect(formWrapper.dive()).toMatchSnapshot();
```
harunurhan: it's OK like that, but usually it's better to separate this to another component (`SignUpForm`) when test gets like this.

drjova:
@harunurhan updated

PR #674, opened 2020-01-23
ui: display banner when job is pending

miguelgrc:
* INSPIR-3092

### DIFF
--- a/ui/src/jobs/components/JobStatusAlert.jsx
+++ b/ui/src/jobs/components/JobStatusAlert.jsx
@@ -1,0 +1,27 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import { Alert } from 'antd';
+
+const ALERT_TYPES_BY_STATUS = { pending: 'warning', closed: 'error' };
+
+function JobStatusAlert({ status }) {
+  const shouldDisplayAlert = ALERT_TYPES_BY_STATUS[status] != null;
+
+  return (
+    shouldDisplayAlert && (
+      <div className="mb2">
+        <Alert
+          type={ALERT_TYPES_BY_STATUS[status]}
+          message={<span>This job is {status}!</span>}
+          showIcon={false}
+        />
+      </div>
+    )
+  );
+}
+
+JobStatusAlert.propTypes = {
+  status: PropTypes.string,
+};
+
+export default JobStatusAlert;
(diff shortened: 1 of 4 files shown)


ui/src/jobs/components/JobStatusAlert.jsx
```diff
@@ -0,0 +1,28 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import { Row, Col, Alert } from 'antd';
+import { JOB_STATUSES } from '../../common/constants';
+
+function JobStatusAlert({ status }) {
+  const types = { pending: 'warning', closed: 'error' };
```
harunurhan: `const ALERT_TYPES_BY_STATUS = ...` and move it above the component.

ui/src/jobs/components/JobStatusAlert.jsx
```diff
@@ -0,0 +1,28 @@
...
+
+  return (
+    (status === 'pending' || status === 'closed') && (
```
harunurhan: extract above as `const shouldDisplayAlert = ALERT_TYPES_BY_STATUS [status] != null`

ui/src/jobs/components/JobStatusAlert.jsx
```diff
@@ -0,0 +1,28 @@
...
+      <Row className="mb2">
+        <Col>
```
harunurhan: `Row` and `Col` is the part of detail page layout, they should stay there, you can add `mb2` to `Alert` directly.

ui/src/jobs/components/JobStatusAlert.jsx
```diff
@@ -0,0 +1,28 @@
...
+            showIcon={false}
+          />
+        </Col>
+      </Row>
+    )
+  );
+}
+
+JobStatusAlert.propTypes = {
+  status: PropTypes.oneOf(JOB_STATUSES),
```
harunurhan: well that's unnecessary to be honest, you can just `PropTypes.string`

ui/src/jobs/components/JobStatusAlert.jsx
miguelgrc: Shouldn't the `mb2` stay in the detail page `Row` then?

PR #678, opened 2020-01-24
search: refactor scripted aggregations

michamos:
* Scripted aggregations now use built-in methods where possible and are
  written on several lines for better readability.

### DIFF
--- a/backend/inspirehep/search/facets.py
+++ b/backend/inspirehep/search/facets.py
@@ -11,6 +11,8 @@ from elasticsearch_dsl.query import Q, Range
 from flask import current_app, request
 from invenio_records_rest.facets import range_filter
 
+from inspirehep.search.utils import minify_painless
+
 
 def range_author_count_filter(field):
     """Range filter for returning records within the corresponding range(s)."""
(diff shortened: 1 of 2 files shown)


backend/inspirehep/search/facets.py
```diff
@@ -310,16 +339,36 @@ def citations_by_year():
         for key, val in current_app.config["HEP_COMMON_FILTERS"].items()
         if key not in excluded_filters
     }
+    map_script = """
+        def add(def x, def y) {
+            x + y
+        }
+        def years = params._source.citations_by_year != null ? params._source.citations_by_year : [];
+        for (element in years) {
+            state.merge(element.year.toString(), element.count, this::add)
```
monaawi: I think it would have looked nicer to use a lambda here instead of ```this:add``` since you are defining add in both places.

michamos: I agree. I initially tried 
```
            state.merge(element.year.toString(), element.count, (old, new) -> old + new)
```
which looks a lot like the 4th example in https://www.elastic.co/guide/en/elasticsearch/painless/7.1/painless-lambdas.html but couldn't make it work (syntax error on the `new` in the lambda argument) so used this as an alternative.

monaawi: I think it is because ```new``` is a reserved word, should have worked otherwise

michamos: :man_facepalming:

PR #682, opened 2020-01-25
files 

monaawi:


backend/inspirehep/files/api/mixin.py
```diff
@@ -0,0 +1,133 @@
...
+            return data
+
+        if "deleted" in data and data["deleted"]:
+            LOGGER.info("Record is deleted, skipping files.", uuid=self.id)
+            return data
+
+        documents = data.pop("documents", [])
+        figures = data.pop("figures", [])
+
+        self.pop("documents", None)
```
drjova: this logic is related to literature only, I'd move it to Literature

backend/inspirehep/files/api/mixin.py
```diff
@@ -0,0 +1,133 @@
...
+                    figure=figure.get("key"),
+                )
+        return builder.record.get("figures")
+
+    @staticmethod
+    def download_file_from_url(url):
+        download_url = (
+            url
+            if url.startswith("http")
+            else f"{current_app.config['PREFERRED_URL_SCHEME']}://{current_app.config['SERVER_NAME']}{url}"
```
drjova: https://github.com/inspirehep/inspirehep/blob/master/backend/inspirehep/utils.py#L17

backend/inspirehep/files/api/mixin.py
```diff
@@ -0,0 +1,133 @@
...
+    ):
+        file_data = self.download_file_from_url(url)
+        new_key = self.hash_data(file_data)
+        mimetype = magic.from_buffer(file_data, mime=True)
+        size = len(file_data)
+        filename = filename or key
+        if current_s3_instance.file_exists(new_key):
+            LOGGER.info("Replacing file metadata", key=new_key)
+            current_s3_instance.replace_file_metadata(
+                new_key, filename, mimetype, "public-read"
```
pazembrz: can you put permissions in config file and not hard-code it?

backend/inspirehep/files/api/mixin.py
```diff
@@ -0,0 +1,133 @@
...
+
+        if "deleted" in data and data["deleted"]:
+            LOGGER.info("Record is deleted, skipping files.", uuid=self.id)
+            return data
+
+        documents = data.pop("documents", [])
+        figures = data.pop("figures", [])
+
+        self.pop("documents", None)
+        self.pop("figures", None)
```
pazembrz: We call self.clear() on InspireRecord.update() which is called after this update so this is unnecessary.

backend/inspirehep/files/api/mixin.py
```diff
@@ -0,0 +1,133 @@
+import hashlib
```
drjova: ``.../api/mixins.py``

backend/tests/integration/files/test_api_mixin.py
```diff
@@ -0,0 +1,382 @@
...
+                "filename": "file1.pdf",
+            }
+        ],
+    }
+    create_record("lit", data=data)
+    assert s3.file_exists(expected_document_key) is False
+
+
+@pytest.mark.vcr()
+def test_update_record_with_documents_and_figures(
```
drjova: what about update which removes some of the documents and figures and an update that adds more?

backend/inspirehep/files/api/mixin.py
pazembrz: That would be terrible behavior. Using logic from inside mixin here? This is why you create mixin to be able to just add/remove it to add/remove functionality without worrying about additional changes in your code...

monaawi: I actually agree with Pawel, I wouldn't put any files logic in literature

drjova: Having a mixin handling just literature related files makes no sense though

pazembrz: so in this case there should be a LiteratureFilesMixin (inheriting from FilesMixin). Otherwise there will be same mess we had in next.

backend/inspirehep/files/api/mixins.py
```diff
@@ -0,0 +1,126 @@
...
+        new_key = self.hash_data(file_data)
+        mimetype = magic.from_buffer(file_data, mime=True)
+        size = len(file_data)
+        filename = filename or key
+        acl = current_app.config["S3_FILE_ACL"]
+        if current_s3_instance.file_exists(new_key):
+            LOGGER.info("Replacing file metadata", key=new_key)
+            current_s3_instance.replace_file_metadata(new_key, filename, mimetype, acl)
+        else:
+            LOGGER.info("Uploading file to s3", key=new_key)
```
drjova: could you please add the record id as well

backend/tests/integration/records/test_utils.py
```diff
@@ -0,0 +1,49 @@
...
+            "http://localhost:5000/record/1759380/files/channelxi3.png",
+            status_code=200,
+            content=expected_content,
+        )
+        result = download_file_from_url(url)
+        assert result == expected_content
+
+
+def test_download_file_from_url_with_full_url(base_app, db):
+    url = "https://inspirehep.net/record/1759380/files/channelxi3.png"
```
drjova: why not record?

PR #683, opened 2020-01-29
accounts: move local login and logout to hep

miguelgrc:
* INSPIR-3088

### DIFF
--- a/backend/inspirehep/config.py
+++ b/backend/inspirehep/config.py
@@ -151,7 +151,7 @@ SECRET_KEY = "CHANGE_ME"
 #: It should be changed before deploying.
 MAX_CONTENT_LENGTH = 100 * 1024 * 1024  # 100 MiB
 #: Max upload size for form data via application/mulitpart-formdata.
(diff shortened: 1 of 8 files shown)


backend/inspirehep/config.py
```diff
@@ -151,7 +151,7 @@ def annotate(self, task):
 #: It should be changed before deploying.
 MAX_CONTENT_LENGTH = 100 * 1024 * 1024  # 100 MiB
 #: Max upload size for form data via application/mulitpart-formdata.
-SESSION_COOKIE_SECURE = True
+SESSION_COOKIE_SECURE = False
```
harunurhan: Can you make sure this is set to `True` in kubernetes repo?

miguelgrc: Looks like it's not

harunurhan: Then you can create a PR there first.

https://gitlab.cern.ch/inspire/kubernetes/blob/master/hep/base/config/inspirehep.cfg

PR #685, opened 2020-01-30
ui: open conferences in new tab from drawer

miguelgrc:
* Fix for INSPIR-2999


ui/src/conferences/components/ConferenceItem.jsx
```diff
@@ -46,7 +46,11 @@ class ConferenceItem extends Component {
       >
         <Row type="flex">
           <Col>
-            <Link className="f5" to={`${CONFERENCES}/${recordId}`}>
+            <Link
```
harunurhan: there is no point to render `Link` if you set the `target` to `_blank`, because it's for client side navigations. But I guess it makes it simpler like that. However make sure this is not dispatching any location change on the parent page (submissions) in that case.

miguelgrc: No, it doesn't dispatch anything on the parent page

PR #686, opened 2020-01-30
ui: hot fix for lit citation summary bug

harunurhan:
CitationSummaryBoxContainer keeps getting rerendering for some reason
(it will be investigated further) when user is logged in. This
rerendering was triggering endless fetch loop.

### DIFF
--- a/ui/src/literature/containers/CitationSummaryBoxContainer.jsx
+++ b/ui/src/literature/containers/CitationSummaryBoxContainer.jsx
@@ -2,7 +2,6 @@ import { connect } from 'react-redux';
 
 import { LITERATURE_NS } from '../../reducers/search';
 import CitationSummaryBox from '../components/CitationSummaryBox';
-import { convertAllImmutablePropsToJS } from '../../common/immutableToJS';
 import { fetchCitationSummary } from '../../actions/citations';
 
 const stateToProps = state => ({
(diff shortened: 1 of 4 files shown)


ui/src/literature/containers/CitationSummaryBoxContainer.jsx
```diff
@@ -2,7 +2,7 @@ import { connect } from 'react-redux';
 
 import { LITERATURE_NS } from '../../reducers/search';
 import CitationSummaryBox from '../components/CitationSummaryBox';
-import { convertAllImmutablePropsToJS } from '../../common/immutableToJS';
+// import { convertAllImmutablePropsToJS } from '../../common/immutableToJS';
```
drjova: even if it's hot fix should we keep it like that?

ui/src/literature/containers/CitationSummaryBoxContainer.jsx
```diff
@@ -11,10 +11,8 @@ const stateToProps = state => ({
...
   onQueryChange(query) {
-    dispatch(fetchCitationSummary(query));
+    dispatch(fetchCitationSummary(query.toJS()));
   },
 });
 
-export default connect(stateToProps, dispatchToProps)(
-  convertAllImmutablePropsToJS(CitationSummaryBox)
-);
+export default connect(stateToProps, dispatchToProps)(CitationSummaryBox);
```
drjova: can we just use ``convertAllImmutablePropsToJS``?

ui/src/literature/containers/CitationSummaryBoxContainer.jsx
harunurhan: ah left over

ui/src/literature/containers/CitationSummaryBoxContainer.jsx
harunurhan: no this is the whole point, we need to pass immutable map which is used for equality check, otherwise these mysterious rerenders ends up generating different object references of the same query and ends up triggering `onQueryChange` callback.

PR #690, opened 2020-02-03
Use explicit end in continuous migration

michamos:
* INSPIR-3142

### DIFF
--- a/backend/inspirehep/migrator/api.py
+++ b/backend/inspirehep/migrator/api.py
@@ -25,39 +25,40 @@ def continuous_migration():
     """Task to continuously migrate what is pushed up by Legacy."""
     # XXX: temp redis url when we use continuous migration in kb8s
     redis_url = current_app.config.get("MIGRATION_REDIS_URL")
-    queue = "continuous_migration"
-
     if redis_url is None:
         redis_url = current_app.config.get("CACHE_REDIS_URL")
 
-    LOGGER.debug("Connected to REDIS", redis_url=redis_url, queue=queue)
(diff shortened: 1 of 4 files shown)


backend/tests/integration-async/migrator/test_migrator_api.py
```diff
@@ -335,19 +338,14 @@ def test_continuous_migration_with_invalid_control_number(
...
+    assert cache.llen("legacy_records") == 3
 
     with pytest.raises(ValueError):
         continuous_migration()
 
-    record_citer = InspireRecord.get_record_by_pid_value(citer_control_number, "lit")
-    record_citer_es = InspireSearch.get_record_data_from_es(record_citer)
-    result_citer_control_number = record_citer_es["control_number"]
-
-    assert citer_control_number == result_citer_control_number
```
michamos: I removed this as the migration of records will be triggered later when encountering the `END` message. I don't think it matters in practice as exceptions at this stage never happen.

backend/inspirehep/migrator/api.py
```diff
@@ -25,39 +25,40 @@ def continuous_migration():
...
+    if not lock.acquire(blocking=False):
+        LOGGER.info("Continuous_migration already executed. Skipping.")
+        return
+
+    try:
+        num_of_records = r.llen("legacy_records")
+        LOGGER.info("Starting migration of records.", records_total=num_of_records)
+
+        while message:
+            if message[0] == b"END":
```
drjova: you may ignore it if you like, but we're using ``message[0]`` in multiple places, maybe assign it?

PR #695, opened 2020-02-04
records: update import view to be case insensitive for DOIs

pazembrz:
* INSPIR-3141

### DIFF
--- a/backend/inspirehep/records/api/literature.py
+++ b/backend/inspirehep/records/api/literature.py
@@ -362,7 +362,7 @@ def import_article(identifier):
 
     elif is_doi(identifier):
         pid_type = "doi"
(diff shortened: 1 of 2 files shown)


backend/inspirehep/records/api/literature.py
```diff
@@ -362,7 +362,7 @@ def import_article(identifier):
 
     elif is_doi(identifier):
         pid_type = "doi"
-        pid_value = normalize_doi(identifier)
+        pid_value = normalize_doi(identifier).lower()
```
monaawi: is it only for doi though? shouldn't it work for arxiv as well? If yes, I would just do ```identifier.lower``` before this

PR #697, opened 2020-02-05
files: change the original_url logic 

monaawi:
* Populate original_url with url only if it is not already populated and if it url is not an s3 url and the url is not relative
* Populating the original_url is used in the afs cache in the migration

### DIFF
--- a/backend/inspirehep/records/api/literature.py
+++ b/backend/inspirehep/records/api/literature.py
@@ -325,14 +325,20 @@ class LiteratureRecord(
             current_s3_instance.upload_file(
                 BytesIO(file_data), new_key, filename, mimetype, acl
             )
-        return {
+        result = {
             "key": new_key,
-            "original_url": original_url or url,
             "filename": filename,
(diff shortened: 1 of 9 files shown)


backend/inspirehep/records/api/literature.py
```diff
@@ -327,7 +327,6 @@ def add_file(
             )
         return {
             "key": new_key,
-            "original_url": original_url or url,
```
michamos: so you never populate `original_url`? not sure I understand.

drjova: in workflows will work, what about migration? Is ``original_url`` populated correctly? cc @michamos

monaawi: no, we leave the ```original_url``` as it is. We don't change the ```original_url``` to be the ```url``` as this wrong - we would be storing afs and ```/api/files..``` in it

monaawi: yeah, I had also checked the workflows, we populate the ```original_url``` there. Not sure about the migration though

PR #700, opened 2020-02-06
files: add create_buckets cli

harunurhan:


backend/inspirehep/files/api/s3.py
```diff
@@ -18,14 +18,28 @@ def __init__(self, client, resource):
...
 
         :param key: the file key
         :return: bucket: The corresponding bucket.
         """
-        bucket = f"{current_app.config.get('S3_BUCKET_PREFIX')}{key[0]}"
-        return bucket
+        return S3.get_prefixed_bucket(key[0])
+
+    @staticmethod
+    def get_prefixed_bucket(bucket_without_prefix):
```
michamos:
```suggestion
    def get_prefixed_bucket_name(bucket_without_prefix):
```

harunurhan: well bucket name, called `bucket` in other parts, that's why I kept it like that.

backend/inspirehep/files/cli.py
```diff
@@ -0,0 +1,48 @@
...
+
+
+@files.command(help="Creates S3 buckets for [0-9] and [a-f]")
+@with_appcontext
+def create_buckets():
+
+    click.secho("Creating buckets")
+
+    for bucket in BUCKETS:
+        current_s3_instance.create_bucket(bucket)
```
drjova: I'd print also which bucket created

backend/inspirehep/files/cli.py
```diff
@@ -0,0 +1,49 @@
...
+        click.secho(f"Created bucket: ${bucket}")
```
drjova: this is not JS :P

```suggestion
        click.secho(f"Created bucket: {bucket}")
```

PR #702, opened 2020-02-07
ui: update text in user feedback form when there is adblock

monaawi:
* INSPIR-3079


ui/src/common/components/UserFeedback/UserFeedback.jsx
```diff
@@ -100,12 +100,22 @@ class UserFeedback extends Component {
...
               description={
-                <span>
-                  To send us your feedback, please disable your adblocker or
-                  DoNotTrack and refresh the page or send us your feedback using
-                  the{' '}
-                  <ExternalLink href={SURVEY_LINK}>feedback form</ExternalLink>
-                </span>
+                <>
+                  <div>
+                    To send us your feedback, please disable your adblocker or
```
harunurhan: let's change `div`s to `p`s I think it's more appropriate here.

PR #706, opened 2020-02-10
authors: add orcid-push setting

harunurhan:
* INSPIR-3059


backend/inspirehep/accounts/views.py
```diff
@@ -72,3 +74,18 @@ def login():
...
     logout_user()
     return jsonify({"message": "Successfully logged out"}), 200
+
+
+@blueprint.route("/settings/orcid-push", methods=["PUT"])
+@login_required
+def set_orcid_push_setting():
+    orcid_account = get_current_user_remote_orcid_account()
+
+    data = request.get_json()
```
drjova:
```suggestion
    data = request.json
```

backend/tests/integration/accounts/test_views.py
```diff
@@ -20,17 +21,31 @@ def test_me_returns_error_when_not_logged_in(api_client):
...
-    user = create_user(role="user")
+    user = create_user(role="user", orcid="0000-0001-8829-5461")
     login_user_via_session(api_client, email=user.email)
-    payload = {"data": {"email": user.email, "roles": ["user"]}}
+    payload = {
+        "data": {
+            "email": user.email,
+            "roles": ["user"],
+            "orcid": "0000-0001-8829-5461",
+            "allow_orcid_push": None,
```
drjova: this is weird, allow push should be False or True based on the account setup

backend/tests/integration/conftest.py
```diff
@@ -304,8 +304,8 @@ def test_needs_user(base_app, create_user)
             user = create_user()
     """
 
-    def _create_user(role="user", orcid=None, email=None):
-        return UserFactory(role=role, orcid=orcid, email=email)
+    def _create_user(role="user", orcid=None, email=None, allow_push=None):
+        return UserFactory(role=role, orcid=orcid, email=email, allow_push=allow_push)
```
drjova: True or False

ui/src/actions/user.js
```diff
@@ -106,7 +109,7 @@ export function userLocalLogin(credentials) {
       const response = await http.post('/accounts/login', credentials);
       dispatch(userLoginSuccess(response.data));
     } catch (error) {
-      const errorPayload = httpErrorToActionPayload(error)
+      const errorPayload = httpErrorToActionPayload(error);
```
drjova: how this is not caught from the linter?

ui/src/authors/components/OrcidPushSettingMessage.jsx
```diff
@@ -0,0 +1,111 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+
+import ExternalLink from '../../common/components/ExternalLink';
+import OrcidProfileLink from '../../common/components/OrcidProfileLink';
+
+function OrcidPushSettingMessage({ orcid, enabled, authorBAI }) {
```
drjova: just a suggestion since it's mostly hardcoded text, we can separate this if blocks

ui/src/common/components/OrcidProfileLink.jsx
```diff
@@ -0,0 +1,20 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+
+import ExternalLink from './ExternalLink';
+
+function OrcidProfileLink({ children, orcid, className }) {
```
drjova: OrcidProfileLinkOrChildren? something that makes it clearer that you can pass the children

ui/src/literature/components/__tests__/__snapshots__/CiteModalAction.test.jsx.snap
```diff
@@ -31,7 +31,7 @@ exports[`CiteModalAction renders with all props 1`] = `
     title="Cite Article"
     transitionName="zoom"
     visible={false}
-    width="50%"
+    width={520}
```
drjova: why we have to change this?

ui/src/reducers/user.js
```diff
@@ -46,6 +52,25 @@ const userReducer = (state = initialState, action) => {
...
+    case USER_SET_ORCID_PUSH_SETTING_REQUEST:
+      return state
+        .set('isUpdatingOrcidPushSetting', true)
+        .set(
+          'updateOrcidPushSettingError',
+          initialState.get('updateOrcidPushSettingError')
+        );
+    case USER_SET_ORCID_PUSH_SETTING_ERROR:
+      return state
+        .set('isUpdatingOrcidPushSetting', false)
```
drjova:
```suggestion
        .set(initialState.isUpdatingOrcidPushSetting)
``` ?

ui/src/actions/user.js
harunurhan: it's prettier, apparently pre-commit doesn't always work properly :/

ui/src/authors/components/OrcidPushSettingMessage.jsx
harunurhan: I was thinking so, but then thought it's easy to read and understand what's going on, you don't think so?

ui/src/common/components/OrcidProfileLink.jsx
harunurhan: `OrcidProfileLinkOrChildren` indicates that it renders the link or the `children`, which is not the case the here.

ui/src/literature/components/__tests__/__snapshots__/CiteModalAction.test.jsx.snap
harunurhan: I realized that it was crap on mobile, it's better to use antd default size, which is overriden automatically on mobile.

ui/src/reducers/user.js
harunurhan: I think ti's better to be explict that we want it to set false after an error.

drjova: we have this in other places though, just for consistency but that's fine

harunurhan: well we shouldn't, if you check all `loading` states, they are handled like that (explicitly)

PR #708, opened 2020-02-10
s3: small cleanup and limit memory usage by disabling threads on upload_fileobj

pazembrz:


backend/inspirehep/records/api/literature.py
```diff
@@ -305,6 +305,7 @@ def add_file(
         new_key = hash_data(file_data)
         mimetype = magic.from_buffer(file_data, mime=True)
         size = len(file_data)
+        file_data = BytesIO(file_data)
```
michamos: what's the point of this?

pazembrz: Here yes. Main part is disabling threads on `upload_fileobj`.

drjova:
do we have any metrics for this change? How much memory it's saving?

drjova:
TBH I'm not convinced about this change, let's discuss it IRL

PR #709, opened 2020-02-10
files: properly handle DownloadFileError in migration

monaawi:
* handle DownloadFileError in the migration so that if the error is not caused by invalid cache and not being able to download local file, the error will be logged properly (instead of throwing hdel is not receiving all arguments) and the error will be on the mirror record too

### DIFF
--- a/backend/inspirehep/migrator/tasks.py
+++ b/backend/inspirehep/migrator/tasks.py
@@ -444,15 +444,18 @@ def migrate_record_from_mirror(
         )
         prod_record.error = exc
         db.session.merge(prod_record)
-    except DownloadFileError:
+    except DownloadFileError as exc:
         removed_cached_files = remove_cached_afs_file_locations(original_urls)
         if not removed_cached_files:
-            raise
-        return migrate_record_from_mirror(
-            prod_record=prod_record,
-            disable_orcid_push=disable_orcid_push,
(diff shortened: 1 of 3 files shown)


backend/inspirehep/migrator/tasks.py
```diff
@@ -444,15 +444,21 @@ def migrate_record_from_mirror(
...
-        if not removed_cached_files:
-            raise
-        return migrate_record_from_mirror(
-            prod_record=prod_record,
-            disable_orcid_push=disable_orcid_push,
-            disable_relations_update=disable_relations_update,
-        )
+    except DownloadFileError as exc:
+        # The DownloadFileError was not caused by an invalid cache
+        if not original_urls:
```
harunurhan: `is_error_caused_by_invalid_cache = original_urls is not None` could be better just to avoid the comment.

monaawi:
isn't the migration itself tied to legacy?

harunurhan:
Well it is, that's why I believe migration related should be more isolated from other parts, such as files.

monaawi:
well, it is. This is strictly related to migration logic. Files logic isn't here. The cache is for the migration

harunurhan:
ah that's great, didn't realize it was part of the migration task!

PR #711, opened 2020-02-11
search: add default_operator AND

drjova:

### DIFF
--- a/backend/inspirehep/search/api.py
+++ b/backend/inspirehep/search/api.py
@@ -41,7 +41,7 @@ class SearchMixin(object):
         """
         if not query_string:
             return self.query()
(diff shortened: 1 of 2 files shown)


backend/inspirehep/search/api.py
```diff
@@ -117,8 +117,7 @@ def query_from_iq(self, query_string):
             user_query = Q(
                 IQ(query_string, self) & Q("term", _collections="Literature")
             )
-            return self.query(user_query)
-        return self.query(IQ(query_string, self))
+        return self.query(IQ(query_string, self), default_operator="AND")
```
michamos: I don't think it's needed here, `IQ` already handles this.

PR #717, opened 2020-02-13
facets: fix nested facets search results

miguelgrc:
* INSPIR-3162


monaawi:
can you also please add a test to make sure this doesn't happen again in the future?

backend/tests/integration/search/test_api.py
```diff
@@ -144,3 +145,30 @@ def test_citations_query_result(api_client, db, es_clear, create_record):
     assert response.json["metadata"]["citation_count"] == 1
     citation = response.json["metadata"]["citations"][0]
     assert citation["control_number"] == record_citing["control_number"]
+
+
+def test_author_publication_query_result(
```
monaawi: could you split it in 2 tests? something like ```test_self_affiliations_filter``` and ```test_self_author_names_filter```?

backend/tests/integration/search/test_api.py
```diff
@@ -144,3 +145,30 @@ def test_citations_query_result(api_client, db, es_clear, create_record):
...
+    api_client, db, es_clear, create_record, create_user, datadir
+):
+    user = create_user(role="cataloger")
+    login_user_via_session(api_client, email=user.email)
+
+    lit1_data = json.loads((datadir / "787590.json").read_text())
```
monaawi: why did you use entire records like this? For me, it is easier to understand the test if there are just small records with exactly what I need for testing so I don't have to go back and forth between the test and the json files

PR #720, opened 2020-02-17
ui: remove figure labels and fix figure preview

harunurhan:
* INSPIR-2871


ui/src/App.scss
```diff
@@ -165,6 +165,11 @@
   }
 }
 
+// height version of tachyons' `mw5` class
+.mhi5 {
```
drjova: why not ``mh5``?

harunurhan: it's taken, tachyons' margin horizontal

PR #724, opened 2020-02-17
orcid: push user literature when setting enabled

harunurhan:
* INSPIR-3171

### DIFF
--- a/backend/inspirehep/accounts/views.py
+++ b/backend/inspirehep/accounts/views.py
@@ -12,6 +12,8 @@ from invenio_db import db
 from invenio_oauthclient import current_oauthclient
 from sqlalchemy.exc import IntegrityError
 
+from inspirehep.orcid.tasks import push_account_literature_to_orcid
+
 from .api import get_current_user_remote_orcid_account
 from .decorators import login_required
 from .handlers import get_current_user_data
@@ -87,5 +89,18 @@ def set_orcid_push_setting():
 
     db.session.add(orcid_account)
     db.session.commit()
+
+    if allow_push is True:
+        orcid = orcid_account.extra_data["orcid"]
+        tokens = orcid_account.remote_tokens
+
+        if len(tokens) != 1:
+            raise ValueError(
+                f"One token per remote account is expected, but found {len(tokens)} for {orcid}"
+            )
+
+        push_account_literature_to_orcid.apply_async(
+            kwargs={"orcid": orcid, "token": tokens[0].access_token}
+        )
 
     return jsonify({"message": "Successfully changed orcid push setting"}), 200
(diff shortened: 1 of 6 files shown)


backend/inspirehep/orcid/tasks.py
```diff
@@ -268,3 +269,13 @@ def _find_user_matching(orcid, email):
     if user_identity:
         return user_identity.user
     return User.query.filter_by(email=email).one_or_none()
+
+
+@shared_task(ignore_result=False, max_retries=3)
```
drjova: with ``ignore_results=False`` the endpoint will wait to schedule all the tasks first

backend/inspirehep/accounts/views.py
```diff
@@ -88,4 +90,17 @@ def set_orcid_push_setting():
...
+        tokens = orcid_account.remote_tokens
+
+        if len(tokens) != 1:
+            raise ValueError(
+                f"One token per remote account is expected, but found {len(tokens)} for {orcid}"
+            )
+
+        push_account_literature_to_orcid.apply_async(
+            kwargs={"orcid": orcid, "token": tokens[0].access_token}
+        )
```
drjova: Actually I'd just get ``[0]`` and raise on IndexError instead of checking how many tokens we have.

harunurhan: whole thing (and many more things) on relies on the fact that we have a single token per account, and all accounts are `orcid`, that's why there is the check.

backend/inspirehep/orcid/tasks.py
harunurhan: yeah copy-paste, needs `True` indeed.

backend/inspirehep/orcid/tasks.py
```diff
@@ -268,3 +269,13 @@ def _find_user_matching(orcid, email):
     if user_identity:
         return user_identity.user
     return User.query.filter_by(email=email).one_or_none()
+
+
+@shared_task(max_retries=3)
```
drjova: ``max_retries`` alone I think will do nothing you have to ``self.retry`` with ``bind=True`` or you have to specify ``autoretry_for`` and specify the exception (i.e. can be something silly like ``autoretry_for=(Exception,)``) do we really want this?

harunurhan: yeah this one is from a copy-paste, tbh I don't think we need retries for this

backend/tests/helpers/factories/models/user_access_token.py
```diff
@@ -48,12 +49,14 @@ def _create(
...
-            remote_account = RemoteAccount(
+            remote_token = RemoteToken.create(
                 user_id=user.get_id(),
                 client_id="orcid",
+                token=token,
+                secret=None,
                 extra_data={"orcid": orcid, "allow_push": allow_push},
             )
-            db.session.add(remote_account)
+            db.session.add(remote_token)
```
drjova: Does it already https://github.com/inveniosoftware/invenio-oauthclient/blob/master/invenio_oauthclient/models.py#L240

PR #725, opened 2020-02-18
ui: link contributions to literature

miguelgrc:
* INSPIR-3161


ui/src/conferences/utils.js
```diff
@@ -0,0 +1,3 @@
+export function getContributionsQuery(recordId) {
```
harunurhan: `getContributionsQueryString`

drjova:
@miguelgrc it has a weird merge commit

PR #726, opened 2020-02-18
supervisors: remove supervisors from search and facets

monaawi:
* INSPIR-3149


backend/inspirehep/records/marshmallow/literature/es.py
```diff
@@ -148,3 +142,13 @@ def get_bookautocomplete(self, record):
...
+
+    @pre_dump
+    def remove_supervisors_from_authors(self, data):
+        if "authors" in data:
+            data["authors"] = [
+                author
+                for author in data["authors"]
+                if "supervisor" not in author.get("inspire_roles", [])
+            ]
+        return data
```
drjova: can we push this logic on ``AuthorsSchema``?

ui/src/literature/components/SupervisorList.jsx
```diff
@@ -0,0 +1,39 @@
...
+import InlineList from '../../common/components/InlineList';
+import Author from '../../common/components/Author';
+import pluralizeUnlessSingle from '../../common/utils';
+
+function renderSupervisor(supervisor) {
+  return <Author author={supervisor} />;
+}
+
+function extractKeyFromSupervisor(supervisor) {
+  return supervisor.get('uuid');
```
harunurhan: this is isn't present when author is not linked.

monaawi: It is, we are generating it at lit update for every author (linked or unlinked - we use it for the disambiguation to identify signatures).

PR #730, opened 2020-02-19
sitemap: add initial module and cli

harunurhan:
* INSPIR-3152


backend/inspirehep/sitemap/cli.py
```diff
@@ -0,0 +1,69 @@
...
+    return f"sitemap{page}.xml"
+
+
+def get_sitemap_page_absolute_url(page):
+    base_url = get_inspirehep_url()
+    return f"{base_url}/{get_sitemap_page_filename(page)}"
+
+
+def write_sitemap_page_content(page, page_content):
+    path = current_app.config["SITEMAP_FILES_PATH"]  # FIXME: should we mkdir -p?
```
drjova: I'd add if ``os.path.isdir`` and raise if it's not present, but not creating because on prod might hide issues

harunurhan: `write` already raises if dir is not there.

backend/inspirehep/sitemap/cli.py
```diff
@@ -0,0 +1,69 @@
...
+from flask.cli import with_appcontext
+
+from inspirehep.utils import get_inspirehep_url
+
+from .sitemap import generate_sitemap_items
+
+
+@click.group()
+def sitemap():
+    """Command group to handle file operations."""
```
pazembrz: Copy-paste? ;)

backend/inspirehep/sitemap/cli.py
```diff
@@ -0,0 +1,69 @@
...
+    help="Generates sitemaps for records that should be indexed by search engines"
+)
+@with_appcontext
+def generate():
+    page_size = current_app.config["SITEMAP_PAGE_SIZE"]
+
+    sitemap_items = iter(generate_sitemap_items())
+
+    page = 1
+    page_items = list(islice(sitemap_items, page_size))
```
pazembrz: Why not list slice? `sitemap_items[:page_size]` ? will be much faster than list(islice) and result is the same as you are not using iterator here at all.

backend/inspirehep/sitemap/sitemap.py
```diff
@@ -0,0 +1,37 @@
...
+# the terms of the MIT License; see LICENSE file for more details.
+
+from inspirehep.pidstore.api import PidStoreBase
+from inspirehep.utils import get_inspirehep_url
+
+from .collections import get_indexable_record_searches
+
+
+# TODO: maybe move to PidStoreBase ?
+def get_endpoint_from_schema(schema):
```
pazembrz: This definitely should go to PidStoreBase

backend/inspirehep/sitemap/cli.py
```diff
@@ -0,0 +1,69 @@
...
+
+
+@sitemap.command(
+    help="Generates sitemaps for records that should be indexed by search engines"
+)
+@with_appcontext
+def generate():
+    page_size = current_app.config["SITEMAP_PAGE_SIZE"]
+
+    sitemap_items = iter(generate_sitemap_items())
```
pazembrz: Why you are creating iterator from generator? At the end it gives you the same thing.

harunurhan: some lines in this function are adapted from zenodo https://github.com/zenodo/zenodo/blob/44ee8568b0641eea96bdefbb85d50643564df9b2/zenodo/modules/sitemap/tasks.py#L36.

The PR is WIP at the moment, my plan is to write tests first and then make this kind of changes.

backend/inspirehep/sitemap/templates/sitemap/index.xml
```diff
@@ -0,0 +1,11 @@
+<?xml version="1.0" encoding="UTF-8" ?>
```
michamos: I don't think templating XML is a great approach. It might work here as it's rather simple, but in general it's better to use something like `lxml.etree.ElementMaker` to build the element tree (we do this in https://github.com/inspirehep/inspirehep/blob/master/backend/inspirehep/orcid/builder.py).

backend/inspirehep/sitemap/cli.py
pazembrz: :+1:

backend/inspirehep/sitemap/cli.py
```diff
@@ -0,0 +1,69 @@
...
+def get_sitemap_page_absolute_url(page):
+    base_url = get_inspirehep_url()
+    return f"{base_url}/{get_sitemap_page_filename(page)}"
+
+
+def write_sitemap_page_content(page, page_content):
+    path = current_app.config["SITEMAP_FILES_PATH"]  # FIXME: should we mkdir -p?
+    file_path = os.path.join(path, get_sitemap_page_filename(page))
+    with open(file_path, "w+") as file:
+        file.write(page_content)
```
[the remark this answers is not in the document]
harunurhan: since we won't do `mkdir`,  I think this version is simple enough.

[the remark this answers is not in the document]
michamos: I would check with @benjamin-bergia about the `mkdir`. I remember him complaining about having to create directories.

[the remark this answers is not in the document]
harunurhan: these are supposed to be somewhere top-level anyhow, I don't think it will be an issue.

backend/inspirehep/sitemap/templates/sitemap/index.xml
harunurhan: This is actually copy-paste from zenodo.

yeah I see, but that looks like so much hustle for something simple like sitemap.

backend/inspirehep/sitemap/cli.py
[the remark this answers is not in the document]
michamos: BTW How are the sitemaps going to be served? wouldn't it be easier to push everything to a public S3 bucket? all you need to serve the files then is to route requests appropriately.

backend/inspirehep/sitemap/templates/sitemap/index.xml
harunurhan: I added XML (syntax only) validation during tests.

backend/inspirehep/sitemap/collections.py
```diff
@@ -0,0 +1,36 @@
...
+from inspirehep.search.api import (
+    AuthorsSearch,
+    ConferencesSearch,
+    JobsSearch,
+    LiteratureSearch,
+)
+
+
+def jobs():
+    only_open = Q("match", **{"status": "open"})
```
drjova: I'd move this one on JobsSearch

PR #731, opened 2020-02-19
mappings: case insensitive search in conferences

miguelgrc:
* INSPIR-3160


drjova:
@miguelgrc this can be a good use case for the new remap cli cc @pazembrz

PR #735, opened 2020-02-20
v0.5.0

drjova:
Conference release


pazembrz:
To rebuild all cnums correctly it requires remigration from legacy.

monaawi:
For the supervisors task, we need to reindex literature

PR #736, opened 2020-02-20
WIP POC: upload files in threads

pazembrz:
INSPIR-3151

DO NOT MERGE!


michamos:
This seems a better approach to deal with the S3 sessions through the app context: https://github.com/dchevell/flask-executor

pazembrz:
@michamos this flask-executor looks to be dead. And actually Issues reported there are quite important for us :/ https://github.com/dchevell/flask-executor/issues/30

michamos:
I'm not saying we have to use exactly that. But starting a new app context inside the executor might be a good approach.

pazembrz:
I don't feel it. It looks like additional unnecessary complication for simple thing :/

michamos:
The issue is that you need to have one S3 session per thread, so you need to manage them somehow. The proper place is the `files` extension, but that requires doing it through the app context.

pazembrz:
But we are taking care of this already. https://boto3.amazonaws.com/v1/documentation/api/latest/guide/resources.html?highlight=multithreading#multithreading-multiprocessing we have something similar. It does not require to create whole app context.

pazembrz:
@michamos Does it looks better now?

PR #737, opened 2020-02-20
migrator: add timeouts for files downloading request and celery task

monaawi:
* INSPIR-3178


backend/inspirehep/migrator/tasks.py
```diff
@@ -259,7 +260,17 @@ def populate_mirror_from_file(source):
         echo(f"Inserted {inserted_records} records into mirror")
 
 
-@shared_task(ignore_result=False, queue="migrator", acks_late=True)
+@shared_task(
+    ignore_result=False,
+    queue="migrator",
+    acks_late=True,
+    soft_time_limit=60 * 60,
+    time_limit=120 * 60,
```
michamos: these are batches of up to 100 records. Are you sure 2 min is enough?

monaawi: yeah, I know they are batches of records. This is why this is 2 hours, not 2 minutes.

michamos: :man_facepalming:

PR #738, opened 2020-02-21
ui: add meta descriptions for all pages

harunurhan:
* INSPIR-3164


ui/src/common/utils.js
```diff
@@ -224,3 +224,16 @@ export function abbreviateNumber(number) {
   const numberOfFractionDigits = number < 10000 ? 1 : 0;
   return numberAbbreviator.abbreviate(number, numberOfFractionDigits);
 }
+
+const HTML_TAG_REGEXP = /(<([^>]+)>)/ig;
+const MAX_DESCRIPTION_LENGTH = 160
+export function makeCompliantMetaDescription(description = '') {
```
drjova: oh we need that for job description?

ui/src/authors/__tests__/utils.test.js
```diff
@@ -19,4 +19,66 @@ describe('utils', () => {
...
+        control_number: 12345,
+        project_membership: [{ name: 'CERN-LHC-CMS' }, { name: 'CERN-AMS' }],
+        positions: [
+          { institution: 'CERN', current: true },
+          { institution: 'MIT' },
+          { institution: 'DESY', current: true },
+        ],
+        arxiv_categories: ['hep-th', 'hep-ex'],
+      });
+      const expected = 'エドウィン. CERN and DESY. hep-th and hep-ex. CERN-LHC-CMS and CERN-AMS'
```
drjova: what's the case with preferred_name and native_names? do we display both?

harunurhan: nope, only native_names, the real name is in the `<title>`

ui/src/common/utils.js
harunurhan: well for anything... we have quite relaxed schema for all records.

PR #741, opened 2020-02-24
literature: display authors with no hepnames records

monaawi:
* INSPIR-3163


ui/src/common/components/Author.jsx
```diff
@@ -5,9 +5,25 @@ import { Map } from 'immutable';
 import { Tooltip } from 'antd';
 
 import AuthorAffiliationList from './AuthorAffiliationList';
-import { AUTHORS } from '../routes';
+import { AUTHORS, SUBMISSIONS_AUTHOR } from '../routes';
+import ExternalLink from './ExternalLink';
+import { getLinkToSearchWithBAI } from '../utils';
 
 class Author extends Component {
+  getCreateProfileTooltipMessage() {
```
harunurhan: `render...` to indicate that it returns markup

harunurhan: or alternatively you could refactor the whole thing, because it seems like "unlinked author with bai" logic is too much. So you can create a small function components here in this file (as different files but in a new folder called `Author`) `UnlinkedAuthor`, `LinkedAuthor`, `AuthorWithBai` and put the logic to decide which one render in the main `Author` component which will be exported.

ui/src/common/utils.js
```diff
@@ -236,4 +237,17 @@ export function makeCompliantMetaDescription(description = '') {
   }
 
   return withoutHtml;
-}
\ No newline at end of file
+}
+
+export function getLiteratureSearchForAuthorBAIUrl(bai) {
```
harunurhan: `getLiteratureSearchUrlForAuthorBAI`

ui/src/common/components/Author/__tests__/__snapshots__/Author.test.jsx.snap
```diff
@@ -0,0 +1,279 @@
...
+          "Affiliation",
+        ]
+      }
+    />
+    )
+  </span>
+</div>
+`;
+
+exports[`Author renders full_name if first_name and last_name missing 1`] = `
```
harunurhan: this test doesn't make sense now.

harunurhan: there are few like this, you need to remove these and test these cases in `getAuthorName` tests.

ui/src/authors/components/AuthorBAI.jsx
```diff
@@ -0,0 +1,20 @@
...
+  return (
+    <span>
+      Author Identifier:{' '}
+      <Link to={getLiteratureSearchUrlForAuthorBAI(bai)}>{bai}</Link>
+    </span>
+  );
+}
+
+AuthorBAI.propTypes = {
+  bai: PropTypes.instanceOf(Map).isRequired,
```
harunurhan: `PropTypes.string.isRequired`

ui/src/common/components/Author/Author.jsx
```diff
@@ -60,11 +39,14 @@ class Author extends Component {
   }
 
   renderAuthorName() {
-    const recid = this.getAuthorRecid();
-    if (recid != null) {
-      return <Link to={`${AUTHORS}/${recid}`}>{this.getAuthorName()}</Link>;
+    const { author } = this.props;
+    if (author.getIn(['record', '$ref'])) {
```
harunurhan: `hasIn` or actually `has('record')` would be enough here.

PR #742, opened 2020-02-24
detail page loading bug

harunurhan:


ui/src/conferences/containers/DetailPageContainer.jsx
```diff
@@ -23,25 +23,11 @@ import { newSearch } from '../../actions/search';
...
-  useEffect(
-    () => {
-      dispatch(fetchConference(recordId));
-      dispatch(newSearch(CONFERENCE_CONTRIBUTIONS_NS));
-      window.scrollTo(0, 0);
-    },
-    [dispatch, recordId]
-  );
-
+function DetailPage({ record }) {
```
drjova: what about the ``scrollTo`` stuff?

ui/src/common/withRouteDataFetcher.jsx
```diff
@@ -0,0 +1,40 @@
+import React, { useEffect } from 'react';
+import { connect } from 'react-redux';
+import { withRouter } from 'react-router-dom';
+
+import { getWrapperComponentDisplayName } from './utils';
+import LoadingOrChildren from './components/LoadingOrChildren';
+
+export default function withRouteDataFetcher(
```
drjova: the name is a bit confusing tbh, wouldn't be more obvious something like ``withLocationParamsDataFetcher`` or similar?

harunurhan: this is used to `fetch` for `data` dependencies of current `route`. So name made sense 😄

ui/src/conferences/containers/DetailPageContainer.jsx
harunurhan: this is in my notes, only thing missing in the PR right now. I think it's safe to `scrollTo(0)`, whenever pathname changes, what do you think?

ui/src/common/withRouteDataFetcher.jsx
harunurhan: To be honest I am more concern about the fact that some components are dispatching actions which doesn't even use params, so it looks little awkward.

PR #745, opened 2020-02-25
poetry: update dependencies

michamos:
This is required to have inspire-dojson v63.0.5

### DIFF
--- a/backend/pyproject.toml
+++ b/backend/pyproject.toml
@@ -25,7 +25,7 @@ invenio-accounts = {git = "https://github.com/inveniosoftware/invenio-accounts.g
 invenio-app= "==1.2.3"
 invenio-base= "==1.2.0"
 invenio-cache= "==1.0.0"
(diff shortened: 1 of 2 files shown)


backend/pyproject.toml
```diff
@@ -25,7 +25,7 @@ invenio-accounts = {git = "https://github.com/inveniosoftware/invenio-accounts.g
 invenio-app= "==1.2.3"
 invenio-base= "==1.2.0"
 invenio-cache= "==1.0.0"
-invenio-celery= "==1.1.0"
+invenio-celery= "==1.1.2"
```
michamos: got a version conflict without this. @drjova is it the right solution?

drjova: what was the conflict? ``v1.1.2`` seems safe (just pinning deps) but we haven't test it

drjova: https://github.com/inveniosoftware/invenio-celery/commits/master

PR #748, opened 2020-02-26
setup: move to python 3.8

pazembrz:
*INSPIR-3177

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,7 +13,7 @@ git:
   depth: 200
 
 python:
-  - "3.6"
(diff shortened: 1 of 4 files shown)


docker-compose.services.yml
```diff
@@ -13,7 +13,7 @@ services:
       context: backend
       args:
         - POETRY_EXTRA_ARGS=--dev
-    image: inspirehep
+    image: inspirehep:py3.8
```
drjova: it's the local image right? we don't need to tag it IMHO

PR #750, opened 2020-02-26
ui: improve initial assets load speed

harunurhan:
Does not split code into chunks, based on routes anymore. Since this
ends up a lot of dependencies duplicated in multiple bundles. Although
in theory this increases size of assets that need to be downloaded
to render a route, in practice, our preloading strategy does not work
as inteded so browsers ends up downloading most of assets for a route.

So for now admin route and submissions routes are async, others are
not. And there is no preloading as of now, but planned to be introduce
in the future.

(Also adds `yarn analyze` to see report for generated bundle by `yarn
build`)

INSPIR-3232

### DIFF
--- a/ui/src/common/VerticalDivider/VerticalDivider.scss
+++ b/ui/src/common/VerticalDivider/VerticalDivider.scss
@@ -1,4 +1,4 @@
 .__VerticalDivider__ {
-  background: $text-color;
-  margin-left: $padding-md - 1px;
+  background: $text-color !important;
+  margin-left: $padding-md - 1px !important;
 }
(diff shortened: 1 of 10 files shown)


ui/src/common/VerticalDivider/VerticalDivider.scss
```diff
@@ -1,4 +1,4 @@
 .__VerticalDivider__ {
-  background: $text-color;
-  margin-left: $padding-md - 1px;
+  background: $text-color !important;
+  margin-left: $padding-md - 1px !important;
```
drjova: where this is helping? ☺️

harunurhan: before it was working just by chance, but normally if you want to override styles with our self-written styles, on a ant component, it needs `!important`. Because it usually applies `class="class-name-from-prop antd-class ..."`

PR #756, opened 2020-02-28
alembic: migration to remove records_bucket fk

pazembrz:
*INSPIR-3105

NEEDS MIGRATION 42a5817ca169

### DIFF
--- a/backend/inspirehep/alembic/42a5817ca169_remove_records_bucket_entries_to_.py
+++ b/backend/inspirehep/alembic/42a5817ca169_remove_records_bucket_entries_to_.py
@@ -1,0 +1,36 @@
+#
+# This file is part of Invenio.
+# Copyright (C) 2016-2018 CERN.
+#
+# Invenio is free software; you can redistribute it and/or modify it
+# under the terms of the MIT License; see LICENSE file for more details.
+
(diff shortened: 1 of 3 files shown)


backend/inspirehep/alembic/42a5817ca169_remove_fk_from_records_buckets_to_.py
```diff
@@ -0,0 +1,31 @@
...
+revision = "42a5817ca169"
+down_revision = "f563233434cd"
+branch_labels = ()
+depends_on = None
+
+
+def upgrade():
+    """Upgrade database."""
+    op.execute(
+        "ALTER TABLE records_buckets DROP CONSTRAINT fk_records_buckets_record_id_records_metadata"
```
michamos: don't you want to drop the whole table? workflows use `workflows_buckets` instead?

pazembrz: No, we don't as this table is managed by invenio module which is also installed on next. So if invenio would ever do some migration on it we will have conflicts. 
For now I'm actually changing migration to not remove FK but to remove all rows from the table and also remove all rows from files_buckets which are connected to entries in this table.

PR #757, opened 2020-02-28
records: remove previous version exception

drjova:


backend/tests/integration-async/records/api/test_base.py
```diff
@@ -41,6 +41,21 @@ def test_record_versioning(app, clear_environment):
     assert record._previous_version
 
 
+def test_record_previous_version_doesnt_fail_if_does_exist(app, clear_environment):
```
monaawi:
```suggestion
def test_record_previous_version_doesnt_fail_if_previous_version_missing(app, clear_environment):
```

PR #758, opened 2020-02-28
Indexer move to proper

drjova:


backend/inspirehep/cli.py
```diff
@@ -26,4 +27,4 @@
 cli.add_command(orcid)
 cli.add_command(mailing)
 cli.add_command(files)
-cli.add_command(sitemap)
```
monaawi: why did you remove the sitemap cli?

backend/inspirehep/records/indexer/tasks.py
```diff
@@ -17,9 +17,9 @@
 )
 from sqlalchemy.orm.exc import NoResultFound, StaleDataError
 
+from inspirehep.indexer.base import InspireRecordIndexer
+from inspirehep.indexer.utils import get_record
 from inspirehep.records.api import AuthorsRecord, LiteratureRecord
-from inspirehep.records.indexer.base import InspireRecordIndexer
-from inspirehep.records.indexer.utils import get_record
```
monaawi: maybe you should remove this since you moved it to indexer/tasks

backend/inspirehep/cli.py
drjova: my bad

backend/inspirehep/records/indexer/tasks.py
drjova: wdym?

monaawi: You added ```indexer/tasks```, why still keep ```records/indexer/tasks```?

drjova: this is a leftover good catch

PR #759, opened 2020-03-02
minter: add BAI minter

pazembrz:
* INSPIR-3209

### DIFF
--- a/backend/inspirehep/pidstore/api/authors.py
+++ b/backend/inspirehep/pidstore/api/authors.py
@@ -4,7 +4,7 @@
 #
 # inspirehep is free software; you can redistribute it and/or modify it under
 # the terms of the MIT License; see LICENSE file for more details.
-
+from inspirehep.pidstore.minters.bai import BAIMinter
 
 from ..minters.control_number import AuthorsMinter
 from ..minters.orcid import OrcidMinter
@@ -13,4 +13,4 @@ from .base import PidStoreBase
 
 class PidStoreAuthors(PidStoreBase):
 
-    minters = [AuthorsMinter, OrcidMinter]
+    minters = [AuthorsMinter, OrcidMinter, BAIMinter]
--- a/backend/tests/integration/pidstore/minters/test_minters_bai.py
+++ b/backend/tests/integration/pidstore/minters/test_minters_bai.py
@@ -1,0 +1,69 @@
+# -*- coding: utf-8 -*-
+#
+# Copyright (C) 2019 CERN.
+#
+# inspirehep is free software; you can redistribute it and/or modify it under
+# the terms of the MIT License; see LICENSE file for more details.
+
+
+import pytest
+from helpers.providers.faker import faker
+from invenio_pidstore.models import PersistentIdentifier, PIDStatus
+
+from inspirehep.pidstore.errors import PIDAlreadyExists
+from inspirehep.pidstore.minters.bai import BAIMinter
+from inspirehep.records.api import AuthorsRecord
+
+
+def test_minter_bai(base_app, db, es, create_record):
+    data = {"ids": [{"schema": "JACOW", "value": "JACoW-12345678"}]}
+    record_data = faker.record("aut", data=data, other_pids=["bai"])
+    BAIMinter.mint(None, record_data)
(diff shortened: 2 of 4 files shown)


backend/tests/integration/pidstore/minters/test_minters_bai.py
```diff
@@ -0,0 +1,56 @@
...
+
+    assert expected_pids_len == result_pids_len
+
+    pid = result_pids[0]
+    assert expected_pids_provider == pid.pid_provider
+    assert expected_pids_status == pid.status
+    assert pid.pid_value in epxected_pids_values
+
+
+def test_minter_orcid_empty(base_app, db, es, create_record):
```
drjova: orcid?

backend/tests/integration/pidstore/minters/test_minters_bai.py
```diff
@@ -0,0 +1,56 @@
...
+    expected_pids_len = 0
+
+    result_pids_len = PersistentIdentifier.query.filter_by(
+        object_uuid=record.id, pid_type="bai"
+    ).count()
+
+    assert expected_pids_len == result_pids_len
+
+
+def test_minter_orcid_already_existing(base_app, db, es, create_record):
```
drjova: orcid?

backend/inspirehep/pidstore/api/authors.py
```diff
@@ -4,7 +4,7 @@
 #
 # inspirehep is free software; you can redistribute it and/or modify it under
 # the terms of the MIT License; see LICENSE file for more details.
-
+from inspirehep.pidstore.minters.bai import BAIMinter
```
drjova: space between docs :)

backend/tests/integration/pidstore/minters/test_minters_bai.py
```diff
@@ -0,0 +1,56 @@
...
+    record_with_orcid = create_record("aut", other_pids=["bai"])
+    data = {"ids": record_with_orcid["ids"]}
+    with pytest.raises(PIDAlreadyExists):
+        create_record("aut", data)
```
drjova: IMO this is not testing the actual API it's a fixture, we should use Author record class for that

pazembrz: We don't need to test Author record class, as we are interested only in Minter itself which is called correctly there and there are proper tests for this.

backend/inspirehep/pidstore/api/authors.py
pazembrz: This is exactly same size of space we have in every test_minters tests. I'm just trying to be consistent.

backend/tests/integration/pidstore/minters/test_minters_bai.py
pazembrz: :+1:

backend/tests/integration/pidstore/minters/test_minters_bai.py
pazembrz: :+1:

backend/tests/integration/pidstore/minters/test_minters_bai.py
drjova: but we might change the fixture, in that case we should either call minter create or the author class

backend/tests/integration/pidstore/minters/test_minters_bai.py
```diff
@@ -0,0 +1,69 @@
...
+
+
+def test_minter_bai_empty(base_app, db, es, create_record):
+    record_data = faker.record("aut")
+    BAIMinter.mint(None, record_data)
+
+    expected_pids_len = 0
+
+    result_pids_len = PersistentIdentifier.query.filter_by(
+        object_uuid=None, pid_type="bai"
```
drjova:
```suggestion
        pid_type="bai"
```

pazembrz: I know that It's not necessary at all, as still there should be only one BAI in pid_store but I put it in test so it could be clearly visible that it is None. It's test. It should be very verbose and readable.

PR #760, opened 2020-03-03
cnum_provider: fix issues with wrong cnums when holes in cnums sequence

pazembrz:
* INSPIR-3236

### DIFF
--- a/backend/inspirehep/pidstore/providers/cnum.py
+++ b/backend/inspirehep/pidstore/providers/cnum.py
@@ -16,6 +16,7 @@ LOGGER = structlog.getLogger()
 
 class InspireCNUMProvider(BaseProvider):
     """CNUM identifier provider."""
+
     pid_type = None
     pid_provider = "cnum"
     default_status = PIDStatus.RESERVED
@@ -50,15 +51,27 @@ class InspireCNUMProvider(BaseProvider):
         opening_date = data.get("opening_date")
         if not opening_date:
             return
-
         adjusted_date = datetime.strptime(opening_date, "%Y-%m-%d").strftime("%y-%m-%d")
         cnum = f"C{adjusted_date}"
-        cnums_count = PersistentIdentifier.query\
-            .filter(PersistentIdentifier.pid_value.like(f"{cnum}%"))\
-            .filter(PersistentIdentifier.pid_type == "cnum")\
-            .count()
+        all_postfixes = []
+        all_cnums = (
+            PersistentIdentifier.query.with_for_update()
+            .filter(PersistentIdentifier.pid_value.like(f"{cnum}%"))
+            .filter(PersistentIdentifier.pid_type == "cnum")
+            .all()
+        )
+        if not all_cnums:
+            return cnum
 
-        if cnums_count > 0:
-            cnum = f"{cnum}.{cnums_count}"
+        for cnum_entry in all_cnums:
+            cnum_value = cnum_entry.pid_value.split(".")
+            if len(cnum_value) > 1:
+                all_postfixes.append(int(cnum_value[-1]))
+        if all_postfixes:
+            postfix = sorted(all_postfixes)[-1] + 1
+        else:
+            postfix = 1
 
-        return cnum
+        full_cnum = f"{cnum}.{postfix}"
+
+        return full_cnum
--- a/backend/tests/integration/pidstore/minters/test_minters_cnum.py
+++ b/backend/tests/integration/pidstore/minters/test_minters_cnum.py
@@ -189,3 +189,43 @@ def test_minter_mints_cnum_on_update_when_cnum_is_missing_in_db(
     expected_cnum = "C05-01-01"
     assert 1 == len(record_cnums)
     assert record_cnums[0].pid_value == expected_cnum
+
+
+def test_generate_cnum_when_holes_in_cnums_sequence(base_app, db, create_record):
+    data = {"opening_date": "2020-01-01"}
+    expected_cnum = "C20-01-01.2"
+    rec1 = create_record("con", data)
+    rec2 = create_record("con", data)
+    rec1.hard_delete()
+    rec3 = create_record("con", data)
+    assert rec3.get("cnum") == expected_cnum
+
+
+def test_generate_cnum_when_holes_in_cnums_sequence_and_weird_creation_order(
+    base_app, db, create_record
+):
+    data = {"opening_date": "2020-01-01", "cnum": "C20-01-01.2"}
+    rec1 = create_record("con", data)
+    data["cnum"] = "C20-01-01.1"
+    rec2 = create_record("con", data)
+    data["cnum"] = "C20-01-01"
+    rec3 = create_record("con", data)
+    del data["cnum"]
+    rec2.hard_delete()
+    rec4 = create_record("con", data)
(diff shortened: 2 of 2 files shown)


backend/inspirehep/pidstore/providers/cnum.py
```diff
@@ -53,12 +55,41 @@ def next(cls, data):
...
+                )
+                .filter(PersistentIdentifier.pid_type == "cnum")
+                .count()
+            )
+            if cnum_count == 0:
+                break
+            postfix += 1
 
-        return cnum
+        return full_cnum
```
drjova: IMO we should get the latest prefix and + 1 instead of this retry

monaawi: He added this retry because, apparently, we can have a cnum with prefix 2 created before a cnum with prefix 1 (which I don't understand why). But I also agree that this retry is ugly. I think we should query and find the biggest prefix and then do +1, the latest prefix won't work

pazembrz: Latest prefix is not correct as many CNUMS were created in weird order. 
Check: 
SELECT * from pidstore_pid where pid_type='cnum' ORDER BY pid_value LIMIT 1000
"2020-02-21 12:07:05.623265"	"2020-02-21 12:07:05.62327"	22160560	"cnum"	"C00-01-04"
"2020-02-21 12:06:45.515772"	"2020-02-21 12:06:45.515778"	22157023	"cnum"	"C00-01-04.1"

So here if we would take latest cnum it would return "C00-01-04.1" as the next one.

backend/inspirehep/pidstore/providers/cnum.py
```diff
@@ -53,12 +55,41 @@ def next(cls, data):
...
-            cnum = f"{cnum}.{cnums_count}"
+        retry_count = 0
+        try:
+            # Try to get postfix number from latest pid
+            last_cnum = (
+                PersistentIdentifier.query.filter(
+                    PersistentIdentifier.pid_value.like(f"{cnum}%")
+                )
+                .filter(PersistentIdentifier.pid_type == "cnum")
+                .order_by(PersistentIdentifier.created.desc())
```
michamos: I think you're overcomplicating this. `created` is not very correlated with CNUM value ordering due to the migration inserting things in random order. What you want is to order by CNUM value, which happens to match lexicographic ordering on `pid_value`. so why don't you simply
```suggestion
                .order_by(PersistentIdentifier.pid_value.desc())
```

backend/inspirehep/pidstore/providers/cnum.py
michamos: I think you need `DESC` also, but there's the issue that ` 10 < 2`, so it's not exactly lexicographic ordering :( if you want to do it in DB, you need to split on `.`, cast to `int` and order based on that. Or you can get everything in Python and compute the next one. In any case there might be (very improbable) collisions in case two concurrent transactions try to mint the same CNUM. The proper way to do this is probably to have another table that maps CNUM prefix to latest CNUM suffix and increment the counter when minting a new CNUM :cry:

backend/inspirehep/pidstore/providers/cnum.py
pazembrz: because pid_value of "C20-01-01.20" would be smaller than "C20-01-01.3"

backend/inspirehep/pidstore/providers/cnum.py
pazembrz: I added row locks for CNUM

backend/inspirehep/pidstore/providers/cnum.py
```diff
@@ -50,15 +52,30 @@ def next(cls, data):
...
         if not opening_date:
             return
-
         adjusted_date = datetime.strptime(opening_date, "%Y-%m-%d").strftime("%y-%m-%d")
         cnum = f"C{adjusted_date}"
-        cnums_count = PersistentIdentifier.query\
-            .filter(PersistentIdentifier.pid_value.like(f"{cnum}%"))\
-            .filter(PersistentIdentifier.pid_type == "cnum")\
-            .count()
+        all_postfixes = []
```
michamos: Minor: they are suffixes, not postfixes

backend/inspirehep/pidstore/providers/cnum.py
```diff
@@ -50,15 +52,30 @@ def next(cls, data):
...
+        all_cnums = (
+            PersistentIdentifier.query.with_for_update()
+            .filter(PersistentIdentifier.pid_value.like(f"{cnum}%"))
+            .filter(PersistentIdentifier.pid_type == "cnum")
+            .all()
+        )
+        if not all_cnums:
+            postfix = None
```
michamos:
```suggestion
            return cnum
```
so you don't have to do the test on L76.

backend/tests/integration/pidstore/minters/test_minters_cnum.py
```diff
@@ -189,3 +189,43 @@ def test_minter_mints_cnum_on_update_when_cnum_is_missing_in_db(
     expected_cnum = "C05-01-01"
     assert 1 == len(record_cnums)
     assert record_cnums[0].pid_value == expected_cnum
+
+
+def test_generate_cnum_when_holes_in_cnums_sequence(base_app, db, create_record):
```
michamos: maybe also add a test where you mint `C20-01-01.1` because only `C20-01-01` exists as it's a different case?

backend/inspirehep/pidstore/providers/cnum.py
pazembrz: due to https://en.wikipedia.org/wiki/Suffix suffix=postfix ?

backend/inspirehep/pidstore/providers/cnum.py
pazembrz: :+1:

backend/tests/integration/pidstore/minters/test_minters_cnum.py
pazembrz: I think that this is actually tested already on the old tests like here: test_minter_mint_cnum_more_than_once

backend/inspirehep/pidstore/providers/cnum.py
```diff
@@ -50,15 +51,27 @@ def next(cls, data):
...
+            return cnum
+        else:
+            for cnum_entry in all_cnums:
+                cnum_value = cnum_entry.pid_value.split(".")
+                if len(cnum_value) > 1:
+                    all_postfixes.append(int(cnum_value[-1]))
+            if all_postfixes:
+                postfix = sorted(all_postfixes)[-1] + 1
+            else:
+                postfix = 1
```
drjova:
```suggestion
        if not all_cnums:
            return cnum
        
            for cnum_entry in all_cnums:
                cnum_value = cnum_entry.pid_value.split(".")
                if len(cnum_value) > 1:
                    all_postfixes.append(int(cnum_value[-1]))
            postfix = 1
            if all_postfixes:
                postfix = sorted(all_postfixes)[-1] + 1
```

pazembrz: :+1:

PR #761, opened 2020-03-03
mappings: Add _created and _updated mappings

pazembrz:
* INSPIR-3233

### DIFF
--- a/backend/inspirehep/search/mappings/v7/records/authors.json
+++ b/backend/inspirehep/search/mappings/v7/records/authors.json
@@ -40,6 +40,12 @@
           }
         }
       },
+      "_created": {
+        "type": "date"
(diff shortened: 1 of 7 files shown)


backend/inspirehep/search/mappings/v7/records/authors.json
```diff
@@ -40,6 +40,12 @@
           }
         }
       },
+      "_created": {
+        "type": "date"
+      },
+      "_updated": {
+        "type": "date"
+      },
```
drjova: ``_update`` ``_created`` is not part of the record metadata

drjova: actually my bad

PR #764, opened 2020-03-03
Upgrade deps (antd, react, react-router)

harunurhan:


michamos:
from the screenshots, it looks like the inter-line spacing increased, so things are less compact vertically. Is that on purpose?

harunurhan:
@michamos yeah antd changed `line-height`, it's `1px` higher now.

harunurhan:
@miguelgrc submissions from mobile devices are quite rare, that's why we don't have ui tests for the mobile layout.

miguelgrc:
@harunurhan In the literature submission "Suggest content" box, the "selection policy" link jumps to a new line (it can be seen in the screenshot).

I was also checking locally and this breaks https://github.com/inspirehep/inspirehep/pull/623 somehow. It only affects small screen sizes, so we can have a look at it later on.

monaawi:
The same issue with the filters for jobs happens for the ```SortBy```, after you click to select a new sorting.

monaawi:
Also, the issue that you once fixed with the search scope selector text not fitting is now back. For example if you select job, all the others will be cut to 4 characters length.

harunurhan:
@monaawi it's already mentioned in the commit message, why this issue came back.

monaawi:
ah ok, I just saw the truncate issue in the commit message. But when I said the same issue as the job filter, I meant the sortby color when you click on it and the same for the search scope selector, not the truncate issue.

harunurhan:
@monaawi that's a new color from `antd` while selecting another option, current selection value greys out, I say we take the default. Placeholder color though was an issue because we explicitly wanted to overwrite that to black for `MultiSelectAggregationFilter`, which stopped working with the new version.

monaawi:
yeah, I saw it was the new color from ```antd```, thought we wanted to overwrite it as well to keep it as before.

PR #765, opened 2020-03-03
facets: refactor facets and display self affiliations for normal users

monaawi:
* INSPIR-3215


backend/inspirehep/search/facets.py
```diff
@@ -130,135 +131,59 @@ def nested_filters(author_recid):
     }
 
 
+def hep_filters(author_recid=None):
+    """author is a query parameter and it looks like 1234_Name%20Surname, thus the splitting."""
```
drjova:
```suggestion
    """Parameter author is a query and it looks like 1234_Name%20Surname, thus the splitting."""
```

backend/tests/integration/search/test_aggregations.py
```diff
@@ -0,0 +1,585 @@
...
+                {"doc_count": 1, "key": 1546300800000, "key_as_string": "2019"},
+            ],
+        }
+        assert response["aggregations"]["earliest_date"] == earliest_date_aggregation
+
+        response = api_client.get("/literature?earliest_date=2018--2019").json
+        assert (
+            response["hits"]["hits"][0]["metadata"]["control_number"]
+            == expected_record["control_number"]
+        )
```
drjova: it doesn't really check if we have only 1 result

drjova: same for the other cases :)

backend/tests/integration/search/test_facets.py
```diff
@@ -1,27 +1,34 @@
...
         },
-        "terms": {
-            "exclude": "Jones, Jessica",
-            "field": "facet_author_name",
-            "size": 20,
-        },
+        "terms": {"field": "facet_author_name", "size": 20},
     }
-    with base_app.test_request_context("?author_recid=Jones, Jessica"):
+    with base_app.test_request_context():
```
drjova: I'd refactor this to use api_client

backend/inspirehep/search/facets.py
```diff
@@ -81,6 +98,10 @@ def inner(values):
     return inner
 
 
+def get_filters_without_excluded(filters, excluded_filters):
+    return {key: val for key, val in filters.items() if key not in excluded_filters}
```
drjova: i'm just wondering, do we really want to filter out the filters? We'll have to keep them in sync and this is what we try to avoid

monaawi: Well, citations summary and citations by year are a special case where we need to filter out some of the filters. The graphs use citeable, refereed, citation count etc. to filter out the results once you click on one of the columns. But then the graph shouldn't update themselves as well. That is why those filters need to not change the citation summary

drjova: 👍 good point

backend/tests/integration/search/test_facets.py
monaawi: I'm not sure I understand what you mean. The reason I added the test request context is because we extract the author from the context for the hep author publications author exclusion and the self affiliations

PR #769, opened 2020-03-04
BibTeX: add "\ " after "." in journal title

MJedr:
INSPIR-3246

### DIFF
--- a/backend/inspirehep/migrator/cli.py
+++ b/backend/inspirehep/migrator/cli.py
@@ -45,7 +45,7 @@ def touch_file(file):
     try:
         os.utime(file)
     except FileNotFoundError:
-        os.mknod(file)
+        open(file, 'w').close()
 
 
 @click.group()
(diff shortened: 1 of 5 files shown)


backend/inspirehep/migrator/cli.py
```diff
@@ -45,7 +45,10 @@ def touch_file(file):
     try:
         os.utime(file)
     except FileNotFoundError:
-        os.mknod(file)
+        try:
+            os.mknod(file)
+        except PermissionError:
+            open(file, 'w').close()
```
michamos: * this is sufficient as it always works:
```suggestion
        open(file, 'w').close()
```
* in principle, it should be in a separate commit, as it has nothing to do with BibTeX. If you don't know how to do, it's OK for this time.

michamos:
Also backend/celerybeat-schedule.db was included in your commit by mistake.

MJedr:
I forgot to rebase and change the commit message in the last commit, sorry

PR #773, opened 2020-03-05
ui: Change citesummary layout

monaawi:
* INSPIR-3213


ui/src/actions/user.js
```diff
@@ -137,6 +138,13 @@ export function setPreferredCiteFormat(format) {
   };
 }
 
+export function setIsCitationSummaryEnabled(isCitationSummaryEnabled) {
```
harunurhan: I was just thinking if it would be a good idea to have `preferences: { }` in `user` state, instead of creating actions for each of them.

ui/src/common/components/CitationSummaryGraph/CitationSummaryGraph.jsx
```diff
@@ -125,11 +120,11 @@ class CitationSummaryGraph extends Component {
   }
 
   onBarMouseHover(hoveredDatapoint, hoveredBar) {
```
harunurhan: `hoveredDatapoint` should be removed from here.

ui/src/literature/components/CitationSummarySwitch.jsx
```diff
@@ -1,23 +1,44 @@
...
 const CHART_ICON = <BarChartOutlined />;
 
-function CitationSummarySwitch(props) {
-  const { checked } = props;
-
+function CitationSummarySwitch({
+  checked,
+  updateHash,
+  onChange,
+  isCitationSummaryEnabled,
```
harunurhan: we should have a clear name to indicate this is user preferences, not the actually `enabled` state

ui/ui-tests/utils/constants.js
```diff
@@ -12,14 +12,14 @@ const routes = {
...
+    literatureSearch: `${appBaseUrl}/literature#with-citation-summary`,
     literatureDetail1472986: `${appBaseUrl}/literature/1472986`,
     authorSearch: `${appBaseUrl}/authors`,
-    authorDetail1274753: `${appBaseUrl}/authors/1274753`,
+    authorDetail1274753: `${appBaseUrl}/authors/1274753#with-citation-summary`,
     jobsSearch: `${appBaseUrl}/jobs`,
     jobDetail1727456: `${appBaseUrl}/jobs/1727456`,
     conferenceSearch: `${appBaseUrl}/conferences`,
-    conferenceDetail1339293: `${appBaseUrl}/conferences/1339293`,
+    conferenceDetail1339293: `${appBaseUrl}/conferences/1339293#with-citation-summary`,
```
harunurhan: let's make this one closed

ui/src/literature/components/CitationSummarySwitch.jsx
```diff
@@ -1,23 +1,44 @@
...
+}) {
+  useEffect(
+    () => {
+      if (isCitationSummaryEnabled) {
```
harunurhan: you can pull this logic out the container and have something like `onCitationSummaryUserPreferenceChange` (yeah kinda long name 😄)

harunurhan: and we will eventually get rid of this effect, when we find a better way.

ui/src/actions/user.js
monaawi: so you want to move also ```preferredCiteFormat``` to these preferences?

PR #776, opened 2020-03-06
views: add redirect views for legacy

miguelgrc:
Redirects links with legacy format to the correct hep endpoint

* INSPIR-3208

**Needs to run alembic**


backend/inspirehep/legacy/views.py
```diff
@@ -0,0 +1,47 @@
+# -*- coding: utf-8 -*-
+#
+# Copyright (C) 2020 CERN.
+#
+# inspirehep is free software; you can redistribute it and/or modify it under
+# the terms of the MIT License; see LICENSE file for more details.
+
+import flask
```
drjova: you don't have to import flask

backend/inspirehep/legacy/views.py
```diff
@@ -0,0 +1,47 @@
...
+from flask import Blueprint, abort, request
```
drjova:
```suggestion
from flask import Blueprint, abort, request, redirect
```

backend/inspirehep/legacy/views.py
```diff
@@ -0,0 +1,47 @@
...
+
+    return flask.redirect(f"/authors/{recid}", 301)
+
+
+@blueprint.route("/search", methods=("GET",))
+def redirect_query():
+    collection_equivalence = {
+        "HEP": "literature",
+        "HepNames": "authors",
+        "Conferences": "conferences",
```
drjova: conferences don't need anymore, we have them on prod now

backend/inspirehep/pidstore/api/base.py
```diff
@@ -142,3 +142,16 @@ def get_pid_from_record_uri(uri):
         except IndexError:
             return None
         return pid_type, pid_value
+
+    @staticmethod
+    def get_pid_type_from_recid(recid):
+        return (
+            PersistentIdentifier.query.with_entities(PersistentIdentifier.pid_type)
+            .filter_by(pid_value=recid)
+            .one()[0]
```
drjova: one() should return the object not an array, also it would be better to catch the exception in case it doesn't exists

backend/inspirehep/legacy/views.py
```diff
@@ -0,0 +1,47 @@
...
+from flask import Blueprint, abort, request
+from sqlalchemy.orm.exc import NoResultFound
+
+from inspirehep.pidstore.api import PidStoreBase
+from inspirehep.records.utils import get_pid_for_pid
+
+blueprint = Blueprint("inspirehep_legacy", __name__, url_prefix="")
+
+
+@blueprint.route("/record/<path:recid>", methods=("GET",))
```
drjova: isn't it ``int``?

backend/inspirehep/legacy/views.py
miguelgrc: But we still want to redirect old links right?

drjova: we didn't do it for Jobs, that's why I think Conferences should follow the same behaviour. WDYT @michamos ?

backend/inspirehep/pidstore/api/base.py
miguelgrc: But with_entities returns a tuple. 
Should I? I think it's better to do it in the view.

backend/inspirehep/legacy/views.py
miguelgrc: But we need to pass it as string for the query.

backend/inspirehep/pidstore/api/base.py
michamos: you can use `.scalar()` which unpacks the element.

backend/inspirehep/legacy/views.py
michamos: I don't think there is any harm in redirecting everything, even if it's probably most useful for HEP.

backend/inspirehep/pidstore/api/base.py
```diff
@@ -142,3 +142,30 @@ def get_pid_from_record_uri(uri):
...
+        """Returns the ``pid_type`` of the provided ``recid``.
+
+        Args:
+            recid (str): the ``recid``
+        Returns:
+            The ``pid_type`` if found. None otherwise.
+        """
+        return (
+            PersistentIdentifier.query.with_entities(PersistentIdentifier.pid_type)
+            .filter_by(pid_value=recid)
```
michamos: there's no index on `pid_value`, so this is a full table scan:
```
explain analyze select pid_type from pidstore_pid where pid_value = '4328';
```
yields:
```
Seq Scan on pidstore_pid  (cost=0.00..118257.46 rows=1 width=4) (actual time=627.701..1416.902 rows=1 loops=1)
  Filter: ((pid_value)::text = '4328'::text)
  Rows Removed by Filter: 3163790
Planning time: 6.127 ms
Execution time: 1416.986 ms
```
we should probably add an index on `pid_type` (the current index is on `(pid_type, pid_value)`, so wrong order for this, and it's useful for the `like` queries on CNUMs).

backend/inspirehep/pidstore/api/base.py
```diff
@@ -142,3 +142,30 @@ def get_pid_from_record_uri(uri):
...
+            .scalar()
+        )
+
+    @staticmethod
+    def get_endpoint_from_recid(recid):
```
michamos:
```suggestion
    def get_endpoint_for_recid(recid):
```

drjova:
@miguelgrc that's an easy fix 
```
/inspirehep/legacy/views.py:9:1: F401 'sqlalchemy.orm.exc.NoResultFound' imported but unused
```

backend/inspirehep/pidstore/api/base.py
miguelgrc: Oh, that's interesting. So for that I should create an alembic migration that creates a `uidx_pid_value` unique index on `pid_value`?

michamos: you should use the proper naming convention (these indexes are old and don't respect it), so `idx_pidstore_pid_pid_value`, and the uniqueness constraint is not needed nor wanted (there might be collisions in theory between `pid_value` with different `pid_type`. It's probably better to create a partial index where `provider=recid` and add that as an extra filter in the query, as we don't need it for the other external pids.

backend/inspirehep/legacy/views.py
```diff
@@ -0,0 +1,46 @@
...
+@blueprint.route("/search", methods=("GET",))
+def redirect_query():
+    collection_equivalence = {
+        "HEP": "literature",
+        "HepNames": "authors",
+        "Conferences": "conferences",
+        "Jobs": "jobs",
+    }
+
+    collection = request.args.get("cc")
```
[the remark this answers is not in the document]
StellaCh: Search in literature or coming soon message won't provide them with the information they want. Since we are keeping legacy for 2 months and we are planning to implement institutions/experiments in the meantime, it'd be better to redirect them to legacy for the time being.

drjova:
@miguelgrc please add in your commit message that we have to run alembic

backend/inspirehep/legacy/views.py
```diff
@@ -0,0 +1,55 @@
...
+    }
+
+    legacy_collection = request.args.get("cc", "HEP")
+    query = request.args.get("p")
+    collection = collection_equivalence.get(legacy_collection)
+    if collection:
+        return redirect(f"/{collection}?q={query}", 301)
+    else:
+        return redirect(
+            f"https://legacy.inspirehep.net/search?cc={legacy_collection}&p={query}",
```
michamos: here we can pass through request args, rather than only `cc` and `p`.

drjova: hmm what's the point? we don't have the same args

michamos: this is going to legacy. Legacy knows what to do with the extra args. Also, you shouldn't hardcode the legacy url, I don't think it's settled which URL we'll use yet.

drjova: my bad, I didn't see the url properly

backend/inspirehep/legacy/views.py
```diff
@@ -0,0 +1,56 @@
...
+
+    legacy_collection = request.args.get("cc", "HEP")
+    query = request.args.get("p")
+
+    collection = collection_equivalence.get(legacy_collection)
+    if collection:
+        return redirect(f"/{collection}?q={query}", 301)
+    else:
+        return redirect(
+            f"{current_app.config['LEGACY_URL']}/search?{request.query_string.decode('utf-8')}",
```
michamos:
```suggestion
            f"{current_app.config['LEGACY_BASE_URL']}/search?{request.query_string.decode('utf-8')}",
```
that's already defined in the config.

miguelgrc: Oops, didn't notice it, I changed it.

backend/tests/integration/legacy/test_views.py
```diff
@@ -0,0 +1,81 @@
...
+    api_client, db, es_clear, create_record, base_app
+):
+    path = "/search?cc=Institutions&p=CERN&whatever=something"
+    response = api_client.get(path)
+
+    response_status_code = response.status_code
+    response_location_header = response.headers.get("Location")
+
+    expected_status_code = 302
+    expected_redirect_url = f"{base_app.config['LEGACY_BASE_URL']}{path}"
```
michamos: you should hardcode this, tests should be as stupid as possible so you avoid potential bugs in tests (as you don't have tests-for-tests).

miguelgrc: Done, thanks for all the comments!

backend/tests/integration/legacy/test_views.py
```diff
@@ -0,0 +1,86 @@
...
+        base_app.config, {"LEGACY_BASE_URL": "https://legacy.inspirehep.net"}
+    ):
+        path = "/search?cc=Institutions&p=CERN&whatever=something"
+        response = api_client.get(path)
+
+        response_status_code = response.status_code
+        response_location_header = response.headers.get("Location")
+
+        expected_status_code = 302
+        expected_redirect_url = f"{base_app.config['LEGACY_BASE_URL']}{path}"
```
michamos:
```suggestion
        expected_redirect_url = "https://legacy.inspirehep.net/search?cc=Institutions&p=CERN&whatever=something"
```
as I said, better be explicit, otherwise you might hide bugs.

PR #779, opened 2020-03-06
facets: reoder doc type facet

monaawi:
* INSPIR-3255

### DIFF
--- a/backend/inspirehep/search/facets.py
+++ b/backend/inspirehep/search/facets.py
@@ -6,6 +6,7 @@
 # the terms of the MIT License; see LICENSE file for more details.
 
 from datetime import datetime
+from itertools import count
 
 from elasticsearch_dsl.query import Q, Range
 from flask import current_app, request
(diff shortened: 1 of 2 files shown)


backend/inspirehep/search/facets.py
```diff
@@ -174,10 +174,10 @@ def records_hep():
         "aggs": {
             **hep_earliest_date_aggregation(order=1),
```
michamos: why not do
```
from itertools import count
order = count(start=1)
```
then
```suggestion
            **hep_earliest_date_aggregation(order=next(order)),
```
etc., instead of hardcoding the order?

backend/inspirehep/search/facets.py
```diff
@@ -300,44 +306,51 @@ def results=[:];
...
+            **hep_collection_aggregation(order=next(order)),
         }
     )
     return records
 
 
 def records_jobs_cataloger():
     records = records_jobs()
-    records["aggs"].update({**jobs_status_aggregation(order=4)})
+    order = count(start=4)
```
monaawi: for the cases of cataloger facets, I want to have a function that gives me the last order of the copied aggregations but then of course, I hit nested aggregations and  would have needed to look into nested dictionaries to retrieve the meta (as I don't know at which level it is for each one), or I could add a useless meta just for this at the level I expect it to be. Do you think it is worth it?

michamos: you can also add `order` as a kwarg for all facet functions instead, like
```
def records_jobs_cataloger(order=count(start=1))
```
In the caller, you can then inject the parent caller if needed.

monaawi: yeah, that's a good idea, thanks. But the count can't be the default argument to the function, I will need to default it to None and add a check 
`    if order is None:
        order = count(start=1)` because counter is mutable

PR #783, opened 2020-03-09
bibtex: change doc_type of cof. papers with journal title to article

MJedr:
* INSPIR-3245

### DIFF
--- a/backend/inspirehep/records/marshmallow/literature/bibtex.py
+++ b/backend/inspirehep/records/marshmallow/literature/bibtex.py
@@ -49,6 +49,21 @@ class BibTexCommonSchema(Schema):
     volume = fields.Method("get_volume")
     year = fields.Method("get_year")
 
+    document_type_map = {
+        "article": "article",
+        "book": "book",
+        "book chapter": "inbook",
+        "proceedings": "proceedings",
+        "report": "techreport",
+        "note": "article",
+        "conference paper": lambda data: "article"
+        if get_value(data, "publication_info.journal_title")
+        else "inproceedings",
+        "thesis": lambda data: "phdthesis"
+        if get_value(data, "thesis_info.degree_type") in ("phd", "habilitation",)
+        else "mastersthesis",
+    }
+
     @staticmethod
     def get_date(data, doc_type):
         publication_year = BibTexCommonSchema.get_best_publication_info(data).get(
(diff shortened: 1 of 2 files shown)


backend/inspirehep/records/marshmallow/literature/bibtex.py
```diff
@@ -79,14 +79,19 @@ def get_document_type(data, doc_type):
...
             "proceedings": "proceedings",
             "report": "techreport",
             "note": "article",
         }
 
         if doc_type in DOCUMENT_TYPE_MAP:
             return DOCUMENT_TYPE_MAP[doc_type]
+        elif doc_type == "conference paper" and any(
+            [entry.get("journal_title") is not None
+             for entry in get_value(data, "publication_info")]):
```
michamos: * you can access nested fields with `get_value` (that's the whole point):
```suggestion
        elif doc_type == "conference paper" and get_value(data, "publication_info.journal_title"):
```
* this `if`...`elif` chain is starting to get a bit unwieldy. The general pattern seems to be: either the type is static, or the type depends on the data. So I would suggest allowing the argument of the mapping to be either a string or a function, so you could have
```python
"conference paper": lambda data: "article" if get_value(data, "publication_info.journal_title") else "inproceedings"
```
You can also move `DOCUMENT_TYPE_MAP` to the class scope (and change to lower-case, this is a leftover from previous refactoring) so it's doesn't have to be evaluated for every record.

backend/inspirehep/records/marshmallow/literature/bibtex.py
```diff
@@ -28,6 +28,16 @@ class BibTexCommonSchema(Schema):
...
     doc_type = fields.Raw()
+    document_type_map = {
+        "article": "article",
+        "book": "book",
+        "book chapter": "inbook",
+        "proceedings": "proceedings",
+        "report": "techreport",
+        "note": "article",
+        "conference paper": lambda data: "article" if get_value(
+            data, "publication_info.journal_title") else "inproceedings"
```
michamos: I guess it wasn't clear, but I suggested to do this because the same pattern is also used for `thesis` in `get_document_type`. You should also put this further down in the class, as it's not part of the serialized fields.

backend/tests/unit/records/marshmallow/literature/test_bibtex_schema.py
```diff
@@ -492,3 +492,16 @@ def test_get_title_with_subtitle():
...
+
+
+def test_bibtex_document_type_conference_paper_with_journal_title():
+    record = {"document_type": ["conference paper"],
+              "publication_info": [
+                  {"pubinfo_freetext": "Geneva, Switzerland: CERN (2002) 401 p"},
+                  {"pubinfo_freetext": "Geneva, Switzerland: CERN (2002) 401 p",
+                   "journal_title": "Nucl.Part.Phys.Proc."}
+              ]
+              }
```
michamos: do you have [pre-commit hooks](https://pre-commit.com/) set up? black should run automatically and format the modified files. Formatting looks off here.

PR #789, opened 2020-03-10
bibtex: remove escaping from math expressions

MJedr:
* INSPIR-3243

### DIFF
--- a/backend/inspirehep/records/serializers/bibtex.py
+++ b/backend/inspirehep/records/serializers/bibtex.py
@@ -4,6 +4,10 @@
 #
 # inspirehep is free software; you can redistribute it and/or modify it under
 # the terms of the MIT License; see LICENSE file for more details.
+import codecs
+import re
+
+import latexcodec  # noqa: F401
 import structlog
 from invenio_records_rest.serializers.response import search_responsify
 from pybtex.database import BibliographyData, Entry, Person
@@ -16,6 +20,21 @@ LOGGER = structlog.getLogger()
 
 
 class BibtexWriter(Writer):
+    math_expression_regex = re.compile(r"(?<!\\)\$.*?\$")
+
+    def _encode(self, text):
+        if "$" not in text:
+            return codecs.encode(text, "ulatex+latin1", "keep")
+        math_expressions = self.math_expression_regex.findall(text)
+        encoded_math_expressions = {
+            codecs.encode(expression, "ulatex+latin1", "keep"): expression
+            for expression in math_expressions
+        }
+        encoded_text = codecs.encode(text, "ulatex+latin1", "keep")
+        for encoded_math, math in encoded_math_expressions.items():
+            encoded_text = encoded_text.replace(encoded_math, math)
+        return encoded_text
+
     def _write_persons(self, stream, persons, role):
         if len(persons) > 10:
             self._write_field(
--- a/backend/tests/integration/records/serializers/test_bibtex.py
+++ b/backend/tests/integration/records/serializers/test_bibtex.py
@@ -4,6 +4,7 @@
 #
 # inspirehep is free software; you can redistribute it and/or modify it under
 # the terms of the MIT License; see LICENSE file for more details.
+import pytest
 
 
 def test_bibtex(api_client, db, es, create_record):
@@ -127,7 +128,9 @@ def test_bibtex_search(api_client, db, es, create_record):
     create_record("lit", data=data_2)
 
     expected_status_code = 200
-    expected_result_1 = "@article{637275237,\n" '    title = "{This is a title.}"\n' "}\n"
+    expected_result_1 = (
+        "@article{637275237,\n" '    title = "{This is a title.}"\n' "}\n"
+    )
     expected_result_2 = (
         "@article{637275232,\n" '    title = "{Yet another title.}"\n' "}\n"
     )
@@ -139,3 +142,54 @@ def test_bibtex_search(api_client, db, es, create_record):
     assert expected_status_code == response_status_code
     assert expected_result_1 in response_data
     assert expected_result_2 in response_data
+
+
+def test_bibtex_doesnt_encode_math_environments(api_client, db, es, create_record):
+    headers = {"Accept": "application/x-bibtex"}
+    data = {
+        "control_number": 637275237,
+        "titles": [
+            {
+                "title": "Low-energy theorem for $\\gamma\\to 3\\pi$: Σ surface terms against $\\pi a_1$-mixing"
+            }
+        ],
+    }
+    record = create_record("lit", data=data)
+    record_control_number = record["control_number"]
+
+    expected_status_code = 200
+    expected_result = '@article{637275237,\n    title = "{Low-energy theorem for $\\gamma\\to 3\\pi$: $\\Sigma$ surface terms against $\\pi a_1$-mixing}"\n}\n'
+    response = api_client.get(
+        "/literature/{}".format(record_control_number), headers=headers
+    )
+
+    response_status_code = response.status_code
+    response_data = response.get_data(as_text=True)
+    assert expected_status_code == response_status_code
+    assert expected_result == response_data
+
+
+@pytest.mark.xfail(reason="latexcodec doesn't know about the special characters")
+def test_bibtex_encodes_unicode_outside_of_math_environments(
+    api_client, db, es, create_record
+):
+    headers = {"Accept": "application/x-bibtex"}
+    data = {
+        "control_number": 637275237,
+        "titles": [
+            {"title": "Core polarization effects up to 12ℏω in 7Li and 10B nuclei"}
+        ],
+    }
+    record = create_record("lit", data=data)
+    record_control_number = record["control_number"]
+
+    expected_status_code = 200
+    expected_result = '@article{637275237,\n    title = "{Core polarization effects up to 12$\hbar\omega$ in 7Li and 10B nuclei}"\n}\n'
+    response = api_client.get(
+        "/literature/{}".format(record_control_number), headers=headers
+    )
+
+    response_status_code = response.status_code
+    response_data = response.get_data(as_text=True)
+    assert expected_status_code == response_status_code
+    assert expected_result == response_data


backend/inspirehep/records/serializers/bibtex.py
```diff
@@ -16,6 +19,24 @@
...
+        import latexcodec  # NOQA
+
+        if "$" not in text:
+            return codecs.encode(text, "ulatex+{}".format(self.encoding))
+        else:
+            found_math_expressions = [
+                f"{expression}" for expression in re.findall(r"\$.*\$", text)
+            ]
+            encoded_text = codecs.encode(text, "ulatex+{}".format(self.encoding))
+            for list_nb, _math_expression in enumerate(
```
michamos: * Better to use [`zip`](https://docs.python.org/3.6/library/functions.html#zip) than fiddling with list indices for this kind of things. 
* Why the leading `_`?
* Nice trick, I had something much more complex in mind (namely splitting the string on math boundaries while keeping track of which mode every part is, and only encoding those that need encoding). Although it shouldn't happen, there's the risk that you get a different number of math segments in the non-encoded and encoded strings.
I would add an assertion like
```python
assert len(found_math_expressions) == len(encoded_math_expressions), f"number of math parts in LaTeX-encoded string is different for '{text}'."
```
before to make sure this doesn't happen.

backend/tests/integration/records/serializers/test_bibtex.py
```diff
@@ -139,3 +141,29 @@ def test_bibtex_search(api_client, db, es, create_record):
     assert expected_status_code == response_status_code
     assert expected_result_1 in response_data
     assert expected_result_2 in response_data
+
+
+def test_bibtex_with_math_environment_in_title(api_client, db, es, create_record):
```
michamos: can you add tests for the corner case I mentioned, and a test checking that text is correctly encoded? These are strings you can use:
* `"Low-energy theorem for $\\gamma\\to 3\\pi$: Σ surface terms against $\\pi a_1$-mixing"` (this one is a bit fake, I added the `Σ`)
* `"Core polarization effects up to 12ℏω in 7Li and 10B nuclei"` (this one is real but latexcodec doesn't know about these characters, so it's not encoded properly, you should give the correct expected output and xfail the test).

backend/tests/integration/records/serializers/test_bibtex.py
```diff
@@ -139,3 +142,55 @@ def test_bibtex_search(api_client, db, es, create_record):
...
+                "title": "Low-energy theorem for $\\gamma\\to 3\\pi$: Σ surface terms against $\\pi a_1$-mixing"
+            }
+        ],
+    }
+    record = create_record("lit", data=data)
+    record_control_number = record["control_number"]
+
+    expected_status_code = 200
+    expected_etag = '"application/x-bibtex@v1"'
+    expected_result = '@article{637275237,\n    title = "{Low-energy theorem for $\\gamma\\to 3\\pi$: Σ surface terms against $\\pi a_1$-mixing}"\n}\n'
```
michamos: what I expect is
```suggestion
    expected_result = '@article{637275237,\n    title = "{Low-energy theorem for $\\gamma\\to 3\\pi$: $\Sigma$ surface terms against $\\pi a_1$-mixing}"\n}\n'
```

backend/inspirehep/records/serializers/bibtex.py
```diff
@@ -16,6 +19,30 @@
...
+            encoded_math_expressions = re.findall(math_expression_regex, encoded_text)
+            for encoded_math_expression, found_math_expression in zip(
+                encoded_math_expressions, found_math_expressions
+            ):
+                encoded_text = encoded_text.replace(
+                    encoded_math_expression, found_math_expression
+                )
+            assert len(found_math_expressions) == len(
+                encoded_math_expressions
+            ), f"number of math parts in LaTeX-encoded string is different for '{text}'."
```
michamos: in general it's better to put assertions doing sanity checks before the code that uses the assumption, so
1. you don't do useless work in case it fails
2. you can rely on the assumption being true when you need it

backend/inspirehep/records/serializers/bibtex.py
```diff
@@ -16,6 +19,30 @@
 
 
 class BibtexWriter(Writer):
+    def _encode(self, text):
+        import latexcodec  # NOQA
+
+        if "$" not in text:
+            return codecs.encode(text, "ulatex+{}".format(self.encoding))
+        else:
+            math_expression_regex = re.compile(r"(?<!\\)\$.*?\$")
```
michamos: it's better to put this kind of constants out of function bodies, so you don't need to evaluate it on every function call and it's more visible. Here you can put it on the class.

backend/inspirehep/records/serializers/bibtex.py
```diff
@@ -16,6 +19,27 @@
...
+            found_math_expressions = re.findall(math_expression_regex, text)
```
michamos: you can also do
```suggestion
            found_math_expressions = math_expression_regex.findall(text)
```
and maybe `found_math_expressions` is a bit long, what about `math_expressions`?

backend/inspirehep/records/serializers/bibtex.py
```diff
@@ -16,6 +19,27 @@
...
+            encoded_text = codecs.encode(text, "ulatex+{}".format(self.encoding))
+            encoded_math_expressions = re.findall(math_expression_regex, encoded_text)
```
michamos: same
```suggestion
            encoded_math_expressions = math_expression_regex.findall(encoded_text)
```

backend/inspirehep/records/serializers/bibtex.py
```diff
@@ -16,6 +19,27 @@
...
+            assert len(found_math_expressions) == len(
+                encoded_math_expressions
+            ), f"number of math parts in LaTeX-encoded string is different for '{text}'."
+            for encoded_math_expression, found_math_expression in zip(
```
michamos: again, variable names are a bit long here I think. What about
```suggestion
            for encoded_math, original_math in zip(
```

backend/tests/integration/records/serializers/test_bibtex.py
michamos: this is important, otherwise it seems you're not encoding anything.

backend/tests/integration/records/serializers/test_bibtex.py
```diff
@@ -139,3 +142,55 @@ def test_bibtex_search(api_client, db, es, create_record):
     assert expected_status_code == response_status_code
     assert expected_result_1 in response_data
     assert expected_result_2 in response_data
+
+
+def test_bibtex_with_math_environment_in_title(api_client, db, es, create_record):
```
michamos:
```suggestion
def test_bibtex_doesnt_encode_math_environments(api_client, db, es, create_record):
```

backend/tests/integration/records/serializers/test_bibtex.py
```diff
@@ -139,3 +142,55 @@ def test_bibtex_search(api_client, db, es, create_record):
...
+    )
+
+    response_status_code = response.status_code
+    response_data = response.get_data(as_text=True)
+    assert expected_status_code == response_status_code
+    assert expected_result == response_data
+
+
+@pytest.mark.xfail(reason="latexcodec doesn't know about the special characters")
+def test_bibtex_with_not_encoded_math_environment_in_title_fails(
```
michamos:
```suggestion
def test_bibtex_encodes_unicode_outside_of_math_environments(
```

backend/tests/integration/records/serializers/test_bibtex.py
```diff
@@ -139,3 +142,55 @@ def test_bibtex_search(api_client, db, es, create_record):
...
+        "control_number": 637275237,
+        "titles": [
+            {"title": "Core polarization effects up to 12ℏω in 7Li and 10B nuclei"}
+        ],
+    }
+    record = create_record("lit", data=data)
+    record_control_number = record["control_number"]
+
+    expected_status_code = 200
+    expected_result = '@article{637275237,\n    title = "{Low-energy theorem for $\\gamma\\to 3\\pi$: $\Sigma$ surface terms against $\\pi a_1$-mixing}"\n}\n'
```
michamos:
```suggestion
    expected_result = '@article{637275237,\n    title = "{Core polarization effects up to 12$\hbar\omega$ in 7Li and 10B nuclei}"\n}\n'
```

backend/inspirehep/records/serializers/bibtex.py
```diff
@@ -16,6 +19,27 @@
 
 
 class BibtexWriter(Writer):
+    def _encode(self, text):
+        import latexcodec  # NOQA
+
+        if "$" not in text:
+            return codecs.encode(text, "ulatex+{}".format(self.encoding))
+        else:
```
michamos: also [flat is better than nested](https://www.python.org/dev/peps/pep-0020/): you don't need the else here, as the function `return`s if it takes the `if` branch.

backend/inspirehep/records/serializers/bibtex.py
```diff
@@ -16,6 +19,31 @@
 
 
 class BibtexWriter(Writer):
+    math_expression_regex = re.compile(r"(?<!\\)\$.*?\$")
+
+    def _encode(self, text):
+        import latexcodec  # NOQA
```
michamos: you can also move the import out of the method.

backend/inspirehep/records/serializers/bibtex.py
```diff
@@ -16,6 +20,25 @@
 
 
 class BibtexWriter(Writer):
+    math_expression_regex = re.compile(r"(?<!\\)\$.*?\$")
+
+    def _encode(self, text):
+        if "$" in text:
```
michamos: That's not what I meant :sweat_smile: 
```
if "$" not in text:
    return ...
# rest of the function body without else
```

backend/tests/integration/records/serializers/test_bibtex.py
```diff
@@ -139,3 +142,54 @@ def test_bibtex_search(api_client, db, es, create_record):
...
     assert expected_result_2 in response_data
+
+
+def test_bibtex_doesnt_encode_math_environments(api_client, db, es, create_record):
+    headers = {"Accept": "application/x-bibtex"}
+    data = {
+        "control_number": 637275237,
+        "titles": [
+            {
+                "title": "Low-energy theorem for $\\gamma\\to 3\\pi$: $\Sigma$ surface terms against $\\pi a_1$-mixing"
```
michamos:
```suggestion
                "title": "Low-energy theorem for $\\gamma\\to 3\\pi$: Σ surface terms against $\\pi a_1$-mixing"
```
The test needs to assert that if you have a unicode character, it gets encoded properly:
```
In [1]: import latexcodec                                                       

In [2]: import codecs                                                           

In [3]: codecs.encode("Σ", "ulatex")                                            
Out[3]: '$\\Sigma$'
```
If that doesn't happen, there's a bug in your logic, which this test is meant to prevent.

backend/inspirehep/records/serializers/bibtex.py
michamos: you didn't move it, you copied it.

backend/inspirehep/records/serializers/bibtex.py
```diff
@@ -4,6 +4,10 @@
 #
 # inspirehep is free software; you can redistribute it and/or modify it under
 # the terms of the MIT License; see LICENSE file for more details.
+import codecs
+import re
+
+import latexcodec
```
michamos: you need to silence `flake8` here (http://flake8.pycqa.org/en/3.1.1/user/ignoring-errors.html#in-line-ignoring-errors).

backend/inspirehep/records/serializers/bibtex.py
```diff
@@ -16,6 +20,25 @@
...
 
 class BibtexWriter(Writer):
+    math_expression_regex = re.compile(r"(?<!\\)\$.*?\$")
+
+    def _encode(self, text):
+        import latexcodec  # NOQA
+
+        if "$" not in text:
+            return codecs.encode(text, "ulatex+{}".format(self.encoding))
+        math_expressions = BibtexWriter.math_expression_regex.findall(text)
```
michamos: attributes and methods on the current class can be accessed through `self`:

```suggestion
        math_expressions = self.math_expression_regex.findall(text)
```

backend/inspirehep/records/serializers/bibtex.py
```diff
@@ -16,6 +20,25 @@
...
+        encoded_text = codecs.encode(text, "ulatex+{}".format(self.encoding))
+        encoded_math_expressions = BibtexWriter.math_expression_regex.findall(
```
michamos:
```suggestion
        encoded_math_expressions = self.math_expression_regex.findall(
```

PR #792, opened 2020-03-11
migrator: reraise all exceptions on migrate record

pazembrz:


backend/inspirehep/migrator/tasks.py
```diff
@@ -291,13 +291,14 @@ def create_records_from_mirror_recids(self, recids):
...
-            continue
+            raise
     try:
         db.session.commit()
     except Exception:
         LOGGER.exception(
             "Cannot commit migrated records", processed_records=list(processed_records)
         )
+        raise
     models_committed.connect(index_after_commit)
```
michamos:
```suggestion
    finally:
        models_committed.connect(index_after_commit)
```

pazembrz: I'm not sure if this is needed as after raising this db connection will be killed. But better safe than sorry. :+1:

backend/inspirehep/migrator/tasks.py
```diff
@@ -291,14 +291,16 @@ def create_records_from_mirror_recids(self, recids):
             raise
         except Exception:
             LOGGER.exception("Cannot migrate record", recid=recid)
-            continue
+            raise
     try:
         db.session.commit()
     except Exception:
```
michamos: we re-raise for any `Exception`, but we don't retry for `Exception`. We probably should catch and re-raise separately the errors that actually occur on `commit` and retry those, keeping the generic `Exception` catching to avoid the chord blowing up on unexpected exceptions.

backend/inspirehep/migrator/tasks.py
```diff
@@ -291,14 +291,16 @@ def create_records_from_mirror_recids(self, recids):
...
         LOGGER.exception(
             "Cannot commit migrated records", processed_records=list(processed_records)
         )
-    models_committed.connect(index_after_commit)
+        raise
+    finally:
```
michamos: actually, there should be a `try`...`finally` around the whole block. If we `raise` from one of the previous exceptions, the signal won't be reattached. This doesn't matter currently as workers only process the migration and we index records separately, but will cause subtle bugs once we also do disambiguation in the workers.

pazembrz: I think this should resolve the issue

backend/inspirehep/migrator/tasks.py
pazembrz: :+1:

backend/inspirehep/migrator/tasks.py
```diff
@@ -275,30 +276,41 @@ def create_records_from_mirror_recids(self, recids):
...
+                    LOGGER.warning("Record is empty", recid=recid)
+            except OperationalError:
+                LOGGER.exception("Got operational error", recid=recid)
+                raise
+            except Exception:
+                LOGGER.exception("Cannot migrate record", recid=recid)
+                raise
+
         db.session.commit()
+    except (InvalidRequestError, OperationalError, StatementError):
```
michamos: :+1: you also need to update https://github.com/inspirehep/inspirehep/pull/792/files#diff-2718fbe727bdf849c6819bf9d3ffa4d6R268

pazembrz: :+1:

PR #796, opened 2020-03-12
facets: Add exclude RPP facet

monaawi:
* INSPIR-3239

### DIFF
--- a/backend/inspirehep/records/config.py
+++ b/backend/inspirehep/records/config.py
@@ -23,6 +23,7 @@ from inspirehep.access_control import (
     api_access_permission_check,
     session_cataloger_permission_factory,
 )
+from inspirehep.search.aggregations import hep_rpp
 from inspirehep.search.api import (
     AuthorsSearch,
     ConferencesSearch,
@@ -38,6 +39,7 @@ from inspirehep.search.facets import (
     citations_by_year,
     conferences_date_range_contains_other_conferences,
     conferences_start_date_range_filter,
+    filter_from_filters_aggregation,
     hep_author_publications,
     hep_author_publications_cataloger,
     hep_conference_contributions,
@@ -414,6 +416,7 @@ HEP_FILTERS = {
     "collection": must_match_all_filter("_collections"),
     "subject": must_match_all_filter("facet_inspire_categories"),
     "arxiv_categories": must_match_all_filter("facet_arxiv_categories"),
+    "rpp": filter_from_filters_aggregation(hep_rpp(order=1)),
 }
 
 JOBS_FILTERS = {
--- a/backend/inspirehep/search/aggregations.py
+++ b/backend/inspirehep/search/aggregations.py
@@ -202,3 +202,34 @@ def hep_collection_aggregation(order, title="Collection", agg_type="checkbox"):
             "meta": {"title": title, "order": order, "type": agg_type},
         }
     }
+
+
+def hep_rpp(order, title="Exclude RPP", agg_type="checkbox"):
+    return {
+        "rpp": {
+            "filters": {
+                "filters": {
+                    "Exclude Review Of Particle Physics": {
+                        "bool": {
+                            "must_not": [
+                                {
+                                    "match": {
+                                        "titles.full_title": {
+                                            "query": "RPP",
+                                            "operator": "and",
+                                        }
+                                    }
+                                }
+                            ]
+                        }
+                    }
+                }
+            },
+            "meta": {
+                "title": title,
+                "order": order,
+                "type": agg_type,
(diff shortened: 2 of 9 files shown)


backend/inspirehep/search/aggregations.py
```diff
@@ -202,3 +202,34 @@ def hep_collection_aggregation(order, title="Collection", agg_type="checkbox"):
...
+        "rpp": {
+            "filters": {
+                "filters": {
+                    "Exclude Review Of Particle Physics": {
+                        "bool": {
+                            "must_not": [
+                                {
+                                    "match": {
+                                        "titles.full_title": {
+                                            "query": "Review Of Particle Physics",
```
michamos:
```suggestion
                                            "query": "RPP",
```
your current search is wrong, it matches, e.g. https://labs.inspirehep.net/literature/1714863. We put `RPP` as an extra title for the Review of Particle Physics, so that works.

backend/inspirehep/search/aggregations.py
```diff
@@ -202,3 +202,34 @@ def hep_collection_aggregation(order, title="Collection", agg_type="checkbox"):
             "meta": {"title": title, "order": order, "type": agg_type},
         }
     }
+
+
+def hep_rpp(order, title="Exclude RPP", agg_type="checkbox"):
+    return {
+        "rpp": {
+            "filters": {
```
michamos: wouldn't it be easier to use `filter` given you've only one filter? in that way things would be less nested too I think.

backend/inspirehep/serializers.py
```diff
@@ -127,3 +130,23 @@ def _unnest_aggregations(aggregations):
                 new_aggs[agg_key] = agg_value
 
         return new_aggs
+
+    @staticmethod
+    def create_buckets_for_filter_aggregations(aggregations):
```
michamos: would it be possible to combine this with the `_unnest_aggregations` function so you don't have to loop twice on aggregations?

backend/inspirehep/search/aggregations.py
monaawi: yeah, I knew the query was matching other things as well but I didn't know about the extra ```RPP``` in the title :+1:

backend/inspirehep/records/config.py
```diff
@@ -414,6 +416,7 @@
     "collection": must_match_all_filter("_collections"),
     "subject": must_match_all_filter("facet_inspire_categories"),
     "arxiv_categories": must_match_all_filter("facet_arxiv_categories"),
+    "rpp": filter_from_filters_aggregation(hep_rpp(1)),
```
michamos:
```suggestion
    "rpp": filter_from_filters_aggregation(hep_rpp(order=1)),
```

backend/inspirehep/serializers.py
```diff
@@ -105,14 +105,16 @@ def serialize_search(self, pid_fetcher, search_result, **kwargs):
...
+            "aggregations"
+        ] = self.unnest_aggregations_and_create_buckets_for_filter_aggregations(
             search_result.get("aggregations", {})
         )
 
         return json.dumps(search_result)
 
     @staticmethod
-    def _unnest_aggregations(aggregations):
+    def unnest_aggregations_and_create_buckets_for_filter_aggregations(aggregations):
```
michamos: maybe simply
```suggestion
    def flatten_aggregations(aggregations):
```

backend/inspirehep/serializers.py
```diff
@@ -105,14 +105,16 @@ def serialize_search(self, pid_fetcher, search_result, **kwargs):
...
         """Flatten the aggregation dict in case there are nested aggregations."""
```
michamos: out of date

backend/tests/integration/records/views/test_views_literature.py
```diff
@@ -407,29 +409,16 @@ def test_literature_facets_author_count_returns_non_empty_bucket(
...
-        "doc_type",
-        "earliest_date",
-        "subject",
-        "collaboration",
-    ]
-    expected_facet_keys.sort()
-    response_data_facet_keys.sort()
 
     assert expected_status_code == response_status_code
-    assert expected_facet_keys == response_data_facet_keys
```
michamos: why?

monaawi: It was a leftover. But now that I looked into the test, it didn't actually make sense to me. It seems to be testing the exact same thing as [test_literature_facets|https://github.com/inspirehep/inspirehep/blob/fe96ade1a8ca881747a2c56ef55c9dfd4fbb5de3/backend/tests/integration/records/views/test_views_literature.py#L328]

PR #798, opened 2020-03-13
bibtex: get year from earliest date too

MJedr:
* INSPIR-3242

### DIFF
--- a/backend/inspirehep/records/api/base.py
+++ b/backend/inspirehep/records/api/base.py
@@ -230,6 +230,13 @@ class InspireRecord(Record):
             return cls.get_records_by_pids(pids)
         return iter([])
 
+    def copy(self):
+        """Copy the record metadata.
+
+            This is needed because the default ``dict.copy`` always returns a ``dict``.
+        """
+        return type(self)(self)
+
     def get_linked_pids_from_field(self, path):
         """Return a list of (pid_type, pid_value) tuples for all records referenced
         in the field at the given path
--- a/backend/inspirehep/records/marshmallow/literature/bibtex.py
+++ b/backend/inspirehep/records/marshmallow/literature/bibtex.py
@@ -66,17 +66,20 @@ class BibTexCommonSchema(Schema):
 
     @staticmethod
     def get_date(data, doc_type):
+
         publication_year = BibTexCommonSchema.get_best_publication_info(data).get(
             "year"
         )
         thesis_date = get_value(data, "thesis_info.date")
         imprint_date = get_value(data, "imprints.date[0]")
-
-        if doc_type.endswith("thesis"):
-            date_choice = thesis_date or publication_year or imprint_date
-        else:
-            date_choice = publication_year or thesis_date or imprint_date
-
+        earliest_date = data.earliest_date
+        date_map = {
+            "mastersthesis": thesis_date,
+            "phdthesis": thesis_date,
+            "book": imprint_date,
+            "inbook": imprint_date,
+        }
+        date_choice = date_map.get(doc_type) or publication_year or earliest_date
         if date_choice:
             return PartialDate.loads(str(date_choice))
 
@@ -142,13 +145,13 @@ class BibTexCommonSchema(Schema):
 
     def get_month(self, data):
         doc_type = data.get("doc_type")
-        date = BibTexCommonSchema.get_date(data, doc_type)
+        date = self.get_date(data, doc_type)
         if date:
             return date.month
 
     def get_year(self, data):
         doc_type = data.get("doc_type")
-        date = BibTexCommonSchema.get_date(data, doc_type)
+        date = self.get_date(data, doc_type)
         if date:
             return date.year
 
--- a/backend/tests/integration/records/api/test_api_base.py
+++ b/backend/tests/integration/records/api/test_api_base.py
@@ -18,6 +18,7 @@ from invenio_records.models import RecordMetadata
 
 from inspirehep.records.api import InspireRecord, LiteratureRecord
 from inspirehep.records.errors import MissingSerializerError, WrongRecordSubclass
+from inspirehep.records.marshmallow.literature.bibtex import BibTexCommonSchema
 
 
 def test_base_get_record(base_app, db, es, create_record_factory):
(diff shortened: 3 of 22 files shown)


backend/inspirehep/records/marshmallow/literature/bibtex.py
```diff
@@ -71,11 +71,16 @@ def get_date(data, doc_type):
         )
         thesis_date = get_value(data, "thesis_info.date")
         imprint_date = get_value(data, "imprints.date[0]")
+        earliest_date = get_value(data, "earliest_date")
```
michamos: there is no explicit `earliest_date` in the record. It's computed from all dates in the record, so it's an attribute on `data` (which is an instance of `LiteratureRecord`): https://github.com/inspirehep/inspirehep/blob/3efec10df15ee432f61aa13d57269568f29a9758/backend/inspirehep/records/api/literature.py#L77

backend/inspirehep/records/marshmallow/literature/bibtex.py
```diff
@@ -71,11 +71,16 @@ def get_date(data, doc_type):
...
 
         if doc_type.endswith("thesis"):
-            date_choice = thesis_date or publication_year or imprint_date
+            date_choice = (
+                thesis_date or publication_year or imprint_date or earliest_date
+            )
         else:
-            date_choice = publication_year or thesis_date or imprint_date
+            date_choice = (
+                publication_year or thesis_date or imprint_date or earliest_date
```
michamos:
```suggestion
                publication_year or imprint_date or earliest_date
```

backend/inspirehep/records/marshmallow/literature/bibtex.py
```diff
@@ -66,17 +66,24 @@ class BibTexCommonSchema(Schema):
...
+            "masterthesis": thesis_date,
+            "book": imprint_date,
+            "book chapter": imprint_date,
+        }
+        if doc_type in date_map:
+            date_choice = date_map[doc_type] or earliest_date
         else:
-            date_choice = publication_year or thesis_date or imprint_date
-
+            date_choice = publication_year or earliest_date
```
michamos: this is the same as the more concise
```python
date_choice = date_map.get(doc_type, publication_year) or earliest_date
```

backend/inspirehep/records/marshmallow/literature/bibtex.py
```diff
@@ -66,17 +66,24 @@ class BibTexCommonSchema(Schema):
...
         imprint_date = get_value(data, "imprints.date[0]")
-
-        if doc_type.endswith("thesis"):
-            date_choice = thesis_date or publication_year or imprint_date
+        earliest_date = LiteratureRecord(data).earliest_date
+        date_map = {
+            "phdthesis": thesis_date,
+            "masterthesis": thesis_date,
+            "book": imprint_date,
+            "book chapter": imprint_date,
```
michamos: are these really possible `doc_type`s?

backend/inspirehep/records/marshmallow/literature/bibtex.py
```diff
@@ -66,17 +66,24 @@ class BibTexCommonSchema(Schema):
 
     @staticmethod
     def get_date(data, doc_type):
+        from inspirehep.records.api import LiteratureRecord
```
michamos: you should put imports at the module scope in principle

MJedr: in that case there is an error, caused by (I guess) circular import, so when I move it on the top it does not work

michamos: I see. Actually you don't need this, as `data` should already be a `LiteratureRecord` in real cases. You should use the `create_record` fitxure in the tests to build a record.

backend/inspirehep/records/marshmallow/literature/bibtex.py
MJedr: Yes, since we take them from data, the same case as with
`document_type_map = {
"article": "article",  
"book": "book",  
"book chapter": "inbook"}`

michamos: but `phdthesis` and `masterthesis` are not INSPIRE document types. So either the first two are wrong, or the last one is.

backend/inspirehep/records/api/literature.py
```diff
@@ -144,6 +144,9 @@ def update_refs_to_conferences(data):
                 continue
             conference["conference_record"] = get_ref_from_pid("con", pid)
 
+    def copy(self):
+        return type(self)(self)
```
michamos: this should be on `InspireRecord` instead, it's generally useful.

backend/inspirehep/records/marshmallow/literature/bibtex.py
```diff
@@ -141,14 +143,14 @@ def get_doi(self, data):
...
-        date = BibTexCommonSchema.get_date(data, doc_type)
+        doc_type = get_value(data, "document_type[0]")
+        date = self.get_date(data.copy(), doc_type)
         if date:
             return date.month
 
     def get_year(self, data):
-        doc_type = data.get("doc_type")
-        date = BibTexCommonSchema.get_date(data, doc_type)
+        doc_type = get_value(data, "document_type[0]")
```
michamos: this is wrong, the first document type is not necessarily the most significant one. It's better to use `doc_type`, as that's what will be used by BibTeX.

backend/inspirehep/records/marshmallow/literature/bibtex.py
```diff
@@ -141,14 +143,14 @@ def get_doi(self, data):
...
+        date = self.get_date(data.copy(), doc_type)
```
michamos: why the `copy`?

backend/inspirehep/records/api/literature.py
```diff
@@ -144,6 +144,9 @@ def update_refs_to_conferences(data):
                 continue
             conference["conference_record"] = get_ref_from_pid("con", pid)
 
+    def copy(self):
```
michamos: you don't need this, using the parent method is the default if not defined on the subclass.

backend/inspirehep/records/api/base.py
```diff
@@ -230,6 +230,9 @@ def get_linked_records_from_dict_field(cls, data, path):
             return cls.get_records_by_pids(pids)
         return iter([])
 
+    def copy(self):
```
michamos: I would add a docstring explaining why this is needed:
```suggestion
    def copy(self):
    """Copy the record metadata.
    
    This is needed because the default ``dict.copy`` always returns a ``dict``.
    """
```

backend/inspirehep/records/marshmallow/literature/bibtex.py
michamos: you didn't address this. Please use the bibtex `doc_type` instead (this affects also the map in `get_date` which has to use those).

backend/tests/unit/records/marshmallow/literature/test_bibtex_schema.py
```diff
@@ -119,20 +119,6 @@ def test_bibtex_document_type_prefers_article():
     assert expected == result
 
 
-def test_get_year_from_thesis_when_pubinfo_present():
```
michamos: why did you remove this?

MJedr: I put it in unit tests

backend/inspirehep/records/marshmallow/literature/bibtex.py
```diff
@@ -66,17 +66,19 @@ class BibTexCommonSchema(Schema):
...
-        if doc_type.endswith("thesis"):
-            date_choice = thesis_date or publication_year or imprint_date
-        else:
-            date_choice = publication_year or thesis_date or imprint_date
-
+        earliest_date = data.earliest_date
+        date_map = {
+            "thesis": thesis_date,
+            "book": imprint_date,
+            "book chapter": imprint_date,
```
michamos: this is wrong now, as `doc_type` is a bibtex `doc_type`. You should have tests catching this.

backend/tests/integration/records/api/test_api_base.py
```diff
@@ -360,3 +361,30 @@ def test_update_record_with_different_control_number(app, db, create_record):
...
+    expected_year = 2000
+    schema = BibTexCommonSchema()
+
+    result = schema.dump(record).data
+    result_year = result["year"]
+
+    assert expected_year == result_year
+
+
+def test_get_year_from_thesis_when_pubinfo_present(app, db, create_record):
```
michamos: this test should have a different date somewhere that is earlier, so you can check that the date is taken from the right field. That part of the logic is currently not tested (and likely doesn't work correctly). You should also add a similar test for books and book chapters.

PR #804, opened 2020-03-17
marshmallow: Include all links which are not hidden to fulltext_links

pazembrz:
*INSPIR-3268

### DIFF
--- a/backend/inspirehep/records/marshmallow/literature/ui.py
+++ b/backend/inspirehep/records/marshmallow/literature/ui.py
@@ -133,7 +133,7 @@ class LiteratureDetailSchema(CatalogerCanEditMixin, LiteraturePublicSchema):
             return missing
         description = data.get("description") or "fulltext"
         url = data.get("url")
-        if url and not data.get("hidden", False) and data.get("fulltext"):
+        if url and not data.get("hidden", False):
             return {"description": description, "value": url}
         return missing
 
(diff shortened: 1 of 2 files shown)


backend/inspirehep/records/marshmallow/literature/ui.py
```diff
@@ -129,11 +129,12 @@ def get_arxiv_fulltext_link(self, data):
         return missing
 
     def get_internal_fulltext_link(self, data):
+        # Assume all non hidden are fulltext - INSPIR-3268
```
monaawi: I don't think this is needed as we can trace back the change to the commit and you have the JIRA task in the commit message

backend/tests/unit/records/marshmallow/literature/test_ui.py
```diff
@@ -76,8 +76,16 @@ def test_internal_fulltext_hidden(current_app_mock):
 
 
 @mock.patch("inspirehep.records.marshmallow.literature.ui.current_app")
-def test_internal_not_fulltext(current_app_mock):
+def test_not_marked_as_fulltext_in_fulltext(current_app_mock):
```
monaawi: maybe ```test_non_fulltext_documents_appear_in_fulltext_links_field```?

backend/inspirehep/records/marshmallow/literature/ui.py
pazembrz: :+1:

backend/tests/unit/records/marshmallow/literature/test_ui.py
pazembrz: :+1:

PR #805, opened 2020-03-17
setup: bump inspire-dosjon

monaawi:
* INSPIR-3291
* Needs remigrating

### DIFF
--- a/backend/tests/integration/migrator/test_migrator_tasks/1313624.xml
+++ b/backend/tests/integration/migrator/test_migrator_tasks/1313624.xml
@@ -1559,7 +1559,6 @@
     <subfield code="n">arXiv:1409.0794</subfield>
     <subfield code="r"></subfield>
     <subfield code="s">2015-01-12 03:41:58</subfield>
-    <subfield code="t">arXiv</subfield>
     <subfield code="v">2</subfield>
     <subfield code="z"></subfield>
   </datafield>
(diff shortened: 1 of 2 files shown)


backend/tests/integration/migrator/test_migrator_tasks/1313624.xml
```diff
@@ -1559,10 +1559,9 @@
     <subfield code="n">arXiv:1409.0794</subfield>
     <subfield code="r"></subfield>
     <subfield code="s">2015-01-12 03:41:58</subfield>
-    <subfield code="t">arXiv</subfield>
```
michamos: what is this?

monaawi: This record xml is used in a bunch of tests and some tests are with documents and they fail because this has source arxiv and will be hidden and not processed through our file adding. So I thought it would be easier to just remove the source arxiv instead of updating all the tests or adding a new record to test with

pazembrz:
I know that this is done in dojson but if you are changing tests, add one simple test which will test behavior with arxiv source

monaawi:
I don't think we should test inspire-dojson things.Basically I would be testing that inspire-dojson is giving me hidden true for source arxiv (which is tested in inspire-dojson). We already have tests to check that hidden documents are not displayed and not added to s3.

PR #806, opened 2020-03-17
orcid: add missing test

MJedr:
* INSPIR - 3051


michamos:
Travis is not reporting the status correctly, but it failed: https://travis-ci.org/github/inspirehep/inspirehep/builds/666857371

backend/inspirehep/orcid/domain_models.py
```diff
@@ -128,28 +128,21 @@ def push(self):
...
-                if not putcode:
-                    msg = (
-                        "No putcode was found in ORCID API for orcid={} and recid={}."
-                        " And the POST has previously failed for the same recid because"
-                        " the work had already existed".format(self.orcid, self.recid)
-                    )
-                    raise exceptions.PutcodeNotFoundInOrcidException(msg)
-                putcode = self._post_or_put_work(putcode)
+            if self.pushing_duplicated_identifier:
+                raise exceptions.DuplicatedExternalIdentifierPusherException
```
michamos:
```suggestion
                try:
                    putcode = self._cache_putcodes_and_put_work()
                except orcid_client_exceptions.WorkAlreadyExistsException:
                     raise exceptions.DuplicatedExternalIdentifierPusherException
```
as you might be missing the putcode when pushing the clashing identifier (you should add a test for this!)

backend/inspirehep/orcid/domain_models.py
```diff
@@ -256,7 +252,7 @@ def _push_work_with_clashing_identifier(self):
 
             if not putcode or not recid:
                 continue
-            if recid == self.recid:
+            if int(recid) == self.recid:
```
michamos: how do you know `self.recid` is an `int`? it's whatever type is passed to the `__init__` method, you should also make it an `int` when setting `self.recid` there.

MJedr: The point is that some tests are passing text as a wrong recid, so it cause an error.

michamos: then you need to change those tests

backend/inspirehep/orcid/domain_models.py
```diff
@@ -125,42 +125,56 @@ def push(self):
...
             # We POSTed the record as new work, but it failed because the work
-            # already exists (identified by the external identifiers).
+            # with the same identifier is already in ORCID.
             # This means we do not have the putcode, thus we cache all
-            # author's putcodes and PUT the work again.
-            try:
-                if self.do_fail_if_duplicated_identifier:
+            # author's putcodes. If the putcode wasn't found we try to push
+            # once again works with clashing identifier to update them and
+            # resolve conflict.
```
michamos: The explanation is not very clear. What about something like

We POSTed the record as new work, but it failed because a work with the same identifier is already in ORCID. This can mean two things:
1. the record itself is already in ORCID, but we don't have the putcode;
2. a different record with the same external identifier is already in ORCID.
We first try to fix 1. by caching all author's putcodes and PUT the work again. If the putcode wasn't found we are probably facing case 2. so we try to push once again works with clashing identifiers to update them and resolve the potential conflict.

backend/inspirehep/orcid/domain_models.py
```diff
@@ -125,42 +125,56 @@ def push(self):
...
+            if self.pushing_duplicated_identifier:
+                # In case the WorkAlreadyExistsException
+                # exception was obtained when pushing records
+                # with clashing identifiers (what should never happen)
+                # DuplicatedExternalIdentifierPusherException is raised.
```
michamos: I don't think this comment is helpful, you simply repeat what is going on (which can be understood by reading the code)

backend/inspirehep/orcid/domain_models.py
```diff
@@ -125,42 +125,56 @@ def push(self):
...
+            putcode = self._cache_all_author_putcodes()
+            if not putcode:
+                try:
+                    self._push_work_with_clashing_identifier()
+                    putcode = self._post_or_put_work(putcode)
+                except orcid_client_exceptions.WorkAlreadyExistsException:
+                    # If the putcode wasn't found, but we obtained
+                    # WorkAlreadyExistsException (e.g. records have the same doi)
+                    # it means that the external identifiers are duplicated
+                    # and we can't do anything about it.
```
michamos:
```suggestion
                    # The PUT/POST failed despite pushing works with clashing identifiers
                    # and we can't do anything about this.
```

backend/inspirehep/orcid/domain_models.py
```diff
@@ -125,42 +125,56 @@ def push(self):
...
             self.converter = OrcidConverter(
                 record=self.inspire_record,
                 url_pattern=current_app.config["LEGACY_RECORD_URL_PATTERN"],
                 put_code=None,
             )
-            putcode = self._post_or_put_work()
+            try:
+                putcode = self._post_or_put_work()
+            except orcid_client_exceptions.WorkAlreadyExistsException:
+                self.push()
```
michamos: Here you might end in a loop again. What about doing simply
```suggestion
            putcode = self._cache_all_author_putcodes()
            putcode = self._post_or_put_work(putcode)
```
(and update the comment in this branch)?

PR #807, opened 2020-03-17
ui: add commas to numbers

monaawi:
* INSPIR-3267


ui/src/common/utils.js
```diff
@@ -251,3 +251,7 @@ export function getAuthorName(author) {
   }
   return author.get('full_name');
 }
+
+export function addCommasToNumber(number) {
+  return number && number.toLocaleString();
```
harunurhan: this doesn't do `add commas to number` necessarily, the behavior of this function depends on the user's locale so in `China` or `UAE` this will return differently.

if what we want is to format the number according to the user locale, then that's fine we can just change the function name, but I don't think that's case. Because we don't intl. anywhere in our app.

So it would be better to pass the locale explicitly, most likely `en-US`

ui/src/common/components/CitationSummaryTable/CitationSummaryTable.jsx
```diff
@@ -94,14 +103,20 @@ class CitationSummaryTable extends Component {
...
+                      ).toFixed(1)}
+                    </FormattedNumber>
                   </td>
                   <td>
-                    {(
-                      publishedBucket.getIn(['average_citations', 'value']) || 0
-                    ).toFixed(1)}
+                    <FormattedNumber>
+                      {(
+                        publishedBucket.getIn(['average_citations', 'value']) ||
```
harunurhan: it's not introduced in this PR, but just realized `||`, `getIn` already has `unSetValue`, these could be written as `.getIn([...], 0)`

incase you'd like to improve.

ui/src/common/components/FormattedNumber.jsx
```diff
@@ -0,0 +1,17 @@
...
+  return addCommasToNumber(children);
+}
+
+FormattedNumber.propTypes = {
+  children: PropTypes.oneOfType([PropTypes.string, PropTypes.number])
+    .isRequired,
+};
+
+FormattedNumber.defaultProps = {
+  children: null,
```
harunurhan: no need for that, especially when the prop is `required`

PR #808, opened 2020-03-17
literature: Do not re-download files which are already on S3

pazembrz:
*INSPIR-3290


backend/inspirehep/files/config.py
```diff
@@ -12,3 +12,4 @@
 S3_BUCKET_PREFIX = "inspire-files-"
 S3_FILE_ACL = "public-read"
 FILES_DOWNLOAD_TIMEOUT = 60
+FORCE_PROCESS_S3_FILES = False
```
monaawi: maybe ```UPDATE_S3_FILES_METADATA```?

monaawi: Maybe ```UPDATE_S3_FILES_METADATA```?

backend/tests/integration/records/api/test_api_literature.py
```diff
@@ -1608,3 +1608,118 @@ def test_literature_updates_refs_to_known_and_unknown_conference_when_ref_alread
...
+                "key": "some_figure_key_on_s3",
+                "original_url": "http://inspire-afs-web.cern.ch/var/data/files/g188/3771220/content.png%3B2",
+                "source": "arxiv",
+                "url": "https://s3.cern.ch/inspire-qa-files-s/some_figure_key_on_s3",
+            }
+        ],
+    }
+    with mock.patch.object(
+        s3, "replace_file_metadata"
+    ) as mocked_s3_reploace_metadata_method:
```
monaawi: ```mocked_s3_replace_metadata```

backend/tests/integration/records/api/test_api_literature.py
```diff
@@ -1608,3 +1608,118 @@ def test_literature_updates_refs_to_known_and_unknown_conference_when_ref_alread
...
+        record = create_record("lit", data=data)
+        mocked_s3_reploace_metadata_method.assert_not_called()
+    expected_documents = [
```
monaawi: aren't these the same as the initial documents and figures? You could avoid duplicating this

backend/tests/integration/records/api/test_api_literature.py
```diff
@@ -1608,3 +1608,118 @@ def test_literature_updates_refs_to_known_and_unknown_conference_when_ref_alread
...
+            "original_url": "http://inspire-afs-web.cern.ch/var/data/files/g188/3771220/content.png%3B2",
+            "source": "arxiv",
+            "url": "https://s3.cern.ch/inspire-qa-files-s/some_figure_key_on_s3",
+        }
+    ]
+    assert record["figures"] == expected_figures
+    assert record["documents"] == expected_documents
+
+
+def test_if_processing_files_which_are_already_on_s3_when_forced(
```
monaawi: ```test_files_metadata_is_replaced_when_replacing_metadata_is_enabled```

backend/inspirehep/records/api/literature.py
```diff
@@ -317,6 +318,31 @@ def add_figures(self, figures):
...
+            }
+            LOGGER.info(
+                "File already on S3 - Skipping",
+                url=url,
+                key=key,
+                recid=self.get("control_number"),
+                uuid=self.id,
+            )
+            return strip_empty_values(result)
+        LOGGER.info(
```
monaawi: i don't think this log is necessary as we have the other 2 with replacing file metadata and adding file to s3 already

backend/inspirehep/records/api/literature.py
```diff
@@ -317,6 +318,31 @@ def add_figures(self, figures):
     def add_file(
         self, url, original_url=None, key=None, filename=None, *args, **kwargs
     ):
+        if current_s3_instance.is_s3_url(url) and not current_app.config.get(
+            "FORCE_PROCESS_S3_FILES", False
+        ):
+            result = {
```
monaawi: During the migration, we use the afs cache from redis to replace the afs url with the corresponding s3 url. Therefore, here you can get a file with an s3 url but without the correct key (the hash of the content) and we also still need the logic of having the key moved to the filename.

backend/inspirehep/files/config.py
pazembrz: :+1:

backend/tests/integration/records/api/test_api_literature.py
pazembrz: :+1:

backend/tests/integration/records/api/test_api_literature.py
pazembrz: Yes, you are right.

backend/inspirehep/records/api/literature.py
pazembrz: :+1:

backend/inspirehep/records/api/literature.py
pazembrz: :+1:

backend/inspirehep/records/api/literature.py
```diff
@@ -317,6 +318,27 @@ def add_figures(self, figures):
...
+        ):
+            if key not in url:
+                filename = filename or key
+                key = url.split("/")[-1]
+            result = {
+                "key": key,
+                "filename": filename,
+                "url": url,
+                "original_url": original_url,
+                **kwargs,
```
monaawi: I know this is not related to your PR but I just realized we are also returning mimetype and size and these aren't on the schema nor added by the builder so they are returned for nothing, could you remove those too please?

backend/inspirehep/records/api/literature.py
monaawi: I just realized this should have been caught by a test. We have a test checking the usage of the file cache but it isn't checking all fields in the file. Now that we have different behavior for when the url is an s3 url, it would be good to assert also the other fields in the afs cache test. Check out ```test_migrate_record_from_mirror_uses_local_cache_for_afs_files```

backend/inspirehep/records/api/literature.py
pazembrz: :+1:

backend/tests/integration/records/api/test_api_literature.py
```diff
@@ -1608,3 +1608,166 @@ def test_literature_updates_refs_to_known_and_unknown_conference_when_ref_alread
...
+        files_count = 2
+        with mock.patch.dict(base_app.config, {"UPDATE_S3_FILES_METADATA": True}):
+            with mock.patch.object(
+                s3, "replace_file_metadata"
+            ) as mocked_s3_replace_metadata:
+                record.update(record_data)
+                assert mocked_s3_replace_metadata.call_count == files_count
+
+
+def test_files_data_updates_when_url_from_cache_is_served_without_proper_key(
```
monaawi: maybe ```test_adding_files_with_s3_url_but_wrong_key```

pazembrz: :+1:

PR #800, opened 2020-03-16
ui: new header, footer, homepage

harunurhan:

PR #809, opened 2020-03-17
ui: make citesummary graph numbers more visible

miguelgrc:
* INSPIR-3266


ui/src/App.scss
```diff
@@ -24,7 +24,7 @@
 
     // :fist-child to avoid setting padding for `UserFeedback`
     & > *:first-child:not(.__Jobs__) {
-      @media (min-width: $screen-xs-max) {
+      @media (min-width: $screen-lg-max) {
```
harunurhan: ?

ui/src/common/components/CitationSummaryGraph/CitationSummaryGraph.jsx
```diff
@@ -22,8 +22,8 @@ import styleVariables from '../../../styleVariables';
 import { shallowEqual, abbreviateNumber } from '../../utils';
 import { browser } from '../../browser';
 
-const BAR_WIDTH = 0.5;
-const GRAPH_MARGIN = { left: 40, right: 20, top: 10, bottom: 40 };
+const BAR_WIDTH = 0.75;
+const GRAPH_MARGIN = { left: 42, right: 10, top: 30, bottom: 40 };
```
harunurhan: I understand the `top` maybe, but why others?

ui/src/common/components/CitationSummaryGraph/CitationSummaryGraph.scss
```diff
@@ -5,33 +5,47 @@
...
     text {
       fill: $secondary-color;
+      font-size: 14px;
     }
   }
 
   .rv-xy-plot__series--label-text {
     color: $text-color;
-    font-size: 0.68rem;
+    font-size: 13px;
```
harunurhan: `font-size`s should be set as `rem` or `em` when it makes sense

ui/src/common/components/CitationSummaryGraph/CitationSummaryGraph.jsx
miguelgrc: On top I increased it since I moved the y-axis label to the top of the graph (check screenshots). Left is because the numbers in the axis were cut in some screen sizes. On the right we don't need that much margin since I moved the label as well to make it more readable.

ui/src/App.scss
miguelgrc: I discussed with Stella about removing the margin in the sides for tablets so that the graph numbers don't overlap.

harunurhan: we shouldn't remove those paddings in whole pages just because of a graph. I'd say leave this. And I am already playing with those in #800, we can decide later if that's really needed or not.

@StellaCh

ui/src/common/components/CitationSummaryGraph/CitationSummaryGraph.jsx
harunurhan: ah ok, saw that.

ui/src/common/utils.js
```diff
@@ -221,7 +221,8 @@ export function addOrdinalSuffix(i) {
 }
 
 const numberAbbreviator = new NumberAbbreviator(['K', 'M', 'B', 'T']);
-export function abbreviateNumber(number) {
+export function abbreviateNumber(number, biggerThan = 999) {
+  if (number <= biggerThan) return number;
```
harunurhan: please move that logic where it's needed, we can change the common function when it's needed for at least 2 usecases.

harunurhan: also make sure this is change is tested.

ui/src/common/components/CitationSummaryGraph/CitationSummaryGraph.jsx
```diff
@@ -138,6 +138,12 @@ class CitationSummaryGraph extends Component {
     return GRAY;
   }
 
+  static getCountLabel(docCount) {
+    if (docCount === 0) return null;
+    if (docCount < 10000) return docCount.toString();
```
monaawi: I know it isn't part of your task necessarily, but can you add ```<FormattedNumber>``` here? for cases under 10000. Discussed it with Stella.

miguelgrc: We're very limited in space in the graph so this makes the numbers overlap again in tablets and smaller screens, are we sure we need this here @StellaCh ? We only have 4 digit numbers so they're not hard to read.

StellaCh: Well, we had this discussion before and we said that we should give it a try and see if it affects the display that much. Can you try and share the screenshots?

miguelgrc: So after discussing with Stella we're not going to add the separator here to avoid overlapping. I saw that it's missing in the table for h-index though, I added them here to avoid creating a PR just for that.

miguelgrc:
I didn't notice that, I removed it.

harunurhan:
can you rebase @miguelgrc ?

PR #813, opened 2020-03-19
ui: underline active collection

harunurhan:
It renders hover styles for collection menu item depending on the
pathname.


ui/src/common/utils.js
```diff
@@ -255,3 +255,7 @@ export function getAuthorName(author) {
 export function addCommasToNumber(number) {
   return number && Number(number).toLocaleString('en-US');
 }
+
+export function getRootOfLocationPathname(pathname) {
+  return pathname.split('/', 2)[1];
```
drjova: beautiful to fail :P

harunurhan: why would it be?

harunurhan: just removing the leftover `, 2`

PR #819, opened 2020-03-20
literature: use threads to process files on record update

pazembrz:
*INSPIR-3302


drjova:
Please **DO NOT MERGE BEFORE THE FINAL RELEASE**

backend/inspirehep/records/api/literature.py
```diff
@@ -267,106 +269,149 @@ def add_files(self, data):
...
+                        LOGGER.exception(
+                            "Duplicated file found",
+                            recid=self.get("control_number"),
+                            uuid=self.id,
+                            document=file_data.get("key"),
+                        )
             try:
-                builder.add_figure(**figure)
-            except ValueError:
+                for future in as_completed(
```
drjova: do we want to handle them one by one? or just use ``wait`` instead?

backend/inspirehep/records/api/literature.py
```diff
@@ -267,106 +269,149 @@ def add_files(self, data):
...
-                    document=document.get("key"),
-                )
-        return builder.record.get("documents")
-
-    def add_figures(self, figures):
-        builder = LiteratureBuilder()
-        for figure in figures:
-            new_file_data = self.add_file(**figure)
-            figure.update(new_file_data)
+    def add_files_data(self, files, builder_add_file):
```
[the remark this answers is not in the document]
drjova: Also, we're loosing the approach to handle documents and figures separetly, the logic now is mixed. ``add_documents`` & ``add_figures`` they can just return the tasks for the executor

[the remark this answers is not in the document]
pazembrz: Problem is that logic for the whole time was identical, only difference was to check 'hidden' key on documents, which actually is never provided with figures. 
If there will be any requirement in the future to process them i different way it's easy to separate them, but for now why you want to keep 2 almost identical functions?

For duplicated parts in the function, I will merge them :+1:

backend/inspirehep/records/api/literature.py
pazembrz: Even if we will use wait we would like to handle them one by one at the end to update values in the record. With this approach we are handling everything together easily.

backend/inspirehep/records/api/literature.py
```diff
@@ -267,106 +269,161 @@ def add_files(self, data):
...
-                document.update(new_file_data)
-            try:
-                builder.add_document(**document)
-            except ValueError:
-                LOGGER.exception(
-                    "Duplicated document found",
-                    recid=self.get("control_number"),
-                    uuid=self.id,
-                    document=document.get("key"),
+    def _generate_tasks(self, files, executor, add_file_func):
```
drjova:
```suggestion
    def _generate_future_tasks(self, files, executor, add_file_func):
```

backend/inspirehep/records/api/literature.py
```diff
@@ -267,106 +269,161 @@ def add_files(self, data):
...
-        if (
-            url.startswith("http")
-            and not current_s3_instance.is_s3_url(url)
-            and not original_url
-        ):
-            result["original_url"] = url
-        return result
+            raise
+
+    def add_files_data(self, files, builder_add_file):
```
drjova: This can be moved to ``add_files`` and have only one threadpool for both documents and figures

backend/inspirehep/records/api/literature.py
```diff
@@ -267,106 +269,161 @@ def add_files(self, data):
...
-        new_key = hash_data(file_data)
-        mimetype = magic.from_buffer(file_data, mime=True)
-        file_data = BytesIO(file_data)
-        filename = filename or key
-        acl = current_app.config["S3_FILE_ACL"]
-        if current_s3_instance.file_exists(new_key):
-            LOGGER.info(
-                "Replacing file metadata",
-                key=new_key,
+    def _update_file_entry(self, file_data, add_file_func):
```
drjova: Why we need a serate function to just update the file? Why we need a serate function to just update the file? It was much nicer to keep figures  and documents logic seprated

backend/inspirehep/records/api/literature.py
pazembrz: Sorry I'm not convinced to this name. It sounds weird..

backend/inspirehep/records/api/literature.py
pazembrz: We use the same thing in 2 separate places. With the same try/except. It's just logical to keep it as separate function than have identical code in 2 places

backend/inspirehep/records/api/literature.py
pazembrz: One thread pool for both figures and documents is not good idea IMHO as results from them are going to separate keys in the record. I could put them in one but then instead of having two simple loops for putting results in record there will be one more complex loop with if/else and results themselves would be put in more complex structure to hold information where results should be put.

backend/inspirehep/records/api/literature.py
drjova: what is the ``_generate_tasks``? celery tasks? ``_geenerate_tasks`` is too generic IMHO

pazembrz: Yeah, but generate_future_task is also ambiguous. Maybe _generate_files_tasks or smth?

PR #824, opened 2020-03-23
mappings: add dot to whitespace filter

drjova:


backend/inspirehep/search/mappings/v7/records/hep.json
```diff
@@ -1282,6 +1283,7 @@
             "initial_normalizer",
             "lowercase_normalizer"
           ],
+          "char_filter": ["dot_to_whitespace_normalizer"],
```
michamos: does this run before or after the tokenizer? is it the right order to do what you want?

drjova: 👍 I'll read more about it, on my tests it worked as expected

drjova: I run before the tokenizer
> Character filters are used to preprocess the stream of characters before it is passed to the tokenizer.

https://www.elastic.co/guide/en/elasticsearch/reference/current/analysis-charfilters.html

PR #825, opened 2020-03-23
ui: change help blog url

harunurhan:
This will break the link until the release (labs/beta becomes master)

INSPIR-3308


michamos:
This is not urgent at all, as we will redirect `labs.inspirehep.net` -> `inspirehep.net`.

harunurhan:
ah right, then this PR can wait.

monaawi:
why does it need to wait?

harunurhan:
it would break `help` link on `labs ` for no reason.

drjova:
@harunurhan could you please rebase?

PR #826, opened 2020-03-23
ui: Fix error when numbers in the CitationSummaryTable are null

monaawi:
https://sentry.inspirehep.net/inspire-prod/ui/issues/17676/


ui/src/common/utils.js
```diff
@@ -259,3 +259,7 @@ export function addCommasToNumber(number) {
 export function getRootOfLocationPathname(pathname) {
   return pathname.split('/')[1];
 }
+
+export function ToFixedNumber(numberOrNull) {
```
harunurhan: lowercase `t` in the function name.

harunurhan: also, it's OK but you don't have to put this necessarily in `utils.js`, you can just put in the only file where it is used.

PR #827, opened 2020-03-23
literature: hide hidden collections from non curators

pazembrz:
*INSPIR-3324

### DIFF
--- a/backend/inspirehep/access_control.py
+++ b/backend/inspirehep/access_control.py
@@ -11,45 +11,31 @@ from inspirehep.accounts.decorators import login_required_with_roles
 from inspirehep.accounts.roles import Roles
 
 
-def check_oauth2(can_method):
-    """Base permission factory that check OAuth2 scope.
-
-    Args:
-     can_method: Permission check function that accept a record in input
-        and return a boolean.
-
-    Returns:
-         class:`flask_principal.Permission` factory.
-    """
-
-    def check(record, *args, **kwargs):
-        @require_api_auth()
-        def can(self):
-            return can_method(record)
-
-        return type("CheckOAuth2Scope", (), {"can": can})()
-
-    return check
-
-
-api_access_permission_check = check_oauth2(lambda self: True)
-
-
-class SessionCatalogerPermission:
-    @login_required_with_roles([Roles.cataloger.value])
-    def can(self):
-        return True
-
-
-def session_cataloger_permission_factory(*args, **kwargs):
-    return SessionCatalogerPermission()
-
-
-class LiteraturePermissionCheck:
+class InspireBasePermissionCheck:
     def __init__(self, record, *args, **kwargs):
         self.record = record
 
     def can(self):
+        raise NotImplementedError
+
+
+class APIAccessPermissionCheck(InspireBasePermissionCheck):
+    @require_api_auth()
+    def can(self):
+        return True
+
+
+class SessionCatalogerPermission(InspireBasePermissionCheck):
+    @login_required_with_roles([Roles.cataloger.value])
+    def cataloger_check(self):
+        return True
+
+    def can(self):
+        return self.cataloger_check()
+
+
+class LiteraturePermissionCheck(SessionCatalogerPermission):
+    def can(self):
         if "Literature" not in self.record.get("_collections", []):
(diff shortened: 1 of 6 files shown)


backend/inspirehep/access_control.py
```diff
@@ -43,3 +43,13 @@ def can(self):
...
+
+
+class LiteraturePermissionCheck:
+    def __init__(self, record, *args, **kwargs):
+        self.record = record
+
+    def can(self):
+        if "Literature" not in self.record.get("_collections", []):
+            return SessionCatalogerPermission().can()
+        return True
```
drjova: why we did it like a class? I'd follow the invenio docs just to be aligned with them https://github.com/inveniosoftware/invenio-records-rest/blob/69c8d45361e898a2689f7c1442395261f701dfbb/invenio_records_rest/config.py#L71

pazembrz: It looks like class, it's called like class: https://labs.inspirehep.net/literature/1767665 But in Invenio they decided that it's better to create small ugly thing... Both are callables, and it will work both with function and class but class here is much more clear.

Like for real... Is this more readable in your opinion:
```python
def permission_check_factory():
        def check_permission(record, *args, **kwargs):
            def can(self):
                if "Literature" not in self.record.get("_collections", []):
                    return SessionCatalogerPermission().can()
            return type('Check', (), {'can': can})()
```
than this
```python
class LiteraturePermissionCheck:
    def __init__(self, record, *args, **kwargs):
        self.record = record

    def can(self):
        if "Literature" not in self.record.get("_collections", []):
            return SessionCatalogerPermission().can()
        return True
```

michamos: so instead of having a static class that is configurable enough for their needs, they have a function that
1. builds a function to do the permission check
2. builds a dynamic class with that function as a method and instantiate it
:man_facepalming:

pazembrz: Yup... probably for "better code quality"...

backend/inspirehep/access_control.py
```diff
@@ -43,3 +43,13 @@ def can(self):
 
 def session_cataloger_permission_factory(*args, **kwargs):
```
michamos: I would also change this while you're at it, by having a dummy `SessionCatalogerPermission.__init__`.

pazembrz: @drjova Can I? :pray: :smile:

drjova: 🚀

backend/inspirehep/access_control.py
harunurhan: "hey, dynamic classes are cool.. so let's use them!" 😄 

btw I did `SessionCatalogerPermission` with a proper class, exactly for this reason because the other way is retarded (at least for our usecase)

backend/inspirehep/access_control.py
harunurhan: I agree `__init__` is better than a separate function.

backend/inspirehep/access_control.py
```diff
@@ -11,35 +11,31 @@
...
+    def can(self):
+        return self.cataloger_check()
 
 
-class SessionCatalogerPermission:
-    @login_required_with_roles([Roles.cataloger.value])
+class LiteraturePermissionCheck(SessionCatalogerPermission):
     def can(self):
+        if "Literature" not in self.record.get("_collections", []):
+            return self.cataloger_check()
```
michamos: I would do composition over inheritance here, but this works for now.

backend/inspirehep/accounts/decorators.py
```diff
@@ -34,6 +36,8 @@ def wrapped_function(*args, **kwargs):
                 abort(401)
 
             if roles:
+                # superuser can do everything
+                roles.append(Roles.superuser.value)
```
michamos: What about
```
if not roles:
    roles = []
# Rest of the function body without `ìf`
```
so you can decorate with `@login_required_with_roles()` to allow only `superuser`?

pazembrz: No, you have to specify at least 1 role. I didn't change behavior when you are not specifying any role as this would break something which we are maybe not testing ;) Better safe than sorry.

PR #828, opened 2020-03-23
WIP: remove citation summary hash from url in other page

miguelgrc:
* INSPIR-3262


ui/src/common/containers/SearchBoxContainer.jsx
```diff
@@ -15,6 +16,8 @@ const stateToProps = state => ({
 
 export const dispatchToProps = dispatch => ({
   onSearch(namespace, value) {
+    // This avoids carrying the hash to other searches when changing collection
+    if (namespace !== 'literature') dispatch(setHash(''));
```
harunurhan: we don't to one line `if`s and more important use constant for search namespace
```
if (namespace !== LITERATURE_NS) {
  dispatch(setHash(''));
}
```

ui/src/common/containers/SearchBoxContainer.jsx
```diff
@@ -15,6 +16,8 @@ const stateToProps = state => ({
 
 export const dispatchToProps = dispatch => ({
   onSearch(namespace, value) {
+    // This avoids carrying the hash to other searches when changing collection
```
harunurhan: could you prefix it like `// HACK: ...`

ui/src/common/containers/__tests__/SearchBoxContainer.test.jsx
```diff
@@ -55,4 +55,67 @@ describe('SearchBoxContainer', () => {
...
+    });
+    const wrapper = mount(
+      <Provider store={store}>
+        <SearchBoxContainer />
+      </Provider>
+    );
+    const onSearch = wrapper.find(SearchBox).prop('onSearch');
+    onSearch(newNamespace, 'test');
+
+    const expectedAction = {
```
harunurhan: do `const expectedAction = setHash('')`

ui/src/common/containers/__tests__/SearchBoxContainer.test.jsx
```diff
@@ -55,4 +55,67 @@ describe('SearchBoxContainer', () => {
...
+      payload: {
+        args: [{ hash: '' }],
+        method: 'replace',
+      },
+    };
+
+    expect(store.getActions()).toContainEqual(expectedAction);
+  });
+
+  it('does not reset the hash on search in literature', async () => {
```
harunurhan: this is already tested in the first test case.

ui/src/common/containers/__tests__/SearchBoxContainer.test.jsx
miguelgrc: What do you mean?

ui/src/common/containers/__tests__/SearchBoxContainer.test.jsx
miguelgrc: Is it? The other one does the search in authors, this one does it in literature. The behaviour should be different.

ui/src/common/containers/__tests__/SearchBoxContainer.test.jsx
harunurhan: sorry, I meant `replace` you can use `replace` action creator directly from `react-router-redux` instead of hardcoding the type here.

PR #834, opened 2020-03-24
more cypress tests

harunurhan:


smoke-tests/cypress/support/commands.js
```diff
@@ -1,25 +1,16 @@
...
+  const optionSelector = `[data-test-id="${dropdownId}-option-${option}"]`;
+  cy.get(dropdownSelector).click();
+  cy.get(optionSelector).click();
+});
+
+Cypress.Commands.add('text', { prevSubject: true }, subject => {
+  return subject.text();
+});
+
+Cypress.Commands.overwrite('visit', (originalVisit, relativeUrl, options) => {
```
drjova: ❤️

drjova: can we do the same for the ``viewport``?

harunurhan: I guess we can create like `desktop`, `mobile` short hands.

PR #835, opened 2020-03-24
urls: Change published url

monaawi:
* INSPIR-3340

### DIFF
--- a/backend/inspirehep/search/aggregations.py
+++ b/backend/inspirehep/search/aggregations.py
@@ -31,7 +31,7 @@ def hep_doc_type_aggregation(order, title="Document Type", agg_type="checkbox"):
                 "bucket_help": {
                     "published": {
                         "text": "Published papers are believed to have undergone rigorous peer review.",
-                        "link": "https://inspirehep.net/info/faq/general#published",
+                        "link": "https://inspirehep.net/help/knowledge-base/faq/#faq-published",
                     }
                 },
             },
--- a/ui/src/common/constants.js
+++ b/ui/src/common/constants.js
@@ -36,6 +36,7 @@ export const CONTENT_POLICY_URL = `${KNOWLEDGE_BASE_URL}/content-policy`;
 export const PRIVACY_POLICY_URL = `${KNOWLEDGE_BASE_URL}/privacy-policy`;
 export const TERMS_OF_USE_URL = `${KNOWLEDGE_BASE_URL}/terms-of-use`;
 export const FAQ_URL = `${KNOWLEDGE_BASE_URL}/faq`;
+export const PUBLISHED_URL = `${FAQ_URL}/#faq-published`;
 export const PAPER_SEARCH_URL = `${KNOWLEDGE_BASE_URL}/inspire-paper-search`;
 
 export const DATE_RANGE_FORMAT = 'YYYY-MM-DD';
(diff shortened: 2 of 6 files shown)


backend/inspirehep/search/aggregations.py
```diff
@@ -31,7 +32,7 @@ def hep_doc_type_aggregation(order, title="Document Type", agg_type="checkbox"):
                 "bucket_help": {
                     "published": {
                         "text": "Published papers are believed to have undergone rigorous peer review.",
-                        "link": "https://inspirehep.net/info/faq/general#published",
+                        "link": f"{get_inspirehep_url()}/help/knowledge-base/faq/#faq-published",
```
monaawi: here I'm not sure if I should construct the url using the server name as this wouldn't work on QA since we have a qa faq published. wdyt?

drjova: Actually leaving it ``https://inspirehep.net/help/help/knowledge-base/faq/#faq-published`` would work for both. Do we use it anywhere else? if yes, maybe we can add ```KNOWLEDGE_BASE_URL``` as the ui

monaawi: it's not used anywhere else. Yeah, I could change the link to remain with ```inspirehep.net``` but I still need to change to point to the new knowledge base. I was trying to avoid changing it so that in the future we wouldn't come fix hardcoded server names. But this won't work until we switch to the new domain.

monaawi: I updated it

backend/inspirehep/search/aggregations.py
```diff
@@ -31,7 +32,7 @@ def hep_doc_type_aggregation(order, title="Document Type", agg_type="checkbox"):
                 "bucket_help": {
                     "published": {
                         "text": "Published papers are believed to have undergone rigorous peer review.",
-                        "link": "https://inspirehep.net/info/faq/general#published",
+                        "link": f"https://inspirehep.net/help/knowledge-base/faq/#faq-published",
```
drjova: you don't need the ``f``

monaawi: leftover :+1:

ui/src/common/constants.js
```diff
@@ -36,6 +36,7 @@ export const CONTENT_POLICY_URL = `${KNOWLEDGE_BASE_URL}/content-policy`;
 export const PRIVACY_POLICY_URL = `${KNOWLEDGE_BASE_URL}/privacy-policy`;
 export const TERMS_OF_USE_URL = `${KNOWLEDGE_BASE_URL}/terms-of-use`;
 export const FAQ_URL = `${KNOWLEDGE_BASE_URL}/faq`;
+export const PUBLISHED_URL = `${KNOWLEDGE_BASE_URL}/faq/#faq-published`;
```
harunurhan: you can use `FAQ_URL` instead of appending `/faq`

backend/inspirehep/search/aggregations.py
```diff
@@ -4,6 +4,7 @@
 #
 # inspirehep is free software; you can redistribute it and/or modify it under
 # the terms of the MIT License; see LICENSE file for more details.
+from inspirehep.utils import get_inspirehep_url
```
drjova: unused import

PR #836, opened 2020-03-24
integration-async: improve flaky tests

pazembrz:
Refactored most flaky test 'test_indexer_updates_authors_papers_when_name_changes' which was causing 60% of failures il last week.
Other tests also updated and refactored to be more stable.
`retry_until_matched` tool for tests refactored and updated to be more reliable and to return meaningful exceptions.
Added tests for `retry_until_matched` tool

*INSPIR-3313

### DIFF
--- a/backend/tests/integration-async/conftest.py
+++ b/backend/tests/integration-async/conftest.py
@@ -106,49 +106,61 @@ def retry_until_matched():
                             'expected_key': 'expected_key_name',
                             'expected_result': 'expected_result_data'
                         }
-                    }
+                    },
+                    {
+                        'step': None,  # If step is None it means reuse result fom previous step command (you can ignore `step` key)
+                        # expected_result dict can be skipped if expected_key will be provided.
+                        'expected_key': "a.b[0].c",
+                        'expected_result': "some value",
+                    },
                 ]
         """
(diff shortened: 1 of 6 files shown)


drjova:
I'd rather try to find a better approach than using  and extending ``retry_until_matched``.

What about using something like https://www.elastic.co/guide/en/elasticsearch/reference/7.x/docs-refresh.html to make sure we have the results available to search https://elasticsearch-py.readthedocs.io/en/master/api.html#elasticsearch.Elasticsearch.index

pazembrz:
From what I can see in docs this refresh has to be added in put/post sent to ES for which you want to wait. So we would have to change functions which we want to test for this. Also it wont work for all citation/references updates which is happening in celery tasks for which we made async-tests at the first place because we don't know when they will start exactly and when they will finish.

drjova:
We can explore other options, such as to check the ``health`` of the cluster etc? Have we explored this kind of solutions?

pazembrz:
Yes, nothing guarantees us results as indexing happens in few separate celery tasks. Which can be started independently, ES has nothing to it. Even if ES could index without any delay, we still have celery tasks which won't run immediately in which we are processing citations, relations etc.

PR #839, opened 2020-03-24
ui: fix dropdowns in modals

monaawi:
* INSPIR-3334

### DIFF
--- a/ui/src/App.scss
+++ b/ui/src/App.scss
@@ -195,7 +195,11 @@ body {
 
 .ant-modal-wrap {
   padding-top: $layout-header-height;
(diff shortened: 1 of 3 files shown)


ui/src/App.scss
```diff
@@ -195,7 +195,11 @@ body {
 
 .ant-modal-wrap {
   padding-top: $layout-header-height;
-  z-index: $zindex-header + 1 !important;
+  z-index: $zindex-footer + 1 !important;
```
harunurhan: shall we move all these to `styleVariables` so that we know what is on top of what?

PR #844, opened 2020-03-26
ui: add guide modal

harunurhan:


ui/src/reducers/ui.js
```diff
@@ -1,15 +1,21 @@
...
 
-import { UI_CLOSE_BANNER } from '../actions/actionTypes';
+import {
+  UI_CLOSE_BANNER,
+  UI_CHANGE_GUIDE_MODAL_VISIBILITY,
+} from '../actions/actionTypes';
 
 export const initialState = fromJS({
   closedBannersById: {},
+  guideModalVisibility: null,
```
drjova: why not just ``true`` ``false``

drjova:
@harunurhan is this the final?

ui/src/reducers/ui.js
harunurhan: null has a meaning, it means it was never displayed before.

harunurhan:
@drjova no waiting for actual step images, will merge immediately when this comes

PR #845, opened 2020-03-26
bibtex: use article type for report

miguelgrc:
* INSPIR-3346

### DIFF
--- a/backend/inspirehep/records/marshmallow/literature/bibtex.py
+++ b/backend/inspirehep/records/marshmallow/literature/bibtex.py
@@ -54,13 +54,13 @@ class BibTexCommonSchema(Schema):
         "book": "book",
         "book chapter": "inbook",
         "proceedings": "proceedings",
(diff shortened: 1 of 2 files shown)


backend/inspirehep/records/marshmallow/literature/bibtex.py
```diff
@@ -54,13 +54,15 @@ class BibTexCommonSchema(Schema):
         "book": "book",
         "book chapter": "inbook",
         "proceedings": "proceedings",
-        "report": "techreport",
+        "report": lambda data: "article"
+        if get_value(data, "publication_info.journal_title")
+        else "techreport",
```
michamos: sorry, change of plans. Actually `techreport` doesn't make sense as it requires an `institution` field which we don't have. Let's make it `article` always instead.

miguelgrc: No worries, changed

PR #849, opened 2020-03-27
ui: increase modal padding for big modal + banner

harunurhan:
This is naive solution which we will revert when we have implement a
new Carousel that doesn't use Modal


harunurhan:
FYI: there are better solutions to this, but this is the quickest one.

PR #852, opened 2020-03-30
ui: disable browser autocomplete for DateField

miguelgrc:
* INSPIR-3351

### DIFF
--- a/ui/src/submissions/common/components/DateField.jsx
+++ b/ui/src/submissions/common/components/DateField.jsx
@@ -21,7 +21,9 @@ class DateField extends Component {
   }
 
   render() {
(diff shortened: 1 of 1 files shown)


ui/src/submissions/common/components/DateField.jsx
```diff
@@ -21,7 +21,7 @@ class DateField extends Component {
   }
 
   render() {
-    const { value, ...otherProps } = this.props;
+    const { value, name, ...otherProps } = this.props;
```
harunurhan: a comment might be a good idea why we're not passing this

```
// HACK: don't pass the `name` in order to disable browser form field autocomplete
// until <ant issue link> is resolved
```

PR #856, opened 2020-04-01
New carousel

harunurhan:


ui/src/literature/components/Figures/Figure.jsx
```diff
@@ -5,36 +5,39 @@ import Image from 'react-image';
...
+          pointer: onClick,
+        })}
+        src={url}
+        container={renderWithContainer}
+        unloaderContainer={renderWithoutContainer}
+        alt="Figure"
+        loader={LOADER}
+      />
+      {caption && (
+        <figcaption>
```
drjova, line 37: In order to make semantic sense we should wrap this in
``<figure>``

drjova: my bad you got it already

drjova: but the caption will not be inside I'd change this https://github.com/inspirehep/inspirehep/pull/856/files#diff-1493d7d484f303eaf6e326eca51c36fbR24 into ``figure``

ui/src/common/components/CarouselModal/CarouselModal.jsx
```diff
@@ -0,0 +1,151 @@
...
+import Modal from 'react-modal';
+import { Carousel, Row, Col, Button } from 'antd';
+
+import './CarouselModal.scss';
+import { LeftOutlined, RightOutlined, CloseOutlined } from '@ant-design/icons';
+import useRefOrThis from '../../hooks/useRefOrThis';
+import { useGlobalEvent } from '../../hooks/useGlobalEvent';
+import useResponsiveCheck from '../../hooks/useResponsiveCheck';
+
+const CarouselModal = forwardRef(
```
drjova, line 12: it's quite big, if we add more logic it would be better to split it

ui/src/literature/components/Figures/Figure.jsx
harunurhan: Yep, `figcaption` should be inside `figure`.. I thought it was already but apparently not. I'll do that, I will also try to avoid using `react-image` container API, which makes it hard understand how underlying DOM structure of the component.

ui/src/common/components/CarouselModal/CarouselModal.jsx
harunurhan: I don't think it's big.. but I am happy to hear ideas how to split it, I think it wouldn't make so much difference and would be lots of prop/callback passing.

PR #859, opened 2020-04-02
views: redirect /collection and profile management from/to legacy

miguelgrc:
* INSPIR-3390

### DIFF
--- a/backend/inspirehep/legacy/views.py
+++ b/backend/inspirehep/legacy/views.py
@@ -23,8 +23,17 @@ def redirect_record(record_path):
     return redirect(f"/{endpoint}/{recid}", 301)
 
 
-@blueprint.route("/author/profile/<path:bai>", methods=("GET",))
+@blueprint.route("/author/<bai>", methods=("GET",))
 def redirect_author(bai):
+    return _redirect_author(bai)
+
+
+@blueprint.route("/author/profile/<bai>", methods=("GET",))
+def redirect_author_profile(bai):
+    return _redirect_author(bai)
+
+
+def _redirect_author(bai):
     recid = get_pid_for_pid("bai", bai, "recid")
     if not recid:
         abort(404)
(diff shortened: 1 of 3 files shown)


backend/tests/integration/legacy/test_views.py
```diff
@@ -85,15 +85,40 @@ def test_redirects_query_from_legacy_url_with_empty_query(
 def test_redirects_query_from_legacy_url_not_in_labs(
     api_client, db, es_clear, create_record, base_app
 ):
-    with patch.dict(
-        base_app.config, {"LEGACY_BASE_URL": "https://legacy.inspirehep.net"}
-    ):
+    with patch.dict(base_app.config, {"LEGACY_BASE_URL": "https://old.inspirehep.net"}):
```
michamos, line 88: not needed anymore as you changed the global config.

backend/tests/integration/legacy/test_views.py
```diff
@@ -85,15 +85,40 @@ def test_redirects_query_from_legacy_url_with_empty_query(
...
+def test_redirects_collections_from_legacy_url(api_client, db, es_clear, create_record):
+    response = api_client.get("/collection/HepNames")
+
+    response_status_code = response.status_code
+    response_location_header = response.headers.get("Location")
+
+    expected_status_code = 301
+    expected_redirect_path = "/authors"
+    assert expected_status_code == response_status_code
+    assert response_location_header.endswith(expected_redirect_path)
```
michamos, line 109: why no assertion on the whole string?

backend/tests/integration/legacy/test_views.py
```diff
@@ -85,15 +85,40 @@ def test_redirects_query_from_legacy_url_with_empty_query(
...
+
+
+def test_redirects_collections_from_legacy_url_not_in_labs(
+    api_client, db, es_clear, create_record, base_app
+):
+    with patch.dict(base_app.config, {"LEGACY_BASE_URL": "https://old.inspirehep.net"}):
```
michamos, line 115: same as before

backend/tests/integration/legacy/test_views.py
miguelgrc: You said it was better to do it like this to make the tests independent from the actual config, right?

backend/tests/integration/legacy/test_views.py
miguelgrc: Since the host was labs before, I did it like this to avoid the tests failing after releasing and redirecting labs to inspirehep.net. Yeah, now I can check the whole url.

backend/tests/integration/legacy/test_views.py
drjova: 👍  we don't have to patch it

backend/tests/integration/legacy/test_views.py
drjova: Indeed, as you do https://github.com/inspirehep/inspirehep/pull/859/files#diff-cca37f1084b473645317310c897326efR114

backend/inspirehep/legacy/views.py
```diff
@@ -23,34 +31,76 @@ def redirect_record(record_path):
...
+        302,
+    )
+
+
+@blueprint.route("/collection/<legacy_collection>", methods=("GET",))
+def redirect_collection(legacy_collection):
+    # TODO: Delete redirection to legacy when Institutions and Experiments are
+    # on labs and legacy UI is completely shut down
+
+    collection = COLLECTION_EQUIVALENCE.get(legacy_collection)
```
drjova, line 95: Can we move it to the global config? In that way we can update it easily

backend/tests/integration/legacy/test_views.py
miguelgrc: Then the URL would depend on the environment. I guess I should use the [app_config fixture](https://github.com/inspirehep/inspirehep/blob/cdc5c7f0e21522223a2178f4388af0ad002925ed/backend/tests/integration/conftest.py#L43), right?

miguelgrc: Then the URL would depend on the environment. I guess I should use the [app_config fixture](https://github.com/inspirehep/inspirehep/blob/cdc5c7f0e21522223a2178f4388af0ad002925ed/backend/tests/integration/conftest.py#L43), right?

drjova: yes exactly, it will be the same for all  tests

PR #861, opened 2020-04-02
ui: polyfill Object.entries

miguelgrc:
* INSPIR-3365


ui/src/index.js
```diff
@@ -1,3 +1,6 @@
+// Fixing Object.entries in older browsers
```
harunurhan, line 1: well this line is obviously a polyfill import, so no need for a comment 😄

PR #863, opened 2020-04-03
POC:self-citations backend

pazembrz:
* INSPIR-3235


drjova:
@pazembrz can we close this?

PR #867, opened 2020-04-06
search: increase maxrecursiondepth for resolving query string

pazembrz:
* INSPIR-3377

### DIFF
--- a/backend/inspirehep/search/factories/query.py
+++ b/backend/inspirehep/search/factories/query.py
@@ -4,16 +4,18 @@
 #
 # inspirehep is free software; you can redistribute it and/or modify it under
 # the terms of the MIT License; see LICENSE file for more details.
-
-
 import inspire_query_parser
 from elasticsearch_dsl import Q
+from flask import current_app
+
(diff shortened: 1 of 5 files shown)


backend/inspirehep/search/factories/query.py
```diff
@@ -4,16 +4,27 @@
...
 
     def inspire_query(query_string, search):
-        return Q(inspire_query_parser.parse_query(query_string))
+        original_recursion_limit = sys.getrecursionlimit()
+        sys.setrecursionlimit(
+            current_app.config.get("SEARCH_MAX_RECURSION_LIMIT", 5000)
+        )
+        try:
+            query = Q(inspire_query_parser.parse_query(query_string))
+        except:
```
michamos, line 24: the `except` is useless here.

pazembrz: :+1: I forgot to remove it ;)

pazembrz:
Decorator? I'm not fan of decorators...  Maybe as context manager instead?

michamos:
context manager would be good :+1:

PR #868, opened 2020-04-06
ui: display search page size changer

harunurhan:
INSPIR-3352


drjova:
@harunurhan maybe i missed that but in the embed search, at least in the screenshots, there is not a test with the page size selection

harunurhan:
@drjova because this changes the default size for those from `10` to `25`

drjova:
could we use another record that has more than 25? To make sure we render properly the dropdown in embeded search as well

harunurhan:
we could, but screenshots are not intended to capture all cases (currently at least).

drjova:
Sure, IMHO it's a core feature and we should test both cases i.e. single page and multiple pages.

harunurhan:
pagination is core feature for search, and right now both cases are tested.. but if you want to test both cases for each collection, that could get unproductive very easily.

drjova:
I don't want to test each collection, I just want to make sure the select box for search results size is rendered properly with the correct style.

harunurhan:
Well there are some search pages that have it, no?

drjova:
I'm talking about the embed search, please correct me if I'm wrong but I didn't see any screenshot for this

harunurhan:
embedded literature search doesn't have its own layout.

PR #869, opened 2020-04-06
submission: check also official_name of country not only 'name'

pazembrz:
* INSPIR-3403

### DIFF
--- a/backend/inspirehep/submissions/marshmallow/conferences.py
+++ b/backend/inspirehep/submissions/marshmallow/conferences.py
@@ -43,16 +43,18 @@ class Conference(Schema):
         builder = ConferenceBuilder()
         builder.add_title(title=data.get("name"), subtitle=data.get("subtitle"))
         builder.set_short_description(value=data.get("description", ""))
-        builder.set_opening_date(get_value(data, 'dates[0]'))
-        builder.set_closing_date(get_value(data, 'dates[1]'))
+        builder.set_opening_date(get_value(data, "dates[0]"))
+        builder.set_closing_date(get_value(data, "dates[1]"))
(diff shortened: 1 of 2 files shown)


backend/inspirehep/submissions/marshmallow/conferences.py
```diff
@@ -43,16 +43,18 @@ def build_conference(self, data) -> dict:
...
         builder.add_inspire_categories(data.get("field_of_interest", []))
         builder.add_public_note(value=data.get("additional_info", ""))
         builder.add_series(
-            name=data.get("series_name"),
-            number=data.get("series_number")
+            name=data.get("series_name"), number=data.get("series_number")
         )
         for address in data.get("addresses"):
-            country = pycountry.countries.get(name=address.get("country"))
+            try:
```
monaawi, line 54: the country options we display in the ui conference form are in https://github.com/inspirehep/inspirehep/blob/master/ui/src/submissions/conferences/schemas/constants.js, are these using the name or the official name? Maybe this should be aligned with the backend as well

pazembrz: Micha was not sure about it as some of the official names are `ugly` 
```
Micha Moskovic16:10
I guess also here the official name would be better. IIRC, there were ugly things things like Congo, Republic of in the official name. Or maybe that was a different lib, I'm not completely sure.
```
So I created Spike for this https://its.cern.ch/jira/browse/INSPIR-3405

monaawi: but the list I gave you is for the submission form, not the detail/search page

PR #874, opened 2020-04-07
bump query parser

MJedr:


backend/poetry.lock
```diff
@@ -2667,7 +2667,7 @@ version = "2.8.4"
 [[package]]
 category = "main"
 description = "Run a subprocess in a pseudo terminal"
-marker = "python_version >= \"3.4\" and sys_platform != \"win32\" or sys_platform != \"win32\""
+marker = "python_version >= \"3.4\" and sys_platform != \"win32\" or sys_platform != \"win32\" or python_version >= \"3.4\" and sys_platform != \"win32\" and (python_version >= \"3.4\" and sys_platform != \"win32\" or sys_platform != \"win32\")"
```
monaawi, line 2670: hm, it's odd it is adding this for you. It doesn't add it when I run ```poetry update```. What version of poetry are you using?

monaawi:
Also, can you change the commit message to be ```setup: bump inspire...``` and add the JIRA task in the message as well?
This is typically the format we use for commit messages:
{area changed} : {description of the change}
* {JIRA TASK}
* {any needed operations like migrating, reindexing etc.}
* {extra info}

backend/poetry.lock
MJedr: I use poetry 1.0.5 and ```poetry update query-parser``` was the only command I run.

monaawi: Can you try with 1.0.0?

PR #876, opened 2020-04-07
files: make `inline`

harunurhan:
So that they can be display inside browsers, not downloaded immediately.

### DIFF
--- a/backend/inspirehep/files/api/s3.py
+++ b/backend/inspirehep/files/api/s3.py
@@ -99,7 +99,21 @@ class S3:
 
     @staticmethod
     def get_content_disposition(filename):
-        return f'inline; filename="{secure_filename(filename)}"'
+        without_subformat = S3.remove_subformat(filename)
+        return f'inline; filename="{secure_filename(without_subformat)}"'
+
+    @staticmethod
+    def remove_subformat(filename):
+        parts = filename.split(".")
+
+        if len(parts) == 0:
+            return filename
+
+        ext = parts.pop()
+        ext_without_subformat = ext.split(";")[0]
+        filename = ".".join(parts)
+
+        return f"{filename}.{ext_without_subformat}"
 
     def replace_file_metadata(self, key, filename, mimetype, acl):
         """Updates the metadata of the given file.
--- a/backend/inspirehep/files/api/s3.py
+++ b/backend/inspirehep/files/api/s3.py
@@ -99,7 +99,7 @@ class S3:
 
     @staticmethod
     def get_content_disposition(filename):
-        return f'attachment; filename="{secure_filename(filename)}"'
+        return f'inline; filename="{secure_filename(filename)}"'
 
     def replace_file_metadata(self, key, filename, mimetype, acl):
         """Updates the metadata of the given file.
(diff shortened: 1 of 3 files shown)


monaawi:
How are we going to handle updating the metadata of the current files? Will we update them manually or run a migration with the update metadata flag true? Anyhow, an operation is needed to have the metadata updated, can you add it in the commit message?

harunurhan:
@monaawi I guess we'll run migration, I'll update the commit message, I'll also push the other task here too.

backend/inspirehep/files/api/s3.py
```diff
@@ -99,7 +99,21 @@ def get_file_url(key):
 
     @staticmethod
     def get_content_disposition(filename):
-        return f'attachment; filename="{secure_filename(filename)}"'
+        without_subformat = S3.remove_subformat(filename)
```
monaawi, line 102: maybe ```filename_without_subformat```?

backend/tests/integration/files/test_api_s3.py
```diff
@@ -119,6 +119,18 @@ def test_get_file_url(base_app, appctx):
...
 def test_get_content_disposition_with_broken_filename(base_app, appctx):
-    expected_content_disposition = 'attachment; filename="Mass_mm_.s"'
-    result = current_s3_instance.get_content_disposition("Mass_mm\n.s\n")
+    expected_content_disposition = 'inline; filename="Mass_mm_.s"'
+    result = current_s3_instance.get_content_disposition("Mass_mm\n.s;b\n")
+    assert result == expected_content_disposition
+
+
+def test_get_content_disposition_with_subformat(base_app, appctx):
+    expected_content_disposition = 'inline; filename="filefulltext.pdf"'
```
monaawi, line 128: is this what we want? with ```secure_filename```, it would have still had the ```.``` between full and text and you are removing it. I think it's fine but I just wanted to put the question it out there.

harunurhan: (this is a behaviour from before) `werkzeug` thinks they are insecure, and I don't think we're losing anything useful there anyway.

backend/inspirehep/files/api/s3.py
harunurhan: i think, one understands that it's a file name, while reading this line.

monaawi: It is understandable although I think ```filename_without_subformat``` is clearer and more understandable. I don't have a strong opinion about it, keep it like this

backend/tests/integration/files/test_api_s3.py
monaawi: Actually, that's what I'm saying, ```werkzeug``` was keeping the ```.``` before so it isn't a behavior from before. ```secure_filename('full.text.pdf')``` would be ```full.text.pdf'``` whereas now you are removing the ```.```. I don't think it is relevant though, I don't know what weird filenames we have

harunurhan: true, you're right I forgot with `"."` back.. it's probably not important but better to keep.

backend/inspirehep/files/api/s3.py
```diff
@@ -99,7 +99,21 @@ def get_file_url(key):
...
+    @staticmethod
+    def remove_subformat(filename):
+        parts = filename.split(".")
+
+        if len(parts) == 0:
+            return filename
+
+        ext = parts.pop()
+        ext_without_subformat = ext.split(";")[0]
+        filename = "".join(parts)
```
monaawi, line 114:
```suggestion
        filename = ".".join(parts)
```

PR #879, opened 2020-04-08
bibtex: don't escape spaces in journals

MJedr:
* INSPIR-3416


backend/inspirehep/records/marshmallow/literature/bibtex.py
```diff
@@ -283,8 +283,8 @@ def get_journal(self, data):
         return (
             BibTexCommonSchema.get_best_publication_info(data)
             .get("journal_title")
-            .replace(".", ".\\ ")
-            .rstrip("\\ ")
+            .replace(".", r". ")
+            .rstrip(r" ")
```
michamos, line 287: you don't need r-strings here.

backend/inspirehep/records/marshmallow/literature/latex.py
```diff
@@ -46,9 +46,9 @@ def get_publication_info(self, data):
             return missing
 
         if "journal_title" in publication_info:
-            publication_info["journal_title"] = publication_info[
-                "journal_title"
-            ].replace(".", ".\\ ")
+            publication_info["journal_title"] = (
+                publication_info["journal_title"].replace(".", r". ").rstrip(r" ")
```
michamos, line 50: same

michamos:
Note that this requires reindexing (I don't remember where this should be mentioned exactly).

PR #880, opened 2020-04-08
search: add function to return list of dicts from ES hits

monaawi:
* INSPIR-3375

### DIFF
--- a/backend/inspirehep/records/marshmallow/institutions/ui.py
+++ b/backend/inspirehep/records/marshmallow/institutions/ui.py
@@ -84,9 +84,9 @@ class InstitutionsDetailSchema(InstitutionsBaseSchema):
             .query(query)
             .params(size=10000, _source=["legacy_ICN", "control_number"])
         )
-        results = [
-            query_result.to_dict() for query_result in query_results.execute().hits
-        ]
+        results = InstitutionsSearch.get_dict_results_from_hits(
+            query_results.execute().hits
+        )
         return results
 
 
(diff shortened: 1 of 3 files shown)


backend/inspirehep/records/marshmallow/institutions/ui.py
```diff
@@ -84,9 +84,7 @@ def get_subsidiary_institutions(data):
             .query(query)
             .params(size=10000, _source=["legacy_ICN", "control_number"])
         )
-        results = [
-            query_result.to_dict() for query_result in query_results.execute().hits
-        ]
+        results = InspireSearch.get_dict_results_from_hits(query_results.execute().hits)
```
michamos, line 87: isn't an `InstitutionSearch` already an `InspireSearch`?

backend/inspirehep/records/views.py
```diff
@@ -44,7 +44,9 @@ def get(self, pid, record):
 
         citing_records_results = LiteratureSearch.citations(record, page, size)
         citing_records_count = citing_records_results.total["value"]
-        citing_records = [citation.to_dict() for citation in citing_records_results]
+        citing_records = InspireSearch.get_dict_results_from_hits(
```
michamos, line 47: same question

pazembrz:
I think I would put it in some helper, as now it's counter intuitive to call static method from class which we already used before. Or change `citations` method to return data already processed by `get_dict_results_from_hits`

monaawi:
I can't call it inside the citations because I also need access to the ```Hit``` to get the total of the results. I think it's the same thing a helper or inside the search, maybe it is better inside the search as we seem to have other helper functions there too

PR #882, opened 2020-04-08
institutions: modify address display

monaawi:
* INSPIR-3413

### DIFF
--- a/ui/src/institutions/components/InstitutionAddress.jsx
+++ b/ui/src/institutions/components/InstitutionAddress.jsx
@@ -1,0 +1,40 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import { Map, List } from 'immutable';
+import InlineList, {
+  InlineUL,
+  SEPARATOR_COMMA,
+} from '../../common/components/InlineList';
+
+function postalAddressesContainWord(postalAddresses, word) {
+  return postalAddresses.some(
+    address => address.toLowerCase().indexOf(word.toLowerCase()) > -1
+  );
+}
+
+function InstitutionAddress({ address }) {
+  const postalAddresses = address.get('postal_address', List());
+  const city = address.getIn(['cities', 0]);
+  const country = address.get('country');
+  return (
+    <span>
+      <InlineUL separator={SEPARATOR_COMMA} wrapperClassName="di">
+        {postalAddresses.size > 0 && <InlineList items={postalAddresses} />}
+        {city &&
+          !postalAddressesContainWord(postalAddresses, city) && (
+            <span>{city}</span>
+          )}
+        {country &&
+          !postalAddressesContainWord(postalAddresses, country) && (
+            <span>{country}</span>
+          )}
+      </InlineUL>
+    </span>
+  );
+}
+
+InstitutionAddress.propTypes = {
+  address: PropTypes.instanceOf(Map).isRequired,
+};
+
+export default InstitutionAddress;
--- a/ui/src/institutions/components/InstitutionAddressList.jsx
+++ b/ui/src/institutions/components/InstitutionAddressList.jsx
@@ -1,0 +1,34 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import { List } from 'immutable';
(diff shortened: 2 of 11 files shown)


ui/src/institutions/components/InstitutionAddress.jsx
```diff
@@ -0,0 +1,39 @@
...
+import PropTypes from 'prop-types';
+import { Map, List } from 'immutable';
+import InlineList, {
+  InlineUL,
+  SEPARATOR_COMMA,
+} from '../../common/components/InlineList';
+
+function postalAddressContainsString(postalAddress, string) {
+  const postalAddressString = postalAddress.join(' ');
+  return postalAddressString.toLowerCase().indexOf(string.toLowerCase()) > -1;
```
harunurhan, line 11:
```js
postalAddress
  .some(address => address.toLowerCase().indexOf(string.toLowerCase()) > -1)
```

ui/src/institutions/components/InstitutionAddress.jsx
```diff
@@ -0,0 +1,39 @@
...
+}
+
+function InstitutionAddress({ address }) {
+  const postalAddress = address.get('postal_address', List());
+  const city = address.getIn(['cities', 0]);
+  const country = address.get('country');
+  return (
+    <span>
+      <InlineUL separator={SEPARATOR_COMMA} wrapperClassName="di">
+        {postalAddress.size > 0 && <InlineList items={postalAddress} />}
```
harunurhan, line 21: this check should be already in `InlineList`

ui/src/institutions/components/InstitutionAddressList.jsx
```diff
@@ -0,0 +1,34 @@
...
+function renderAddress(address) {
+  return <InstitutionAddress address={address} />;
+}
+
+function InstitutionAddressList({ addresses }) {
+  return (
+    <InlineList
+      wrapperClassName="di"
+      items={addresses}
+      // FIXME: set extractKey explictly
```
harunurhan, line 19: how about fixing this?

ui/src/institutions/components/InstitutionAddress.jsx
```diff
@@ -0,0 +1,39 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import { Map, List } from 'immutable';
+import InlineList, {
+  InlineUL,
+  SEPARATOR_COMMA,
+} from '../../common/components/InlineList';
+
+function postalAddressContainsString(postalAddress, string) {
```
harunurhan, line 9: it's better not use keywords like `string` and `type`, incase we'd like to add some type checking with `TypeScript`.

ui/src/institutions/components/InstitutionAddress.jsx
```diff
@@ -0,0 +1,39 @@
...
+  const postalAddressString = postalAddress.join(' ');
+  return postalAddressString.toLowerCase().indexOf(string.toLowerCase()) > -1;
+}
+
+function InstitutionAddress({ address }) {
+  const postalAddress = address.get('postal_address', List());
```
harunurhan, line 15: let's always call this `postalAddresses`

ui/src/institutions/components/InstitutionAddress.jsx
monaawi: It is already in ```InlineList``` but if I don't add this check also here then it doesn't seem to work with the ```InlineUL```. If I have an empty list and then something after it, it will still add the separator in the middle unless I add the condition here as well

ui/src/institutions/components/InstitutionAddressList.jsx
monaawi: I would but I'm not sure what to add to the key. Nothing in the address is unique/mandatory

harunurhan: I see, it's not nice like that, It's calling `Map.toString()` all the time to get the key, we could memoize that but I guess it's not worth it at the moment.

PR #883, opened 2020-04-09
bibtex: add editors from parent record

MJedr:
* INSPIR-3399

### DIFF
--- a/backend/inspirehep/records/marshmallow/literature/bibtex.py
+++ b/backend/inspirehep/records/marshmallow/literature/bibtex.py
@@ -126,7 +126,14 @@ class BibTexCommonSchema(Schema):
         return self.get_authors_with_role(data.get("authors", []), "author")
 
     def get_authors_with_role_editor(self, data):
-        return self.get_authors_with_role(data.get("authors", []), "editor")
+        editors = self.get_authors_with_role(data.get("authors", []), "editor")
+        if not editors and data.get("doc_type") in [
+            "inbook",
+            "inproceedings",
+            "article",
+        ]:
+            editors = self.get_book_editors(data)
+        return editors
 
     def get_eprint(self, data):
         return get_value(data, "arxiv_eprints.value[0]", default=None)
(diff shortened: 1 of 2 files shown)


backend/inspirehep/records/marshmallow/literature/bibtex.py
```diff
@@ -126,7 +127,14 @@ def get_authors_with_role_author(self, data):
...
 
     def get_authors_with_role_editor(self, data):
-        return self.get_authors_with_role(data.get("authors", []), "editor")
+        editors = self.get_authors_with_role(data.get("authors", []), "editor")
+        if not editors and data.get("doc_type") in [
+            "inbook",
+            "inproceedings",
+            "article",
+        ]:
+            editors = self.get_book_editors(data)
```
drjova, line 136: I'm a bit confused, this is calling ``get_book_editors`` and https://github.com/inspirehep/inspirehep/pull/883/files#diff-9574106e620d1306e43f0ca0466a6edaR283 it calls ``get_authors_with_role_editor``

MJedr: The case is that for book chapters, conference papers and some articles (which are actually conference papers with journal_title in publication_info) the information about editors should be taken from the parent record. The reason why I added that:
```
if not editors and data.get("doc_type") in [
            "inbook",
            "inproceedings",
            "article",
        ]:
             editors = self.get_book_editors(data)
```
is that some articles are conference papers and they have information about editors hidden in parent record, so they need to be treated differently, and some articles are normal articles and they have info about editors in the authors field.

drjova: Please correct me if I'm wrong but you are already passing the parent record by calling ``bookeditors = fields.Method("get_book_editors")`` from the serializer:

```python
    def get_book_editors(self, data):
        parent_record = self.get_parent_record(data)
        parent_editors = self.get_authors_with_role_editor(parent_record)
```

So the flow is the following: You have a record with ``doc_type: 'article'``,  the serializer will call ``get_book_editors`` which will call the parent of the record and get the ``editor`` from authors. If you need the editors only for ``inbook, inproceedings, article`` then the check should be inside the ``get_book_editors``.


```python
    def get_book_editors(self, data):
         if not editors and data.get("doc_type") in [ "inbook", "inproceedings", "article",]:
            parent_record = self.get_parent_record(data)
            return self.get_authors_with_role_editor(parent_record)
         return None
```

Please note the serializer with ```fields.Method("get_book_editors")``` will always call ``get_book_editors``.

MJedr: Yes, sorry, I forgot to remove this field (I added it at the beginning, but then I realised that it would be better to have one field related to editors - authors_with_role_editor, no matter if it is a book chapter or book), so that is the reason why ```get_book_editors``` looked this way. Of course, if it is better we can do 2 separate fields, but then some changes should be done also in bibtex and latex display.

drjova: LGTM then you just need to remove the field

PR #888, opened 2020-04-15
ui: add local search history and suggestions

harunurhan:
INSPIR-3408


ui/src/common/components/SearchBox/searchHistory.js
```diff
@@ -0,0 +1,42 @@
+import { SEARCH_BOX_NAMESPACES } from '../../../reducers/search';
+import LRASet from './LRASet';
+
+const STORAGE_KEY = 'search-box-history';
+const HISTORY_LIMIT = 25;
+
+function runWhenIdle(task) {
```
drjova, line 7: idle js everywhere

harunurhan: this is when js engine is idle, not the user so unfortunately we can't use `idlejs` 😭  😃

harunurhan:
Btw, I guess we don't care perfect multitab support, right? because currently it reads from storage only at the start and keeps write on it. 

If user was to search via multiple tabs, history would be overwritten always by the tab where the latest search happened and we wouldn't able to save searches from the other tab.

drjova:
@harunurhan that's a good point, let's see if we get any compains about that

PR #892, opened 2020-04-17
search: addition of query parser endpoint

drjova:


backend/inspirehep/search/views.py
```diff
@@ -0,0 +1,24 @@
...
+
+from flask import Blueprint, abort, current_app, jsonify, request
+from inspire_query_parser import parse_query
+
+blueprint = Blueprint("inspirehep_search", __name__, url_prefix="/search")
+
+
+@blueprint.route("/query-parser", methods=["GET"])
+def query_parser():
+    if not current_app.config.get("FEATURE_FLAG_ENABLE_QUERY_PARSER_ENTPOINT", True):
```
michamos, line 16:
```suggestion
    if not current_app.config.get("FEATURE_FLAG_ENABLE_QUERY_PARSER_ENDPOINT", True):
```

backend/inspirehep/search/views.py
```diff
@@ -0,0 +1,24 @@
...
+        abort(404)
+
+    try:
+        query = request.values.get("query", "", type=str)
```
michamos, line 20:
```suggestion
        query = request.values.get("q", "", type=str)
```
so you can just edit the URL to get the query parser.

drjova:
@michamos done

backend/tests/integration/__init__.py
```diff
@@ -0,0 +1,6 @@
+# -*- coding: utf-8 -*-
+#
+# Copyright (C) 2020 CERN.
+#
+# inspirehep is free software; you can redistribute it and/or modify it under
+# the terms of the MIT License; see LICENSE file for more details.
```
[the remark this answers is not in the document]
drjova: we need for pytests, pytest if there is no ``__init__.py`` import the tests as top-level modules and since we're not using unique names for the tests (we organize them per module, hence we have ``search/test_views.py`` and ``records/test_views.py`` for example) pytest compains. More info: https://docs.pytest.org/en/latest/goodpractices.html#tests-outside-application-code

PR #893, opened 2020-04-17
orcid: handle multiple ORCID putcodes for the same record

MJedr:
* INSPIR-3383

### DIFF
--- a/backend/inspirehep/orcid/domain_models.py
+++ b/backend/inspirehep/orcid/domain_models.py
@@ -42,6 +42,7 @@ class OrcidPusher(object):
         self.lock_name = "orcid:{}".format(self.orcid)
         self.client = OrcidClient(self.oauth_token, self.orcid)
         self.converter = None
+        self.cached_author_putcodes = {}
 
     @time_execution
     def _get_inspire_record(self):
@@ -195,6 +196,7 @@ class OrcidPusher(object):
         xml_element = self.converter.get_xml(do_add_bibtex_citation=True)
         # ORCID API allows 1 non-idempotent call only for the same orcid at
         # the same time. Using `distributed_lock` to achieve this.
+
         with utils.distributed_lock(self.lock_name, blocking=True):
             if putcode:
                 response = self.client.put_updated_work(xml_element, putcode)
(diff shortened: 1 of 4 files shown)


backend/inspirehep/orcid/domain_models.py
```diff
@@ -204,16 +206,30 @@ def _post_or_put_work(self, putcode=None):
         response.raise_for_result()
         return response["putcode"]
 
+    def _delete_works_with_duplicated_putcodes(self, cached_putcodes_recids):
+        unique_putcodes_recids = {}
+        for fetched_putcode, fetched_recid in cached_putcodes_recids:
+            if fetched_recid in unique_putcodes_recids.values():
```
michamos, line 212: accessing a dictionary by keys is fast (O(1)), accessing a dictionary by values is slow (O(len(dict))). So it's better to invert the dict here so you can do
```suggestion
            if fetched_recid in unique_recids_putcodes:
```

backend/inspirehep/orcid/domain_models.py
```diff
@@ -204,16 +206,30 @@ def _post_or_put_work(self, putcode=None):
...
+                self._delete_work(fetched_putcode)
+            else:
+                unique_putcodes_recids[fetched_putcode] = fetched_recid
+        return [(putcode, recid) for putcode, recid in unique_putcodes_recids.items()]
```
michamos, line 216: why not use the previously built dict as a cache instead?

backend/tests/integration/orcid/test_orcid_domain_models.py
```diff
@@ -265,6 +264,20 @@ def test_push_new_work_already_existing_duplicated_external_identifier_exception
...
+    ):
+        with override_config(
+            ORCID_APP_CREDENTIALS={"consumer_key": "0000-0001-8607-8906"}
+        ):
+            self.add_work(
+                "test_orcid_domain_models_TestOrcidPusherPostNewWork_conflicting_no_recid.xml"
+            )
+            pusher = domain_models.OrcidPusher(self.orcid, self.recid, self.oauth_token)
+            pusher.push()
+            assert mock.called
```
michamos, line 279: why mock this? I would do it like the other tests and check the cassette to make sure the DELETE is sent and the response is ok, instead you should add 
```suggestion
            assert self.cache.has_work_content_changed(self.inspire_record)
```

MJedr: I am not sure if ```self.cache.has_work_content_changed(self.inspire_record)``` will be good there. We are randomly choosing the record which is going to be deleted and those records might change. It is why I preferred to check if one of the records was deleted.

michamos: at the end of the day, the push should succeed, so the record will have been pushed and the cache will agree with the current version.

MJedr: So it should be ```assert not self.cache.has_work_content_changed(self.inspire_record)``` as the record doesn't change.

michamos: indeed, I forgot the `not` :facepalm:

backend/tests/integration/orcid/cassettes/TestOrcidPusherPostNewWork.test_push_new_work_already_existing_and_delete_duplicated_records_different_putcodes.yaml
```diff
@@ -0,0 +1,411 @@
...
+      - nginx/1.16.1
+      X-Content-Type-Options:
+      - nosniff
+      X-Frame-Options:
+      - DENY
+      X-XSS-Protection:
+      - 1; mode=block
+    status:
+      code: 409
+      message: Conflict
```
michamos, line 118: I had a second thought about this, but after thinking about it a bit more (and remembering tests are not isolated), it's fine :+1:

MJedr: Yes, the thing is that we are pushing the work which already exists (twice) in orcid, so to cache all author putcodes and delete duplicated record from orcid there must be 409 (WorkAlreadyExists) before.

MJedr: Yes, the thing is that we are pushing the work which already exists (twice) in orcid, so to cache all author putcodes and delete duplicated record from orcid there must be 409 (WorkAlreadyExists) before.

PR #895, opened 2020-04-20
mappings: case insensitive search for dois in hep mapping

MJedr:
* INSPIR-3418

### DIFF
--- a/backend/inspirehep/search/mappings/v7/records/hep.json
+++ b/backend/inspirehep/search/mappings/v7/records/hep.json
@@ -939,6 +939,7 @@
               "dois": {
                 "fields": {
                   "raw": {
+                    "normalizer": "lowercase_normalizer",
                     "type": "keyword"
                   }
                 },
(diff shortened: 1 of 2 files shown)


pazembrz:
Would be nice if you could add simple test. It's more for the future changes in mappings to be sure that no one will break this ;)

MJedr:
Fine, I found a very similar [commit](https://github.com/inspirehep/inspirehep/commit/615673848d68819c79862c858303b1c893654bf6) without tests, so it is why I thought that it is going to be fine

drjova:
We don't have tests for mapping analyzers and normilizers at the moment, this can be merged without. @pazembrz has a good point, we can add e2e for the most important user cases.

monaawi:
Correct me if I'm wrong, but why would this be difficult to test for this particular case? Basically, we want to make sure that we have case insensitive search for dois so we can simply test that searching for a doi with a different case gives us a correct result.

pazembrz:
> We don't have tests for mapping analyzers and normilizers at the moment
Because no one ever bothered to create them, but most of them are quite simple to create.

drjova:
I didn't say that it's hard to test, conceptually this kind of tests don't belong to the intergration suite, it's like we are testing the ``analysis-icu`` elasticsearch plugin. These are user cases that should be tested e2e IMHO. Mappings and its analyzers is a configuration for an external service which we don't have direct control.

pazembrz:
I believe at least simple tests should be in integration suite, as this is actually an integration between services. And I'm not saying to test if mapping is correct etc, but just check if response for simple query is ok. Like here, add one record to ES and check if you can query with uppercase DOI and lowercase DOI. just like test for DB for DOI.

PR #900, opened 2020-04-21
mixins: add self-citations backend

pazembrz:
* INSPIR-3235


backend/inspirehep/records/api/mixins.py
```diff
@@ -228,6 +346,46 @@ def update(self, data, disable_relations_update=False, *args, **kwargs):
...
+                for result in RecordsAuthors.query.filter(
+                    RecordsAuthors.record_id != self.id
+                )
+                .filter(
+                    and_(RecordsAuthors.author_id.in_(diff)),
+                    RecordsAuthors.id_type == "INSPIRE BAI",
+                )
+                .all()
+            ]
+        return differed_papers_uuids
```
michamos, line 367: Not sure if I'm stupid or this is crazy, but this looks like you're going to reindex all ATLAS papers as soon as a single ATLAS paper is touched. A self-citation is a citation, so it can only affect citing or cited papers. Cited papers get reindexed already in any case to update the citation count, so that leaves citing papers. Can't you use a query on the citation table you've updated previously to find those?

backend/inspirehep/records/api/mixins.py
```diff
@@ -203,17 +232,106 @@ def update_refs_in_citation_table(self, save_every=100):
                         citer_id=self.model.id,
                         cited_id=reference,
                         citation_date=citation_date,
+                        citation_type=RecordCitationType.citation,
```
michamos, line 235: so you treat this as a default, I think it makes sense to make it explicit at the DB level as suggested above.

michamos, line 235: so you treat this as a default, I think it makes sense to make it explicit at the DB level as suggested above.

backend/inspirehep/records/api/mixins.py
```diff
@@ -203,17 +232,106 @@ def update_refs_in_citation_table(self, save_every=100):
...
+        RecordCitations.query.filter(
+            and_(
+                RecordCitations.cited_id == uuid,
+                not_(RecordCitations.citer_id.in_(self_citations)),
+                RecordCitations.citation_type == RecordCitationType.self_citation,
+            )
+        ).update(
+            {RecordCitations.citation_type: RecordCitationType.citation},
+            synchronize_session=False,
         )
```
michamos, line 335: I think it would be easier to do the whole update in DB by joining on the query to get self-citations, as this is inefficient and you run into the risk of exceeding the query limit length if there are lots of self-citations.

michamos, line 335: I think it would be easier to do the whole update in DB by joining on the query to get self-citations, as this is inefficient and you run into the risk of exceeding the query limit length if there are lots of self-citations.

michamos, line 335: I think it would be easier to do the whole update in DB by joining on the query to get self-citations, as this is inefficient and you run into the risk of exceeding the query limit length if there are lots of self-citations.

backend/inspirehep/records/api/mixins.py
michamos, line 367: Not sure if I'm stupid or this is crazy, but this looks like you're going to reindex all ATLAS papers as soon as a single ATLAS paper is touched. A self-citation is a citation, so it can only affect citing or cited papers. Cited papers get reindexed already in any case to update the citation count, so that leaves citing papers. Can't you use a query on the citation table you've updated previously to find those?

backend/inspirehep/records/api/mixins.py
```diff
@@ -228,6 +346,46 @@ def update(self, data, disable_relations_update=False, *args, **kwargs):
...
+                for result in RecordsAuthors.query.filter(
+                    RecordsAuthors.record_id != self.id
+                )
+                .filter(
+                    and_(RecordsAuthors.author_id.in_(diff)),
+                    RecordsAuthors.id_type == "collaboration",
+                )
+                .all()
+            ]
+        return differed_papers_uuids
```
michamos, line 387: same question

backend/inspirehep/alembic/5a0e2405b624_add_citation_type_to_citation_table.py
```diff
@@ -0,0 +1,36 @@
...
+# revision identifiers, used by Alembic.
+revision = "5a0e2405b624"
+down_revision = "595c36d68964"
+branch_labels = ()
+depends_on = None
+
+
+def upgrade():
+    """Upgrade database."""
+    op.execute("CREATE TYPE enum_citation_type as ENUM ('citation', 'self_citation')")
```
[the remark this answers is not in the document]
pazembrz: As I remember there was a discussion that self-citation won't be only citation type, so there was a decision that this fiels should be a name of citation type. If it's not valid anymore, I will change it.

backend/inspirehep/records/api/mixins.py
```diff
@@ -103,12 +107,17 @@ def update(self, *args, **kwargs):
 
 
 class CitationMixin(PapersAuthorsExtensionMixin):
-    def _citation_query(self):
+    def _citation_query(self, exclude_self_citations=False):
         """Prepares query with all records which cited this one
         Returns:
             query: Query containing all citations for this record
```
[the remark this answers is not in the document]
pazembrz: :+1:

backend/inspirehep/records/api/mixins.py
```diff
@@ -203,17 +232,106 @@ def update_refs_in_citation_table(self, save_every=100):
...
+                        citation_type=RecordCitationType.citation,
                     )
                 )
             if len(references_waiting_for_commit) >= save_every:
                 db.session.bulk_save_objects(references_waiting_for_commit)
                 references_waiting_for_commit = []
         if references_waiting_for_commit:
             db.session.bulk_save_objects(references_waiting_for_commit)
+
+        _start = datetime.now()
```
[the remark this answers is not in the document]
pazembrz: Leftover from POC :+1:

backend/inspirehep/records/api/mixins.py
```diff
@@ -203,17 +232,106 @@ def update_refs_in_citation_table(self, save_every=100):
...
+        return collaboration_list
+
+    def get_self_cited_referenced_papers(self):
+        uuid = str(self.model.id)
+        sql_query = f"""
+            SELECT papers_from_authors.record_id
+            FROM
+                (SELECT DISTINCT a1.record_id FROM {RecordsAuthors.__tablename__} a1, {RecordsAuthors.__tablename__} a2
+                WHERE
+                    a2.record_id ='{uuid}'
```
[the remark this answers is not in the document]
pazembrz: You mean table names or UUID?
Table names are defined by us, user never have access to them so no chance that names will be replaced, UUID is also internal thing, verified by Postgres. it's not possible in normal circumstances that user will pass malicious string to perform SQL injection in there.
Maybe if there will be some bug which allows user to modify memory in python interpreter, but then we would have bigger problems than SQL injection...
Do i miss something?

backend/inspirehep/records/api/mixins.py
```diff
@@ -203,17 +232,106 @@ def update_refs_in_citation_table(self, save_every=100):
...
             uuid=str(self.id),
+            self_citations_update=(datetime.now() - _start).total_seconds(),
+        )
+
+    def get_authors_bais(self):
+        authors_list = []
+        for author in self.get_value("authors.ids", []):
+            for entry in author:
+                if entry.get("schema") == "INSPIRE BAI":
+                    authors_list.append(entry["value"])
```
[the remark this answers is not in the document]
pazembrz: We have so many helpers. I was not even aware that this one exists. Thanks :+1:

backend/inspirehep/records/api/mixins.py
```diff
@@ -203,17 +232,106 @@ def update_refs_in_citation_table(self, save_every=100):
...
+        return authors_list
+
+    def get_collaborations_values(self):
+        collaboration_list = []
+        for collaborations_values in self.get_value("collaborations.value", []):
+            for collaboration in collaborations_values:
+                collaboration_list.append(collaboration)
+        return collaboration_list
```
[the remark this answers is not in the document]
pazembrz: I want to extract list of lists to one flat list. I can do it better. You are right.

backend/inspirehep/records/api/mixins.py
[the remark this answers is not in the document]
pazembrz: Default in db is actually `citation`, but I prefer to specify it  here so it will be more readable.

backend/inspirehep/records/api/mixins.py
pazembrz: Hmm. This is how I understand this requirement:
```
 Change get modified references to return also for changed authors or collaborations.
```
But it might be too wide, you are right.

pazembrz: Ah, ok I get it.

pazembrz: :+1:

backend/inspirehep/records/api/mixins.py
pazembrz: :+1:

backend/inspirehep/records/api/mixins.py
pazembrz: Making this update only in db it will make code less readable. Now it's really fast and reliable.
Still isn't the limit of PostgreSQL query a 1GB or even more? To reach near this limit we should have a paper with at least 16M self-citations.

backend/inspirehep/records/api/mixins.py
```diff
@@ -203,17 +232,106 @@ def update_refs_in_citation_table(self, save_every=100):
...
+                        ),
+                    ),
+                    RecordCitations.citation_type == RecordCitationType.citation,
+                )
+            ).update(
+                {RecordCitations.citation_type: RecordCitationType.self_citation},
+                synchronize_session=False,
+            )
+        # update not-self_citations
+        RecordCitations.query.filter(
```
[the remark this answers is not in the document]
pazembrz: Actually there might be (ex) self-citations when current record is cited so this still is needed to update them to normal citations.

backend/inspirehep/records/api/mixins.py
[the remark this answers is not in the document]
pazembrz: Argh, you are right. It's simpler construct than in authors.ids, get_values is good enough.

backend/inspirehep/alembic/5a0e2405b624_add_citation_type_to_citation_table.py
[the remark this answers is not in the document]
michamos: I have no idea what else you had in mind, and if that's indeed the case, it's better to have multiple columns as flags for the different types, because otherwise you need to have something like an array if a given citation has multiple types.

backend/inspirehep/records/api/mixins.py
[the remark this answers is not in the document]
michamos: I mean UUIDs mostly. Currently there's probably no way a user can inject a user id, but
1. you still have to think about it and prove that's indeed the case
2. you don't know how things might evolve, and if it's safe now doesn't mean it will remain safe forever.

On the other hand, if you use `text`, you know you can't have SQL injections, so I really don't understand why we're not doing the correct thing and make this query obviously safe (following the principle of [defense in depth](https://en.wikipedia.org/wiki/Defense_in_depth_(computing)).

backend/inspirehep/records/api/mixins.py
michamos: then why do we have the batching in https://github.com/inspirehep/inspirehep/blob/e6004615e0171eefd05778c3e222ab73161b5241/backend/inspirehep/records/api/base.py#L88-L97 ?

backend/inspirehep/alembic/5a0e2405b624_add_citation_type_to_citation_table.py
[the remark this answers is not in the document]
pazembrz: :+1:

backend/inspirehep/records/api/mixins.py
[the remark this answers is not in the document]
pazembrz: :+1:

backend/inspirehep/records/api/mixins.py
pazembrz: Discussed IRL

backend/inspirehep/records/api/mixins.py
```diff
@@ -228,6 +343,56 @@ def update(self, data, disable_relations_update=False, *args, **kwargs):
...
+                    and_(RecordsAuthors.author_id.in_(diff)),
+                    RecordsAuthors.id_type == "collaboration",
+                )
+                .filter(
+                    RecordsAuthors.record_id.in_(
+                        self.get_self_cited_referenced_papers()
+                    )
+                )
+                .all()
+            ]
```
michamos, line 393: same comment

backend/inspirehep/search/facets.py
```diff
@@ -218,15 +218,25 @@ def records_hep(order=None):
...
-        if (doc.refereed.length >0 && doc.refereed[0]) {
-            state.citations_refereed.add(doc.citation_count[0])
-        } else {
-            state.citations_non_refereed.add(doc.citation_count[0])
-        }
-    """
+    if request.values.get("exclude-self-citations", False):
+        field = "citation_count_without_self_citations"
+    else:
+        field = "citation_count"
```
michamos, line 231:
```suggestion
    exclude_sef_citations = request.values.get("exclude-self-citations", False)
    field = "citation_count_without_self_citations" if exclude_self_citations else "citation_count"
```

backend/inspirehep/search/facets.py
```diff
@@ -271,7 +281,7 @@ def flattened_refereed = [];
                         "aggs": {
                             "citation_buckets": {
                                 "range": {
-                                    "field": "citation_count",
+                                    "field": field,
```
michamos, line 284: do we really want to change the name here? we're not changing it for the stats computed from it. If you want to communicate the info that self-citations are excluded, maybe add a meta field `excludes-self-citations: true/false`?

backend/inspirehep/search/facets.py
```diff
@@ -298,17 +308,24 @@ def citations_by_year():
...
-        def years = params._source.citations_by_year != null ? params._source.citations_by_year : [];
-        for (element in years) {
-            state.merge(element.year.toString(), element.count, (x, y) -> x + y)
-        }
-    """
+
+    if request.values.get("exclude-self-citations", False):
+        field = "citations_by_year_without_self_citations"
+    else:
+        field = "citations_by_year"
```
michamos, line 321: same comment as previously

backend/inspirehep/search/facets.py
```diff
@@ -322,7 +339,7 @@ def results=[:];
         "filters": {**filters},
         "filter": {"term": {"citeable": "true"}},
         "aggs": {
-            "citations_by_year": {
+            field: {
```
michamos, line 342: how will this work on the UI? are we displaying both on one graph (in which case it's probably better to get both in one request)? or are we doing a toggle on the graph?

backend/inspirehep/alembic/5a0e2405b624_add_citation_type_to_citation_table.py
```diff
@@ -0,0 +1,30 @@
...
+down_revision = "595c36d68964"
+branch_labels = ()
+depends_on = None
+
+
+def upgrade():
+    """Upgrade database."""
+    op.add_column(
+        "records_citations",
+        sa.Column("is_self_citation", sa.Boolean, nullable=True, default=False),
```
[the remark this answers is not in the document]
pazembrz: When we will Create this column it will be null at start. Setting default value in there won't help. To overcome this we would havge first to create migration with null=True then migrate ALL citations with some value, then apply another migration which will disallow null values.

It can be done in one migration with something like:
```python
op.add_column(
    "records_citations",
    sa.Column("is_self_citation", sa.Boolean, nullable=True, default=False)
)
op.execute("UPDATE records_citations SET is_self_citation = false")
op.alter_column('records_citations', 'is_self_citation', nullable=False)
```

But this migration would take some time and during this process continuous_migration, record updates, indexing, etc would be broken and won't work at all :/

backend/inspirehep/indexer/tasks.py
```diff
@@ -73,20 +73,21 @@ def process_references_for_record(record):
         LOGGER.info(
             f"Found {len(uuids)} references changed, indexing them", uuid=str(record.id)
         )
-        return batch_index(uuids)
+        return uuids
     LOGGER.info("No references changed", uuid=str(record.id))
+    return []
 
 
 def process_author_papers_if_author_changed_name(record):
```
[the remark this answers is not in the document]
pazembrz: :+1:

backend/inspirehep/indexer/tasks.py
```diff
@@ -55,15 +56,14 @@ def process_references_for_record(record):
     """Tries to find differences in record references.
```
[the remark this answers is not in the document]
pazembrz: :+1:

backend/inspirehep/alembic/5a0e2405b624_add_citation_type_to_citation_table.py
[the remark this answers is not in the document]
michamos: in the update functions you assume that's it's either set to `true` or `false` for existing record. So I believe nothing will be updated for those records.

backend/inspirehep/records/api/mixins.py
```diff
@@ -203,19 +238,99 @@ def update_refs_in_citation_table(self, save_every=100):
...
+                UNION
+                SELECT citer_id AS record_id
+                  from {RecordCitations.__tablename__}
+                where cited_id = :uuid) as cited_papers
+            WHERE
+            cited_papers.record_id=papers_from_authors.record_id
+        """
+        )
+        return [
+            str(row[0])
```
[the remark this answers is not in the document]
pazembrz: :+1:

backend/inspirehep/records/api/mixins.py
```diff
@@ -203,19 +238,99 @@ def update_refs_in_citation_table(self, save_every=100):
...
+        ]
+
+    def update_self_citations(self):
+        self_citations = self.get_self_cited_referenced_papers()
+        LOGGER.info(
+            "Self-citated_papers",
+            self_citations_count=len(self_citations),
+            recid=self.get("control_number"),
+        )
+        uuid = str(self.model.id)
```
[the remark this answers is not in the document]
pazembrz: :+1:

backend/inspirehep/records/api/mixins.py
```diff
@@ -228,6 +343,56 @@ def update(self, data, disable_relations_update=False, *args, **kwargs):
         else:
             self.update_refs_in_citation_table()
 
+    def get_all_connected_papers_of_modified_authors(self):
```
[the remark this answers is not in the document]
pazembrz: Ok, I moved this functions from indexer to Proper classes as they using private methods from record they should be part of proper class.

backend/inspirehep/records/api/mixins.py
```diff
@@ -228,6 +343,56 @@ def update(self, data, disable_relations_update=False, *args, **kwargs):
...
+        diff = list(current_authors.symmetric_difference(old_authors))
+        differed_papers_uuids = []
+        if diff:
+            differed_papers_uuids = [
+                result.record_id
+                for result in RecordsAuthors.query.filter(
+                    RecordsAuthors.record_id != self.id
+                )
+                .filter(
+                    and_(RecordsAuthors.author_id.in_(diff)),
```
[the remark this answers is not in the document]
pazembrz: :+1:

backend/inspirehep/records/api/mixins.py
```diff
@@ -203,19 +238,99 @@ def update_refs_in_citation_table(self, save_every=100):
...
+            FROM
+                (SELECT DISTINCT a1.record_id FROM {RecordsAuthors.__tablename__} a1, {RecordsAuthors.__tablename__} a2
+                WHERE
+                    a2.record_id = :uuid
+                    AND a1.author_id=a2.author_id
+                    AND (
+                        (a1.id_type='INSPIRE BAI' AND a2.id_type='INSPIRE BAI')
+                         OR (a1.id_type='collaboration' AND a2.id_type='collaboration')
+                    )
+                )AS papers_from_authors,
```
michamos, line 281:
```suggestion
                ) AS papers_from_authors,
```

backend/inspirehep/records/api/mixins.py
```diff
@@ -203,19 +238,99 @@ def update_refs_in_citation_table(self, save_every=100):
...
+                (SELECT cited_id as record_id FROM
+                  {RecordCitations.__tablename__} WHERE
+                citer_id = :uuid
+                UNION
+                SELECT citer_id AS record_id
+                  from {RecordCitations.__tablename__}
+                where cited_id = :uuid) as cited_papers
```
michamos, line 288:
```suggestion
                  FROM {RecordCitations.__tablename__}
                WHERE cited_id = :uuid) as cited_papers
```

backend/inspirehep/records/api/mixins.py
```diff
@@ -228,6 +343,56 @@ def update(self, data, disable_relations_update=False, *args, **kwargs):
...
+                    and_(RecordsAuthors.author_id.in_(diff)),
+                    RecordsAuthors.id_type == "INSPIRE BAI",
+                )
+                .filter(
+                    RecordsAuthors.record_id.in_(
+                        self.get_self_cited_referenced_papers()
+                    )
+                )
+                .all()
+            ]
```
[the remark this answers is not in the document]
pazembrz: :+1:

backend/inspirehep/records/api/mixins.py
pazembrz: :+1:

backend/inspirehep/search/facets.py
pazembrz: Sorry but what's the difference except that your code is a little less readable? ;)

michamos: That's why it says "Suggested change", it's a suggestion :)

backend/inspirehep/search/facets.py
pazembrz: We have to as citation_count without self citations is stored in this field, and we want to generate bucket aggregation for it isn't it? 
Why we want to add this meta gield? When user requests for citation count without self-citations he know what he requested already isn't he?

backend/inspirehep/search/facets.py
pazembrz: Same answer

backend/inspirehep/search/facets.py
pazembrz: We will have toggle. No one ever said about displaying both on one graph.

backend/inspirehep/records/api/mixins.py
pazembrz: :+1:

backend/inspirehep/records/api/mixins.py
pazembrz: :+1:

backend/inspirehep/alembic/5a0e2405b624_add_citation_type_to_citation_table.py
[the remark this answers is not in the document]
pazembrz: Until first migration will be done they will be just ignored. That's true. But it won't break anything and will still work properly with FEATURE_FLAG set to False

backend/inspirehep/search/facets.py
pazembrz: :+1:

backend/inspirehep/search/facets.py
```diff
@@ -283,13 +293,14 @@ def flattened_refereed = [];
...
-                            "citations_count": {"sum": {"field": "citation_count"}},
-                            "average_citations": {"avg": {"field": "citation_count"}},
+                            "citations_count": {"sum": {"field": field}},
+                            "average_citations": {"avg": {"field": field}},
                         },
                     },
                 },
             }
         },
+        "excludes-self-citations": "false" if field == "citation_count" else "false",
```
michamos, line 303: is this supposed to be in reaction to my suggestion? I don't think this is in the right place (plus it's always `"false"`), and anyway you're probably right that it's not really needed.

backend/tests/integration/alembic/test_alembic.py
```diff
@@ -200,9 +213,14 @@ def _get_table_names(db_alembic):
...
 
 def _get_custom_enums(db_alembic):
-    query = """ SELECT pg_type.typname AS enumtype,
-        pg_enum.enumlabel AS enumlabel
-        FROM pg_type
-        JOIN pg_enum
+    query = """ SELECT pg_type.typname AS enumtype,  
+        pg_enum.enumlabel AS enumlabel 
+        FROM pg_type  
+        JOIN pg_enum
```
michamos, line 219: you've added a bunch of trailing whitespace here.

backend/tests/integration/records/api/test_api_literature.py
```diff
@@ -1996,3 +1996,267 @@ def test_hard_deleting_record_removes_entries_in_authors_records_table(
         RecordsAuthors.query.filter_by(record_id=record.id).count()
         == expected_authors_entries_count
     )
+
+
+def test_self_citations_on_authors_callculated_on_record_creation(
```
michamos, line 2001:
```suggestion
def test_self_citations_on_authors_calculated_on_record_creation(
```

backend/tests/integration/records/api/test_api_literature.py
```diff
@@ -1996,3 +1996,267 @@ def test_hard_deleting_record_removes_entries_in_authors_records_table(
...
+    assert rec1.citation_count_without_self_citations == 1
+
+    assert rec2.citation_count == 1
+    assert rec2.citation_count_without_self_citations == 1
+
+    assert rec3.citation_count == 0
+    assert rec3.citation_count_without_self_citations == 0
+
+
+def test_mixed_self_citations_on_authors_callculated_on_record_creation(
```
michamos, line 2030:
```suggestion
def test_mixed_self_citations_on_authors_calculated_on_record_creation(
```

backend/tests/integration/records/api/test_api_literature.py
```diff
@@ -1996,3 +1996,267 @@ def test_hard_deleting_record_removes_entries_in_authors_records_table(
...
+    assert rec1.citation_count_without_self_citations == 1
+
+    assert rec2.citation_count == 1
+    assert rec2.citation_count_without_self_citations == 0
+
+    assert rec3.citation_count == 0
+    assert rec3.citation_count_without_self_citations == 0
+
+
+def test_self_citations_on_authors_callculated_on_record_update(
```
michamos, line 2071:
```suggestion
def test_self_citations_on_authors_calculated_on_record_update(
```

backend/tests/integration/records/api/test_api_literature.py
```diff
@@ -1996,3 +1996,267 @@ def test_hard_deleting_record_removes_entries_in_authors_records_table(
...
+    assert rec1.citation_count_without_self_citations == 0
+
+    assert rec2.citation_count == 1
+    assert rec2.citation_count_without_self_citations == 0
+
+    assert rec3.citation_count == 0
+    assert rec3.citation_count_without_self_citations == 0
+
+
+def test_self_citations_on_authors_callculated_on_other_record_update(
```
michamos, line 2113:
```suggestion
def test_self_citations_on_authors_calculated_on_other_record_update(
```

backend/tests/integration/records/api/test_api_literature.py
```diff
@@ -1996,3 +1996,267 @@ def test_hard_deleting_record_removes_entries_in_authors_records_table(
...
+    rec1.update(rec1_data)
+
+    assert rec1.citation_count == 1
+    assert rec1.citation_count_without_self_citations == 1
+
+    assert rec2.citation_count == 0
+    assert rec2.citation_count_without_self_citations == 0
+
+
+def test_self_citations_on_collaborations_callculated_on_record_creation(
```
michamos, line 2148:
```suggestion
def test_self_citations_on_collaborations_calculated_on_record_creation(
```

backend/tests/integration/records/api/test_api_literature.py
```diff
@@ -1996,3 +1996,267 @@ def test_hard_deleting_record_removes_entries_in_authors_records_table(
...
+    assert rec1.citation_count_without_self_citations == 1
+
+    assert rec2.citation_count == 1
+    assert rec2.citation_count_without_self_citations == 0
+
+    assert rec3.citation_count == 0
+    assert rec3.citation_count_without_self_citations == 0
+
+
+def test_self_citations_on_collaborations_callculated_on_record_update(
```
michamos, line 2202:
```suggestion
def test_self_citations_on_collaborations_calculated_on_record_update(
```

backend/tests/integration/records/api/test_api_literature.py
```diff
@@ -1996,3 +1996,267 @@ def test_hard_deleting_record_removes_entries_in_authors_records_table(
...
+    assert rec1.citation_count_without_self_citations == 0
+
+    assert rec2.citation_count == 1
+    assert rec2.citation_count_without_self_citations == 0
+
+    assert rec3.citation_count == 0
+    assert rec3.citation_count_without_self_citations == 0
+
+
+def test_self_citations_on_collaborations_callculated_on_other_record_update(
```
michamos, line 2237:
```suggestion
def test_self_citations_on_collaborations_calculated_on_other_record_update(
```

backend/inspirehep/search/facets.py
pazembrz: Argh, this is mistake. It was just a test and I didn't want to put it there. Sorry.

backend/tests/integration/alembic/test_alembic.py
pazembrz: It's black :/

backend/tests/integration/records/api/test_api_literature.py
pazembrz: :+1:

backend/tests/integration/alembic/test_alembic.py
pazembrz: Ok, it was my IDE, fixed.

backend/inspirehep/indexer/tasks.py
```diff
@@ -144,7 +86,19 @@ def index_record(self, uuid, record_version=None, force_delete=None):
...
+                record.get_all_connected_papers_of_modified_authors()
+            )
+            papers_to_reindex.extend(
+                record.get_all_connected_papers_of_modified_collaborations()
+            )
     if isinstance(record, AuthorsRecord):
-        process_author_papers_if_author_changed_name(record=record)
+        papers_to_reindex.extend(
+            record.get_linked_author_papers_if_author_changed_name()
+        )
```
michamos, line 102: I would suggest going all the way and encapsulating this at the record level, so you can do here something like:
```python
records_to_reindex = record.get_linked_records_to_reindex()
```
and have a dummy implementation on `InspireRecord` that always returns an empty list and override it on `AuthorsRecord` and `LiteratureRecord`. WDYT?

backend/inspirehep/records/api/mixins.py
```diff
@@ -203,19 +234,99 @@ def update_refs_in_citation_table(self, save_every=100):
...
+        )
+        return [
+            row.record_id
+            for row in db.session.execute(sql_query, {"uuid": uuid}).fetchall()
+        ]
+
+    def update_self_citations(self):
+        self_citations = self.get_self_cited_referenced_papers()
+        LOGGER.info(
+            "Self-citated_papers",
```
michamos, line 297:
```suggestion
            "Self-cited papers",
```

backend/inspirehep/search/facets.py
```diff
@@ -298,17 +308,19 @@ def citations_by_year():
...
-            state.merge(element.year.toString(), element.count, (x, y) -> x + y)
-        }
-    """
+
+    map_script = (
+        f"def years = params._source.citations_by_year != null ? params._source.citations_by_year : [];"
+        "for (element in years) {"
+        "    state.merge(element.year.toString(), element.count, (x, y) -> x + y)"
+        "}"
+    )
```
michamos, line 323: you can revert this part, now you don't need to interpolate stuff

backend/inspirehep/indexer/tasks.py
pazembrz: I prefer not to do this as there are already complains that we have to much logic inside InspireRecords opbjects.

pazembrz: But personally I agree that it's logical to move it there.

backend/inspirehep/search/facets.py
pazembrz: :+1:

backend/inspirehep/records/api/mixins.py
pazembrz: :+1:

PR #904, opened 2020-04-22
v1.0.9

drjova:


pazembrz:
Requires to run alembic upgrade (which will take around 5 minutes as it's creating new index on records_citations table)

PR #906, opened 2020-04-22
login: orcid login in new tab

miguelgrc:
* INSPIR-3349


backend/inspirehep/accounts/views.py
```diff
@@ -32,14 +32,12 @@ def me():
...
-    payload = get_current_user_data()
-    return render_template("accounts/postmessage.html", payload=payload)
+    return redirect(session["next_url"])
 
 
 @blueprint.route("/signup", methods=["GET"])
 def sign_up_required():
-    payload = {"user_needs_sign_up": True}
-    return render_template("accounts/postmessage.html", payload=payload)
+    return redirect(session["signup_url"])
```
harunurhan, line 40: we don't need this, since `signup_url` is static, we can just set them as `signup_redirect_url` in config, rather than passing first from the client and then redirecting here.

backend/inspirehep/accounts/views.py
```diff
@@ -57,8 +55,15 @@ def sign_up_user():
         return jsonify(data_current_user), 200
 
 
-@blueprint.route("/login", methods=["POST"])
+@blueprint.route("/login", methods=["GET"])
 def login():
+    session["next_url"] = request.args.get("next")
+    session["signup_url"] = request.args.get("signup_url")
+    return redirect("/api/oauth/login/orcid")
```
harunurhan, line 62: why do we need this new `view` ?

backend/inspirehep/accounts/views.py
miguelgrc: That URL points to the frontend's SignUpForm. We shouldn't set the URL here because we would be coupling with the frontend

backend/inspirehep/accounts/views.py
miguelgrc: Because we need to save `next_url` and `signup_url` so that we can do the redirection later. If we just pass them as query params to `oauthclient` they're ignored. `oauthclient` only reads `next_url` from config.

harunurhan: so if we didn't need `signup_url`, we wouldn't need this?

backend/inspirehep/accounts/views.py
harunurhan: I think this wouldn't be the kind of coupling that we care, at the moment at least.

backend/inspirehep/accounts/views.py
miguelgrc: We would still need it for the `next_url` since it's dynamic

harunurhan: I haven't checked deeply and `invenio-oauthclient`'s code is not known for its simplicity so I might be wrong but it shouldn't be like that.

harunurhan: https://github.com/inveniosoftware/invenio-oauthclient/blob/e7e4804260d4a4058414415fcb90d024a4dfa8ea/tests/test_views.py#L48

miguelgrc: If I do `/api/oauth/login/orcid?next=/jobs` and don't configure any `authorized_redirect_url` what I get after logging in is `/?message=Successfully+authorized.&code=200#show_login`.

backend/inspirehep/accounts/views.py
drjova: Actually that was one of my points when discussing with Miguel, starting adding UI endpoints to backend change more rule in multiple places (i.e. Traefik). It shouldn't be often but still I don't see any value hardcoding it.

backend/inspirehep/accounts/views.py
drjova: The issue here is that they don't handle the ``next`` url https://github.com/inveniosoftware/invenio-oauthclient/blob/4e1acca086cf894a1a0184e96a30c401754f7317/invenio_oauthclient/handlers/rest.py#L40 in the default handler, we can override it https://github.com/inveniosoftware/invenio-oauthclient/blob/4e1acca086cf894a1a0184e96a30c401754f7317/invenio_oauthclient/config.py#L105

harunurhan: Like it's in the refinement, no?

Ok so seems like there are 2 options:

either we create those dummy views to redirect around which seems extra but kinda easy to read what's going on.

or we just override the handler to use `next_url`.

Which do you think it's better? @drjova @miguelgrc

drjova: let's have the control, they might change handlers again

harunurhan: alright then, @miguelgrc you can go like you did initially.

miguelgrc: It should be ready like this then :)

ui/src/user/components/LoginPage/LoginPage.jsx
```diff
@@ -33,7 +34,12 @@ class LoginPage extends Component {
                 https://orcid.org/register
               </ExternalLink>
             </p>
-            <Button className="login-button h3" onClick={onLoginClick}>
+            <Button
+              className="login-button h3"
+              href={`/api/accounts/login?next=${encodeURIComponent(
+                previousUrl || '/'
```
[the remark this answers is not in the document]
miguelgrc: It shouldn't break anything if we do it, right?

ui/src/user/components/LoginPage/LoginPage.jsx
```diff
@@ -8,13 +8,17 @@ import './LoginPage.scss';
...
+import { USER_SIGNUP } from '../../../common/routes';
 
 const META_DESCRIPTION = 'Log in to your INSPIRE account. Log in with ORCID';
 const TITLE = 'Login';
 
 class LoginPage extends Component {
   render() {
-    const { onLoginClick } = this.props;
+    const { previousUrl } = this.props;
+    const loginHref = `/api/accounts/login?next=${encodeURIComponent(
```
harunurhan, line 19: `previousUrl` should already be uri friendly no?

miguelgrc: Yeah actually it's useless, I changed it.

drjova:
LGTM

PR #910, opened 2020-04-23
cli: overwrite Invenio reindex and remap with Inspire commands

pazembrz:
INSPIR-3432

### DIFF
--- a/backend/inspirehep/indexer/cli.py
+++ b/backend/inspirehep/indexer/cli.py
@@ -17,16 +17,12 @@ from flask.cli import with_appcontext
 from invenio_db import db
 from invenio_pidstore.models import PersistentIdentifier, PIDStatus
 from invenio_search import current_search
+from invenio_search.cli import index
 
 from inspirehep.indexer.tasks import batch_index
 from inspirehep.records.api import InspireRecord
 
 LOGGER = structlog.getLogger()
-
-
-@click.group()
-def indexer():
-    """Command group Inspire indexing & remap operations. (DO NOT USE ``index`` from Invenio)"""
 
 
 def next_batch(iterator, batch_size):
@@ -72,7 +68,7 @@ def _prepare_logdir(log_path):
         makedirs(path.dirname(log_path))
 
 
-@indexer.command("reindex")
+@index.command("reindex")
 @click.option("--all", is_flag=True, help="Reindex all the records.", show_default=True)
 @click.option(
     "-p",
@@ -122,7 +118,7 @@ def _prepare_logdir(log_path):
 def reindex_records(
     ctx, all, pidtype, pid, queue_name, batch_size, db_batch_size, log_path
 ):
-    """Reindex records in ElasticSearch.
+    """(Inspire) Reindex records in ElasticSearch.
 
     This command indexes the all the records related to the given PIDs in batches, asynchronously,
     by sending celery tasks to the specified queue. Indexing errors logged to file into the `log-path` folder.
@@ -246,9 +242,9 @@ def reindex_records(
             LOGGER.warning(failure)
 
 
-@indexer.command(
+@index.command(
     "remap",
-    help="Remaps specified indexes. Removes all data from index during this process.",
+    help="(Inspire) Remaps specified indexes. Removes all data from index during this process.",
 )
 @click.option("--yes-i-know", is_flag=True)
 @click.option(
--- a/backend/inspirehep/cli.py
+++ b/backend/inspirehep/cli.py
@@ -12,7 +12,7 @@ from invenio_base.app import create_cli
 from inspirehep.factory import create_app
 from inspirehep.files.cli import files
 from inspirehep.fixtures import fixtures
-from inspirehep.indexer.cli import indexer
+from inspirehep.indexer.cli import index
 from inspirehep.mailing.cli import mailing
 from inspirehep.orcid.cli import orcid
 from inspirehep.records.cli import citations, importer, jobs
@@ -28,4 +28,4 @@ cli.add_command(orcid)
 cli.add_command(mailing)
 cli.add_command(files)
 cli.add_command(sitemap)
-cli.add_command(indexer)
+cli.add_command(index)
--- a/backend/tests/integration/indexer/test_cli.py
+++ b/backend/tests/integration/indexer/test_cli.py
@@ -8,12 +8,11 @@
 import random
 
 from click.testing import CliRunner
-from helpers.providers.faker import faker
-from invenio_search import current_search, current_search_client
+from invenio_search import current_search
 from invenio_search.utils import build_index_name
 from mock import patch
 
-from inspirehep.indexer.cli import indexer
+from inspirehep.cli import cli
 from inspirehep.search.api import (
     AuthorsSearch,
     ConferencesSearch,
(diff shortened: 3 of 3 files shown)


backend/inspirehep/indexer/cli.py
```diff
@@ -24,6 +24,11 @@
 LOGGER = structlog.getLogger()
 
 
+@click.group()
+def index():
+    """*INACTIVE* Deprecated command from INVENIO. *INACTIVE*"""
```
michamos, line 29: what about all the other index management commands, like `init`, `delete`, etc.?

pazembrz: index is whole group all commands like init delete etc are inside it. They won't be available at all when overwriting this group.

pazembrz:
```bash
[pazembrz@laptop backend]$ poetry run inspirehep --help
Usage: inspirehep [OPTIONS] COMMAND [ARGS]...

  Command Line Interface for Invenio.

Options:
  --version  Show the flask version
  --help     Show this message and exit.

Commands:
  access     Account commands.
  alembic    Perform database migrations.
  citations  Command for citations
  db         Database commands.
  files      Command group to handle file operations.
  fixtures   Command related to records in inspire
  importer   Command to import records.
  index      *INACTIVE* Deprecated command from INVENIO.
  indexer    Command group Inspire indexing & remap...
  instance   Instance commands.
  jobs       Command for jobs
  mailing    Command to handle mailing.
  migrate    Commands to migrate records from legacy.
  orcid      Commands to run ORCID tasks.
  pid        PID-Store management commands.
  records    Records management.
  roles      Role commands.
  routes     Show the routes for the app.
  run        Run a development server.
  shell      Runs a shell in the app context.
  sitemap    Command group to handle sitemap operations.
  tokens     OAuth2 server token commands.
  users      User commands.
  webpack    Webpack commands.
```

```bash
[pazembrz@laptop backend]$ poetry run inspirehep index --help
Usage: inspirehep index [OPTIONS] COMMAND [ARGS]...

  *INACTIVE* Deprecated command from INVENIO. *INACTIVE*

Options:
  --help  Show this message and exit.
```

drjova: I'd not deprecate the whole group unless we move all these commands to our ``indexer`` module.

pazembrz: :+1:

backend/inspirehep/indexer/cli.py
```diff
@@ -17,13 +17,22 @@
...
 
 LOGGER = structlog.getLogger()
 
 
+@index.command("reindex")
+def reindex_overwrite(*args, **kwargs):
+    """WARNING: Invenio reindex command is outdated. Use `indexer reindex` command instead."""
+    click.echo(
+        "*INACTIVE* Invenio reindex command is outdated. Use `indexer reindex` command instead. *INACTIVE*"
+    )
```
michamos, line 33: what's the point of doing this, why don't we simply define this to be our own `reindex`, instead of having a different group for that?

drjova:
Actually I just checked the changes, this solution will not work. It is an entry point  ordering issue, if we don't resolve the  entrypoint in the right order the ``invenio_indexer.cli`` will never be overriden by ``inspirehep.indexer.cli``. By solving the loading order then we just override ``invenio_indexer`` as @michamos suggesting. We'll probably need to override the loader and make sure our entrypoints are loaded after the invenio https://github.com/inveniosoftware/invenio-base/blob/master/invenio_base/app.py#L169

backend/inspirehep/indexer/cli.py
pazembrz: This task is for blocking invenio reindex command. Our own reindex in in indexer reindex. I think that it's done like that to easily distinguish our indexer commands from invenio ones but I wasn't one who separated them and I don't recall when it happened. Please discuss things like that on Jira tasks, not during reviews, as it's a little too late already ;)

michamos: It was done like this with a separate command as a workaround for the invenio indexer. I agree that it should have been raised during the refinement, but I didn't take part in it.

pazembrz:
As it was mentioned in refinement already, when we are overwriting index group in inspirehep.cli it's already AFTER app is initialized so loading order doesn't matter here anymore.

backend/inspirehep/indexer/cli.py
drjova: > As it was mentioned in refinement already, when we are overwriting index group in inspirehep.cli it's already AFTER app is initialized so loading order doesn't matter here anymore.

Actually, this is exactly how it was before, we're overriding the ``reindex`` command but it didn't work https://github.com/inspirehep/inspirehep/blob/1defd3f6849dba6ae43253af2cb1bd64c7ac98ce/backend/inspirehep/indexer/cli.py#L71 It was the actual reason why we moved the commands in a sepate group (bad solution),  the  ``reindex`` command was pointing to ``invenio_inexer.cli`` instead of ``inspirehep.indexer.cli`` due to wrong entrypoints loading.

We have to solve the order issue and just add/override the ``invenio_indexer.cli``,  to have a unified cli. As we have discussed during the refinement the main problem is the entrypoint loading order. We have to just fix the order, add the ``remap`` command and override the ``reindex`` command , which was acutally a very valid point @pazembrz raised during the refiment. 

Also even if it's in the refinement, if we find during the reivew that something is not optimal or wrong we don't have to merge it. We can simply adapt or close the PR.

pazembrz: But we never put this overwritten index group in inspirehep.cli, this is why it was depending form load order.

drjova: I'm not sure I understand, maybe I'm missing something. The code seems exactly the same as before https://github.com/inspirehep/inspirehep/blob/1defd3f6849dba6ae43253af2cb1bd64c7ac98ce/backend/inspirehep/indexer/cli.py#L71

pazembrz: please check inspirehep/cli.py

backend/inspirehep/indexer/cli.py
```diff
@@ -122,7 +118,7 @@ def _prepare_logdir(log_path):
 def reindex_records(
     ctx, all, pidtype, pid, queue_name, batch_size, db_batch_size, log_path
 ):
-    """Reindex records in ElasticSearch.
+    """(Inspire) Reindex records in ElasticSearch.
```
drjova, line 121: unecessary :)

backend/inspirehep/indexer/cli.py
```diff
@@ -72,7 +68,7 @@ def _prepare_logdir(log_path):
         makedirs(path.dirname(log_path))
 
 
-@indexer.command("reindex")
+@index.command("reindex")
```
drjova, line 71: maybe add a help to indicates that this is the inspire one

pazembrz: This is why I added this (Inspire) Thing so it will be easy distinguishable that it's inspire one ;) - First sentence from doc-string is displayed as a description for the command.
```bash
[pazembrz@laptop backend]$ poetry run inspirehep index --help
Usage: inspirehep index [OPTIONS] COMMAND [ARGS]...

  Manage search indices.

Options:
  --help  Show this message and exit.

Commands:
  check    Check Elasticsearch version.
  create   Create a new index.
  delete   Delete index by its name.
  destroy  Destroy all indexes.
  init     Initialize registered aliases and mappings.
  list     List indices.
  put      Index input data.
  queue    Manage indexing queue.
  reindex  (Inspire) Reindex records in ElasticSearch.
  remap    (Inspire) Remaps specified indexes.
  run      Run bulk record indexing.
```

pazembrz: @drjova

backend/inspirehep/indexer/cli.py
pazembrz: Please read comment below.

PR #911, opened 2020-04-24
travis: push commits to prod sentry

michamos:
This pushes commit info to prod sentry and allows it to trace new issues back to recent commits, see for example https://sentry.inspirehep.net/inspire-prod/hep/issues/18917/?query=is%3Aunresolved (I did the procedure manually to test that it works correctly).


build-and-deploy.sh
```diff
@@ -105,9 +111,13 @@ main() {
...
-  deployQA "hep"
-  sentryQA
-  maybeDeploySmokeTestsQA
+  if [ -z "${TRAVIS_TAG}" ]; then
+    deployQA "ui"
+    deployQA "hep"
+    sentryQA
+    maybeDeploySmokeTestsQA
+  else
+    sentryPROD
```
michamos, line 120: question @benjamin-bergia @drjova wouldn't it be convenient to also change the tags in the prod environment automatically through a trigger? given we're using a prod branch anyway for the kubernetes repo, there is no danger of things getting auto-deployed to prod and would simplify things as you don't have to actively watch the image creation anymore.

drjova: That's a good point, actually. 👍

michamos: I added a commit to do that @drjova

PR #913, opened 2020-04-24
orcid: account for async push of clashing identifiers

MJedr:
* INSPIR-3469


backend/tests/integration/orcid/test_orcid_domain_models.py
```diff
@@ -467,21 +468,39 @@ def test_happy_flow_put(self):
...
 
     def test_duplicated_external_identifier_pusher_exception(self):
-        self.factory_clashing.record_metadata.json["titles"] = [
-            {"source": "submitter", "title": "title1"}
-        ]
         with override_config(
             FEATURE_FLAG_ENABLE_ORCID_PUSH=True,
             FEATURE_FLAG_ORCID_PUSH_WHITELIST_REGEX=".*",
             ORCID_APP_CREDENTIALS={"consumer_key": "0000-0001-8607-8906"},
         ), pytest.raises(exceptions.DuplicatedExternalIdentifierPusherException):
```
michamos, line 493: can you move the `pytest.raises` to the place it actually should raise? so we make sure it doesn't raise too early but exactly where it's expected.

backend/tests/integration/orcid/test_orcid_domain_models.py
```diff
@@ -467,21 +468,39 @@ def test_happy_flow_put(self):
...
-            result_putcode = pusher.push()
-
-        assert self.putcode == result_putcode
-        assert not self.cache.has_work_content_changed(self.inspire_record)
+            with mock.patch(
+                "inspirehep.orcid.tasks.orcid_push.retry"
+            ) as celery_task_with_retry:
+                celery_task_with_retry.side_effect = (
+                    Retry(exc=exceptions.DuplicatedExternalIdentifierPusherException),
+                )
```
michamos, line 476: why is this needed? because of eager mode?

backend/inspirehep/orcid/domain_models.py
```diff
@@ -153,10 +153,9 @@ def push(self):
             # of the clashing work in ORCID and push a fresh version of that
             # record.
             # This scenario might be triggered by a merge of 2 records in Inspire.
-            if self.pushing_duplicated_identifier:
-                raise exceptions.DuplicatedExternalIdentifierPusherException
-            self._push_work_with_clashing_identifier()
-            putcode = self._post_or_put_work(putcode)
+            if not self.pushing_duplicated_identifier:
+                self._push_work_with_clashing_identifier()
+            raise exceptions.DuplicatedExternalIdentifierPusherException
```
michamos, line 158: maybe add a comment here to say this causes it to be retried in the celery task?

backend/tests/integration/orcid/test_orcid_domain_models.py
MJedr: Yes, it's same as here https://github.com/inspirehep/inspirehep/blob/642836ac47b99bab9bc578b382c10ca6892ceae2/backend/tests/integration/orcid/test_orcid_domain_models.py#L448-L460

michamos: it's not the same. There you want to simulate exceeding the time limit. There's no such simulation happening here, this should be the standard behavior of the celery task. Did you try without the mocking?

backend/tests/integration/orcid/cassettes/TestOrcidPusherDuplicatedIdentifier.test_duplicated_external_identifier_pusher_exception.yaml
```diff
@@ -299,44 +474,42 @@ interactions:
...
     uri: https://api.orcid.org/v2.0/0000-0003-1134-6827/work
   response:
     body:
-      string: '{"response-code":409,"developer-message":"409 Conflict: You have already
-        added this activity (matched by external identifiers), please see element
-        with put-code 71115035. If you are trying to edit the item, please use PUT
-        instead of POST.","user-message":"There was an error when updating the record.
-        Please try again. If the error persists, please contact INSPIRE-HEP for assistance.","error-code":9021,"more-info":"https://members.orcid.org/api/resources/troubleshooting"}'
+      string: "{\n  \"error\" : \"invalid_token\",\n  \"error_description\" : \"Invalid\
+        \ access token: mytoken\"\n}"
```
michamos, line 485: this seems incorrect.

backend/tests/integration/orcid/test_orcid_domain_models.py
```diff
@@ -467,26 +468,39 @@ def test_happy_flow_put(self):
...
+            with pytest.raises(exceptions.DuplicatedExternalIdentifierPusherException):
+                tasks.orcid_push.apply_async(
+                    queue="orcid_push_legacy_tokens",
+                    kwargs={
+                        "orcid": self.orcid,
+                        "rec_id": self.recid,
+                        "oauth_token": self.oauth_token,
+                    },
+                    throw=True,
+                )
```
michamos, line 480: what about calling the push again here manually, so the happy flow succeeds (with a comment that celery will do that)? then you don't have to modify the next test.

PR #919, opened 2020-04-28
ui: add citation_* meta tags for literature

harunurhan:
(Used by Google Scholar)

INSPIR-3479

### DIFF
--- a/ui/src/literature/components/LiteratureDocumentHead.jsx
+++ b/ui/src/literature/components/LiteratureDocumentHead.jsx
@@ -1,0 +1,113 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import { Map } from 'immutable';
+import moment from 'moment';
+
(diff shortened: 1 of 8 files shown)


ui/src/literature/components/LiteratureDocumentHead.jsx
```diff
@@ -0,0 +1,110 @@
...
+      <meta name="citation_title" content={title} />
+      <meta name="citation_online_date" content={onlineDate} />
+
+      {abstract && <meta name="citation_abstract" content={abstract} />}
+
+      {date && (
+        <meta name="citation_publication_date" content={publicationDate} />
+      )}
+      {arxiv && (
+        <meta name="citation_technical_report_number" content={arxiv} />
```
michamos, line 72: I would put an `arXiv` in front, so `arXiv:2004.012345`

PR #920, opened 2020-04-28
Test for INSPIR-3425

pazembrz:
For now only inspirehep/records/api is changed. Please check if it is ok for everyone.

### DIFF
--- a/backend/tests/integration/conftest.py
+++ b/backend/tests/integration/conftest.py
@@ -23,6 +23,7 @@ from helpers.factories.models.records import RecordMetadataFactory
 from helpers.utils import get_test_redis, override_config
 from moto import mock_s3
 
+from inspirehep.cli import cli
 from inspirehep.factory import create_app as inspire_create_app
 from inspirehep.files.api.s3 import S3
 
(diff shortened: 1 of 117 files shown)


pazembrz:
Ignore failing tests for now, They will fail as I made many changes in fixtures and only `inspirehep/records/api` was updated.

backend/tests/integration/conftest.py
```diff
@@ -345,9 +234,5 @@ def app_cli_runner(appctx):
 
 
 @pytest.fixture(scope="function")
-def redis(base_app):
-    redis_url = current_app.config.get("CACHE_REDIS_URL")
-    r = StrictRedis.from_url(redis_url, decode_responses=True)
-    r.flushall()
-
-    yield r
+def app_clean(base_app, db, es_clear, vcr_config):
```
drjova, line 237: I thought we ``es_clear`` was slow and we'll not add it to the tests, does it slow down the tests a lot?

pazembrz: At the end it added 5 seconds for all 200 changed tests from ~65 seconds to ~70 seconds on my machine 
By average it's +0.025s per test

drjova: that's cool, great work! can we test if for all tests? also if we use at the end only ``es_clear`` I'd call the fixture just ``app``

pazembrz: To make it work with all the tests I have to go through all tests. You can notice that many fixtures like for example `create_record` wont' be a fixtures anymore. So I have to go every test, change fixtures and import all required ex-fixtures from test utils ;) This will take some time and this is why I asked for your comments as I don't want to go through all tests more than once if something should be changed or done differently in your opinion.

PR #923, opened 2020-04-28
self-citations: quick update for self-citations

pazembrz:
*Added indexes for is_self_citation column in records_citations table
 *Fix: When feature flag for self-citations is disabled, ignore self-citations queries to speedup migration.
 *Removed unnecessary flag from _citations_by_year method

### DIFF
--- a/backend/inspirehep/alembic/8ba47044154a_add_indexed_to_records_citations_on_is_.py
+++ b/backend/inspirehep/alembic/8ba47044154a_add_indexed_to_records_citations_on_is_.py
@@ -1,0 +1,41 @@
+# -*- coding: utf-8 -*-
+#
+# Copyright (C) 2019 CERN.
+#
+# inspirehep is free software; you can redistribute it and/or modify it under
+# the terms of the MIT License; see LICENSE file for more details.
+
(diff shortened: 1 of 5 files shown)


backend/inspirehep/alembic/8ba47044154a_add_indexed_to_records_citations_on_is_.py
```diff
@@ -0,0 +1,41 @@
...
+        "records_citations",
+        ["cited_id", "is_self_citation"],
+        unique=False,
+    )
+    op.create_index(
+        "ix_records_citations_citer_id_citation_type",
+        "records_citations",
+        ["citer_id", "is_self_citation"],
+        unique=False,
+    )
```
michamos, line 31: don't you want to extend the existing indexes to add `is_self_citation` instead?

pazembrz: Currently we have two indexes, with citer_id, cited_id and opposite one. They are quite big(2G). With our infrastructure additional small indexes would be more efficient than extending existing ones.

PR #924, opened 2020-04-28
facets: fix citation_summary facet

pazembrz:
*FIX: citation_summary is not reading 'exclude-self-citations' parameter properly
*mappings: Remove unnecessary mapping for citations_by_year_without_self_citations

### DIFF
--- a/backend/inspirehep/search/facets.py
+++ b/backend/inspirehep/search/facets.py
@@ -4,7 +4,6 @@
 #
 # inspirehep is free software; you can redistribute it and/or modify it under
 # the terms of the MIT License; see LICENSE file for more details.
-
 from datetime import datetime
 from itertools import count
 
@@ -225,7 +224,7 @@ def citation_summary():
         "citation_count_without_self_citations",
     ]
     filters = get_filters_without_excluded(hep_filters(), excluded_filters)
-    if request.values.get("exclude-self-citations", False):
+    if "exclude-self-citations" in request.values:
         field = "citation_count_without_self_citations"
     else:
         field = "citation_count"
--- a/backend/inspirehep/search/mappings/v7/records/hep.json
+++ b/backend/inspirehep/search/mappings/v7/records/hep.json
@@ -585,17 +585,6 @@
         },
         "type": "object"
       },
-      "citations_by_year_without_self_citations": {
-        "properties": {
-          "year": {
-            "type": "integer"
-          },
-          "count": {
-            "type": "integer"
-          }
-        },
-        "type": "object"
-      },
       "citeable": {
         "type": "boolean"
       },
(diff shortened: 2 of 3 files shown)


backend/inspirehep/search/facets.py
```diff
@@ -225,7 +225,7 @@ def citation_summary():
         "citation_count_without_self_citations",
     ]
     filters = get_filters_without_excluded(hep_filters(), excluded_filters)
-    if request.values.get("exclude-self-citations", False):
+    if json.loads(request.values.get("exclude-self-citations", "false")):
```
drjova, line 228: what will happen if ``exclude-self-citations=+/1``? Can we use somehting like this https://github.com/inspirehep/inspirehep/blob/master/backend/inspirehep/search/views.py#L20 which we specify the type?

pazembrz: By default all parameters in request.values are string already. 1/0 will work as True/False as it will be parsed by json.load as int. `+` would not work. It will return `500` as loading json will just raise

pazembrz: Ah ok , I get what you meant. You are right. It's better this way. :+1:

backend/inspirehep/search/facets.py
```diff
@@ -225,7 +225,7 @@ def citation_summary():
         "citation_count_without_self_citations",
     ]
     filters = get_filters_without_excluded(hep_filters(), excluded_filters)
-    if request.values.get("exclude-self-citations", False):
+    if request.values.get("exclude-self-citations", False, type=bool):
```
michamos, line 228: `bool("false") == True`. Why not simply
```suggestion
    if "exclude-self-citations" in request.values:
```
then it doesn't matter what you set it to: if it's present, it's interpreted as `True` whatever it's set to, otherwise as `False`.

pazembrz: IMHO it makes code less readable.

pazembrz: IDK what others thinks?

michamos: It's not about readability but about behavior. I think it's surprising that `/facets?exclude-self-citations=` doesn't exclude self-citations, but `/facets?exclude-self-citations=false` does.

pazembrz: but `/facets?exclude-self-citations=false` also won't exclude self-citations.

pazembrz: No, you are right. Now any value in the parameter is True :/ So it's not what I wanted, I will go back to json.load then or I will go with your suggestion.

pazembrz: :+1:

backend/inspirehep/search/facets.py
pazembrz: :-1:

backend/inspirehep/search/facets.py
drjova: nice idea, IMHO this is not how you standarize an api and it's more towards legacy; send a value and we'll guess what to do with it. ``true`` & ``false`` are universal and objective, they can be validated and documented easily with standard tools.

PR #927, opened 2020-04-30
setup: upgrade lxml

MJedr:
* upgrade orcid tests which were failing because of switch to newer version (different canonicalisation method was used since 4.4.)
* makes inspire compatibile with inspire-utils
* INSPIR-3477

### DIFF
--- a/backend/pyproject.toml
+++ b/backend/pyproject.toml
@@ -51,8 +51,7 @@ pybtex = "*"
 cryptography = "<2.6"
 flask-kvsession = {git = "https://github.com/inspirehep/flask-kvsession.git"}
 python-redis-lock = "^3.3"
-# FIXME: orcid tests are failing with `4.4` because of OrderedDict changes
-lxml = "~=4.3,<4.4"
+lxml = "~=4.4"
 fqn-decorators = "~=1.2,>=1.2.3"
 editdistance = "^0.5.3"
 munkres = "^1.0"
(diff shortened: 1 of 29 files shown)


backend/pyproject.toml
```diff
@@ -52,7 +52,7 @@ cryptography = "<2.6"
 flask-kvsession = {git = "https://github.com/inspirehep/flask-kvsession.git"}
 python-redis-lock = "^3.3"
 # FIXME: orcid tests are failing with `4.4` because of OrderedDict changes
-lxml = "~=4.3,<4.4"
+lxml = "4.4"
```
pazembrz, line 55: Is 4.5 incompatible?
```suggestion
lxml ~= "4.4"
```
So we could at least update to latest fix on 4.4. branch.

MJedr: Ok, I think that it should (there are no significant changes in 4.5 comparing to 4.4, just some bug fixes and one new feature)

backend/tests/integration/orcid/cassettes/TestOrcidPusherPutUpdatedWork.test_push_updated_work_invalid_data_putcode.yaml
```diff
@@ -53,118 +55,6 @@ interactions:
...
-      - nginx/1.16.1
-      X-Content-Type-Options:
-      - nosniff
-      X-Frame-Options:
-      - DENY
-      X-XSS-Protection:
-      - 1; mode=block
-    status:
-      code: 409
-      message: Conflict
```
michamos, line 167: what happened here?

MJedr: To me the new one looks correctly:

https://github.com/inspirehep/inspirehep/blob/6ce3689b5c64f02bec3aa91ff84a70f77a1a2795/backend/inspirehep/orcid/domain_models.py#L160-L173

There is a sequence:
- _post_or_put_work() (PUT) --> PutcodeNotFoundPutException 404
- _cache_all_author_putcodes() (GET) --> 200
- _post_or_put_work() (PUT) --> 201.

PR #932, opened 2020-05-04
ui: exclude self-citations checkbox

miguelgrc:
* INSPIR-3263


ui/src/literature/containers/CitationSummaryBoxContainer.jsx
```diff
@@ -2,14 +2,25 @@ import { connect } from 'react-redux';
...
+  excludeSelfCitations: state.ui.get('excludeSelfCitations'),
 });
 
 const dispatchToProps = dispatch => ({
-  onQueryChange(query) {
-    dispatch(fetchCitationSummary(query.toJS()));
+  onQueryChange(query, excludeSelfCitations) {
+    if (excludeSelfCitations) {
+      dispatch(
+        fetchCitationSummary(query.set('exclude-self-citations', true).toJS())
```
harunurhan, line 16: why not make `exclude-self-citations` part of the `query`. And maybe even going further it can be part of the search query in state, so you don't have to keep it on `ui` state? would that break anything?

miguelgrc: You mean basically dispatching `searchQueryUpdate instead of `changeExcludeSelfCitations`? It wouldn't work because we only want to refresh the citation summary, and not the search results.

harunurhan: that's unfortunate.. the refactor that I have been working on could have fixed that too :/

PR #934, opened 2020-05-04
institutions: enable ui tests and remove beta ribbon

monaawi:
* INSPIR-3507


ui/src/common/routes.js
```diff
@@ -44,5 +44,5 @@ export const EDIT_CONFERENCE = '/editor/record/conferences';
 export const EDIT_INSTITUTION = '/editor/record/institutions';
 
 export function isBetaRoute(locationPathname) {
-  return locationPathname.startsWith(INSTITUTIONS);
+  return locationPathname.startsWith(SEMINARS);
```
monaawi, line 47: I put seminars here because I would have had to add it soon anyhow

PR #936, opened 2020-05-05
authors: fix when indexing author which name didn't changed

pazembrz:


monaawi:
I just had a look on how you refactored the ```process_references_for_papers``` and I don't  like the approach. Why did you separate the modified authors/modified collaboration from the rest? Is it because the rest mostly use "linked" institutions/conferences etc.? The naming for that is very confusing ```get_linked_papers_if_reference_changed```, they are not even papers! Some are returning institutions to reindex. It's messy.
I would have a function ```get_records_to_reindex_for_lit_record``` or something to signify that it doesn't return just papers and reference is changed is too vague. And in that function, have all the calls we use to retrieve the records which need to be reindexed. Then you could test this by calling that function and checking it return empty list instead of calling indexing with delay False and seeing it doesn't throw exception. And then you can have another function for author records (to get records to reindex)
We also need to rename the variables ```papers_to_reindex``` is wrong. Before it was called ```uuids``` which made sense more or less.

pazembrz:
> Why did you separate the modified authors/modified collaboration from the rest?
I didn't separated them. They are new.

> The naming for that is very confusing `get_linked_papers_if_reference_changed`, they are not even papers! 
Ok, I'm not best with names. I will change this.

> I would have a function `get_records_to_reindex_for_lit_record` or something to signify that it doesn't return just papers and reference is changed is too vague. And in that function, have all the calls we use to retrieve the records which need to be reindexed.
I'm not sure about this. Where you would preffer to put this? In Lit record? Why lit record should care about this? Maybe in reindex? But then we would create new function just to call 3 methods from Literature Record?

And what with author record? Should it stay unchanged? Then our records API would be inconsistent. So Maybe we should create method `get_records_uuids_to_reindex` for all Classes which would return proper uuids  But i believe that this would be useful if this would be more complex, as creating this to replace 2 ifs in indexer is a little to much right now I believe. 

> We also need to rename the variables `papers_to_reindex` is wrong. Before it was called `uuids` which made sense more or less.
Agree

monaawi:
> > Why did you separate the modified authors/modified collaboration from the rest?
> 
> I didn't separated them. They are new.

Well, they still should be with the other function calls 

> I would have a function `get_records_to_reindex_for_lit_record` or something to signify that it doesn't return just papers and reference is changed is too vague. And in that function, have all the calls we use to retrieve the records which need to be reindexed.

It's not just 3 function calls, it's 5 (the other 2 should be here too). I think it makes sense to be a separate function as the code will be more readible and you would be able to test the functionality of reindexing records based on this lit record in an easier way.

> And what with author record? Should it stay unchanged? Then our records API would be inconsistent. So Maybe we should create method `get_records_uuids_to_reindex` for all Classes which would return proper uuids But i believe that this would be useful if this would be more complex, as creating this to replace 2 ifs in indexer is a little to much right now I believe.

Why is it inconsistent? You just have a function that is giving you uuids to reindex. You can even do one function and put the ifs in there as well and just call it. I just think it would be good to have it separated and not all the logic inside the indexing task so it can be tested in a better way and modified collaborations and authors should go with the other as well. Right now, we have a function that gets us half of the records to be indexed. Then we call another 2 functions to get the rest inside the indexing task which is messy.

pazembrz:
Discussed IRL with @monaawi . I created new task to move logic out of index_record task. https://its.cern.ch/jira/browse/INSPIR-3518

PR #938, opened 2020-05-05
login: fix orcid login session issues

miguelgrc:
* Fix for INSPIR-3349

The problem is that the session is reset on rare occasions probably due to some issues with Redis, so in that situation a KeyError was being thrown when reading `next_url` and `signup_url` from session.

### DIFF
--- a/backend/inspirehep/accounts/views.py
+++ b/backend/inspirehep/accounts/views.py
@@ -32,12 +32,7 @@ def me():
 
 @blueprint.route("/login-success")
 def login_success():
-    return redirect(session["next_url"])
-
-
-@blueprint.route("/signup", methods=["GET"])
-def sign_up_required():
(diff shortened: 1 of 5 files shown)


backend/inspirehep/accounts/views.py
```diff
@@ -32,12 +32,13 @@ def me():
...
 def login_success():
-    return redirect(session["next_url"])
+    return redirect(session.get("next_url", "/"))
 
 
 @blueprint.route("/signup", methods=["GET"])
 def sign_up_required():
-    return redirect(session["signup_url"])
+    # UI signup form url
+    return redirect("/user/signup")
```
harunurhan, line 41: at the end, you can remove this view (`/signup`) and add this (`/user/signup`) directly to the config.

PR #945, opened 2020-05-11
seminars: create tickets for submissions

harunurhan:
* Also sends confirmation email to the user when a new seminar is submitted.

### DIFF
--- a/backend/inspirehep/mailing/api/seminars.py
+++ b/backend/inspirehep/mailing/api/seminars.py
@@ -1,0 +1,43 @@
+# -*- coding: utf-8 -*-
+#
+# Copyright (C) 2019 CERN.
+#
+# inspirehep is free software; you can redistribute it and/or modify it under
+# the terms of the MIT License; see LICENSE file for more details.
+
+import structlog
+from flask import current_app, render_template
+
+from inspirehep.mailing.providers.flask_mail import send_email
+
+LOGGER = structlog.getLogger()
+
+
+def get_submissions_confirmation_email_html_content(seminar):
+    host = current_app.config["SERVER_NAME"]
+    recid = seminar["control_number"]
+    seminar_url = f"{host}/seminars/{recid}"
+    seminar_edit_url = f"{host}/submissions/seminars/{recid}"
+    return render_template(
+        "mailing/seminars/confirmation_new.html",
+        seminar_url=seminar_url,
+        seminar_edit_url=seminar_edit_url,
+    )
+
+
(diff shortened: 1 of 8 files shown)


backend/inspirehep/mailing/config.py
```diff
@@ -15,6 +15,8 @@
 CONFERENCES_CONFIRMATION_EMAIL_ADDRESS = "conf@inspirehep.net"
 """Email address for sending confirmation email for conferences"""
 
+SEMINARS_CONFIRMATION_EMAIL_ADDRESS = "seminars@inspirehep.net"
```
drjova, line 18: do we even have this kind of emails?

backend/inspirehep/mailing/api/seminars.py
```diff
@@ -0,0 +1,43 @@
...
+import structlog
+from flask import current_app, render_template
+
+from inspirehep.mailing.providers.flask_mail import send_email
+
+LOGGER = structlog.getLogger()
+
+
+def get_submissions_confirmation_email_html_content(seminar):
+    host = current_app.config["SERVER_NAME"]
```
drjova, line 17:
```suggestion
    host = get_inspirehep_url()
```

harunurhan: full url looks ugly in the email, isn't `https` by default anyhow?

backend/inspirehep/mailing/config.py
harunurhan: I am not sure, that was the requirement. but it will work either way right?

StellaCh: see last messages on https://inspirehep.zulipchat.com/#narrow/stream/195298-experts/topic/virtual.20seminars.20in.20INSPIRE

backend/inspirehep/mailing/api/seminars.py
drjova: it's using https https://github.com/inspirehep/inspirehep/blob/b6aa07abab5504eb5744ef74d161a1f5c1e2f208/backend/inspirehep/utils.py#L18

backend/inspirehep/mailing/config.py
drjova: 👍

backend/inspirehep/mailing/api/seminars.py
harunurhan: No no I mean, the load balancer.. it's routing to `https` by default and `protocol` seems ugly in the mail content.

PR #947, opened 2020-05-13
migrator: autoretry on OperationalError in update_relations task

pazembrz:
*INSPIR-3531


michamos:
What about the case where this happens on a record update (say from a workflow), not through the migrator?

pazembrz:
Then this will just raise and fail workflow then workflow can be restarted. We don't have there any specific tasks, so there is no way to easily auto-retry this

PR #950, opened 2020-05-13
seminars: fix meta description

monaawi:


ui/src/seminars/containers/DetailPageContainer.jsx
```diff
@@ -53,7 +53,7 @@ function DetailPage({ record }) {
     <>
       <DocumentHead
         title={title.get('title')}
-        description={makeCompliantMetaDescription(abstract)}
+        description={makeCompliantMetaDescription(abstract.get('value'))}
```
harunurhan, line 56: oh, a bit late for this, but `abstract` is not a required field

PR #951, opened 2020-05-13
submissions: improve author suggester

MJedr:
* INSPIR-3502

### DIFF
--- a/backend/inspirehep/records/marshmallow/authors/es.py
+++ b/backend/inspirehep/records/marshmallow/authors/es.py
@@ -32,4 +32,5 @@ class AuthorsElasticSearchSchema(ElasticSearchBaseSchema, AuthorsRawSchema):
                 force_list(record.get_value(path, default=[])) for path in paths
             )
         )
+
         return {"input": input_values}
--- a/backend/tests/integration/records/api/test_api_authors.py
+++ b/backend/tests/integration/records/api/test_api_authors.py
@@ -4,8 +4,7 @@
 #
 # inspirehep is free software; you can redistribute it and/or modify it under
 # the terms of the MIT License; see LICENSE file for more details.
-
-
+import json
 import uuid
 
 import pytest
--- a/ui/src/submissions/common/getSuggestionSourceNameValue.js
+++ b/ui/src/submissions/common/getSuggestionSourceNameValue.js
@@ -1,0 +1,10 @@
+export default function renderAuthorSuggestion(suggestion) {
+  const name = suggestion._source.name.value;
+  const currentPosition = suggestion._source.positions
+    .filter(position => position.current)
+    .map(position => position.institution)[0];
+  if (currentPosition) {
+    return `${name} (${currentPosition})`;
+  }
+  return name;
+}
--- a/ui/src/submissions/jobs/components/JobForm.jsx
+++ b/ui/src/submissions/jobs/components/JobForm.jsx
@@ -3,6 +3,7 @@ import PropTypes from 'prop-types';
 import { Field, Form } from 'formik';
 import { Row } from 'antd';
 
+
 import ExternalLink from '../../../common/components/ExternalLink';
 import TextField from '../../common/components/TextField';
 import SelectField from '../../common/components/SelectField';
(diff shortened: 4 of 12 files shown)


ui/src/submissions/common/getSuggestionSourceNameValue.js
```diff
@@ -0,0 +1,17 @@
+/* eslint-disable no-unused-vars */
```
harunurhan, line 1: yo shouldn't disable that.

ui/src/submissions/common/getSuggestionSourceNameValue.js
```diff
@@ -0,0 +1,17 @@
...
+function checkIfPositionCurrent(position) {
```
harunurhan, line 2: no need to be a function, you can just `.filter(position => position.current)`

ui/src/submissions/common/getSuggestionSourceNameValue.js
```diff
@@ -0,0 +1,17 @@
...
+  return 'current' in position;
+}
+
+export default class NameSuggester {
+  static getSuggestionSourceNameValue(suggestion) {
```
harunurhan, line 7: you can just make this a `function` rather than wrapping in a class.

ui/src/submissions/common/getSuggestionSourceNameValue.js
```diff
@@ -0,0 +1,17 @@
...
+    const name = suggestion._source.name.value;
+    const currentPosition = suggestion._source.positions
+      .filter(checkIfPositionCurrent)
+      .map(position => position.institution)[0];
+    if (typeof currentPosition !== 'undefined') {
```
harunurhan, line 12: `if(currentPosition)`

ui/src/submissions/seminars/components/SeminarForm.jsx
```diff
@@ -72,7 +69,9 @@ function SeminarForm({ values, isSubmitting, isValid, isValidating }) {
                 placeholder="Family name, first name"
                 pidType="authors"
                 suggesterName="author"
-                extractItemCompletionValue={getSuggestionSourceNameValue}
+                extractItemCompletionValue={
```
harunurhan, line 72: you should leave `extractItemCompletionValue` as it is. Because it's what is filled in the form and sent as author name to the backend. We don't wanna send `Urhan, Harun (CERN)` as the author name. Institution should be only for display.

That's why you should this function pass `renderResultItem` prop instead.

ui/src/submissions/common/getSuggestionSourceNameValue.js
harunurhan: in that case, it's `renderAuthorSuggestion`

backend/tests/integration/records/api/test_api_authors.py
```diff
@@ -191,3 +190,14 @@ def test_get_author_papers(inspire_app):
...
+
+def test_dump_for_es(inspire_app):
+    additional_fields = {
+        "positions": [{"institution": "CERN", "current": True}, {"institution": "DESY"}]
+    }
+    data = faker.record("aut", data=additional_fields)
+    record = AuthorsRecord.create(data)
+    dump = record.serialize_for_es()
+
+    assert "CERN" in dump["author_suggest"]["input"]
```
harunurhan, line 203: do we really want that? what goes into `input` used for `autocomplete`, so we want to autocomplete authors when user types an institutions. (I am not sure if they will, unless we tell them)

harunurhan: if current `position` is just for display putting into `_source` and displaying on the UI is enough.

MJedr: I agree, but that was a part of refinement. I think it is not useful, as we have to do it in UI

backend/inspirehep/records/marshmallow/authors/es.py
```diff
@@ -32,4 +32,5 @@ def get_author_suggest(self, record):
                 force_list(record.get_value(path, default=[])) for path in paths
             )
         )
+
```
harunurhan, line 35: redundant

ui/src/submissions/common/getSuggestionSourceNameValue.js
```diff
@@ -0,0 +1,10 @@
+export default function renderAuthorSuggestion(suggestion) {
```
harunurhan, line 1: the file name

ui/src/submissions/seminars/components/SeminarForm.jsx
```diff
@@ -72,7 +69,10 @@ function SeminarForm({ values, isSubmitting, isValid, isValidating }) {
                 placeholder="Family name, first name"
                 pidType="authors"
                 suggesterName="author"
-                extractItemCompletionValue={getSuggestionSourceNameValue}
+                renderResultItem={renderAuthorSuggestion}
+                extractItemCompletionValue={
+                  renderAuthorSuggestion.renderResultItem
```
harunurhan, line 74: this is not gonna work, perhaps it's better to do local testing first.

harunurhan: also this shouldn't be changed and still use the previous `getSuggestionSourceNameValue`

ui/src/submissions/jobs/components/JobForm.jsx
```diff
@@ -29,14 +29,6 @@ class JobForm extends Component {
     return suggestion._source.legacy_ICN;
   }
 
-  static getSuggestionSourceLegacyName(suggestion) {
```
harunurhan, line 32: why removing this?

ui/src/submissions/common/getSuggestionSourceNameValue.js
harunurhan: also this needs tests.

PR #953, opened 2020-05-13
seminars: fix submission timezone conversion

harunurhan:


backend/inspirehep/submissions/marshmallow/seminars.py
```diff
@@ -30,14 +30,14 @@ def utc_offset_from(timezone):
 def local_form_datetime_to_iso_utc(local_form_datetime_str, local_timezone):
     local_datetime_obj = datetime.strptime(local_form_datetime_str, FORM_DATE_FORMAT)
     utc_offset = utc_offset_from(local_timezone)
-    utc_datetime_obj = local_datetime_obj + utc_offset
+    utc_datetime_obj = local_datetime_obj - utc_offset
```
michamos, line 33: I don't think this is the right approach, it will be wrong when there's a change of offset between now and the seminar. Why don't you do simply as explained in the `pytz` docs?
```
timezone = pytz.timezone(local_timezone)
local_datetime_obj = timezone.localize(local_datetime_obj)
utc_datetime_obj = local_datetime_obj.astimezone("UTC")
```

PR #955, opened 2020-05-14
seminars: fix addressless update bug

harunurhan:
Also adds tests for submitting and updating only with minimum amount of fields (required)

### DIFF
--- a/backend/tests/integration/submissions/test_submissions_views.py
+++ b/backend/tests/integration/submissions/test_submissions_views.py
@@ -8,6 +8,7 @@
 import json
 from copy import deepcopy
 
+import pytest
 from flask import current_app, url_for
 from freezegun import freeze_time
 from helpers.utils import (
@@ -1284,6 +1285,14 @@ def test_conference_raise_loader_error(ticket_mock, inspire_app):
     assert response.status_code == 400
 
 
+REQUIRED_SEMINAR_RECORD_DATA = {
+    "title": {"title": "The Cool Seminar"},
+    "inspire_categories": [{"term": "Accelerators"}],
+    "speakers": [{"name": "Urhan, Ahmet"},],
+    "end_datetime": "2020-05-06T12:30:00.000000",
+    "start_datetime": "2020-05-06T06:30:00.000000",
+    "timezone": "Europe/Zurich",
+}
 SEMINAR_RECORD_DATA = {
     "title": {"title": "The Cool Seminar"},
     "public_notes": [{"value": "A public note"}],
@@ -1318,6 +1327,13 @@ SEMINAR_RECORD_DATA = {
     "timezone": "Europe/Zurich",
 }
 
+REQUIRED_SEMINAR_FORM_DATA = {
+    "name": "The Cool Seminar",
+    "timezone": "Europe/Zurich",
+    "dates": ["2020-05-06 08:30 AM", "2020-05-06 02:30 PM"],
+    "speakers": [{"name": "Urhan, Ahmet"},],
+    "field_of_interest": ["Accelerators"],
+}
 SEMINAR_FORM_DATA = {
     "name": "The Cool Seminar",
     "timezone": "Europe/Zurich",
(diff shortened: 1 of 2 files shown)


backend/tests/integration/submissions/test_submissions_views.py
```diff
@@ -1284,6 +1285,14 @@ def test_conference_raise_loader_error(ticket_mock, inspire_app):
...
 
+REQUIRED_SEMINAR_RECORD_DATA = {
+    "title": {"title": "The Cool Seminar"},
+    "inspire_categories": [{"term": "Accelerators"}],
+    "speakers": [{"name": "Urhan, Ahmet"},],
+    "end_datetime": "2020-05-06T12:30:00.000000",
+    "start_datetime": "2020-05-06T06:30:00.000000",
+    "timezone": "Europe/Zurich",
+}
 SEMINAR_RECORD_DATA = {
```
monaawi, line 1296: you seem to have the same data for ```REQUIRED_SEMINAR_RECORD_DATA``` for the required fields, you could use them from there and not repeat them

backend/tests/integration/submissions/test_submissions_views.py
```diff
@@ -1318,6 +1327,13 @@ def test_conference_raise_loader_error(ticket_mock, inspire_app):
...
 
+REQUIRED_SEMINAR_FORM_DATA = {
+    "name": "The Cool Seminar",
+    "timezone": "Europe/Zurich",
+    "dates": ["2020-05-06 08:30 AM", "2020-05-06 02:30 PM"],
+    "speakers": [{"name": "Urhan, Ahmet"},],
+    "field_of_interest": ["Accelerators"],
+}
 SEMINAR_FORM_DATA = {
     "name": "The Cool Seminar",
```
monaawi, line 1338: same

backend/tests/integration/submissions/test_submissions_views.py
```diff
@@ -1446,39 +1479,49 @@ def test_new_seminar_submission_with_cataloger_login(
...
     assert response.status_code == 200
 
     payload = json.loads(response.data)
     seminar_id = payload["pid_value"]
     seminar_record = SeminarsRecord.get_record_by_pid_value(seminar_id)
     seminar_record_data = {
-        key: value
-        for (key, value) in seminar_record.items()
-        if key in SEMINAR_RECORD_DATA
+        key: value for (key, value) in seminar_record.items() if key in record_data
```
monaawi, line 1524: why is this needed? what extra keys does the seminar_record have?

backend/tests/integration/submissions/test_submissions_views.py
harunurhan: well some are same, some different like `field_of_interest`, so I don't wanna make this 
```
{
**REQUIRED_SEMINAR_RECORD_DATA,
override some present fields,
new fields,
}
```

or
```
{
"present_field": REQUIRED_SEMINAR_RECORD_DATA["present_field"]
}
```
for all.

harunurhan: for me that way is more clear for test input.

backend/tests/integration/submissions/test_submissions_views.py
harunurhan: this is for partial assertion, keys like `control_number`

monaawi: I think it would be better to remove the fields you are sure you want to remove from the seminar_record_data or add field to the expected_record_data. For example. for the case with ```REQUIRED_SEMINAR_RECORD_DATA```, you are only asserting those fields, it would be good to assert the entire record in case there are some weird bugs and some fields that shouldn't be there are there or some that should have been removed are still there.

harunurhan: yeah it would be and we have a plan to either write a helper for this or use that is already built-in to `pytest`. it wasn't straightforward that's why initially I went with this after discussing with @drjova 

anyhow this is a bug fix.

PR #958, opened 2020-05-14
Form fixes

harunurhan:


backend/tests/integration/mailing/test_api_seminars.py
```diff
@@ -3,14 +3,15 @@
...
+from inspirehep.utils import get_inspirehep_url
 
 
 @patch("inspirehep.mailing.api.seminars.send_email")
 def test_send_seminar_confirmation_email(mock_send_email, base_app, db, es_clear):
     expected_recipient = "johndoe@gmail.com"
     seminar = {"control_number": "1234"}
     expected_subject = "Your seminar (1234) has been successfully submitted!"
-    host = current_app.config["SERVER_NAME"]
+    host = get_inspirehep_url()
```
michamos, line 14: I would hardcode this

ui/src/submissions/seminars/schemas/seminar.js
```diff
@@ -15,7 +16,7 @@ const seminarSchema = object().shape({
     .required()
     .label('Seminar Name'),
   dates: array()
-    .of(date())
+    .of(date(SEMINAR_DATETIME_FORMAT))
```
drjova, line 19: can we tests this?

backend/tests/integration/mailing/test_api_seminars.py
harunurhan: because we don't send emails from qa?

ui/src/submissions/seminars/schemas/seminar.js
harunurhan: it's tested but I can add failure case for dates of different format, which won't happen because UI component is enforcing the format.

backend/tests/integration/mailing/test_api_seminars.py
harunurhan: @michamos I left it like this, since I don't think it's important and we do have same for all other tickets/emails.

PR #963, opened 2020-05-15
refersto: refersto tab in author profile

miguelgrc:
* INSPIR-3430

Also modified the title of the date facet to Papers per year everywhere except in literature search (and in this new tab) after discussing with Stella.
Added the `New` feature tag to the new tab.

### DIFF
--- a/backend/inspirehep/search/aggregations.py
+++ b/backend/inspirehep/search/aggregations.py
@@ -6,7 +6,7 @@
 # the terms of the MIT License; see LICENSE file for more details.
 
 
-def hep_earliest_date_aggregation(order, title="Date of paper", agg_type="range"):
+def hep_earliest_date_aggregation(order, title="Papers per year", agg_type="range"):
     return {
         "earliest_date": {
             "date_histogram": {
@@ -157,7 +157,7 @@ def hep_self_author_affiliations_aggregation(
         "self_affiliations": {
             "nested": {"path": "authors"},
             "aggs": {
-                "nested": {
+                "nested_agg": {
                     "filter": {"term": {"authors.record.$ref": author_recid}},
                     "aggs": {
                         "self_affiliations": {
@@ -174,6 +174,23 @@ def hep_self_author_affiliations_aggregation(
     }
 
 
+def hep_author_affiliations_aggregation(
+    order, title="Affiliations", agg_type="checkbox"
+):
+    return {
+        "affiliations": {
+            "nested": {"path": "authors"},
+            "aggs": {
+                "affiliations": {
+                    "terms": {"field": "authors.affiliations.value.raw", "size": 20},
+                    "meta": {"title": title, "order": order, "type": agg_type},
+                    "aggs": {"affiliations": {"reverse_nested": {}}},
+                }
+            },
+        }
+    }
+
+
 def hep_self_author_names_aggregation(
     order, author_recid, title="Name variations", agg_type="checkbox"
 ):
@@ -181,7 +198,7 @@ def hep_self_author_names_aggregation(
         "self_author_names": {
             "nested": {"path": "authors"},
             "aggs": {
-                "nested": {
+                "nested_agg": {
                     "filter": {"term": {"authors.record.$ref": author_recid}},
                     "aggs": {
                         "self_author_names": {
--- a/ui/src/authors/containers/AuthorCitationsContainer.jsx
+++ b/ui/src/authors/containers/AuthorCitationsContainer.jsx
@@ -1,0 +1,35 @@
+import React, { useMemo } from 'react';
+import { connect } from 'react-redux';
+import PropTypes from 'prop-types';
+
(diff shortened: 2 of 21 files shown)


ui/src/authors/utils.js
```diff
@@ -5,6 +5,10 @@ export function getCurrentAffiliationsFromPositions(positions) {
   return positions.filter(position => position.get('current'));
 }
 
+export function getAuthorCitationsQueryString(bai) {
```
harunurhan, line 8: doesn't have to here, if it's only used in single place.

backend/inspirehep/search/aggregations.py
```diff
@@ -174,14 +174,40 @@ def hep_self_author_affiliations_aggregation(
...
+                    "terms": {"field": "authors.affiliations.value.raw", "size": 20},
+                    "meta": {"title": title, "order": order, "type": agg_type},
+                    "aggs": {
+                        "affiliations": {
+                            "reverse_nested": {},
+                            "aggs": {
+                                "inner_agg": {
+                                    "terms": {"field": "authors.full_name.raw"}
+                                }
+                            },
```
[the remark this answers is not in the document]
miguelgrc: You're totally right, that's not needed

ui/src/authors/containers/DetailPageContainer/DetailPageContainer.jsx
```diff
@@ -145,23 +152,51 @@ function DetailPage({
...
-            <Col xs={24} md={12} lg={8}>
-              <ContentBox subTitle="Citations per year">
-                <EmptyOrChildren data={publications} title="0 Research works">
-                  <CitationsByYearGraphContainer />
-                </EmptyOrChildren>
-              </ContentBox>
+          </Row>
+          <Row>
+            <Col span={24}>
+              <Tabs type="card" tabBarStyle={{ marginBottom: 0 }}>
```
harunurhan, line 158: probably this `Tabs` pattern can be a component like `ContentBoxTabs`, same thing exist in literature detail page too.

miguelgrc: But that wouldn't simplify it that much, right? We'd need a `ContentBoxTabs` which should then have as children a `ContentTab` or something like that which gets as props `tooltip`, `name`, `count`, `key`, `forceRender`, `children`...

harunurhan: only for `Tabs` that will have to style which you copy pasted from literature detail page to here too, and these props. it won't do anything else, basically ` <Tabs ...props and class name...>{children}</Tabs>`

ui/src/authors/containers/AuthorCitationsContainer.jsx
```diff
@@ -0,0 +1,34 @@
...
+    () => ({
+      q: `refersto a ${authorBai}`,
+    }),
+    [authorBai]
+  );
+
+  return (
+    <LiteratureSearchContainer
+      namespace={AUTHOR_CITATIONS_NS}
+      baseQuery={baseQuery}
```
harunurhan, line 19: don't we need `baseAggregationsQuery` too, if we have to author facet, we need to exclude that name right?

miguelgrc: No, in this tab we don't need to exclude the author in any facet

ui/src/authors/containers/DetailPageContainer/DetailPageContainer.jsx
miguelgrc: It would only extract the `type="card"` and `tabBarStyle`. `className` is set to `Tabs` only in literature. In authors I added it individually to the `ContentBox` in each tab since otherwise it'd also remove the top border of the cards inside the tab (which we don't have in literature).

harunurhan: I guess it's not difficult make a style that works both though.. but on the other hand why do we need ContentBox for each tab? we can just wrap whole `Tabs` maybe?

miguelgrc: Yeah we can with an uglier selector like `.remove-top-border-of-card > .ant-tabs-content > .ant-tabs-tabpane > .__ContentBox__ > .ant-card-bordered`. We could use a `Card` instead of a `ContentBox` but then we'd need to tweak the margins and so. How do you want to wrap it anyway?

harunurhan: wrapping that I meant.
```
<ContentBox>
  <Tabs..
</ContentBox>
```

from there, you probably need less css to make it look like we want. I can't see exactly without playing with it locally but give it a try, if not that's OK too I guess.

miguelgrc: If I do this then the tabs are inside the white `ContentBox` and they look super weird :(

PR #965, opened 2020-05-18
seminars: fix series link

monaawi:


ui/src/common/components/__tests__/EventSeries.test.jsx
```diff
@@ -21,7 +26,16 @@ describe('EventSeries', () => {
       { name: 'Conference 2', number: 10 },
       { name: 'Conference 3' },
     ]);
-    const wrapper = shallow(<EventSeries series={series} />);
+    const wrapper = shallow(
+      <EventSeries series={series} pidType={CONFERENCES_PID_TYPE} />
+    );
+    expect(wrapper.dive()).toMatchSnapshot();
+  });
+  it('renders using pidType', () => {
```
harunurhan, line 34: don't all of them use `pidType` already?, I think just passing `SEMINARS` in one of them is fine.

monaawi: Actually, I wanted to test also the part where it says which conference/seminar it is in the series, this is why I added a second one with name and number but with seminars. I'll change the test name though

harunurhan: then just have `renders conference series with name and number`  and `renders seminar series with name and number`.. to make it more obvious.

PR #971, opened 2020-05-18
migrator: fix stop_after_one_step fails as task_uuid is passed

pazembrz:
*INSPIR-3534

### DIFF
--- a/backend/tests/integration/migrator/test_migrator_cli.py
+++ b/backend/tests/integration/migrator/test_migrator_cli.py
@@ -10,7 +10,7 @@ import os
 
 import pkg_resources
 import pytest
-from helpers.utils import override_config
+from helpers.utils import create_record, override_config
 from invenio_db import db
 from invenio_search import current_search
 from mock import patch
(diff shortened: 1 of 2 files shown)


backend/tests/integration/migrator/test_migrator_cli.py
```diff
@@ -220,3 +220,15 @@ def test_migrate_continuously(mock_migration, mock_handler, inspire_app, cli):
...
+def test_regression_migration_step_exits_without_error_when_steps_ends(
+    inspire_app, cli
+):
+    create_record("lit")
+    result = cli.invoke(["migrate", "mirror_step", "-s", "2"])
+    assert 0 == result.exit_code
+    assert (
+        "stop_after_one_step() takes 0 positional arguments but 1 was given"
+        not in result.output
+    )
```
drjova, line 234: IMO just ``assert 0 == result.exit_code`` is enough, WDYT?

pazembrz: :+1:

PR #973, opened 2020-05-22
authentication: token authentication check for roles

MJedr:
* INSPIR-3545


drjova:
@MJedr could we add tests to ensure only tokens for superusers can create and update records?

PR #974, opened 2020-05-25
ui: add option to see more references

MJedr:
* INSPIR-3569

### DIFF
--- a/ui/src/common/components/ListWithPagination.jsx
+++ b/ui/src/common/components/ListWithPagination.jsx
@@ -12,6 +12,7 @@ const GRID_CONFIG = {
   xl: 4,
 };
 
+const PAGE_SIZE_OPTIONS = ['25', '50', '100', '250'];
 
 /**
  * Only displays given page items at once and pagination ui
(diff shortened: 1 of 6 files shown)


ui/src/common/components/ListWithPagination.jsx
```diff
@@ -41,14 +50,18 @@ class ListWithPagination extends Component {
         pageSize={pageSize}
         loading={loading}
         showTotal={ListWithPagination.getPaginationRangeInfo}
+        onShowSizeChange={onSizeChange}
```
harunurhan, line 53: below props should only be set if this is one (`onShowSizeChange`) is present, or at least do `showSizeChanger={onShowSizeChange != null}`, otherwise size changer will be displayed no matter what.

ui/src/common/components/ListWithPagination.jsx
```diff
@@ -70,8 +83,9 @@ ListWithPagination.propTypes = {
   pageItems: PropTypes.instanceOf(Immutable.List).isRequired,
   renderItem: PropTypes.func.isRequired,
   onPageChange: PropTypes.func.isRequired,
-  pageSize: PropTypes.number.isRequired,
-  page: PropTypes.number.isRequired,
+  onSizeChange: PropTypes.func,
```
harunurhan, line 86: we need a test for this as well, and make sure `onSizeSize` is called when internal `Pagination.onShowSizeChange` triggered.

PR #975, opened 2020-05-25
Search ui refactor

harunurhan:


ui/src/search/queryChange.js
```diff
@@ -0,0 +1,121 @@
...
+import { fetchSearchResults, fetchSearchAggregations } from '../actions/search';
+import { fetchCitationSummary } from '../actions/citations';
+import { isCitationSummaryEnabled } from '../literature/containers/CitationSummarySwitchContainer';
+
+export function onLiteratureQueryChange(helper, dispatch) {
+  const isInitialQueryUpdate = helper.isInitialQueryUpdate();
+  if (isInitialQueryUpdate || helper.hasQueryChanged()) {
+    const searchUrl = `${helper.getPathname()}?${helper.getQueryString()}`;
+    const urlWithHash = `${searchUrl}${helper.state.router.location.hash}`;
+    // FIXME: this logic exist on others too
```
monaawi, line 12: You are right, there is some duplicated logic around here. There are actually only few differences between these functions and they are cases we had identified before. One is whether it is embedded or not for which we use the ```initialQueryUpdate```, then we have the aggregation fetch modes, the urlHash and the citation summary. The way it is now, I like that the methods are clear and separate and easy to follow but I don't like the fact that if we need to add a new namespace with its specific fetchmode or it also has citation summary or uses hash, we would need to look into these methods and pick which parts we need for our new collection and duplicate the code again. I liked that before we could specify these options for every namespace and it worked out of the box. I think we could do something similar here. I think these options are similar to the ```persistedQueryParamsDuringNewSearch``` and I think they would go well in the search config. And then all of these functions could be reduced to just one, something like:
```javascript
const isInitialQueryUpdate = helper.isInitialQueryUpdate() && searchConfig[helper.namespace].isEmbedded;
if ( isInitialQueryUpdate || helper.hasQueryChanged()) {
	const searchUrl = `${helper.getPathname()}?${helper.getQueryString()}`;

	if(searchConfig[helper.namespace].withHash){
		const searchUrl = `${searchUrl}${helper.state.router.location.hash}`;
	}

	if (isInitialQueryUpdate) {
		 dispatch(replace(searchUrl));
	} else {
		dispatch(push(searchUrl));
	}

	dispatch(fetchSearchResults(helper.namespace, searchUrl));
}

if (
  (searchConfig[helper.namespace].aggregationsFetchMode === FETCH_MODE_ALWAYS ||
  (aggregationsFetchMode === FETCH_MODE_INITIAL && isAggregationsEmpty))
  && (isInitialQueryUpdate || helper.hasQueryChangedExceptSortAndPagination())
) {
	const aggsQueryString = searchConfig[helper.namespace].aggregationsFetchMode === FETCH_MODE_INITIAL ?
'' : helper.getAggregationsQueryString();
    const searchAggsUrl = `${helper.getPathname()}/facets?${aggsQueryString}`;
    dispatch(fetchSearchAggregations(helper.namespace, searchAggsUrl));
    
    if (searchConfig[helper.namespace].withCitationSummary && isCitationSummaryEnabled(helper.state)) {
      dispatch(fetchCitationSummary(helper.namespace));
    }
}
```
It is a bit similar to what we had before but it isn't scattered around between the dispatcher and the search actions. This is just a rough idea. Wdyt?

harunurhan: I would rather to duplicate than have this, whole idea was to go have flexible side effects for each collection when it's needed.


we can have more helper methods to reduce duplication to minimum and have even more readable `onQueryChange` functions without many `if`s.

ui/src/actions/search.js
```diff
@@ -167,11 +92,32 @@ export function fetchSearchAggregationsForCurrentQuery(namespace) {
   };
 }
 
-// this then MAY trigger search and aggregations request in `searchDispatcher.js`
+function getQueryFromState(namespace, state) {
```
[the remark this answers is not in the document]
harunurhan: although we can get around that, as of now... helper is created with previous and next state which requires `QUERY_UPDATE` to be dispatched but `NEW_SEARCH` should be dispatched before the `QUERY_UPDATE`.

In the future we should probably get rid of this `NEW_SEARCH` and this logic should be part of `QUERY_UPDATE` in the reducer.

ui/src/search/queryChange.js
monaawi: Actually, I find it easier to see what is happening to each namespace by looking at its config and seeing what sort of aggregations it has or if it is embedded and things like that. I also liked the easiness of adding a new namespace and just selecting its config.
The code with ifs is just to show you that the different parts are easily covered by the config we had before. Of course, it could be thought about in a nicer way and written to be more readable, I wrote it is a rough idea. 
But to me, many of these functions are just full of duplications and when we need to add a new one, one probably has to go through these functions, see if there is one that has all of the elements he needs, if not, create a new ```onQueryChange``` and find the pieces of the puzzle that he needs for his specific config which I think is just cumbersome.

harunurhan: How come it's easier read a function with bigger `if` conditions, more lines, and a config that is on the other file than reading a single function?

Anyway I don't find it easier read, also this function will grow for sure and get worse, as we add more feature. That's why I don't agree with putting all things in the same blob of logic.

We should just focus on reducing duplication and making it more readable for each namespace, so we can try make a simpler query change functions like:
```
onXQueryChange(helper) {
  helper.search();
  helper.updateLocation(); // would be here only for some namespaces
  
  helper.searchAggregations(); // would be here only for some namespaces

  // other side effects depending on the name space
}
```

monaawi: I didn't say it would be a big function with ```if``` conditions, I said ```The code with ifs is just to show you that the different parts are easily covered by the config we had before```, that code is just to show that the entire logic is coordinated by those elements - the way it fetches aggregation, it is embedded or not and so on. 
We should focus on removing duplication but I still like the configurable part that we had before and I think it was more reusable. Right now, it is mostly copy pasted code spread around in 4-5 different functions which will get spread around even more. What if that part changes? we go and change it in all places?

harunurhan: you don't need to say it, It becomes a bigger function when you put it all in single function, so as the `if(conditions)`'s conditions getting bigger.

PR #977, opened 2020-05-26
experiments: Add experiments search and detail pages

monaawi:
* INSPIR-3353
* Needs remapping of experiments


backend/inspirehep/records/marshmallow/experiments/ui.py
```diff
@@ -0,0 +1,82 @@
...
+    predecessor_experiments = NestedWithoutEmptyObjects(
+        PredecessorExperimentSchemaV1,
+        default=[],
+        dump_only=True,
+        many=True,
+        attribute="related_records",
+    )
+
+    @staticmethod
+    def get_subsidiary_experiments(data):
```
drjova, line 51: We could move it to https://github.com/inspirehep/inspirehep/blob/master/backend/inspirehep/search/api.py#L261

ui/src/experiments/components/ExperimentAssociatedArticles.jsx
```diff
@@ -0,0 +1,21 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+
+import { Link } from 'react-router-dom';
+import { LITERATURE } from '../../common/routes';
+import { getPapersQueryString } from '../utils';
+
+function ExperimentAssociatedArticles({ recordId, legacyName }) {
```
harunurhan, line 8: `ExperimentAssociatedArticlesLink` might be better when I saw it in another component to be used I thought it was a list component displaying the articles.

ui/src/experiments/components/ExperimentCollaborationArticles.jsx
```diff
@@ -0,0 +1,19 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+
+import { Link } from 'react-router-dom';
+import { LITERATURE } from '../../common/routes';
+
+function ExperimentCollaborationArticles({ recordId }) {
```
harunurhan, line 7: same

ui/src/experiments/components/ExperimentCollaborationMembers.jsx
```diff
@@ -0,0 +1,19 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+
+import { Link } from 'react-router-dom';
+import { AUTHORS } from '../../common/routes';
+
+function ExperimentCollaborationMembers({ recordId }) {
```
harunurhan, line 7: same

ui/src/experiments/components/ExperimentDates.jsx
```diff
@@ -0,0 +1,52 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import moment from 'moment';
+import { InlineUL } from '../../common/components/InlineList';
+import { hasDayMonthAndYear, hasMonthAndYear } from '../../common/utils';
+
+function getDisplayFormatForDateString(date) {
+  if (hasDayMonthAndYear(date)) {
+    return 'YYYY-MM-DD';
```
harunurhan, line 9: this is I think the forth or third different date format , should we consider unifying all/some if makes sense? @StellaCh 

(literature `YYYY/MM/DD`, conferences `D MMMM YYYY`)

ui/src/experiments/components/ExperimentDates.jsx
```diff
@@ -0,0 +1,52 @@
...
+  dateStarted,
+  dateProposed,
+  dateApproved,
+  dateCompleted,
+  dateCancelled,
+  wrapperClassName,
+}) {
+  return (
+    <InlineUL wrapperClassName={wrapperClassName}>
+      {dateProposed && `Proposed: ${getFormattedDate(dateProposed)}`}
```
harunurhan, line 33:
```jsx
dateProposed  && <span>Proposed: {getFormattedDate(dateProposed)}</span>
```
( you don't need template string in the markup)

ui/src/experiments/containers/DetailPageContainer.jsx
```diff
@@ -0,0 +1,220 @@
...
+import { EXPERIMENT_PAPERS_NS } from '../../reducers/search';
+import { newSearch } from '../../actions/search';
+import ExperimentPapers from './ExperimentPapers';
+import { makeCompliantMetaDescription } from '../../common/utils';
+
+function DetailPage({ record }) {
+  const metadata = record.get('metadata');
+
+  const legacyName = metadata.get('legacy_name');
+  const institutions = metadata.get('institutions', List());
```
harunurhan, line 34: why default was necessary? you could do `institutions &&` rather than `institutions.size > 0`

ui/src/experiments/components/ExperimentDates.jsx
monaawi: yup, it is a new date format. It was in the requirements, this is why I couldn't reuse others.

ui/src/common/components/RelatedRecordsList.jsx
```diff
@@ -2,14 +2,14 @@ import React from 'react';
...
-import pluralizeUnlessSingle from '../../common/utils';
+import InlineList from './InlineList';
+import { INSTITUTIONS } from '../routes';
+import pluralizeUnlessSingle from '../utils';
 
 function renderRelatedRecord(relatedRecord) {
   return (
     <Link to={`${INSTITUTIONS}/${relatedRecord.get('control_number')}`}>
-      {relatedRecord.get('legacy_ICN')}
+      {relatedRecord.get('legacy_ICN') || relatedRecord.get('legacy_name')}
```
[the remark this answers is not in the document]
monaawi: yeah, now I added also pidType to it (I had missed the link) and made it have only experiments and institutions. But I don't know what name to give it

ui/src/experiments/components/ExperimentDates.jsx
StellaCh: We can reuse the literature one here

ui/src/common/components/RelatedRecordsList.jsx
```diff
@@ -0,0 +1,44 @@
...
+      extractKey={extractKeyFromRelatedRecord}
+      renderItem={renderRelatedRecord}
+    />
+  );
+}
+
+RelatedRecordsList.propTypes = {
+  relatedRecords: PropTypes.instanceOf(List),
+  relationType: PropTypes.string.isRequired,
+  label: PropTypes.string.isRequired,
```
harunurhan, line 39: I think label can derive from the pidType, but that's kinda of an optional thing.

We can just have a global maps later `pidType <-> pathname <-> namespace <-> label`.

ui/src/experiments/components/ExperimentDates.jsx
harunurhan: Cool I guess now it could be shared then right? somehow.

ui/src/search/config.js
```diff
@@ -72,7 +74,15 @@ const searchConfig = {
...
-  } 
+  },
+  [EXPERIMENTS_NS]: {
+    persistedQueryParamsDuringNewSearch: defaultPersistedQueryParamsDuringNewSearch,
+    onQueryChange: onAggregationlessCollectionQueryChange,
+  },
+  [EXPERIMENT_PAPERS_NS]: {
+    persistedQueryParamsDuringNewSearch: persistedQueryParamsDuringNewSearchForEvents,
+    onQueryChange: onEmbeddedLiteratureQueryChange,
+  },
```
harunurhan, line 85: see, no copy paste 😄

ui/src/experiments/components/ExperimentDates.jsx
monaawi: Actually no, the date in literature is already formatted in the backend. How about the date format in conferences? that would be very easy to reuse. @StellaCh

ui/src/common/components/RelatedRecordsList.jsx
monaawi: yeah, it can be derived from the pidType but I'll leave it like this for now

ui/src/experiments/components/ExperimentDates.jsx
StellaCh: We need to unify them, it doesn't make much sense to have different formats. The main difference is that for seminars and conferences DD is important, so it should be in the beginning, whereas for experiments we don't care so much about the specific day, we should rather focus on the year and month

StellaCh: What are the options for experiment date?

monaawi: It can be anything. We can even keep it as it is now and as it is in the mockup. It's just that currently we have different formats across collections. If you say they are different use cases and that they shouldn't be the same, we can keep them different.
For literature, we have something like ```Nov 28, 1997```, for conferences ```1 June 2020``` and experiments in the mockup and how they are currently implemented have ```1992-10-01```

StellaCh: let's have maximum 2 ways of displaying dates. Therefore, we keep Nov 28, 1997 and 1 June 2020 and we reuse the 1st option for experiments (Nov 28, 1997)

PR #981, opened 2020-05-27
records: fix revisions revert

pazembrz:
*INSPIR-3486

### DIFF
--- a/backend/inspirehep/records/api/base.py
+++ b/backend/inspirehep/records/api/base.py
@@ -13,6 +13,7 @@ from itertools import chain
 
 import structlog
 from elasticsearch import NotFoundError
+from flask import current_app
 from inspire_dojson.utils import strip_empty_values
 from inspire_schemas.api import validate as schema_validate
 from inspire_utils.record import get_value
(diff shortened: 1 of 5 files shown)


backend/inspirehep/records/api/base.py
```diff
@@ -451,3 +454,49 @@ def serialize_for_es(self):
...
+
+
+class InspireRevisionsIterator(RevisionsIterator):
+    def __getitem__(self, revision_id):
+        """Get a specific revision.
+
+        Revision id is always smaller by 1 from version_id.
+        Replaces RevisionsIterator from Invenio until they merge proper patch
+        """
+        if revision_id < 0:
```
drjova, line 491: shouldn't this be `< 1`?

pazembrz: no, as usually revision 0 is correct. But in Inspire it's impossible to have revision 0 as we are doing increment by 2 when we create record.

pazembrz:
I have to change this PR as Invenio already merged my change, so I can remove like half of this PR. Don't approve yet please.

pazembrz:
Ok, After discussion with Harrun I won't update invenio-records and we will stay with fix hard-coded in our code ad I would have to update invenio-records from 1.0.2 to 1.3.2

drjova:
@pazembrz I'd still bump it if it doesn't require to bump everything.

backend/tests/integration-async/records/api/test_base.py
```diff
@@ -140,3 +138,57 @@ def test_get_modified_references_returns_no_references_when_non_impacting_metada
     db.session.commit()
 
     assert citing_record.get_modified_references() == []
+
+
+def test_revert_revision_works_correctly_and_runs_update(inspire_app):
```
drjova, line 143: Shall we test also that the data is correct after the revert?

pazembrz: Check is already there (check line 194)

PR #982, opened 2020-05-27
views: redirect data record to legacy

monaawi:
* INSPIR-3568

### DIFF
--- a/backend/inspirehep/legacy/views.py
+++ b/backend/inspirehep/legacy/views.py
@@ -20,6 +20,9 @@ def redirect_record(record_path):
     if not endpoint:
         abort(404)
 
+    if endpoint == "data":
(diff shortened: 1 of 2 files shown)


backend/inspirehep/legacy/views.py
```diff
@@ -20,6 +20,9 @@ def redirect_record(record_path):
     if not endpoint:
         abort(404)
 
+    if endpoint == "data":
+        return redirect(f"{current_app.config['LEGACY_BASE_URL']}/record/{recid}", 301)
```
michamos, line 24: Maybe make it a temporary redirect?
```suggestion
        return redirect(f"{current_app.config['LEGACY_BASE_URL']}/record/{recid}", 302)
```

PR #986, opened 2020-05-28
api: add public rest apis

pazembrz:
*INSPIR-3448

### DIFF
--- a/backend/inspirehep/records/config.py
+++ b/backend/inspirehep/records/config.py
@@ -24,6 +24,7 @@ from inspirehep.access_control import (
     SessionCatalogerPermission,
     SessionSuperuserPermission,
 )
+from inspirehep.records.links import build_citation_search_link
 from inspirehep.search.aggregations import hep_rpp
 from inspirehep.search.api import (
     AuthorsSearch,
@@ -66,7 +67,7 @@ RECORD = {
     "pid_fetcher": "recid",
     "default_endpoint_prefix": True,
     # XXX decide about the links
-    "links_factory_imp": lambda links: {},
+    "links_factory_imp": "inspirehep.records.links:inspire_detail_links_factory",
     "indexer_class": RecordIndexer,
     "search_type": None,
     "search_factory_imp": "inspirehep.search.factories.search:search_factory_without_aggs",
@@ -83,6 +84,8 @@ RECORD = {
     "update_permission_factory_imp": deny_all,
     "delete_permission_factory_imp": deny_all,
     "list_permission_factory_imp": allow_all,
+    "record_serializers_aliases": {"json": "application/json"},
+    "search_serializers_aliases": {"json": "application/json"},
 }
 
 LITERATURE = deepcopy(RECORD)
@@ -121,6 +124,18 @@ LITERATURE.update(
                 "_source": ["control_number", "self", "titles", "authors"],
                 "completion": {"field": "bookautocomplete"},
             },
+        },
+        "record_serializers_aliases": {
+            "bibtex": "application/x-bibtex",
+            "latex-eu": "application/vnd+inspire.latex.eu+x-latex",
+            "latex-us": "application/vnd+inspire.latex.us+x-latex",
+            "json": "application/json",
+        },
+        "search_serializers_aliases": {
+            "bibtex": "application/x-bibtex",
+            "latex-eu": "application/vnd+inspire.latex.eu+x-latex",
+            "latex-us": "application/vnd+inspire.latex.us+x-latex",
+            "json": "application/json",
         },
     }
 )
@@ -592,3 +607,6 @@ LITERATURE_SOURCE_EXCLUDES_BY_CONTENT_TYPE = {
         "_bibtex_display",
     ]
 }
+
+
+ADDITIONAL_LINKS = {"LITERATURE": {"citations": build_citation_search_link}}
--- a/backend/inspirehep/records/links.py
+++ b/backend/inspirehep/records/links.py
@@ -1,0 +1,78 @@
+# -*- coding: utf-8 -*-
+#
+# Copyright (C) 2019 CERN.
+#
+# inspirehep is free software; you can redistribute it and/or modify it under
+# the terms of the MIT License; see LICENSE file for more details.
+from flask import current_app, request, url_for
+from invenio_records_rest import current_records_rest
+
+from inspirehep.pidstore.api import PidStoreBase
+
+
+def inspire_detail_links_factory(pid, record=None, record_hit=None, *args, **kwargs):
+    links = {}
+    endpoint = find_record_endpoint(pid, record_hit)
+    endpoint_item = "invenio_records_rest.{0}_item".format(endpoint)
+    self = url_for(endpoint_item, pid_value=pid.pid_value, _external=True)
+
+    serializers_endpoints = current_app.config[endpoint.upper()].get(
+        "record_serializers_aliases", {}
+    )
+    for serializer_endpoint in serializers_endpoints.keys():
+        links[serializer_endpoint] = f"{self}?format={serializer_endpoint}"
+
+    additional_endpoints = current_app.config.get("ADDITIONAL_LINKS", {}).get(
+        endpoint.upper(), {}
+    )
+    for additional_endpoint_name, additional_endpoint in additional_endpoints.items():
+        links[additional_endpoint_name] = (
+            url_for(additional_endpoint, pid_value=pid.pid_value, _external=True)
+            if isinstance(additional_endpoint, str)
+            else additional_endpoint(pid)
+        )
+
+    return links
+
+
+def inspire_search_links(links):
+    self_url = None
+    if links:
+        self_url = links.get("self")
+    if not self_url:
+        return {}
+    endpoint = request.path[1:]
+    formats = (
+        current_app.config[endpoint.strip("/").upper()]
+        .get("search_serializers_aliases", {})
+        .keys()
+    )
+    links.update(
+        {format_name: f"{self_url}&format={format_name}" for format_name in formats}
+    )
+
+    return links
+
+
+def find_record_endpoint(pid, record_hit=None, **kwargs):
+    """gets endpoint from pid type or from `$schema` if record_data is from search results,
+    as all pid_types from search_result are `recid`.
+    If both ways of resolving endpoint are not available gets it from pid_value - additional db query"""
+    if pid.pid_type != "recid":
+        return current_records_rest.default_endpoint_prefixes[pid.pid_type]
+    elif record_hit and "$schema" in record_hit.get("_source", {}):
+        return PidStoreBase.get_endpoint_from_pid_type(
+            PidStoreBase.get_pid_type_from_schema(record_hit["_source"]["$schema"])
+        )
+    else:
+        return PidStoreBase.get_endpoint_from_pid_type(
+            PidStoreBase.get_pid_type_from_recid(pid.pid_value)
+        )
+
+
+def build_citation_search_link(pid):
+    return url_for(
+        "invenio_records_rest.literature_list",
+        _external=True,
+        q=f"refersto:recid:{pid.pid_value}",
+    )
--- a/backend/inspirehep/search/api.py
+++ b/backend/inspirehep/search/api.py
@@ -10,12 +10,13 @@
 
 from elasticsearch import RequestError
 from elasticsearch_dsl.query import Match, Q
-from flask import current_app
+from flask import current_app, request
 from invenio_search import current_search_client as es
 from invenio_search.api import DefaultFilter, RecordsSearch
 
 from inspirehep.accounts.api import is_superuser_or_cataloger_logged_in
 from inspirehep.pidstore.api import PidStoreBase
+from inspirehep.search.errors import MaximumSearchPageSizeExceeded
 from inspirehep.search.factories import inspire_query_factory
 from inspirehep.search.utils import RecursionLimit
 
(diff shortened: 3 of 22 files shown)


backend/inspirehep/records/config.py
```diff
@@ -118,6 +120,18 @@
                 "completion": {"field": "bookautocomplete"},
             },
         },
+        "record_serializers_aliases": {
+            "bibtex": "application/x-bibtex",
+            "latex_eu": "application/vnd+inspire.latex.eu+x-latex",
+            "latex_us": "application/vnd+inspire.latex.us+x-latex",
```
michamos, line 126:
```suggestion
            "latex-eu": "application/vnd+inspire.latex.eu+x-latex",
            "latex-us": "application/vnd+inspire.latex.us+x-latex",
```

backend/inspirehep/records/config.py
```diff
@@ -118,6 +120,18 @@
...
+            "json": "application/json",
+        },
+        "search_serializers_aliases": {
+            "bibtex": "application/x-bibtex",
+            "latex_eu": "application/vnd+inspire.latex.eu+x-latex",
+            "latex_us": "application/vnd+inspire.latex.us+x-latex",
```
michamos, line 132:
```suggestion
            "latex-eu": "application/vnd+inspire.latex.eu+x-latex",
            "latex-us": "application/vnd+inspire.latex.us+x-latex",
```

backend/inspirehep/records/links.py
```diff
@@ -0,0 +1,65 @@
...
+    )
+    links.update(
+        {format_name: f"{self_url}&format={format_name}" for format_name in formats}
+    )
+
+    return links
+
+
+def find_record_endpoint(pid, record_hit=None, **kwargs):
+    if pid.pid_type != "recid":
```
michamos, line 56: do we ever have `pid_type == "recid"`?

backend/inspirehep/records/links.py
```diff
@@ -0,0 +1,65 @@
...
+        return current_records_rest.default_endpoint_prefixes[pid.pid_type]
+    elif record_hit and "$schema" in record_hit.get("_source", {}):
```
michamos, line 58: why is this branch needed? is the same function called on search results and on detailed records? AFAICT it's only used for individual records that have a `pid.pid_type` set that could be used.

backend/tests/integration/records/serializers/json/test_authors.py
```diff
@@ -317,3 +317,25 @@ def test_authors_search_json_with_logged_in_cataloger(inspire_app):
 
     assert expected_status_code == response_status_code
     assert expected_result == response_data_hits_metadata
+
+
+def test_authors_detail_links(inspire_app):
+    expected_status_code = 200
+    record = create_record("aut")
+    expected_links = {
+        "json": f"http://localhost:5000/authors/{record['control_number']}?format=json"
```
michamos, line 326: * Shouldn't this contain a `/api/` at the beginning? same question for all examples. I guess it's because you're using the  `inspire_app.test_client` and not the `test_client` of the `api` Flask application.
* Why isn't there a `self` link here? (again, same question for all record types)

backend/tests/integration/records/serializers/json/test_literature.py
```diff
@@ -900,3 +900,31 @@ def test_record_link_to_hepdata_in_detail_view(inspire_app):
...
     )
+
+
+def test_literature_detail_links(inspire_app):
+    expected_status_code = 200
+    record = create_record("lit")
+    cn = record["control_number"]
+    expected_links = {
+        "bibtex": f"http://localhost:5000/literature/{cn}?format=bibtex",
+        "citations": f"http://localhost:5000/literature/{cn}/citations",
```
michamos, line 911: I thought we wanted to redirect to a search with `q=refersto:recid:{cn}` here, as the `citations` endpoint is a specific serialization for the frontend (that might be removed soon).

backend/tests/integration/records/serializers/test_bibtex.py
```diff
@@ -10,7 +10,7 @@
 
 def test_bibtex(inspire_app):
     headers = {"Accept": "application/x-bibtex"}
-    data = {"control_number": 637275237, "titles": [{"title": "This is a title."}]}
+    data = {"control_number": 637_275_237, "titles": [{"title": "This is a title."}]}
```
michamos, line 13: thanks, black, very useful :rofl:

backend/tests/integration/records/serializers/test_bibtex.py
```diff
@@ -193,3 +196,13 @@ def test_bibtex_encodes_unicode_outside_of_math_environments(inspire_app):
...
+
+
+def test_literature_detail_bibtex_link_alias_format(inspire_app):
+    expected_status_code = 200
+    record = create_record("lit")
+    expected_content_type = "application/x-bibtex"
+    with inspire_app.test_client() as client:
+        response = client.get(f"/literature/{record['control_number']}?format=bibtex")
+    assert response.status_code == expected_status_code
+    assert response.content_type == expected_content_type
```
michamos, line 208: can you maybe test that it actually returns a bibtex payload?

backend/inspirehep/search/api.py
```diff
@@ -100,6 +101,10 @@ def source_for_content_type(self, content_type):
         return self
 
     def execute(self, *args, **kwargs):
+        if request:
+            size = request.args.get("size", default=25, type=int)
+            if size > current_app.config.get("SEARCH_MAX_SEARCH_PAGE_SIZE", 500):
+                raise MaximumSearchPageSizeExceeded()
```
michamos, line 107: I couldn't find a test for this. What status code are we returning? I think `400` would be appropriate.

backend/inspirehep/search/errors.py
```diff
@@ -0,0 +1,13 @@
...
+#
+# inspirehep is free software; you can redistribute it and/or modify it under
+# the terms of the MIT License; see LICENSE file for more details.
+from inspirehep.errors import BaseRestError
+
+
+class MaximumSearchPageSizeExceeded(BaseRestError):
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self.description = f"Maximum search page size exceeded."
```
michamos, line 13: This is not useful, you should say what the maximum is.

backend/tests/integration/records/serializers/json/test_literature.py
```diff
@@ -900,3 +900,31 @@ def test_record_link_to_hepdata_in_detail_view(inspire_app):
...
+        "json": f"http://localhost:5000/literature/{cn}?format=json",
+        "latex_eu": f"http://localhost:5000/literature/{cn}?format=latex_eu",
+        "latex_us": f"http://localhost:5000/literature/{cn}?format=latex_us",
+    }
```
michamos, line 915: could you add tests for `/api/arxiv` and `/api/orcid`?

backend/inspirehep/records/config.py
pazembrz: :+1:  but then we have error in task requirements ;)

backend/inspirehep/records/links.py
pazembrz: Yes, when record is serialized from Elasticsearch (search results) then it's pid_type is always recid.

backend/inspirehep/records/links.py
pazembrz: Yes, same function is called for search results and detailed records

backend/tests/integration/records/serializers/json/test_authors.py
pazembrz: In test we can add /api/ or ignore it, then it will produce links accordingly. I can add it here but all requests in this test file are made without /api/ so I tired to be consistent with previous tests.

pazembrz: self link in requirements was mentioned only on search result response not on record detail or search record details. If it should be also on record detail please update requirements ;)

backend/tests/integration/records/serializers/test_bibtex.py
pazembrz: Yeah, he loves to do this ;)

backend/tests/integration/records/serializers/test_bibtex.py
pazembrz: Is it necessary? We have only one way to return bibtex content_type. And we are testing it's output in other tests. Don't forget that this `format=` stuff is only a shortcut.

backend/inspirehep/search/api.py
pazembrz: Ah right. I missed this. Thanks :+1:

backend/inspirehep/records/links.py
michamos: can you add a comment about that? it's far from obvious.

backend/inspirehep/search/errors.py
pazembrz: :+1:

backend/tests/integration/records/serializers/json/test_literature.py
pazembrz: :+1:

backend/tests/integration/records/serializers/json/test_literature.py
pazembrz: This was not mentioned in requirements nor in Tech notes.

backend/inspirehep/records/links.py
pazembrz: :+1:

backend/tests/integration/records/serializers/json/test_literature.py
pazembrz: Actually in requirements this specific link was mentioned as example for citations `"https://inspirehep.net/api/literature/451647/citations?page=1&size=25"`

pazembrz: And from what I can see this endpoint will stay there for now. Only UI won't call it anymore.

backend/tests/integration/records/api/test_api_authors.py
```diff
@@ -190,3 +190,20 @@ def test_get_author_papers(inspire_app):
     author_papers = author.get_papers_uuids()
     assert str(lit_1.id) in author_papers
     assert str(lit_2.id) not in author_papers
+
+
+def test_arxiv_url_also_supports_format_alias(inspire_app):
```
michamos, line 195:
```suggestion
def test_orcid_url_also_supports_format_alias(inspire_app):
```

backend/tests/integration/records/serializers/json/test_authors.py
michamos: I think it makes sense also for the detailed records for consistency. What do you think?

backend/tests/integration/records/serializers/json/test_literature.py
michamos: But that endpoint returns only the UI data, not a full record. I think that's confusing. IIRC, we discussed this with @drjova and decided to use the search for this. WDYT @drjova ?

backend/tests/integration/records/serializers/json/test_authors.py
pazembrz: Please note that `self` will be actually the same link as `json` and it will add additional complexity to links generator to not add this self link in search record detail view.

michamos: All I'm saying is we should do `self` everywhere or nowhere. I don't think that's the case currently.

backend/tests/integration/records/serializers/json/test_literature.py
pazembrz: Why stuff like this is always appearing during review... Here it won't be too difficult to change it but still it's wrong. If requirements was accepted already, and refinement for the task was done and task was almost finished and it's now on review, we shouldn't change requirements but bugfix task should be created for this.

backend/tests/integration/records/serializers/json/test_authors.py
pazembrz: currently we have self only on `search result page` not on `search result item` nor on `detail record view`. Adding it to `detail view` will automatically add it to `search result item`, so there should be added additional check to not add them there as this will be messy if every record on `search page` will also have `self`. 
Additionally `self` is the same as `json` link, so maybe if we really want `self` everywhere we can replace `json` to `self`, but this looks messy IMHO.

backend/tests/integration/records/api/test_api_authors.py
pazembrz: :+1:

backend/tests/integration/records/serializers/json/test_literature.py
drjova: @pazembrz this is totally not your fault and I know it's frustrating. It's good that we cautch such things during review and we have to be more careful with the requirements in the future. This is my fault, @michamos you're right we dicided to go for the search results and I didn't update the US afterwards.

pazembrz: I have weird feeling that you are trying to calm me down. Was I aggressive in the previous post? If I was I'm sorry. It wasn't my intention. I was trying to say that our workflow is weird and we shouldn't look who's fault was that, but on the moment when we see bug/problem like that a new bug should be created in Jira. It could be fixed by me, even in same PR but in separate commit so we could keep cohesion (is this the right word?) between Jira tasks and our code. So in the future we could understand what happened here (why task requirements are different from code) without looking for closed PR and reading all comments in it.

drjova: In that case is a problem that can be solved immidietly in the review, which is a good. I'll update the US with this information.

pazembrz: :+1:

backend/tests/integration/records/serializers/json/test_experiments.py
```diff
@@ -94,186 +94,23 @@ def test_experiments_json_search(inspire_app, datadir):
     assert expected_updated == response_updated
 
 
-def test_experiments_detail(inspire_app, datadir):
```
michamos, line 97: was all of this removed on purpose?

pazembrz: Ouch, this looks like my error during resolving conflict. Thanks for spotting this :+1:

PR #990, opened 2020-06-02
seminars: export to calendar

harunurhan:


ui/src/setupTests.js
```diff
@@ -34,3 +34,11 @@ global.window.matchMedia = jest.fn().mockImplementation(query => ({
   removeEventListener: jest.fn(),
   dispatchEvent: jest.fn(),
 }));
+
+global.window.location = {
+  origin: 'http://localhost:3000',
+  host: 'localhost:3000',
+  protocol: 'http:',
```
drjova, line 41:
```suggestion
  protocol: 'http',
```

harunurhan: actually original `window.location.protocol` has `:` suffix.

PR #994, opened 2020-06-02
bibgen: generate bibliography tool

miguelgrc:
* INSPIR-3420


backend/inspirehep/tools/views.py
```diff
@@ -0,0 +1,58 @@
...
+def allowed_file(filename):
+    return "." in filename and filename.rsplit(".", 1)[1].lower() in ALLOWED_EXTENSIONS
+
+
+@blueprint.route("/bibgen", methods=["POST"])
+def generate_bibliography():
+
+    requested_format = request.args.get("format")
+
+    bytes_file = request.files["file"]
```
michamos, line 30: how does this work? is the request streamed? what happens if I send a 20 GB (so larger than the web node memory) file?

backend/inspirehep/tools/views.py
```diff
@@ -0,0 +1,58 @@
...
+    key = hash_data(file_data)
+    file_data = BytesIO(file_data)
+    filename = get_filename(bytes_file.filename, requested_format)
+    mime_type = get_mimetype(requested_format)
+    bucket = current_app.config.get("S3_BIBGEN_BUCKET")
+
+    current_s3_instance.upload_file(
+        file_data, key, filename, mime_type, current_app.config["S3_FILE_ACL"], bucket
+    )
+    download_url = f"{current_app.config.get('S3_HOSTNAME')}/{bucket}/{key}"
```
michamos, line 49: this is incompatible with [INSPIR-3586](https://its.cern.ch/jira/browse/INSPIR-3586), so either we serve these files from S3 directly and not like the other files, or we need to amend that user story to handle these files.

backend/inspirehep/tools/views.py
```diff
@@ -0,0 +1,58 @@
...
+
+    for error in errors:
+        error = {
+            "message": f"Reference to {error['ref']} not found on line {error['line']}."
```
michamos, line 53: Confusing error mesage, as sometimes there are too many found references. These two cases should be distinguished.

backend/inspirehep/tools/views.py
```diff
@@ -0,0 +1,58 @@
...
+from inspirehep.records.utils import hash_data
+from inspirehep.tools.utils import get_filename, get_mimetype, get_references
+
+blueprint = Blueprint("inspirehep_tools", __name__, url_prefix="/tools")
+
+ALLOWED_EXTENSIONS = {"tex"}
+
+
+def allowed_file(filename):
+    return "." in filename and filename.rsplit(".", 1)[1].lower() in ALLOWED_EXTENSIONS
```
michamos, line 22: so we validate on filename? I thought we wanted to select also a MIME-type optionally in the browser (which doesn't necessarily agree with the file extension).

backend/inspirehep/tools/views.py
```diff
@@ -0,0 +1,58 @@
...
+
+from io import BytesIO, TextIOWrapper
+
+from flask import Blueprint, abort, current_app, jsonify, request
+
+from inspirehep.files.api import current_s3_instance
+from inspirehep.records.utils import hash_data
+from inspirehep.tools.utils import get_filename, get_mimetype, get_references
+
+blueprint = Blueprint("inspirehep_tools", __name__, url_prefix="/tools")
```
michamos, line 16: note that we have https://inspirehep.net/tools` going to `next` currently. If you do the same path on the UI this will make for fun routing.

backend/inspirehep/tools/views.py
miguelgrc: Can we have such a case? We could save the file and work with it, then upload it to S3 and delete it from the node's storage. Is it worth it?

backend/inspirehep/tools/views.py
miguelgrc: I guess we can just serve them from S3 directly. In any case it wouldn't be a big change to serve them from inspire once we do it for the other files.

backend/inspirehep/tools/views.py
miguelgrc: The refinement didn't differentiate between both cases, but I can change this, I also think it'd be better and easy.

backend/inspirehep/tools/views.py
miguelgrc: As @drjova said allowing just certain extensions is the common thing to do. If we want to filter by MIME-type we can do it in the frontend. We could allow some formats such as as .tex or .txt but I don't think it makes sense to allow a .js or something else. What do you think?

backend/inspirehep/tools/views.py
michamos: no, I'm not saying we'll have actual tex files containing 20GB of data. All I'm saying is that I can create such a file, slap a `.tex` extension on it and submit it. So you should probably limit file size to some reasonable size (10MB seems plenty, even less should be ok).

backend/inspirehep/tools/views.py
```diff
@@ -0,0 +1,69 @@
...
+
+ALLOWED_EXTENSIONS = {"tex"}
+
+
+def allowed_file(filename):
+    return "." in filename and filename.rsplit(".", 1)[1].lower() in ALLOWED_EXTENSIONS
+
+
+@blueprint.route("/bibgen", methods=["POST"])
```
michamos, line 30: why so short? I would suggest something more explicit like
```suggestion
@blueprint.route("/bibliography-generation", methods=["POST"])
```

backend/inspirehep/tools/views.py
michamos: I think you should add something to that user story then about this, so we don't lose track of it.

backend/inspirehep/tools/views.py
```diff
@@ -0,0 +1,69 @@
...
+    file_data = "\n\n".join(references).encode("utf-8")
+    key = hash_data(file_data)
+    file_data = BytesIO(file_data)
+    filename = get_filename(bytes_file.filename, requested_format)
+    mime_type = get_mimetype(requested_format)
+    bucket = current_app.config.get("S3_BIBGEN_BUCKET")
+
+    current_s3_instance.upload_file(
+        file_data, key, filename, mime_type, current_app.config["S3_FILE_ACL"], bucket
+    )
```
michamos, line 54: what's the plan for garbage collecting these temporary files? we ignore the issue? we add an S3 lifecycle to the bucket to expire objects after some time? https://docs.aws.amazon.com/AmazonS3/latest/dev/object-lifecycle-mgmt.html

miguelgrc: We could set an expiration policy of one day for example. I'll have a look at it.

backend/inspirehep/tools/views.py
miguelgrc: I limited it to 10MB for this endpoint. Request size is limited to 100MB for the whole application anyway.

backend/inspirehep/tools/utils.py
```diff
@@ -0,0 +1,95 @@
...
+            continue
+        cites = re.findall(r"\\cite\s*\{(.*?)\}", line, re.DOTALL)
+        for ref_line in cites:
+            ref_list = ref_line.split(",")
+            for one_ref in ref_list:
+                one_ref = re.sub(r"\s", "", one_ref)
+                if re.match(r"^#\d{1,2}$", one_ref):
+                    continue
+                if one_ref not in references:
+                    references.append({"value": one_ref, "line": num})
```
michamos, line 39: I would:
* use a tuple here, so you can unpack the tuple easily while iterating
* transform this into a generator (documenting it in the function name/docstring), so you don't need to build the whole reference list ahead of time. Remove L25 and
```suggestion
                    yield (one_ref, num)
```

backend/inspirehep/tools/utils.py
```diff
@@ -0,0 +1,95 @@
...
+
+
+def find_references(references, requested_format):
+    display_format = FORMAT_TO_SOURCE_FIELD[requested_format]
+
+    ret = []
+    errors = []
+    for r in references:
+        ref = r["value"]
+        line = r["line"]
```
michamos, line 51: then you can do
```suggestion
    for ref, line in references:
```

backend/inspirehep/tools/utils.py
```diff
@@ -0,0 +1,95 @@
...
+
+def get_mimetype(requested_format):
+    if requested_format == "bibtex":
+        return "application/x-bibtex"
+    return "application/x-latex"
+
+
+def get_filename(filename, requested_format):
+    extension = ".bib" if requested_format == "bibtex" else "-references.tex"
+    name = filename.rsplit(".", 1)[0]
```
michamos, line 94: I don't think you want to split on the first dot. What about `my.perfectly.fine.filename.tex`? this is not JS so you don't need to invent this kind of stuff, Python has a [standard library](https://docs.python.org/3/library/os.path.html#os.path.splitext) :wink:

backend/inspirehep/tools/views.py
michamos: See previous comment, splitting on the first dot is not a good idea.

backend/tests/integration/tools/test_views.py
```diff
@@ -0,0 +1,123 @@
...
+        "%\\cite{1721863}\n\\bibitem{1721863}\nJ.~R.~Ellis,\n%``Baryon Number Generation in Grand Unified Theories,''\n%0 citations counted in INSPIRE as of 12 Jun 2020",
+        "%\\cite{Beacom:2004yd}\n\\bibitem{Beacom:2004yd}\nJ.~F.~Beacom,\n%``Neutrinoless universe,''\n%0 citations counted in INSPIRE as of 12 Jun 2020",
+        "%\\cite{1721865}\n\\bibitem{1721865}\nZ.~Bern,\n%``On-shell recurrence relations for one-loop QCD amplitudes,''\n[arXiv:hep-th/0501240 [hep-th]].\n%0 citations counted in INSPIRE as of 12 Jun 2020",
+    ]
+
+    expected_references_latex_eu = [
+        "%\\cite{1721863}\n\\bibitem{1721863}\nJ.~R.~Ellis,\n%``Baryon Number Generation in Grand Unified Theories,''\n%0 citations counted in INSPIRE as of 12 Jun 2020",
+        "%\\cite{Beacom:2004yd}\n\\bibitem{Beacom:2004yd}\nJ.~F.~Beacom,\n%``Neutrinoless universe,''\n%0 citations counted in INSPIRE as of 12 Jun 2020",
+        "%\\cite{1721865}\n\\bibitem{1721865}\nZ.~Bern,\n%``On-shell recurrence relations for one-loop QCD amplitudes,''\n[arXiv:hep-th/0501240 [hep-th]].\n%0 citations counted in INSPIRE as of 12 Jun 2020",
+    ]
```
michamos, line 107: I think you need to freeze the time here, as the results contain the date

backend/inspirehep/tools/utils.py
miguelgrc: The problem with this is that I need to check if the reference is repeated.

ui/src/common/routes.js
```diff
@@ -47,6 +47,8 @@ export const EDIT_CONFERENCE = '/editor/record/conferences';
 export const EDIT_INSTITUTION = '/editor/record/institutions';
 export const EDIT_SEMINAR = SUBMISSIONS_SEMINAR;
 
+export const BIBGEN = '/bibliography-generator';
```
drjova, line 50: could we name it `BIBLIOGRAPHY_GENERATOR` instead of `BIBGEN`?

backend/inspirehep/files/config.py
```diff
@@ -11,6 +11,7 @@
 S3_SECRET_KEY = "CHANGE_ME"
 S3_BUCKET_PREFIX = "inspire-files-"
 S3_FILE_ACL = "public-read"
+S3_BIBGEN_BUCKET = "inspire-tmp"
```
drjova, line 14:
```suggestion
S3_BIBLIOGRAPHY_GENERATOR_BUCKET = "inspire-tmp"
```

backend/inspirehep/tools/views.py
```diff
@@ -0,0 +1,77 @@
...
+    get_references,
+)
+
+blueprint = Blueprint("inspirehep_tools", __name__, url_prefix="")
+
+ALLOWED_EXTENSIONS = {".tex"}
+
+
+def allowed_file(filename):
+    return splitext(filename)[1] in ALLOWED_EXTENSIONS
```
drjova, line 28:
```suggestion
    _, extension = splitext(filename)
    return extension  in ALLOWED_EXTENSIONS
```

Just a bit cleaner

backend/inspirehep/tools/views.py
```diff
@@ -0,0 +1,77 @@
...
+
+
+@blueprint.route("/bibliography-generator", methods=["POST"])
+def generate_bibliography():
+    cl = request.content_length
```
drjova, line 33:
```suggestion
    content_length = request.content_length
```

backend/tests/integration/tools/test_views.py
```diff
@@ -0,0 +1,125 @@
...
+
+import pytest
+from freezegun import freeze_time
+from helpers.utils import create_record
+from werkzeug.datastructures import FileStorage
+
+from inspirehep.tools.utils import find_references
+
+
+@pytest.fixture(scope="function")
```
drjova, line 16:
```suggestion
@pytest.fixture(scope="function")
@freeze_time("2020-06-12")
```

intead of ``with`` :)

miguelgrc: It doesn't work if I use both decorators, that's why I had to use the `with`

miguelgrc:
Is it worth it to rename all the variables and files? I think it's understandable like this right?

drjova:
tbh it reminds naming of the legacy version. We have everywhere clear names i.e. `literature` and not `lit` or `literatureAuthor` and not `litAut`, I know it's a bit of extra effort but it will make this feature consistent with the rest.

PR #996, opened 2020-06-02
ui: cancel unncessary http requests

harunurhan:
Adds a wrapper around http client that initially handles `get` requests.
This fixes problems with inconsistent ui especially when user on a
slow network, also avoid unnecessary downlaods.

Example problem:
1. Click on an author on literature page
2. Go back to literature page before citation summary of the author is loaded

This will trigger another citation summary request and
now there will be 2 active citation summary requests.
It will display wrong results if for some reason earlier request
finishes the last.


harunurhan:
@drjova I was thinking about unit testing the wrapper, but then seemed a bit of an overkill but I could give it a try, it could be straight forward.

And you're right, `httpErrorPayload` is missing the cancellation error case.

PR #999, opened 2020-06-03
experiments: add number of papers

MJedr:
* INSPIR-3436
* requires remapping
* requires alembic upgrade
* requires remigration


backend/inspirehep/alembic/afe5f484abcc_add_experiments_literature_table.py
```diff
@@ -0,0 +1,63 @@
+#
+# This file is part of Invenio.
+# Copyright (C) 2016-2018 CERN.
+#
+# Invenio is free software; you can redistribute it and/or modify it
+# under the terms of the MIT License; see LICENSE file for more details.
+
+"""add  experiments_literature table"""
```
monaawi, line 8: ```experiment_literature```

backend/inspirehep/records/api/base.py
```diff
@@ -443,6 +445,15 @@ def linked_institutions_pids(self):
             )
         )
 
+    @property
+    def linked_experiments_pids(self):
+        return set(
+            chain.from_iterable(
```
monaawi, line 451: since it only uses one field, you don't need all of this here. You can simply return the ```get_linked_pids_from_field``` of that field

backend/inspirehep/records/api/mixins.py
```diff
@@ -553,3 +554,70 @@ def get_modified_institutions_uuids(self):
         pids_changed = set.symmetric_difference(set(pids_latest), set(pids_previous))
 
         return list(self.get_records_ids_by_pids(list(pids_changed)))
+
+
+class ExperimentPapersMixin:
```
monaawi, line 559: These relation mixin classes are starting to be exactly the same. We should think of a way to refactor them and extract the common logic. I'll add a separate task for that though

backend/tests/helpers/utils.py
```diff
@@ -97,7 +97,6 @@ def create_record(record_type, data=None, **kwargs):
         )
     """
     accepted_record_types = current_app.config["PID_TYPE_TO_INDEX"].keys()
-
```
monaawi, line 100: leftover

backend/tests/integration-async/migrator/test_migrator_tasks.py
```diff
@@ -593,3 +590,62 @@ def test_migrate_recids_from_mirror_all_only_with_literature_author_and_invalid(
...
+    data = faker.record("lit", with_control_number=True)
+
+    data["accelerator_experiments"] = [
+        {"legacy_name": "LIGO", "record": {"$ref": exp_ref}}
+    ]
+
+    record = InspireRecord.create(data, disable_relations_update=True)
+    db.session.commit()
+
+    task = update_relations.delay([record.id])
```
monaawi, line 614: the test says it is testing the ```process_references_in_records``` not the ```update_relations```. You should also disconnect the signal to ES because otherwise the test is useless. You are asserting that the record is indexed but it would have been indexed from the beginning without calling the reindex task.

backend/tests/integration/alembic/test_alembic.py
```diff
@@ -14,6 +14,15 @@
 def test_downgrade(inspire_app):
     alembic = Alembic(current_app)
 
+    alembic.downgrade(target="afe5f484abcc")
+
```
monaawi, line 18: can you assert also that the ```experiment_literature``` table is not there?

backend/tests/integration/alembic/test_alembic.py
```diff
@@ -14,6 +14,15 @@
...
+    assert "ix_experiment_literature_literature_uuid" not in _get_indexes(
+        "experiment_literature"
+    )
+    assert "ix_experiment_literature_experiment_uuid" not in _get_indexes(
+        "experiment_literature"
+    )
+
     alembic.downgrade(target="8ba47044154a")
     assert "ix_records_authors_id_type_record_id" not in _get_indexes("records_authors")
```
monaawi, line 27: you should also add the test for the upgrade

backend/tests/integration/records/api/test_api_mixins.py
```diff
@@ -437,3 +437,133 @@ def test_self_citations_in_detail_view_not_logged_user(
...
+    experiment_2_papers = experiment_2.model.experiment_papers
+    lit_record_experiments = rec.model.experiments
+
+    assert len(experiment_papers) == 0
+    assert len(experiment_2_papers) == 1
+    assert len(lit_record_experiments) == 1
+    assert lit_record_experiments[0].experiment_uuid == experiment_2.id
+
+
+def test_record_with_experiment_adds_only_linked_ones_in_institution_lit_table(
```
monaawi, line 493: experiment_lit

backend/tests/integration/records/api/test_api_mixins.py
```diff
@@ -437,3 +437,133 @@ def test_self_citations_in_detail_view_not_logged_user(
...
+    inspire_app,
+):
+    ref = "http://localhost:8000/api/experiments/1234"
+    rec_data = {
+        "accelerator_experiments": [{"legacy_name": "LIGO", "record": {"$ref": ref}}]
+    }
+    rec = create_record("lit", rec_data)
+
+    assert len(rec.model.institutions) == 0
```
monaawi, line 502: rec.model.experiments

backend/tests/integration/records/api/test_api_mixins.py
```diff
@@ -437,3 +437,133 @@ def test_self_citations_in_detail_view_not_logged_user(
...
+):
+    experiment = create_record("exp")
+    experiment_control_number = experiment["control_number"]
+    ref = f"http://localhost:8000/api/experiments/{experiment_control_number}"
+    experiment.delete()
+    rec_data = {
+        "accelerator_experiments": [{"legacy_name": "LIGO", "record": {"$ref": ref}}]
+    }
+    rec = create_record("lit", rec_data)
+    assert len(rec.model.institutions) == 0
```
monaawi, line 516: rec.model.experiments

ui/src/experiments/components/ExperimentItem.jsx
```diff
@@ -29,6 +34,19 @@ function ExperimentItem({ metadata }) {
           )}
         </>
       }
+      rightActions={
+        <ListItemAction>
+          <Link to={`${EXPERIMENTS}?q=${getPapersQueryString(recordId)}`}>
```
monaawi, line 39: this link should be to LITERATURE

ui/src/experiments/components/ExperimentItem.jsx
```diff
@@ -29,6 +34,19 @@ function ExperimentItem({ metadata }) {
...
+            <IconText
```
monaawi, line 40: we could also extract this in a component which takes the literature query, we have it in multiple places already

ui/src/experiments/components/__tests__/__snapshots__/ExperimentItem.test.jsx.snap
```diff
@@ -17,6 +17,18 @@ exports[`ExperimentItem renders with all props set 1`] = `
...
     </React.Fragment>
   }
+  rightActions={
+    <ListItemAction>
+      <ForwardRef
+        to="/experiments?q=accelerator_experiments.record.$ref:1234"
+      >
+        <IconText
+          icon={<ForwardRef(LoginOutlined) />}
+          text="0 papers"
```
monaawi, line 27: can you add to one of the tests metadata a number of papers?

backend/tests/integration/records/api/test_api_experiments.py
```diff
@@ -165,3 +166,67 @@ def test_aut_citation_count_property_blows_up_on_wrong_pid_type(inspire_app):
 
     with pytest.raises(AttributeError):
         record.citation_count
+
+
+def test_deleted_institution_deletes_relations_in_institution_literature_table(
```
monaawi, line 171: experiment

backend/tests/integration/records/api/test_api_experiments.py
```diff
@@ -165,3 +166,67 @@ def test_aut_citation_count_property_blows_up_on_wrong_pid_type(inspire_app):
...
+    inspire_app,
+):
+    experiment = create_record("exp")
+    institution_control_number = experiment["control_number"]
```
monaawi, line 175: same

ui/src/experiments/components/__tests__/ExperimentItem.test.jsx
```diff
@@ -8,11 +8,12 @@ describe('ExperimentItem', () => {
     const metadata = fromJS({
       legacy_name: 'Experiment new',
       control_number: 1234,
+      number_of_papers: 99,
       institutions: [
         {
           value: 'CERN',
           record: {
-            $ref: 'https://inspirehep.net/api/institutions/902725',
+            $ref: 'https://inspirehep.net/experiments/1318020',
```
monaawi, line 16: why did you change this?

ui/src/common/components/LiteratureQueryLink.jsx
```diff
@@ -0,0 +1,16 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+
+import { Link } from 'react-router-dom';
+import { LITERATURE } from '../routes';
+
+const LiteratureQueryLink = ({ children, recordId }) => (
```
monaawi, line 7: I would make a component that has also the IconText part since it is also shared and just get the text as prop

monaawi: Actually, I just saw your comment that it is the same as ```CitationLinkAction```. I think you can generalize that one, we actually could use the formattedNumber in the other places as well

MJedr: Yes, I think that's the best idea

ui/src/common/components/CitationsPapersLink.jsx
```diff
@@ -0,0 +1,31 @@
...
+import { LoginOutlined } from '@ant-design/icons';
+
+import { Link } from 'react-router-dom';
+import { LITERATURE } from '../routes';
+import IconText from './IconText';
+import FormattedNumber from './FormattedNumber';
+import pluralizeUnlessSingle from '../utils';
+
+const CitationsPapersLink = ({ recordId, itemCount, displayedName }) => (
+  <Link to={`${LITERATURE}?q=refersto:recid:${recordId}`}>
```
harunurhan, line 12: this link will work only for `citations` but it's used for other stuff, if you look at the snapshots from the tests, we need different queries.

ui/src/common/components/CitationsPapersIcon.jsx
```diff
@@ -0,0 +1,26 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import { LoginOutlined } from '@ant-design/icons';
+
+import IconText from './IconText';
+import FormattedNumber from './FormattedNumber';
+import pluralizeUnlessSingle from '../utils';
+
+const CitationsPapersIcon = ({ itemCount, displayedName }) => (
```
harunurhan, line 9: name doesn't make sense for `experiments papers` right? or `conference papers` which is not using at the moment but it should.

On the other hand better approach would be to single component for all.
(name could be different)
```
<IncomingLiteratureReferencesLinkAction
  count={1234}
  referenceType="contribution or paper or citation etc"
  linkQuery={...the query that will be appended to the literature link...}
/>
```

ui/src/literature/components/CitationsLinkAction.jsx
```diff
@@ -1,30 +1,19 @@
...
-            text={
-              <>
-                <FormattedNumber>{citationCount}</FormattedNumber>{' '}
-                {pluralizeUnlessSingle('citation', citationCount)}
-              </>
-            }
-            icon={<LoginOutlined />}
-          />
-        </Link>
+        <IncomingLiteratureReferencesLinkAction
```
monaawi, line 12: you should move the ```EventTracker``` and the ```ListItemAction``` inside the ```IncomingLiteratureReferencesLinkAction```. Even though ```EventTracker``` wasn't added in all the places you are using this, it would be good to add it.

ui/src/common/components/__tests__/IncomingLiteratureReferencesLinkAction.test.jsx
```diff
@@ -0,0 +1,20 @@
+import React from 'react';
+import { shallow } from 'enzyme';
+import IncomingLiteratureReferencesLinkAction from '../IncomingLiteratureReferencesLinkAction';
+
+describe('IncomingLiteratureReferencesLinkAction', () => {
+  it('renders with required props', () => {
```
monaawi, line 6: can add a test that has a single itemCount so we know it is using the ```pluralizeUnlessSingle```?

ui/src/common/components/__tests__/IncomingLiteratureReferencesLinkAction.test.jsx
```diff
@@ -0,0 +1,20 @@
...
+    const itemCount = 29;
+    const displayedName = 'citation';
+    const recordId = 888;
+    const linkQuery = `refersto:recid:${recordId}`;
+    const wrapper = shallow(
+      <IncomingLiteratureReferencesLinkAction
+        itemCount={itemCount}
+        linkQuery={linkQuery}
+        displayedName={displayedName}
```
monaawi, line 15: Also, I would rename this to ```referenceType``` as Harun suggested. It goes better with the name of the component

PR #1000, opened 2020-06-03
self-citations: fix in tooltip info and add new feature flag

MJedr:
* INSPIR-3557


ui/src/common/components/NewFeatureTag/NewFeatureTag.jsx
```diff
@@ -1,17 +1,26 @@
...
 
 import './NewFeatureTag.scss';
 
 const GREEN = styleVariables['success-color'];
 
-function NewFeatureTag() {
+function NewFeatureTag({ className }) {
   return (
-    <Tag className="__NewFeatureTag__" color={GREEN}>
+    <Tag className={className} color={GREEN}>
```
harunurhan, line 12: This is how you add `className` like props:

https://github.com/inspirehep/inspirehep/blob/a59416dab0e592c35b2b38d84412ad0cdd75016d/ui/src/common/components/InlineList/InlineList.jsx#L27

So here would be:

`className={classnames(''__NewFeatureTag__", className)`

and you define the actual class that you're passing, in the parent.

ui/src/common/components/NewFeatureTag/NewFeatureTag.jsx
```diff
@@ -1,17 +1,27 @@
...
       New
     </Tag>
   );
 }
 
+NewFeatureTag.propTypes = {
+  className: PropTypes.string,
+};
+
+NewFeatureTag.defaultProps = {
```
harunurhan, line 23: no needed

ui/src/common/components/NewFeatureTag/NewFeatureTag.scss
```diff
@@ -9,3 +9,7 @@
     opacity: 1;
   }
 }
+
+.NoMargin {
```
harunurhan, line 13: so this class should in the component where you're setting not here.

also `without-margin-left` would be a better name.

ui/src/common/components/NewFeatureTag/NewFeatureTag.jsx
```diff
@@ -1,17 +1,27 @@
...
   return (
-    <Tag className="__NewFeatureTag__" color={GREEN}>
+    <Tag className={classnames('__NewFeatureTag__', className)} color={GREEN}>
       New
     </Tag>
   );
 }
 
+NewFeatureTag.propTypes = {
+  className: PropTypes.string,
```
harunurhan, line 20: although there is not a lot of value in that, but you can create a new test now like "renders with class name" and snapshot `<NewFeatureTag className="test"...`

PR #1001, opened 2020-06-03
ui: simplify onQueryChange callbacks

harunurhan:
Also fixes going back issue when a search link is clicked

### DIFF
--- a/ui/src/actions/search.js
+++ b/ui/src/actions/search.js
@@ -92,7 +92,11 @@ export function fetchSearchAggregations(namespace, url) {
   };
 }
 
-export function searchQueryUpdate(namespace, query) {
+export function searchQueryUpdate(
+  namespace,
+  query,
+  dueToNavigationToSearchPage = false
+) {
   return async (dispatch, getState) => {
     const currentQuery = getState().search.getIn([
       'namespaces',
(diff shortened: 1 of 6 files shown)


ui/src/search/helper.js
```diff
@@ -32,6 +42,35 @@ export default class SearchHelper {
     );
   }
 
+  updateLocation(urlSuffix = '') {
+    const searchUrl = `${this.getPathname()}?${this.getQueryString()}${urlSuffix}`;
+    if (this.isInitialQueryUpdate() || this.dueToNavigationToSearchPage) {
```
monaawi, line 47: So in the end we check for ```initialQueryUpdate``` even if it is embedded search? Like I said, it is worth it to be able to get rid of the code duplication. It looks better :+1:

ui/src/actions/search.js
```diff
@@ -116,8 +120,18 @@ export function searchQueryUpdate(namespace, query) {
...
+      namespace,
+      prevState,
+      nextState,
+      dispatch,
+      dueToNavigationToSearchPage
+    );
+    searchConfig[namespace].onQueryChange(
+      helper,
+      dispatch,
+      dueToNavigationToSearchPage
```
monaawi, line 133: how come we didn't need this before for the synclocation? I didn't understand what change broke this

ui/src/search/helper.js
harunurhan: no we don't.. embedded ones don't update location (yet)

ui/src/actions/search.js
harunurhan: https://github.com/inspirehep/inspirehep/blob/e233cb503fdf9e205aaa2c43197b59c96a681b99/ui/src/actions/search.js#L53-L64

That one.

PR #1002, opened 2020-06-03
tests: fix cassettes for mailing

drjova:


michamos:
So we did have a cassette already. Do you understand why it stopped matching?

drjova:
@michamos actually it was missing, I assert now that the cassette is played.

backend/tests/integration/mailing/test_api_jobs.py
```diff
@@ -54,13 +55,15 @@ def test_render_jobs_weekly_campaign_job_record_template_only(create_jobs):
 
 @pytest.mark.vrc()
 def test_subscirbe_to_the_list(inspire_app, vcr_cassette):
-    result = subscribe_to_jobs_weekly_list("luke@cage.com", "Luke", "Cage")
+    subscribe_to_jobs_weekly_list("luke@cage.com", "Luke", "Cage")
     assert vcr_cassette.all_played
 
 
-def test_subscirbe_to_the_list_with_invalid_email(inspire_app):
+@pytest.mark.vrc()
```
michamos, line 62: shouldn't this be `vcr` instead of `vrc`?

backend/tests/integration/mailing/test_api_jobs.py
```diff
@@ -54,13 +55,15 @@ def test_render_jobs_weekly_campaign_job_record_template_only(create_jobs):
...
+def test_subscirbe_to_the_list_with_invalid_email(inspire_app, vcr_cassette):
```
michamos, line 63:
```suggestion
def test_subscribe_to_the_list_with_invalid_email(inspire_app, vcr_cassette):
```

backend/tests/integration/mailing/test_api_jobs.py
```diff
@@ -54,13 +55,15 @@ def test_render_jobs_weekly_campaign_job_record_template_only(create_jobs):
 
 @pytest.mark.vrc()
 def test_subscirbe_to_the_list(inspire_app, vcr_cassette):
```
michamos, line 57:
```suggestion
def test_subscribe_to_the_list(inspire_app, vcr_cassette):
```

backend/tests/integration/mailing/test_api_jobs.py
```diff
@@ -54,13 +55,15 @@ def test_render_jobs_weekly_campaign_job_record_template_only(create_jobs):
...
+    subscribe_to_jobs_weekly_list("luke@cage.com", "Luke", "Cage")
     assert vcr_cassette.all_played
 
 
-def test_subscirbe_to_the_list_with_invalid_email(inspire_app):
+@pytest.mark.vrc()
+def test_subscirbe_to_the_list_with_invalid_email(inspire_app, vcr_cassette):
     with pytest.raises(ValueError):
         subscribe_to_jobs_weekly_list("luke", "Luke", "Cage")
+    assert vcr_cassette.all_played
```
michamos, line 66: maybe we should have something like this in the `conftest`? we do for `orcid` tests.

backend/tests/integration/mailing/test_api_jobs.py
drjova: good catch, this was the actual reason why there was no recording until I added the  ``vcr_cassette`` fixture.

PR #1008, opened 2020-06-08
ui: add minimal ics utility to replace `ics` package

harunurhan:
`ics` package introduces lots of dependencies that are not relevent to our use case


ui/package.json
```diff
@@ -2,7 +2,7 @@
   "name": "inspire-next-react",
   "license": "GPL-2.0",
   "version": "0.1.19",
-  "proxy": "http://localhost:8000",
+  "proxy": "https://inspirebeta.net",
```
drjova, line 5: leftover?

PR #1010, opened 2020-06-09
serializers: do not display search_options on non ui requests

pazembrz:
*INSPIR-3605

### DIFF
--- a/backend/tests/integration/records/serializers/json/test_authors.py
+++ b/backend/tests/integration/records/serializers/json/test_authors.py
@@ -266,11 +266,10 @@ def test_authors_search_json_does_not_have_sort_options(inspire_app):
         response = client.get("/authors", headers=headers)
 
     response_status_code = response.status_code
-    response_data = json.loads(response.data)
-    sort_options = response_data["sort_options"]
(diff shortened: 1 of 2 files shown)


backend/tests/integration/records/serializers/json/test_authors.py
```diff
@@ -266,11 +266,10 @@ def test_authors_search_json_does_not_have_sort_options(inspire_app):
         response = client.get("/authors", headers=headers)
 
     response_status_code = response.status_code
-    response_data = json.loads(response.data)
-    sort_options = response_data["sort_options"]
+    response_data = response.json
 
     assert expected_status_code == response_status_code
-    assert expected_sort_options == sort_options
+    assert "sort_options" not in response_data
```
drjova, line 272: also a test for `ui` serializer, to make sure it contains the ``sort_options``

pazembrz: We already have this tests for every type which should have sort_options https://github.com/inspirehep/inspirehep/search?q=sort_options&unscoped_q=sort_options

PR #1011, opened 2020-06-09
ui: separate storage

harunurhan:
This is prior work before we move to IndexedDB or some wrapper like
`localForage`.

Current blocker is rehydration of redux state async, requires more work
and can wait until we really needed.

### DIFF
--- a/ui/src/common/storage.js
+++ b/ui/src/common/storage.js
@@ -1,0 +1,52 @@
+// adapted from https://developer.mozilla.org/en-US/docs/Web/API/Web_Storage_API/Using_the_Web_Storage_API#Testing_for_availability
+function isStorageAvailable() {
+  let storage;
+  try {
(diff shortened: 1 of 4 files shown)


ui/src/common/storage.js
```diff
@@ -0,0 +1,52 @@
...
+}
+
+const storage = {
+  isAvailable: isStorageAvailable(),
+
+  getSync(key) {
+    // eslint-disable-next-line react/no-this-in-sfc
+    if (this.isAvailable) {
+      const rawValue = localStorage.getItem(key);
+      return JSON.parse(rawValue);
```
drjova, line 30: `try catch`?

harunurhan: for what? `parse`?

drjova: yes for parse, true there is no much we can do in that case

PR #1012, opened 2020-06-09
hep: use lowercase normalizer for author bais

harunurhan:


michamos:
Note that this will require remapping.

harunurhan:
@michamos I guess I don't have to put this in commit message, since mapping change will be visible

PR #1022, opened 2020-06-11
indexer: reindex records which are citing record with changed BAI

pazembrz:
*INSPIR-3618

### DIFF
--- a/backend/inspirehep/records/api/mixins.py
+++ b/backend/inspirehep/records/api/mixins.py
@@ -347,10 +347,16 @@ class CitationMixin(PapersAuthorsExtensionMixin):
         prev_version = self._previous_version
         current_authors = set(self.get_authors_bais())
         old_authors = set(prev_version.get_authors_bais())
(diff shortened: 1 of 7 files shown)


backend/inspirehep/records/api/mixins.py
```diff
@@ -364,17 +370,25 @@ def get_all_connected_records_uuids_of_modified_authors(self):
...
+        diff = current_collaborations.symmetric_difference(old_collaborations)
+        connected_papers = set()
         if diff:
-            differed_papers_uuids = [
+            citers = {
+                citer[0]
+                for citer in RecordCitations.query.filter_by(cited_id=self.id)
+                .with_entities("citer_id")
+                .all()
+            }
```
michamos, line 390: you're doing the same query twice potentially, also for modified authors (not dramatic, but suboptimal).

pazembrz: This query is small and fast and it's easy to do it this way, at the end there won't be any duplicates as results are de-duplicated

PR #1024, opened 2020-06-12
search: allow author search to perform query string query

MJedr:
* INSPIR-3587

### DIFF
--- a/backend/inspirehep/search/api.py
+++ b/backend/inspirehep/search/api.py
@@ -194,15 +194,22 @@ class AuthorsSearch(InspireSearch):
     def query_from_iq(self, query_string):
         if not query_string:
             return self.query()
-
-        names_analyzed_query = Q("match", names_analyzed=query_string)
-        names_analyzed_initials_query = Q("match", names_analyzed_initials=query_string)
-        query_string = Q("query_string", query=query_string)
-        query = Q(
-            "bool",
-            should=[names_analyzed_query, names_analyzed_initials_query, query_string],
-        )
-        return self.query(query)
+        query_string_query = Q("query_string", query=query_string)
+        if ":" not in query_string:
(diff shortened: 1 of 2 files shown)


backend/inspirehep/search/api.py
```diff
@@ -194,15 +194,22 @@ class Meta:
...
-        names_analyzed_query = Q("match", names_analyzed=query_string)
-        names_analyzed_initials_query = Q("match", names_analyzed_initials=query_string)
-        query_string = Q("query_string", query=query_string)
-        query = Q(
-            "bool",
-            should=[names_analyzed_query, names_analyzed_initials_query, query_string],
-        )
-        return self.query(query)
+        query_string_query = Q("query_string", query=query_string)
+        if ":" in query_string:
```
drjova, line 198: it's the opposite, if there is `:` do a normal query :)

backend/inspirehep/search/api.py
```diff
@@ -194,15 +194,22 @@ class Meta:
...
+            query = Q(
+                "bool",
+                should=[
+                    names_analyzed_query,
+                    names_analyzed_initials_query,
+                    query_string_query,
+                ],
+            )
+            return self.query(query)
+        return self.query(query_string_query)
```
drjova, line 212:
```suggestion
        return self.query("query_string", query=query_string)
```

and remove https://github.com/inspirehep/inspirehep/pull/1024/files#diff-d1f02c1c80adf7a3e7d52ce17ea54705R197

MJedr: But it is also used in here: https://github.com/inspirehep/inspirehep/blob/55be643abe324762b3241c714da7e4508cc1db44/backend/inspirehep/search/api.py#L203, so either I define this query twice for each case, or once as now

drjova: My bad I missed that!

PR #1025, opened 2020-06-14
global: init of matcher

drjova:
* Addition of matcher module to enable search with on publication info
  using GROBID service.

* INSPIR-3444


backend/inspirehep/matcher/__init__.py
```diff
@@ -0,0 +1,7 @@
+# -*- coding: utf-8 -*-
```
michamos, line 1: I don't think this file is needed on Python 3.

backend/inspirehep/matcher/api.py
```diff
@@ -0,0 +1,117 @@
...
+from inspire_matcher import match
+from inspire_utils.dedupers import dedupe_list
+from inspire_utils.record import get_value
+
+from .utils import grobid_to_reference
+
+
+def get_reference_from_grobid(query):
+    data = {"citations": query}
+    url = current_app.config["GROBID_URL"]
```
michamos, line 20: maybe split this into base URL + resource, so we can reuse the GROBID URL for other stuff later (e.g. reference extraction from PDF)?

backend/inspirehep/matcher/api.py
```diff
@@ -0,0 +1,117 @@
...
+    result_xml = requests.post(url, data=data).content
```
michamos, line 21: You should check the HTTP status code here I guess (`response.raise_for_status()`).

backend/inspirehep/matcher/utils.py
```diff
@@ -0,0 +1,110 @@
...
+# the terms of the MIT License; see LICENSE file for more details.
+
+from bs4 import BeautifulSoup
+from inspire_dojson.utils import strip_empty_values
+from inspire_utils.helpers import force_list
+
+
+def get_arxiv_eprint(beatutiful_soup_xml_obj):
+    try:
+        return beatutiful_soup_xml_obj.find("idno", attrs={"type": "arxiv"}).text
```
michamos, line 15: use `xpath` with `parsel` instead of beautiful soup like we usually do, e.g. https://github.com/inspirehep/hepcrawl/blob/master/hepcrawl/parsers/jats.py? this will fail miserably if you have more than one arXiv reference. I would also use the same approach as there by having a class + properties with a `parse` method calling the builder on the properties.

backend/inspirehep/search/api.py
```diff
@@ -323,6 +358,16 @@ class Meta:
...
         doc_types = "_doc"
 
+    def normalize_title(self, journal_title):
+        normalized_journal_title = journal_title
+        hits = self.query("match", lowercase_journal_titles=journal_title).execute()
+        if hits:
+            try:
+                normalized_journal_title = hits[0].short_title
+            except (AttributeError, IndexError):
+                LOGGER.info("Failed to normalize journal title in", result=hits[0])
```
michamos, line 368: why would this happen? `short_title` is required by the schema. I think this is a real exception, not something to basically ignore by logging as `info`.

backend/inspirehep/matcher/api.py
```diff
@@ -0,0 +1,117 @@
...
+    Args:
+        reference (dict): the metadata of the reference.
+        config (dict): the list of inspire-matcher configurations for queries.
+        previous_matched_recid (int): the record id of the last matched
+            reference from the list of references.
+    Returns:
+        dict: the matched reference.
+    """
+    # XXX: avoid this type casting.
+    try:
```
pazembrz, line 44: Shouldn't we fix code which expects year to be string than cast this like that and then recasting it back to int at the end? This is really ugly.

backend/inspirehep/matcher/utils.py
drjova: It's great acutally that we have `ReferenceBuilder`, I should have checked

backend/inspirehep/matcher/api.py
drjova: I changed it a bit

backend/inspirehep/search/api.py
drjova: 👍  tbh I just copy paste the code from https://github.com/inspirehep/inspire-next/blob/d6898dfbe061977208a7a69be0d1e7e0c46a8a7d/inspirehep/utils/normalizers.py#L31. I should have refactored it.

drjova:
@michamos @pazembrz updated

backend/tests/unit/matcher/test_parsers.py
```diff
@@ -0,0 +1,105 @@
...
+<biblStruct >
+    <analytic>
+        <title/>
+        <author>
+            <persName xmlns="http://www.tei-c.org/ns/1.0"><surname>Graff</surname></persName>
+        </author>
+    </analytic>
+    <idno type="DOI">10.1007/JHEP11(2010)093</idno>
+    <idno type="arxiv">1009.1212</idno>
+    <idno type="report">ESS 2013-001</idno>
```
michamos, line 26: * are you sure you get `type="report"`? in the example I looked at it was just a bare `<idno>` without `type` attribute.
* I think you should use more realistic examples, the `idno` tags don't seem to be in the right place (they should be under `analytic` according to examples I looked at).

backend/inspirehep/matcher/parsers.py
```diff
@@ -0,0 +1,77 @@
...
+#
+# inspirehep is free software; you can redistribute it and/or modify it under
+# the terms of the MIT License; see LICENSE file for more details.
+
+from inspire_dojson.utils import strip_empty_values
+from inspire_schemas.api import ReferenceBuilder
+from parsel import Selector
+
+
+class GrobidParser:
```
michamos, line 13:
```suggestion
class GrobidReferenceParser:
```
Maybe also document that this parses a single reference (with root node `biblStruct`).

backend/inspirehep/matcher/config.py
```diff
@@ -0,0 +1,210 @@
+# -*- coding: utf-8 -*-
+#
+# Copyright (C) 2020 CERN.
+#
+# inspirehep is free software; you can redistribute it and/or modify it under
+# the terms of the MIT License; see LICENSE file for more details.
+
+GROBID_URL = "https://grobid.inspirebeta.net"
+GROBID_PROCESS_CITATION_URL = f"{GROBID_URL}/api/processCitation"
```
michamos, line 9: If you do it this way, you can't override only the `GROBID_URL` through the environment/other config file.

backend/inspirehep/matcher/parsers.py
```diff
@@ -0,0 +1,77 @@
...
+
+        for doi in self.dois:
+            self.builder.add_uid(doi)
+
+        self.builder.set_journal_issue(self.journal_issue)
+        self.builder.set_journal_volume(self.journal_volume)
+        self.builder.set_journal_title(self.journal_title)
+        self.builder.set_page_artid(page_start=self.page_start, page_end=self.page_end)
+        self.builder.set_year(self.year)
+        return strip_empty_values(self.builder.obj)
```
michamos, line 33: is the `strip_empty_values` really needed?

backend/inspirehep/matcher/parsers.py
```diff
@@ -0,0 +1,77 @@
...
+        return self.root.xpath("//idno[@type='isbn']/text()").get()
+
+    @property
+    def report_numbers(self):
+        return self.root.xpath("//idno[@type='report']/text()").getall()
+
+    @property
+    def journal_volume(self):
+        return self.root.xpath(
+            "//monogr//imprint//biblScope[@unit='volume']/text()"
```
michamos, line 54: why the double slashes everywhere? If you provide the full path, single slashes should be sufficient.

backend/inspirehep/search/api.py
```diff
@@ -118,19 +123,49 @@ class Meta:
...
+
+        :param query_string: Query string as a user would input in INSPIRE's search box.
+        :type query_string: string
+        :returns: Elasticsearch DSL search class
+        """
+        search_query = self.query_for_superuser_or_users(query_string)
+        search_results = search_query.execute()
+
+        if search_results.hits.total["value"] == 0 and request:
+            reference_match = self.match_reference_from_request_query()
```
michamos, line 145: maybe pass the query string to the method,
```suggestion
            query_string = request.values.get("q", "", type=str)
            reference_match = self.match_reference_from_query(query_string)
```
to make this less magical instead of doing it inside?

backend/inspirehep/search/api.py
```diff
@@ -118,19 +123,49 @@ class Meta:
...
+            if reference_match:
+                return self.query(Q("term", control_number=reference_match))
+        return search_query
+
+    def match_reference_from_request_query(self):
+        query_string = request.values.get("q", "", type=str)
+        if not query_string:
+            return None
+
+        reference = get_reference_from_grobid(query_string)
```
michamos, line 155: what about exceptions raised by wrong GROBID status code here?

backend/inspirehep/matcher/api.py
```diff
@@ -0,0 +1,83 @@
...
+    if len(matched_recids) == 1:
+        return matched_recids[0]
+    return None
+
+
+def match_reference_control_number(reference):
+    """Match a reference using inspire-matcher.
+    Args:
+        reference (dict): the metadata of a reference.
+        previous_matched_recid (int): the record id of the last matched
```
michamos, line 51: wrong docstring. You dropped things that were there previously apparently. Was that done on purpose? what's the plan for the matcher on `next`? is the idea to replace it completely by this one at some point? if so do we really want to drop features that we will need in the future?

backend/tests/unit/matcher/test_parsers.py
drjova: The examples are based on GROBID responses, some of them indeed I have enchached with more metadata to edge more cases but the structure is exaclty as GROBID responses. 

Could you please provide these examples with `analytic`? Please see below some of the queries that I used:

1. 
```bash
curl 'https://grobid.inspirebeta.net/api/processCitation' --data-raw 'citations=[28] K. Ito, H. Nakajima, T. Saka, and S. Sasaki, “N=2 Instanton Effective Action in
Ω-background and D3/D(-1)-brane System in R-R Background,” JHEP 11 (2010) 093,
arXiv:1009.1212 [hep-th].'
```
```xml
<biblStruct >
        <monogr>
                <title level="m" type="main">N=2 Instanton Effective Action in Ω-background and D3/D(-1)-brane System in R-R Background</title>
                <author>
                        <persName xmlns="http://www.tei-c.org/ns/1.0"><forename type="first">K</forename><surname>Ito</surname></persName>
                </author>
                <author>
                        <persName xmlns="http://www.tei-c.org/ns/1.0"><forename type="first">H</forename><surname>Nakajima</surname></persName>
                </author>
                <author>
                        <persName xmlns="http://www.tei-c.org/ns/1.0"><forename type="first">T</forename><surname>Saka</surname></persName>
                </author>
                <author>
                        <persName xmlns="http://www.tei-c.org/ns/1.0"><forename type="first">S</forename><surname>Sasaki</surname></persName>
                </author>
                <idno type="arXiv">arXiv:1009.1212</idno>
                <imprint>
                        <date type="published" when="2010" />
                        <biblScope unit="page">93</biblScope>
                </imprint>
        </monogr>
        <note>hep-th</note>
</biblStruct>
```

2. 

```bash
curl 'https://grobid.inspirebeta.net/api/processCitation' --data-raw 'citations=Phys. Lett. B 704 (2011) 223' 
```

```xml
        <analytic>
                <title/>
        </analytic>
        <monogr>
                <title level="j">Phys. Lett. B</title>
                <imprint>
                        <biblScope unit="volume">704</biblScope>
                        <biblScope unit="page">223</biblScope>
                        <date type="published" when="2011" />
                </imprint>
        </monogr>
</biblStruct>
```

backend/inspirehep/matcher/parsers.py
drjova: Yes, otherwise it returns `None` values in the keys, it should have been handled in the ``ReferenceBuilder`` I guess

drjova:
@michamos updated

backend/inspirehep/matcher/parsers.py
drjova: I wanted to avoid full path

backend/inspirehep/search/api.py
```diff
@@ -118,18 +124,58 @@ class Meta:
...
+
         :param query_string: Query string as a user would input in INSPIRE's search box.
         :type query_string: string
         :returns: Elasticsearch DSL search class
         """
-        if not is_superuser_or_cataloger_logged_in():
-            user_query = Q(
-                IQ(query_string, self) & Q("term", _collections="Literature")
+        search_query = self.query_for_superuser_or_users(query_string)
+        search_results = search_query.execute()
```
michamos, line 143: am I understanding correctly that we're going to dispatch every query twice? there's no better way to do this?

drjova: Yes, otherwise I have to do some really dirty hack in `execute` and I wanted to avoid that.

drjova: @michamos I'll add a feature flag at the moment, I'd like first to see if it actually works as we expecting and If it works I'll find a better solution.

backend/inspirehep/matcher/parsers.py
michamos: :cry:

PR #1026, opened 2020-06-15
indexer: replace _previous_version with _last_indexed

pazembrz:
* replace `_previous_version` with `_last_indexed` which is a db version of last indexed version of record (based on `_update` field)
 * indexer_task: move search of records to re-index before current record re-indexing.
 * integration_async: Update all tests which were checking `_previous_version` as now after re-indexing `_last_indexed` will show current_version (as current will be last indexed).
 * fixed `retry_until_matched` for expected_results which are logical False

* INSPIR-3618

### DIFF
--- a/backend/inspirehep/indexer/tasks.py
+++ b/backend/inspirehep/indexer/tasks.py
@@ -77,15 +77,6 @@ def index_record(self, uuid, record_version=None, force_delete=None):
     if not force_delete:
         deleted = record.get("deleted", False)
 
-    if force_delete or deleted:
-        try:
-            InspireRecordIndexer().delete(record)
-            LOGGER.debug("Record removed from ES", uuid=str(uuid))
-        except NotFoundError:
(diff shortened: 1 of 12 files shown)


drjova:
@pazembrz maybe I missed parts of your discussions, but what's the purpose of this PR? Seems to be missmatching ES with DB

pazembrz:
This should fix problem described by me here: https://inspirehep.zulipchat.com/#narrow/stream/122202-devel/topic/indexing.20incosistencies/near/200515453

@drjova What you mean by 
>  Seems to be missmatching ES with DB
This should actually help to keep es in sync with DB

michamos:
I'm not sure this makes sense, as you'll run into atomicity issue (the version on ES changes between reading from ES and computing the records to reindex).

pazembrz:
Still this should be less problematic than current way how we are doing this. In worst case we will re-index twice same referenced records. Currently in worst case we are not re-indexing any referenced records.

PR #1029, opened 2020-06-17
docker-compose:speedup local docker build

pazembrz:
docker-compose:speedup local docker build
    
     * Build local web and next from released hep image
     * Update circle-ci config to use more recent ubuntu with latest docker and docker-compose
     * Add --parallel option on building images on circle-ci


run-e2e.sh
```diff
@@ -1,9 +1,9 @@
...
-docker-compose -f ./docker-compose.core.yml exec web-next inspirehep db create
-docker-compose -f ./docker-compose.core.yml exec web inspirehep importer records -d data/records/authors
-docker-compose -f ./docker-compose.core.yml exec web inspirehep importer records -f data/records/conference/7577512.json
-docker-compose -f ./docker-compose.core.yml -f ./e2e/docker-compose.e2e.yml run --rm e2e bash -c "cd /opt/e2e && yarn && yarn test" || rc=$?
+docker-compose up -d --force-recreate
+docker-compose exec web ./scripts/setup
+docker-compose exec web-next inspirehep db create
+docker-compose exec web inspirehep importer records -d data/records/authors
+docker-compose exec web inspirehep importer records -f data/records/conference/7577512.json
+docker-compose -f ./docker-compose.yml -f ./docker-compose.override.yml -f ./e2e/docker-compose.e2e.yml run --rm e2e bash -c "cd /opt/e2e && yarn && yarn test" || rc=$?
```
michamos, line 8: IIUC, `docker-compose.core.yml`->`docker-compose.yml` and `docker-compose.dev.yml` -> `docker-compose.override.yml`. So here it looks like you're adding the `override` file which was not there previously. Is that on purpose?

README.md
```diff
@@ -271,5 +271,5 @@ $ ./docker-inspire run --rm web poetry run inspirehep importer records -d data/r
 # Local
 $ poetry run inspirehep importer records -f data/records/literature/374836.json -f data/records/authors/999108.json
 # Docker
-$ ./docker-inspire run --rm web poetry run inspirehep importer records -f data/records/literature/374836.json -f data/records/authors/999108.json
+$ docker-compose run --rm web poetry run inspirehep importer records -f data/records/literature/374836.json -f data/records/authors/999108.json
```
michamos, line 274: it looks like `web` is defined in `override`. Does this really work?

drjova:
@pazembrz could we also move `kibana` and `flower`?

run-e2e.sh
pazembrz: Yes, override file allows to override config from docker-compose file https://docs.docker.com/compose/extends/#understanding-multiple-compose-files 
Thanks to this change we don't have to use ./docker-inspier script anymore. 
Here I'm providing override as in e2e tests I would like to use quicker build, but If you think that we should use full build then it can be removed.

README.md
pazembrz: yes: https://docs.docker.com/compose/extends/#understanding-multiple-compose-files

pazembrz:
Where you would like them to be moved?

README.md
pazembrz: in short: default `docker-compose` command is actually doing same thing as `docker-compose -f docker-compose.yml -f docker-compose.override.yml`

run-e2e.sh
pazembrz: Ok, I noticed that running e2e with dev dockerfile is not working unfortunately. Restoring to previous e2e setup.

drjova:
@pazembrz do we have an indication how much faster it is? :)

pazembrz:
On local build my Laptop is now building web/worker in around 1 minute before it was taking 10 minutes.
UI - no change as there is nothing to do unfortunately so on my Laptop it still takes 6 minutes

On Circle CI it went down from around 25 minutes  to around 18-19 minutes.

You can also test it on your machine as I know that macs are slower with docker (for Miguel whole build without cache was taking around 1h 30m as I remember while on my Laptop it was taking around 20 minutes)

drjova:
for the e2e we should mount the UI code to `docker-compose.override.yml` so we can run them much faster on local

pazembrz:
Please notice that e2e are not using `docker-compose.override.yml`

drjova:
We have to be able to use it for local development otherwise it'll still be very slow for e2e tests

pazembrz:
Additionally we are actually mounting all the code in `docker-compose.yml` https://github.com/inspirehep/inspirehep/blob/7ef2914e2ac9945280c306157c9462ebefd3bfed/docker-compose.core.yml#L48 file, but it's not giving any speedup.
Ui is built in a way that every change of the code will rebuild image https://github.com/inspirehep/inspirehep/blob/master/ui/Dockerfile and we should probably change this Dockerfile (and image it's using).

pazembrz:
I can try to find a better way of building UI

harunurhan:
Isn't it better to directly use an empty nginx and mount the local build folder to the inner image directory where nginx serve from?

This would require only UI to be build for every change not the whole image, which means no extra `yarn install` while building the image.

pazembrz:
Bu then we will have to use some additional script for building ui on running container, as this cannot be done during build stage as you can mount folders only on working container (not during build). This will increase maintenance complexity.

drjova:
https://github.com/inspirehep/inspirehep/issues/1097

run-e2e.sh
```diff
@@ -1,10 +1,14 @@
...
 rc=0
-docker-compose -f ./docker-compose.core.yml up -d --force-recreate
-docker-compose -f ./docker-compose.core.yml exec web ./scripts/setup
-docker-compose -f ./docker-compose.core.yml exec web-next inspirehep db create
-docker-compose -f ./docker-compose.core.yml exec web inspirehep importer records -d data/records/authors
-docker-compose -f ./docker-compose.core.yml exec web inspirehep importer records -f data/records/conference/7577512.json
-docker-compose -f ./docker-compose.core.yml exec web inspirehep importer records -f data/records/literature/374836.json
-docker-compose -f ./docker-compose.core.yml -f ./e2e/docker-compose.e2e.yml run --rm e2e bash -c "cd /opt/e2e && yarn && yarn test" || rc=$?
+files="-f docker-compose.yml -f ./e2e/docker-compose.e2e.yml"
+#containers="mq-next es db cache mq web worker-next worker web-next ui-build"
```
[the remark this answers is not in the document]
drjova: 👍

ui/Dockerfile
```diff
@@ -1,19 +1,6 @@
...
-ARG VERSION
-ENV REACT_APP_VERSION="${VERSION}"
-RUN yarn run build
-
-
 FROM nginx:1.19-alpine
 
 EXPOSE 8080
-COPY --from=build-stage /usr/src/app/build /usr/share/nginx/html
+#COPY --from=build-stage /usr/src/app/build /usr/share/nginx/html
```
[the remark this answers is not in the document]
drjova: 👍

backend/Dockerfile_local
```diff
@@ -0,0 +1,15 @@
+# -*- coding: utf-8 -*-
+#
```
drjova, line 2: the filenames are inconsistent in some cases are with `_` in some others with `-`

PR #1031, opened 2020-06-18
ui: fix existing conferences drawer visibility issue

miguelgrc:
* INSPIR-3589

### DIFF
--- a/ui/src/submissions/conferences/components/ExistingConferencesDrawer.scss
+++ b/ui/src/submissions/conferences/components/ExistingConferencesDrawer.scss
@@ -1,5 +1,5 @@
 .__ExistingConferencesDrawer__ {
-  z-index: $zindex-header + 1;
+  z-index: $zindex-header + 1 !important;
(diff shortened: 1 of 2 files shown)


ui/src/submissions/conferences/components/ExistingConferencesDrawer.scss
```diff
@@ -1,5 +1,5 @@
 .__ExistingConferencesDrawer__ {
-  z-index: $zindex-header + 1;
+  z-index: $zindex-header + 1 !important;
```
drjova, line 2: do we really need the `!important`?

miguelgrc: Yes, it's necessary, otherwise it's overwritten by https://github.com/inspirehep/inspirehep/blob/335b060eaf1814c4c17708bbad5730aa0ed16fe3/ui/src/App.scss#L182

PR #1033, opened 2020-06-18
journals: fix lowercase_journal_titles type

drjova:

### DIFF
--- a/backend/inspirehep/search/mappings/v7/records/journals.json
+++ b/backend/inspirehep/search/mappings/v7/records/journals.json
@@ -133,8 +133,8 @@
         "type": "object"
       },
       "lowercase_journal_titles": {
-        "analyzer": "lowercase_analyzer",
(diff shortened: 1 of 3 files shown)


backend/inspirehep/search/mappings/v7/records/journals.json
```diff
@@ -245,10 +245,25 @@
       "normalizer": {
         "lowercase_normalizer": {
```
michamos, line 246: this affects other fields too that are using it. Maybe create a new specific normalizer for this instead?

drjova: I was thinking about it, it's only affecting `_collections` which will not be affected but I agree it's cleaner to separate it.

drjova:
@michamos done

drjova:
@michamos I have added a test that covers explicity for white spaces as well

PR #1034, opened 2020-06-18
ui: Unify UrlsAction and WebsitesAction

miguelgrc:
* INSPIR-3594


ui/src/experiments/containers/DetailPageContainer.jsx
```diff
@@ -60,17 +59,7 @@ function DetailPage({ record }) {
...
-                {urls && (
-                  <UrlsAction
-                    urls={urls}
-                    iconText="links"
-                    icon={<LinkOutlined />}
-                  />
-                )}
-              </>
-            }
+            leftActions={<>{urls && <UrlsAction urls={urls} text="links" />}</>}
```
drjova, line 62: do we need the empty ``<>``?

miguelgrc: Oops

PR #1035, opened 2020-06-19
migrator: don't show blacklisted recs in dashboard

michamos:
* INSPIR-3646

### DIFF
--- a/backend/tests/integration/migrator/test_migrator_view.py
+++ b/backend/tests/integration/migrator/test_migrator_view.py
@@ -5,11 +5,10 @@
 # inspirehep is free software; you can redistribute it and/or modify it under
 # the terms of the MIT License; see LICENSE file for more details.
 
-import json
-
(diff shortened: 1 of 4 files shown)


backend/tests/integration/migrator/test_migrator_view.py
```diff
@@ -130,6 +131,62 @@ def test_get_does_not_return_deleted_records(inspire_app, datadir):
...
+                "valid": False,
+                "error": "Error: Least recent error.",
+            },
+        ]
+    }
+
+    response_data = json.loads(response.data)
+
+    assert response.status_code == 200
+    assert expected_data == response_data
```
drjova, line 187:
```suggestion
    assert expected_data == response.json
```

michamos: ok, I copy/pasted this from another test, I'll change all tests to use this.

PR #1037, opened 2020-06-22
ui: add missing tests for sub-collections routes

MJedr:
* INSPIR-3630


monaawi:
I think you missed experiments

PR #1038, opened 2020-06-22
matcher: improve parser

drjova:
* Improves `page_start` when reference has only one page.

* Fixes wrong config.

### DIFF
--- a/backend/inspirehep/matcher/config.py
+++ b/backend/inspirehep/matcher/config.py
@@ -119,19 +119,6 @@ REFERENCE_MATCHER_DEFAULT_PUBLICATION_INFO_CONFIG = {
                     ],
                     "type": "nested",
                 },
-                {
-                    "paths": [
-                        "reference.publication_info.journal_title",
-                        "reference.publication_info.journal_volume",
-                        "reference.publication_info.year",
-                    ],
-                    "search_paths": [
-                        "publication_info.journal_title.raw",
-                        "publication_info.journal_volume",
-                        "publication_info.year",
-                    ],
-                    "type": "nested",
-                },
             ]
         }
     ],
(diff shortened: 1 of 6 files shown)


backend/inspirehep/matcher/config.py
```diff
@@ -119,19 +119,6 @@
                     ],
                     "type": "nested",
                 },
-                {
```
drjova, line 122: my bad, I was playing with the config and this was a left over. Now is the same with https://github.com/inspirehep/inspire-next/blob/master/inspirehep/modules/refextract/config.py#L92

backend/inspirehep/matcher/parsers.py
```diff
@@ -68,9 +68,14 @@ def journal_title(self):
...
     @property
     def page_start(self):
-        return self.root.xpath(
+        page_start = self.root.xpath(
             "/biblStruct/monogr/imprint/biblScope[@unit='page']/@from"
         ).get()
+        if page_start:
+            return page_start
+        return self.root.xpath(
+            "/biblStruct/monogr/imprint/biblScope[@unit='page']/text()"
```
michamos, line 77: alternatively you can select both at the XPath level (only one of those should exist, and you're picking that one with `.get()`):
```suggestion
        page_start = self.root.xpath(
            "/biblStruct/monogr/imprint/biblScope[@unit='page']/(@from | text())"
        ).get()
```

drjova: @michamos nice one!

drjova: though: `ValueError: XPath error: Invalid expression in /biblStruct/monogr/imprint/biblScope[@unit='page']/(@from | text())`

drjova: `(/biblStruct/monogr/imprint/biblScope[@unit='page']/@from | /biblStruct/monogr/imprint/biblScope[@unit='page']/text())` this seems ok

michamos: strange, here it works: https://www.freeformatter.com/xpath-tester.html

PR #1039, opened 2020-06-23
search: improve reference matching on literature search

drjova:


drjova:
Note: I didn't want to hack it more, I could have moved this logic to the serializer and try to wrap the results. IMO this is a tradeoff to keep the logic contained in the search classes.

backend/inspirehep/search/api.py
```diff
@@ -177,7 +176,15 @@ def match_reference(self, query_string):
         except KeyError:
             pass
 
-        return match_reference_control_number(reference)
+        reference_match_control_number = match_reference_control_number(reference)
```
michamos, line 179: you could also get the full result here, as the matching is already doing a search (actually several searches), but it adds more complexity, so this is probably good enough.

drjova: Indeed, but I'd have to modify also the matcher api which we suppose to use them for `next` as well. By default it returns the reference with the record `$ref`

PR #1040, opened 2020-06-23
bibgen: rewrite texkeys and handle 0 refs case

miguelgrc:
* Fix for INSPIR-3420

### DIFF
--- a/backend/inspirehep/tools/utils.py
+++ b/backend/inspirehep/tools/utils.py
@@ -67,7 +67,7 @@ def find_references(references, requested_format):
         results = (
             LiteratureSearch()
             .query_from_iq(f"{keyword}:{ref}")
-            .params(size=2, _source=display_format)
+            .params(size=2, _source=[display_format, "texkeys", "control_number"])
             .execute()
         )
 
@@ -77,7 +77,10 @@ def find_references(references, requested_format):
         elif len(hits) > 1:
             errors.append({"ref": ref, "line": line, "type": "ambiguous"})
         else:
-            ret.append(hits[0]["_source"][display_format])
+            source_field = hits[0]["_source"]
+            control_number = source_field["control_number"]
+            texkey = getattr(source_field, "texkeys", [control_number])[0]
+            ret.append(source_field[display_format].replace(f"{{{texkey}", f"{{{ref}"))
 
     return ret, errors
 
--- a/backend/inspirehep/tools/errors.py
+++ b/backend/inspirehep/tools/errors.py
@@ -1,0 +1,25 @@
+# -*- coding: utf-8 -*-
+#
+# Copyright (C) 2020 CERN.
+#
+# inspirehep is free software; you can redistribute it and/or modify it under
+# the terms of the MIT License; see LICENSE file for more details.
+
+from invenio_rest.errors import RESTException
+
+
+class FileTooBigError(RESTException):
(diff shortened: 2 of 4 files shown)


backend/inspirehep/tools/utils.py
```diff
@@ -77,7 +77,10 @@ def find_references(references, requested_format):
         elif len(hits) > 1:
             errors.append({"ref": ref, "line": line, "type": "ambiguous"})
         else:
-            ret.append(hits[0]["_source"][display_format])
+            source_field = hits[0]["_source"]
+            ret.append(
+                source_field[display_format].replace(source_field["texkeys"][0], ref)
```
michamos, line 82: this fails if there are no texkeys (might happen sometimes, because the legacy algorithm to assign texkeys might fail). I don't think you can do the substitution in that case, but you should handle it gracefully.

miguelgrc: I guess I should keep the Inspire texkey in those cases right? Even though it's not going to work when the users paste them into their documents. Should I show some kind of warning message in the UI?

michamos: I think it's enough if you don't blow up, currently it affects only 30000 records, but maybe check with @StellaCh.

StellaCh: It doesn't affect a big volume of records, but since we have a UI to show errors, displaying a message to users will be helpful to make them understand why this blows up and how they can avoid it.

miguelgrc: What message should I use so that they understand more or less the issue?

backend/inspirehep/tools/utils.py
```diff
@@ -77,7 +77,15 @@ def find_references(references, requested_format):
...
-            ret.append(hits[0]["_source"][display_format])
+            source_field = hits[0]["_source"]
+            replace_string = source_field["control_number"]
+            if "texkeys" in source_field:
+                replace_string = source_field["texkeys"][0]
+            ret.append(
+                source_field[display_format].replace(
+                    "{%s" % replace_string, "{%s" % ref
+                )
+            )
```
michamos, line 88: You can use `get` to replace the `if` and use `{{` to escape the curly bracket instead of using ugly string formatting:
```suggestion
            texkey = source_field.get("texkeys", [control_number])[0]
            ret.append(
                source_field[display_format].replace(
                    f"{{{texkey}", "{{{ref}"
                )
            )
```

miguelgrc: Yes, this looks better, thanks. I have to use `getattr` though since `get` doesn't work with Elasticsearch DSL's `AttrDict`

backend/inspirehep/tools/utils.py
```diff
@@ -77,7 +77,10 @@ def find_references(references, requested_format):
         elif len(hits) > 1:
             errors.append({"ref": ref, "line": line, "type": "ambiguous"})
         else:
-            ret.append(hits[0]["_source"][display_format])
+            source_field = hits[0]["_source"]
+            control_number = source_field["control_number"]
+            texkey = getattr(source_field, "texkeys", [control_number])[0]
```
michamos, line 82: just curious: what's wrong with
```suggestion
            texkey = source_field.get("texkeys", [control_number])[0]
```
?

PR #1042, opened 2020-06-24
search: match reference improve error handling

drjova:


backend/inspirehep/search/api.py
```diff
@@ -127,10 +127,13 @@ class Meta:
...
-            reference_match = self.match_reference(query_string)
-            if reference_match:
-                return reference_match
+            try:
+                query_string = request.values.get("q", "", type=str)
+                reference_match = self.match_reference(query_string)
+                if reference_match:
+                    return reference_match
+            except Exception:
+                LOGGER.exception("Match reference error.", query_string=query_string)
```
pazembrz, line 136: Isn't better to raise here some RestError? so frontend will receive some error which can be displayed to the user later? Otherwise user will get empty results.

drjova: This is exactly what we want to avoid :)

PR #1043, opened 2020-06-24
search: ignore leading zero in `page_artid`

drjova:


backend/inspirehep/search/mappings/v7/records/hep.json
```diff
@@ -1521,6 +1526,11 @@
           "type": "pattern_replace",
           "pattern": "\\.",
           "replacement": " "
+        },
+        "remove_leading_zero_normalizer": {
+          "type": "pattern_replace",
+          "pattern": "^0",
```
michamos, line 1532: maybe even
```suggestion
          "pattern": "^0+",
```

backend/inspirehep/search/mappings/v7/records/hep.json
```diff
@@ -1497,6 +1497,11 @@
           "char_filter": ["dot_to_whitespace_normalizer"],
           "tokenizer": "standard",
           "type": "custom"
+        },
+        "page_artid_analyzer": {
```
michamos, line 1501: I don't think you're configuring the field to use this analyzer

drjova: I just pushed it, you're too fast :P

drjova:
@michamos updated

backend/inspirehep/search/mappings/v7/records/hep.json
```diff
@@ -1497,6 +1498,11 @@
...
+          "char_filter": ["remove_leading_zero_normalizer"],
+          "tokenizer": "standard",
```
michamos, line 1504: do we really want to go through the tokenizer? I think it would be better to keep it as a `keyword` but use a normalizer, otherwise strange things might happen, I don't exactly understand the rules for splitting words containing chars+digits which might occur here.

drjova: I don't think will make any difference in that case, but I've updated it

michamos:
I've seen this type of errors multiple times today: https://travis-ci.org/github/inspirehep/inspirehep/jobs/701661423#L455 looks like the random record generator is a bit too random.

PR #1044, opened 2020-06-24
files: remove exception log for duplicate files

monaawi:
* INSPIR-3611

### DIFF
--- a/backend/inspirehep/records/api/literature.py
+++ b/backend/inspirehep/records/api/literature.py
@@ -309,7 +309,7 @@ class LiteratureRecord(
         try:
             add_file_func(**file_data)
         except ValueError:
-            LOGGER.exception(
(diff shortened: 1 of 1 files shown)


backend/inspirehep/records/api/literature.py
```diff
@@ -309,7 +309,7 @@ def _update_file_entry(self, file_data, add_file_func):
         try:
             add_file_func(**file_data)
         except ValueError:
```
monaawi, line 311: I think it makes sense to keep throwing ValueError from schemas in this case, as we are trying to populate literature with wrong data.
In our case, we want to skip over those files, so we catch it and just log warning

drjova: you are right!

PR #1046, opened 2020-06-25
files: do not use direct s3 links on records

pazembrz:
serve files from inspirehep domain instead.
* INSPIR-3586


backend/inspirehep/files/config.py
```diff
@@ -16,4 +16,5 @@
 FILES_MAX_UPLOAD_THREADS = 5
 FILES_UPLOAD_THREAD_TIMEOUT = 120
 FILES_DOWNLOAD_TIMEOUT = 60
+FILES_PUBLIC_PATH = "/file/"
```
drjova, line 19:
```suggestion
FILES_PUBLIC_PATH = "/files/"
```

backend/inspirehep/files/api/s3.py
```diff
@@ -178,3 +205,32 @@ def create_bucket(self, bucket):
             Bucket=self.get_prefixed_bucket(bucket),
             ACL=current_app.config.get("S3_FILE_ACL"),
         )
+
+    def get_key_from_url(self, url):
+        try:
+            return urlsplit(url).path.split("/")[-1]
+        except (IndexError, AttributeError):
+            LOGGER.exception("incorrect url.", url=url)
```
drjova, line 213:
```suggestion
            LOGGER.exception("S3 incorrect url.", url=url)
```

ui/docker/nginx/config/local.conf
```diff
@@ -40,4 +40,10 @@ server {
     proxy_set_header Host localhost:8000;
     proxy_http_version 1.1;
   }
+
+  location /file/{
```
drjova, line 44:
```suggestion
  location /files/ {
```

backend/inspirehep/files/api/s3.py
```diff
@@ -178,3 +205,32 @@ def create_bucket(self, bucket):
             Bucket=self.get_prefixed_bucket(bucket),
             ACL=current_app.config.get("S3_FILE_ACL"),
         )
+
+    def get_key_from_url(self, url):
+        try:
+            return urlsplit(url).path.split("/")[-1]
+        except (IndexError, AttributeError):
+            LOGGER.exception("S3 incorrect url.", url=url)
+            raise
```
drjova, line 214: Do we want to raise twice? Isn't it the case?

backend/inspirehep/records/api/literature.py
```diff
@@ -388,14 +388,19 @@ def add_file(
             app_context: Original app context should be passed here if running in separate thread
         """
         with app_context.app.app_context():
-            if current_s3_instance.is_s3_url(url) and not current_app.config.get(
-                "UPDATE_S3_FILES_METADATA", False
-            ):
+            if (
+                current_s3_instance.is_s3_url(url)
+                or current_s3_instance.is_public_url(url)
+            ) and not current_app.config.get("UPDATE_S3_FILES_METADATA", False):
```
drjova, line 394:
```suggestion
             is_s3_or_public_url = current_s3_instance.is_s3_url(url) or current_s3_instance.is_public_url(url)
             if is_s3_or_public_url and not current_app.config.get("UPDATE_S3_FILES_METADATA", False):
```

backend/inspirehep/files/api/s3.py
```diff
@@ -22,6 +23,13 @@ def __init__(self, client, resource, config=None):
             config = TransferConfig(max_concurrency=1, use_threads=False)
         self.config = config
 
+    @property
+    def public_file_path(self):
+        return urljoin(
+            f"{current_app.config['PREFERRED_URL_SCHEME']}://{current_app.config['SERVER_NAME']}",
```
drjova, line 29: can't we use https://github.com/inspirehep/inspirehep/blob/e33f5c2f4d55573ff115f154169664020f7883e8/backend/inspirehep/utils.py#L17 ?

backend/inspirehep/files/api/s3.py
pazembrz: Maybe I'm wrong, but here we raise this exception only once. LOGGER.exception is logging it, but without re-raising exception wont' be propagated further.
Do you think it's better to return None here?

drjova: 👍  you're right

PR #1047, opened 2020-06-25
bibliography generator: simplify docs & fix journal key bug

michamos:
* INSPIR-3420

### DIFF
--- a/ui/src/bibliographyGenerator/BibliographyGenerator.jsx
+++ b/ui/src/bibliographyGenerator/BibliographyGenerator.jsx
@@ -17,8 +17,17 @@ const BIBLIOGRAPHY_GENERATOR_FORMATS = [
   { value: 'latex_us', display: 'LaTeX (US)' },
   { value: 'latex_eu', display: 'LaTeX (EU)' },
 ];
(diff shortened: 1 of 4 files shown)


ui/src/bibliographyGenerator/BibliographyGenerator.jsx
```diff
@@ -145,104 +146,111 @@ Finally, let me make a mistake citing this paper \\cite{hep-scifi/0101001}.
...
-
+\\bibitem{hep-th/0412210}
+Z.~Bern, L.~J.~Dixon and D.~A.~Kosower,
+%\`\`All Next-to-maximally-helicity-violating one-loop gluon amplitudes in N=4 super-Yang-Mills theory,''
+Phys. Rev. D \\textbf{72} (2005), 045014
+doi:10.1103/PhysRevD.72.045014
+[arXiv:hep-th/0412210 [hep-th]].
+%136 citations counted in INSPIRE as of 24 Jun 2020
+
+%\\cite{JHEP,0412,015}
```
michamos, line 202: actually there's a bug here, these should be `.`s. I'll add a commit to fix it.

PR #1048, opened 2020-06-25
v1.1.13

drjova:


drjova:
* remap & reindex literature

michamos:
the bibliography generator has a bug fixed in #1047. I think we should either wait for that to be merged or remove/feature flag the bibliography generator.

drjova:
we're not in rush to deploy this, we could wait

drjova:
I'll close it for now, not to waste resource on travis everytime we commit something

PR #1054, opened 2020-06-26
pidstore: do not delete all external pids when it's unecessary

pazembrz:
* Additionally fix cnum provider when creating cnum from data provided.
 * Fix ORCID generator in tests

 * INSPIR-3641

### DIFF
--- a/backend/inspirehep/pidstore/minters/base.py
+++ b/backend/inspirehep/pidstore/minters/base.py
@@ -9,6 +9,7 @@ import structlog
 from inspire_utils.helpers import force_list
 from inspire_utils.record import get_value
 from invenio_pidstore.errors import PIDAlreadyExists
+from invenio_pidstore.models import PersistentIdentifier, PIDStatus
 
 from inspirehep.pidstore.errors import MissingSchema
 from inspirehep.pidstore.providers.external import InspireExternalIdProvider
@@ -48,13 +49,14 @@ class Minter:
         """
         return self.get_pid_values()
 
-    def create(self, pid_value):
+    def create(self, pid_value, **kwargs):
         LOGGER.info(
             "Minting",
             pid_type=self.pid_type,
             recid=pid_value,
             object_type=self.object_type,
             object_uuid=str(self.object_uuid),
+            pid_provider=self.provider.pid_provider,
         )
         try:
             return self.provider.create(
@@ -62,6 +64,7 @@ class Minter:
                 pid_value=pid_value,
                 object_type=self.object_type,
                 object_uuid=self.object_uuid,
+                **kwargs
             )
         except PIDAlreadyExists as e:
             raise PIDAlreadyExistsError(e.pid_type, e.pid_value) from e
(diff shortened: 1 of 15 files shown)


backend/inspirehep/pidstore/minters/base.py
```diff
@@ -78,11 +79,42 @@ def mint(cls, object_uuid, data):
...
+                object_type=minter.object_type,
+            )
+            .filter(PersistentIdentifier.status != PIDStatus.DELETED)
+            .with_entities("pid_value")
+            .all()
+        }
+        pids_requested = minter.get_pid_values()
+        pids_to_delete = pids_in_db - pids_requested
+        pids_to_create = pids_requested - pids_in_db
+        minter.delete(object_uuid, None, pids_to_delete)
```
drjova, line 98: If we don't need to pass `data` anymore let's change the signature of delete, it's used only internaly.

backend/inspirehep/pidstore/minters/doi.py
```diff
@@ -18,4 +18,4 @@ def create(self, pid_value):
 
     def get_pid_values(self):
         pid_values = {str(pid_value).lower() for pid_value in super().get_pid_values()}
-        return pid_values
+        return set(pid_values)
```
drjova, line 21: isn't already a set?

pazembrz: :+1:  Yes, you are right. Here it's already set.

backend/inspirehep/pidstore/minters/base.py
pazembrz: we still use data in `delete` methods in `ControlNumberMinter` and `CNUMMinter`

backend/tests/integration/pidstore/minters/test_minters_arxiv.py
```diff
@@ -110,3 +111,50 @@ def test_mitner_arxiv_eprints_missing_schema(inspire_app):
 
     with pytest.raises(MissingSchema):
         ArxivMinter.mint(record_id, record_data)
+
+
+def test_minter_proper_pid_management(inspire_app):
```
drjova, line 116: could we add a test with more pid types? a combination of internal and external?

pazembrz: All tests for externals are actually tests with internal also (control_number is internal) But I will add few more tests.

backend/tests/helpers/providers/record_provider.py
```diff
@@ -39,12 +39,23 @@ def arxiv():
...
+        val = str(orcid_id).replace("-", "").replace(" ", "").upper()
+        r = 0
+        for x in val:
+            r = (r + int(x)) * 2
+        ck = (12 - r % 11) % 11
+        if ck == 10:
+            ck = "X"
+
+        orcid_id += str(ck)
+        return orcid_id
```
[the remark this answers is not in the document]
pazembrz: This is calculation of checksum which is integral part of ORCID, and it's checked, so without it generated ORCID will be wrong and will result in record creation failure if used with `create_record` helper.

backend/tests/integration/pidstore/minters/test_minters_arxiv.py
```diff
@@ -110,3 +111,50 @@ def test_mitner_arxiv_eprints_missing_schema(inspire_app):
 
     with pytest.raises(MissingSchema):
         ArxivMinter.mint(record_id, record_data)
+
+
+def test_arxiv_minter_proper_pid_management(inspire_app):
```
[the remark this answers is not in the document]
pazembrz: :+1:

backend/tests/integration/pidstore/minters/test_minters_control_number.py
```diff
@@ -326,3 +325,24 @@ def test_control_number_seminars_without_control_number(inspire_app):
...
+
+def test_control_number_minter_proper_pid_management(inspire_app):
+    record = create_record("lit")
+    expected_cn = str(record["control_number"])
+    cn_pid = PersistentIdentifier.query.filter_by(pid_type="lit").one()
+
+    assert cn_pid.pid_value == expected_cn
+
+    data = dict(record)
+    data["control_number"] = f"{expected_cn}1"
```
drjova, line 338: there is `1` at the end, is this correct?

pazembrz: Yes, I Intentionally added 1 to change control_number to be sure that it's different to test behavior when CN changes.

PR #1055, opened 2020-06-29
tests: fix flaky integration test

pazembrz:
Quick fix for flaky integration test

### DIFF
--- a/backend/tests/integration/records/test_utils.py
+++ b/backend/tests/integration/records/test_utils.py
@@ -94,7 +94,8 @@ def test_get_parent_record_when_more_than_one(inspire_app):
     }
     rec = create_record("lit", data=data)
     extracted_parent_record = get_parent_record(rec)
-    assert extracted_parent_record == parent_record
+    parent_records = [parent_record, second_parent_record]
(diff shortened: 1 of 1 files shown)


backend/tests/integration/records/test_utils.py
```diff
@@ -94,7 +94,8 @@ def test_get_parent_record_when_more_than_one(inspire_app):
     }
     rec = create_record("lit", data=data)
     extracted_parent_record = get_parent_record(rec)
-    assert extracted_parent_record == parent_record
+    parent_records = list(parent_record, second_parent_record)
```
drjova, line 97: do I miss something or this should throw an exception? :/

pazembrz: Why do you think that?

drjova:
```
----> 1 list('a', 'b')

TypeError: list() takes at most 1 argument (2 given)
```

pazembrz: Ah, yest... Sorry... Monday morning....

PR #1065, opened 2020-06-30
global: multi worker processes

drjova:


michamos:
I think you need to change some unit tests: https://travis-ci.org/github/inspirehep/inspirehep/jobs/703491286#L394

PR #1075, opened 2020-07-01
seminars: display speaker in the title of calender export

harunurhan:
INSPIR-3608


ui/src/seminars/components/ExportToCalendarAction/common.js
```diff
@@ -0,0 +1,8 @@
+export function getEventTitle(seminar) {
+  const seminarTitle = seminar.getIn(['title', 'title']);
+  const speakers = seminar.get('speakers');
+  const speaker = speakers.get(0);
+  return `${speaker.get('first_name')} ${speaker.get('last_name')} ${
+    speakers.size > 1 ? ' et al. ' : ''
+  } — ${seminarTitle}`;
+}
```
drjova, line 8: don't we have this logic somewhere else?

harunurhan: similar in the author list, but it's for more authors and much more complicated.

PR #1078, opened 2020-07-01
ui: set proper mimetype and extension for cite

miguelgrc:
* INSPIR-3636

### DIFF
--- a/ui/src/common/utils.js
+++ b/ui/src/common/utils.js
@@ -4,7 +4,6 @@ import moment from 'moment-timezone';
 import { Map } from 'immutable';
 import NumberAbbreviator from 'number-abbreviate';
 import { LITERATURE } from './routes';
-import { browser } from './browser';
 
 export function forceArray(maybeArray) {
   return maybeArray === undefined || Array.isArray(maybeArray)
@@ -205,29 +204,22 @@ export function getRecordIdFromRef($ref) {
 
 export function downloadTextAsFile(
   text,
-  type = 'application/txt',
-  extension = 'txt'
+  filename = 'download.txt',
+  type = 'text/plain'
 ) {
   const blob = new Blob([text], { type });
-
   if (window.navigator && window.navigator.msSaveOrOpenBlob) {
     // Edge
-    window.navigator.msSaveOrOpenBlob(blob, `download.${extension}`);
-  } else if (browser.isSafari()) {
-    // Safari
+    window.navigator.msSaveOrOpenBlob(blob, filename);
+  } else {
     const url = URL.createObjectURL(blob);
     const a = document.createElement('a');
     a.style.display = 'none';
     a.href = url;
-    a.download = `download.${extension}`;
+    a.download = filename;
     document.body.appendChild(a);
     a.click();
     window.URL.revokeObjectURL(url);
-  } else {
-    // the rest
-    const url = URL.createObjectURL(blob);
-    window.open(url, '_self');
-    URL.revokeObjectURL(url);
   }
 }
 
--- a/ui/src/literature/components/CiteModalAction.jsx
+++ b/ui/src/literature/components/CiteModalAction.jsx
@@ -12,7 +12,11 @@ import SelectBox from '../../common/components/SelectBox';
 import ListItemAction from '../../common/components/ListItemAction';
 import IconText from '../../common/components/IconText';
 import EventTracker from '../../common/components/EventTracker';
-import { CITE_FORMAT_OPTIONS, CITE_FORMAT_VALUES } from '../constants';
+import {
+  CITE_FORMAT_OPTIONS,
+  CITE_FORMAT_VALUES,
+  CITE_FILE_FORMAT,
+} from '../constants';
 import citeArticle from '../citeArticle';
 import { downloadTextAsFile } from '../../common/utils';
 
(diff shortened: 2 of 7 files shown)


ui/src/literature/components/CiteModalAction.jsx
```diff
@@ -74,7 +78,11 @@ class CiteModalAction extends Component {
       }
     }
 
-    this.setState({ citeContent });
+    this.setState({
+      citeContent,
+      mimetype: CITE_FILE_FORMAT[format].mimetype,
+      extension: CITE_FILE_FORMAT[format].extension,
```
harunurhan, line 84: you can keep the `format` in the state rather than these 2.

ui/src/common/utils.js
```diff
@@ -3,8 +3,8 @@ import cloneDeep from 'lodash.clonedeep';
 import moment from 'moment-timezone';
 import { Map } from 'immutable';
 import NumberAbbreviator from 'number-abbreviate';
+import { saveAs } from 'file-saver';
```
harunurhan, line 6: i don't think this brings anything valuable at the moment, so we can keep our simple function.

ui/src/common/utils.js
```diff
@@ -208,27 +208,9 @@ export function downloadTextAsFile(
...
-    document.body.appendChild(a);
-    a.click();
-    window.URL.revokeObjectURL(url);
-  } else {
-    // the rest
-    const url = URL.createObjectURL(blob);
-    window.open(url, '_self');
-    URL.revokeObjectURL(url);
-  }
+  saveAs(blob, `INSPIRE-Cite_${date}.${extension}`);
```
harunurhan, line 213: this is a generic function used to save anything, `INSPIRE-Cite` doesn't make sense. 

`downloadTextAsFile` can just take file name (together with the extension)

ui/src/common/utils.js
```diff
@@ -208,27 +208,9 @@ export function downloadTextAsFile(
   type = 'application/txt',
   extension = 'txt'
 ) {
+  const date = moment().format('YYYY-MM-DDTHHmmss');
```
harunurhan, line 211: was this a requirement?

ui/src/common/utils.js
miguelgrc: It seemed cleaner and easier since it would work in every browser and display the correct filename and extension.
Otherwise, I'm not sure why we have different code for Safari than for the rest. Why weren't we setting the filename and extension for the rest of the browsers? If we just use that same code for Chrome it works perfectly.

ui/src/common/utils.js
miguelgrc: I discussed it with Stella as a way of differentiating citation files

harunurhan: this is not good, if you use current download date as a name, "Downloads" folder will have duplicates as you download the same stuff and since the name is different for each time, there won't be easy way for user to see which files are duplicates.

ui/src/common/utils.js
harunurhan: "the rest" code was there before everything else, and setting file name was never strong requirement. If filename setting is important you can just keep `IE` and `anchor` based implementation.

Our code works for the browsers we officially support already, and this package does more to handle other cases that we don't care. And there is no guarantee it will load some heavy dependencies on a later release. Therefore it's not worth it.

ui/src/common/utils.js
harunurhan: this is still not addressed.

ui/src/literature/components/CiteModalAction.jsx
harunurhan: neither this.

ui/src/common/utils.js
StellaCh: What do you mean `search` for CiteAll @harunurhan ?

harunurhan: I mean we can't set record id for file that is dowloaded via CiteAll, since there are many records. So we could just name it `search`

StellaCh: There's not much value for this and we can't avoid having some duplicates. So let's do 'INSPIRE-Cite-record' for single ones and 'INSPIRE-CiteAll' for the rest.

miguelgrc: So we go without dates right?

StellaCh: yes

ui/package.json
```diff
@@ -34,6 +34,7 @@
     "collections": "^5.1.10",
     "connected-react-router": "^6.4.0",
     "craco-antd": "1.11.0",
+    "file-saver": "^2.0.2",
```
harunurhan, line 37: left over

ui/yarn.lock
```diff
@@ -3977,6 +3977,11 @@ file-loader@2.0.0:
     loader-utils "^1.0.2"
     schema-utils "^1.0.0"
 
+file-saver@^2.0.2:
```
harunurhan, line 3980: leftover

ui/src/seminars/components/ExportToCalendarAction/ExportToCalendarAction.jsx
```diff
@@ -18,7 +18,7 @@ function ExportToCalendarAction({ seminar }) {
   const onDownloadClick = useCallback(
     () => {
       const fileContent = getIcsFileContent(seminar);
-      downloadTextAsFile(fileContent, 'text/calendar', 'ics');
+      downloadTextAsFile(fileContent, 'INSPIRE-Calendar.ics', 'text/calendar');
```
harunurhan, line 21: could use record id here, and have sth like `INSPIRE-Seminar-recid`.

PR #1083, opened 2020-07-02
literature: improve display of book series

michamos:
* INSPIR-3559

### DIFF
--- a/ui/src/literature/components/BookSeriesInfoList.jsx
+++ b/ui/src/literature/components/BookSeriesInfoList.jsx
@@ -4,14 +4,16 @@ import { List } from 'immutable';
 
 import InlineList, { SEPARATOR_AND } from '../../common/components/InlineList';
 
-function extractChapter(bookChapter) {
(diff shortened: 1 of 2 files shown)


ui/src/literature/components/BookSeriesInfoList.jsx
```diff
@@ -4,25 +4,22 @@ import { List } from 'immutable';
...
 
-function renderSeriesInfo(bookChapter) {
-  return (
-    <span>
-      {bookChapter.get('title')} volume:{bookChapter.get('volume')}
-    </span>
-  );
+function renderSeriesInfo(bookSeries) {
+  const volume = bookSeries.get('volume');
+  return <span>{bookSeries.get('title') + volume && `, ${volume}`}</span>;
```
michamos, line 13: I don't think this is the right way to do it, suggestions welcome :)

harunurhan: sth like that would be better.
```js
<span>{bookSeries.get('title')}{volume && <span>, {volume}</span>}</span>
```

Issue #1070, opened 2020-07-01
Unify the way we check if user is submitter for jobs and seminars and default can edit false if there are db issues


drjova:
Jobs check if the user is the submitter by using the current_user.email, they should use the orcid instead. Change this both in the detail/search serializers and the submissions put condition. (seminars are using the orcid)

Also, catch issues that might come up when retrieving current user's orcid and default can_edit to false so the search doesn't blow up when we have db issues

 

**TECH NOTES:**

In https://github.com/inspirehep/inspirehep/blob/89d8f5289db8ffcccfbc5ed50407befd93bad229/backend/inspirehep/accounts/api.py#L39 catch SQLAlchemyError general exception, log it and return None so no permissions will be granted to user. 

Use these two lines https://github.com/inspirehep/inspirehep/blob/89d8f5289db8ffcccfbc5ed50407befd93bad229/backend/inspirehep/accounts/api.py#L26, https://github.com/inspirehep/inspirehep/blob/89d8f5289db8ffcccfbc5ed50407befd93bad229/backend/inspirehep/accounts/api.py#L29 instead of https://github.com/inspirehep/inspirehep/blob/89d8f5289db8ffcccfbc5ed50407befd93bad229/backend/inspirehep/records/marshmallow/jobs/base.py#L54-L56 but checking for inequality

Unify serializers `can_edit` as well to check only ORCID

In the jobs serializer, don't use the post_dump for the ``can_edit``. It can use a ``simple fields.Method``

PR #1089, opened 2020-07-06
jobs: check if user can edit job with orcid

MJedr:
* INSPIR-3591

### DIFF
--- a/backend/inspirehep/accounts/api.py
+++ b/backend/inspirehep/accounts/api.py
@@ -6,12 +6,15 @@
 # the terms of the MIT License; see LICENSE file for more details.
 
 
+import structlog
 from flask_login import current_user
 from inspire_utils.record import get_value
 from invenio_oauthclient.models import RemoteAccount, UserIdentity
-from sqlalchemy.orm.exc import NoResultFound
+from sqlalchemy.exc import SQLAlchemyError
 
 from inspirehep.accounts.roles import Roles
+
+LOGGER = structlog.getLogger()
 
 
 def is_superuser_or_cataloger_logged_in():
@@ -44,7 +47,8 @@ def get_current_user_orcid():
             .id
         )
         return orcid
-    except NoResultFound:
+    except SQLAlchemyError:
+        LOGGER.info("No access granted to the user.")
         return None
 
 
--- a/backend/inspirehep/records/marshmallow/jobs/base.py
+++ b/backend/inspirehep/records/marshmallow/jobs/base.py
@@ -5,64 +5,20 @@
 # inspirehep is free software; you can redistribute it and/or modify it under
 # the terms of the MIT License; see LICENSE file for more details.
 
-from inspire_utils.record import get_value
+
 from marshmallow import fields
 
-from inspirehep.accounts.api import (
-    is_loggedin_user_email,
-    is_superuser_or_cataloger_logged_in,
-)
 from inspirehep.records.marshmallow.base import RecordBaseSchema
 from inspirehep.records.marshmallow.common import AcceleratorExperimentSchemaV1
-from inspirehep.submissions.utils import has_30_days_passed_after_deadline
 
 
 class JobsRawSchema(RecordBaseSchema):
-    class Meta:
-        exclude = ("can_edit",)
-
     def __init__(self, *args, **kwargs):
         super().__init__()
-        self.post_dumps.append(self.set_editable_post_dump)
 
     accelerator_experiments = fields.Nested(
         AcceleratorExperimentSchemaV1, dump_only=True, many=True
     )
-
-    def set_editable_post_dump(self, object_, original_data):
-        if JobsRawSchema.is_job_editable(object_):
-            object_["can_edit"] = True
-        return object_
-
-    @staticmethod
-    def is_job_editable(data):
-        """Check if the given job is editable
-
-        A job is editable if one of the following is true:
-        * current user is superadmin or cataloger
-        * the job 'status' is not 'closed' and the job's author is logged-in.
-
-        Args:
-            data (dict): the jobs metadata.
-
-        Returns:
-            bool: True if the job can be edited, False otherwise.
-        """
-        if is_superuser_or_cataloger_logged_in():
-            return True
-
-        email = get_value(data, "acquisition_source.email")
-        if not is_loggedin_user_email(email):
-            return False
-
-        status = get_value(data, "status")
-        if status != "closed":
-            return True
-
-        deadline = get_value(data, "deadline_date")
-        if status == "closed" and not has_30_days_passed_after_deadline(deadline):
-            return True
-        return False
 
 
 class JobsAdminSchema(JobsRawSchema):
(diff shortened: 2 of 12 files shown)


drjova:
is this connected to https://github.com/inspirehep/inspirehep/issues/1070 ? If yes could you please added it to your commit message? ([more info](https://docs.github.com/en/github/managing-your-work-on-github/linking-a-pull-request-to-an-issue#linking-a-pull-request-to-an-issue-using-a-keyword))

backend/inspirehep/records/marshmallow/jobs/base.py
```diff
@@ -18,22 +19,15 @@
...
 
 class JobsRawSchema(RecordBaseSchema):
-    class Meta:
-        exclude = ("can_edit",)
-
     def __init__(self, *args, **kwargs):
         super().__init__()
-        self.post_dumps.append(self.set_editable_post_dump)
+
+    can_edit = fields.Method("is_job_editable",)
```
drjova, line 25: It will not work like that, because the Job will be indexed to elasticsearch with this field. We should either move it to the `ui` serializer (preferable) or leave it in the `post_dump`

backend/tests/integration/records/serializers/json/test_jobs.py
```diff
@@ -7,14 +7,20 @@
...
 from freezegun import freeze_time
 from helpers.utils import create_record, create_user, logout
 from invenio_accounts.testutils import login_user_via_session
+from invenio_oauthclient import current_oauthclient
 from marshmallow import utils
 
 from inspirehep.accounts.roles import Roles
 
 
+def return_true():
```
drjova, line 20: ?

backend/tests/integration/records/serializers/json/test_jobs.py
```diff
@@ -91,73 +99,73 @@ def test_jobs_json_author_can_edit_but_random_user_cant(inspire_app, datadir):
...
-    assert "can_edit" not in response_data_metadata
+        assert not response_data_metadata["can_edit"]
 
 
 @freeze_time("2020-02-01")
 def test_jobs_json_author_can_edit_if_closed_and_less_than_30_days_after_deadline(
     inspire_app, datadir
 ):
-    headers = {"Accept": "application/json"}
+    with inspire_app.test_client() as client:
```
drjova, line 115: is it any specific reason why you added everything under the client?

backend/tests/integration/records/serializers/json/test_jobs.py
```diff
@@ -44,45 +46,51 @@ def test_jobs_json(inspire_app, datadir):
 
 
 def test_jobs_json_cataloger_can_edit(inspire_app, datadir):
-    headers = {"Accept": "application/json"}
+    with inspire_app.test_client() as client:
+        headers = {"Accept": "application/vnd+inspire.record.ui+json"}
 
-    data = json.loads((datadir / "955427.json").read_text())
+        data = json.loads((datadir / "955427.json").read_text())
```
drjova, line 52: I think also you are using tabs instead of spaces 😁

backend/inspirehep/records/serializers/json/jobs.py
```diff
@@ -19,3 +20,14 @@
 jobs_json_response = record_responsify(jobs_json, "application/json")
 
 jobs_json_response_search = search_responsify(jobs_json, "application/json")
+
+seminars_json_detail = JSONSerializer(wrap_schema_class_with_metadata(JobsDetailSchema))
```
drjova, line 24: you don't need this here.

backend/tests/integration/records/views/test_views_jobs.py
```diff
@@ -149,7 +149,7 @@ def test_jobs_accelerator_experiments(inspire_app, datadir):
 
 
 def test_jobs_record_search_results_does_not_return_pending_job_to_non_superuser(
-    inspire_app
+    inspire_app,
```
drjova, line 152: you don't need the extra comma

backend/inspirehep/records/serializers/json/jobs.py
MJedr: beside the wrong variable name I think that it is neccessary to have this serialiser because I moved ```can_edit``` method to the ```JobsDetailSchema``` class

backend/tests/integration/records/views/test_views_jobs.py
MJedr: That's pre-commit thing, it's added automatically

backend/inspirehep/accounts/api.py
```diff
@@ -37,15 +40,19 @@ def is_loggedin_user_email(email):
...
-    try:
-        orcid = (
-            UserIdentity.query.filter_by(id_user=current_user.get_id(), method="orcid")
-            .one()
-            .id
-        )
-        return orcid
-    except NoResultFound:
-        return None
+    if current_user:
```
drjova, line 43: the current_user will be always set even if the user is not logged (as Anonymous) we you would like to add this extra if here?

backend/tests/integration/records/serializers/json/test_jobs.py
```diff
@@ -185,35 +193,42 @@ def test_jobs_search_json(inspire_app, datadir):
...
     )
     another_job_metadata = next(
         hit["metadata"]
         for hit in hits
-        if hit["metadata"]["acquisition_source"]["email"] != user.email
+        if hit["metadata"]["acquisition_source"]["orcid"] != user_orcid
     )
 
-    assert "can_edit" not in another_job_metadata
+    assert not another_job_metadata["can_edit"]
```
drjova, line 231: why you changed `assert "can_edit" not in another_job_metadata`? :)

backend/tests/integration/records/serializers/json/test_jobs.py
```diff
@@ -91,22 +99,20 @@ def test_jobs_json_author_can_edit_but_random_user_cant(inspire_app, datadir):
...
-        random_user = create_user(email="random@user.com")
-        login_user_via_session(client, email=random_user.email)
+        random_user = create_user(email="random@user.com", orcid="not_valid")
+        mock_current_user.return_value = random_user
 
         response = client.get(f"/jobs/{record_control_number}", headers=headers)
         response_data_metadata = json.loads(response.data)["metadata"]
 
-    assert "can_edit" not in response_data_metadata
+    assert not response_data_metadata["can_edit"]
```
drjova, line 108: ditto

backend/tests/integration/records/serializers/json/test_jobs.py
MJedr: Cause it's visible (as in seminars; btw. variable names in the test are wrong).
https://github.com/inspirehep/inspirehep/blob/05c189b27610d8665c0aaae982a25734e1bce9dc/backend/tests/integration/records/serializers/json/test_seminars.py#L218-L230

In literature it's not visible but there is used two step serialisation
https://github.com/inspirehep/inspirehep/blob/05c189b27610d8665c0aaae982a25734e1bce9dc/backend/inspirehep/records/marshmallow/literature/ui.py#L191-L205

So should I make it as in literature (or hide it somehow if it's ```False```) or we keep it as in seminars?

drjova: Thanks for explanation, let's keep it as seminars

backend/inspirehep/accounts/api.py
drjova: @MJedr ``current_user`` is a redundant check, why want it here?

MJedr: @drjova yes, you're right. It's updated :)

Issue #1063, opened 2020-06-30
Assign Papers use cases


drjova:
Create an assign papers API for the following use cases and should handle bulk:

**Assign unassigned papers**

    For the given author
        Change papers to curated_relation = true
        If record stub=true, remove stub=true.

**Paper from author1 belongs to author2 (author2 already existing in INSPIRE)**

    change author1 ref to author2
    paper curated_relation = true
    If author2 is stub, unstub it.

**Paper from author1 belongs to author2 (author2 is unknown)**

    Remove curated_relation = true
    Create stub author2
    Add paper ref in the removed papers list of author1 (blacklist)
    Change ref author1 to author2

**This case will not be handled in this US we're missing requirements**
~~Paper from author1 belongs to author2 (author2 is new author)~~

  ~~Create a new author, like in disambiguation with name and name variants but not stubbed~~
    ~~change ref to author2~~
    ~~curated_relation = true~~

### **Tech notes:**

Create a new module `claims` that has the following endpoint:

`POST /assign` which accepts the following payload:

```json
{
  "literature_recids": [ ],
   "to_author_recid": "<pid_value> || null",
   "from_author_recid": "<pid_value> || null"
}
```

**Case** `from_author_recid=null`

```
foreach `literature_recids`, loop over authors, if author id == to_author_recid: 
  Change papers to curated_relation = true
If `to_author_recid` stub=true, remove it.
```

**Case** `to_author_recid=<pid_value>`

```
foreach `literature_recids`, loop over authors, if author id == from_author_recid:
    change `from_author_recid` ref to `to_author_recid`
    add `curated_relation=true`

If `to_author_recid` stub=true, remove it.
```

**Case** `to_author_recid=null`

Create a [stub author](https://github.com/inspirehep/inspirehep/blob/master/backend/inspirehep/disambiguation/utils.py#L75) (`stub_author`) 

```
foreach `literature_recids`, loop over authors, if author id == from_author_recid:
    change `from_author_recid` ref to `stub_author`
    remove `curated_relation=true` 
    add paper ref in the removed papers list of from_author_recid (blacklist)
```

PR #1092, opened 2020-07-07
assign: add initial API

harunurhan:
Ref: #1063

### DIFF
--- a/backend/inspirehep/assign/views.py
+++ b/backend/inspirehep/assign/views.py
@@ -1,0 +1,72 @@
+# -*- coding: utf-8 -*-
+#
+# Copyright (C) 2020 CERN.
+#
+# inspirehep is free software; you can redistribute it and/or modify it under
+# the terms of the MIT License; see LICENSE file for more details.
+
+from flask import Blueprint, jsonify, request
+from inspire_dojson.utils import get_recid_from_ref, get_record_ref
+from invenio_db import db
+
+from inspirehep.accounts.decorators import login_required_with_roles
+from inspirehep.accounts.roles import Roles
+from inspirehep.disambiguation.utils import create_new_empty_author
+from inspirehep.records.api import AuthorsRecord, LiteratureRecord
+
+blueprint = Blueprint("inspirehep_assign", __name__, url_prefix="/assign")
+
+
+def get_literature_records_by_recid(recids):
+    pids = [("lit", str(recid)) for recid in recids]
+    return LiteratureRecord.get_records_by_pids(pids)
+
+
+def get_author_by_recid(literature_record, author_recid):
+    return next(
+        author
+        for author in literature_record.get("authors")
+        if get_recid_from_ref(author.get("record")) == author_recid
+    )
+
+
+def unstub_author_by_recid(author_recid):
+    author = AuthorsRecord.get_record_by_pid_value(author_recid)
+    if author.get("stub") is True:
+        author["stub"] = False
+        author.update(dict(author))
+
+
+def assign_papers(from_author_recid, to_author_recid, literature_recids):
+    for record in get_literature_records_by_recid(literature_recids):
+        from_author = get_author_by_recid(record, from_author_recid)
+        from_author["record"] = get_record_ref(to_author_recid, endpoint="authors")
+        from_author["curated_relation"] = True
+        record.update(dict(record))
+
+
+def assign_to_new_stub_author(from_author_recid, literature_recids):
+    # TODO: differentiate from BEARD created stub author
+    to_author = create_new_empty_author()
+    assign_papers(from_author_recid, to_author["control_number"], literature_recids)
+
+
+def assign_to_author(from_author_recid, to_author_recid, literature_recids):
+    assign_papers(from_author_recid, to_author_recid, literature_recids)
+    unstub_author_by_recid(to_author_recid)
+
+
+@blueprint.route("", methods=["POST"])
+@login_required_with_roles([Roles.cataloger.value])
+def assign_view():
+    body = request.get_json()
+    to_author_recid = body.get("to_author_recid")
+    from_author_recid = body["from_author_recid"]
+    literature_recids = body["literature_recids"]
+    with db.session.begin_nested():
+        if to_author_recid is None:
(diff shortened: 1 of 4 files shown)


backend/inspirehep/assign/views.py
```diff
@@ -0,0 +1,69 @@
...
+    assign_papers(from_author_recid, to_author["control_number"], literature_recids)
+
+
+def assign_to_author(from_author_recid, to_author_recid, literature_recids):
+    assign_papers(from_author_recid, to_author_recid, literature_recids)
+    unstub_author_by_recid(to_author_recid)
+
+
+@blueprint.route("", methods=["POST"])
+def assign_view():
```
drjova, line 58: I'd use with `MethodView` it will organize the code better and we could extend it with more logic for other verbs later.

backend/inspirehep/assign/views.py
```diff
@@ -0,0 +1,69 @@
...
+    body = request.get_json()
+    to_author_recid = body.get("to_author_recid")
+    from_author_recid = body["from_author_recid"]
+    literature_recids = body["literature_recids"]
+    # how about a db transaction?
+    if to_author_recid is None:
+        assign_to_new_stub_author(from_author_recid, literature_recids)
+    else:
+        assign_to_author(from_author_recid, to_author_recid, literature_recids)
```
drjova, line 67: what about error handling? 🤔

backend/inspirehep/assign/views.py
```diff
@@ -0,0 +1,69 @@
...
+def assign_papers(from_author_recid, to_author_recid, literature_recids):
+    for record in get_literature_records_by_recid(literature_recids):
+        from_author = get_author_by_recid(record, from_author_recid)
+        from_author["record"] = get_record_ref(to_author_recid, endpoint="authors")
+        from_author["curated_relation"] = True
+        record.update(dict(record))
+
+
+def assign_to_new_stub_author(from_author_recid, literature_recids):
+    # do we need different data for this?, since it's not by BEARD
```
drjova, line 47: Good point, indeed it would be better to have it with different data.

backend/inspirehep/assign/views.py
```diff
@@ -0,0 +1,69 @@
...
+    unstub_author_by_recid(to_author_recid)
+
+
+@blueprint.route("", methods=["POST"])
+def assign_view():
+    body = request.get_json()
+    to_author_recid = body.get("to_author_recid")
+    from_author_recid = body["from_author_recid"]
+    literature_recids = body["literature_recids"]
+    # how about a db transaction?
```
drjova, line 63: I'd add it's case in a nested session and at the end commit (if that's what you mean).

backend/inspirehep/assign/views.py
harunurhan: do we need other verbs?

backend/inspirehep/assign/views.py
harunurhan: you mean missing input? right know it will raise key error

backend/inspirehep/assign/views.py
harunurhan: yep, just to commit all at once.

backend/inspirehep/assign/views.py
drjova: indeed for the moment we don't, it's just a suggestion and not a blocking comment

backend/inspirehep/assign/views.py
harunurhan: but BEARD needs to be triggered or at least run for this author. Does this require any specific field to be present rather than `stub: true` ?

backend/inspirehep/assign/views.py
drjova: or if something thorws on assign (updating the records etc.), it will return `500` don't we want to handle these cases better?

backend/inspirehep/assign/views.py
drjova: I don't think it's good to commit in that case, we have incomplete information and we should warn the user that the request failed instead of partially failed. wdyt

backend/inspirehep/assign/views.py
harunurhan: I am not sure do we?

backend/inspirehep/assign/views.py
harunurhan: yeah that's what I think too.

backend/inspirehep/assign/views.py
harunurhan: yeah i see, but I don't even see that in the future, so just gonna leave it like that.

backend/tests/integration/assign/test_views.py
```diff
@@ -0,0 +1,126 @@
...
+
+import json
+
+from helpers.utils import create_record
+from inspire_dojson.utils import get_recid_from_ref
+
+from inspirehep.records.api import AuthorsRecord, LiteratureRecord
+
+
+def test_assign_from_an_author_to_another(inspire_app):
```
drjova, line 16: I see some missing tests in the happy flow:
* `to_author` who doesn't have`stub=true`

I know this test is covering most of it, but adding it will test also the other case.

backend/inspirehep/assign/views.py
drjova: if we don't display any specific error messages we can leave it as is, it's for internal use.

backend/inspirehep/assign/views.py
```diff
@@ -0,0 +1,69 @@
...
+    to_author = create_new_empty_author()
+    assign_papers(from_author_recid, to_author["control_number"], literature_recids)
+
+
+def assign_to_author(from_author_recid, to_author_recid, literature_recids):
+    assign_papers(from_author_recid, to_author_recid, literature_recids)
+    unstub_author_by_recid(to_author_recid)
+
+
+@blueprint.route("", methods=["POST"])
```
drjova, line 57: what about only allow it to login user or even superadmins for now? It's dangerous to leave it open until we finilize the whole feature.

harunurhan: yeah for sure that's missing! should only be for catalogers and superusers.

Issue #1053, opened 2020-06-26
Add seminar in literature detailed page


drjova:
In the detailed page of a paper, if it has a linked seminar, add a new tab called Seminars next to Figures with the number of Seminars linked. 
The display should be similar to the search display. We don't need facets or action buttons, just display name of seminar, speaker, affiliation, date and time as in the mockup below:

<img width="408" alt="Screenshot 2020-07-03 at 16 53 19" src="https://user-images.githubusercontent.com/24955201/86480357-28ede700-bd4e-11ea-89e2-18390459df50.png">

The seminar boxes should be clickable and link to the seminar page. The speaker and affiliation are also links to the respective pages.

MJedr:
1.  In literature mapping add the field for referenced seminars 
2.  Create table (SeminarLiterature) and mixin for it, add number of papers as a property of the SeminarRecord, add calls of the mixin methods to reindex modified institutions and also call mixin method to update table when new entries are added. Add the delete and hard delete function to the SeminarRecord (the steps are described in details [here]([https://its.cern.ch/jira/browse/INSPIR-3312](https://its.cern.ch/jira/browse/INSPIR-3312))) The link should be based on the new field in literature mapping

3.  Create a component for that with the query which will get all seminars names and display them as links. Write a helper function to separate links with comma. Use ```pluralizeUnlessSingle``` component to add s to the s letter if plural. 
4.  Add the new component to the LiteratureDetailedPage

michamos:
Do we really need all of this? can't we do a search in seminars at display time instead?

harunurhan:
@michamos I think that would be better, since we expect usually one seminar and few at most. 

We could either delay the page load a bit, and perform the search on backend, on the serializer level or just do a parallel http request on the UI. I'd go with second option since this isn't the most important info for a paper, and a lot of times there is no associated seminar.

PR #1096, opened 2020-07-07
literature: add seminars to lit detail page

monaawi:
Ref: #1053


monaawi:
image

ui/src/literature/containers/DetailPageContainer/DetailPageContainer.jsx
```diff
@@ -232,6 +242,21 @@ function DetailPage({ authors, record, referencesCount, supervisors }) {
...
+                    <TabNameWithCount
+                      loading={seminarsCount == null && loadingSeminars}
+                      name="Seminars"
+                      count={seminarsCount}
+                    />
+                  }
+                  key="3"
+                  forceRender
+                >
+                  <ContentBox className="remove-top-border-of-card">
```
harunurhan, line 256: `className` is redundant

ui/src/literature/containers/LiteratureSeminars.jsx
```diff
@@ -0,0 +1,74 @@
...
+        metadata={result.get('metadata')}
+        selectedTimezone={LOCAL_TIMEZONE}
+        enableActions={false}
+      />
+    ),
+    []
+  );
+
+  return (
+    <Col span={24}>
```
harunurhan, line 42: Is this really needed? it's strange to root of the component to be `Col` without `Row`

ui/src/seminars/components/SeminarItem.jsx
```diff
@@ -32,27 +32,29 @@ function SeminarItem({ metadata, selectedTimezone }) {
...
-              icon={<FileOutlined />}
-              text="material"
-            />
-          )}
-          <ExportToCalendarAction seminar={metadata} />
-          {canEdit && (
-            <EditRecordAction pidType="seminars" pidValue={recordId} />
-          )}
-        </>
+        enableActions && (
```
harunurhan, line 35: problem with this approach (just hiding with flag), we deliver more data, and do more serializing than we need.. but I guess this is not significant at this point.

Issue #1100, opened 2020-07-09
Implement publications selection for author profile


harunurhan:
## SS

<img width="404" alt="Screenshot 2020-07-09 at 18 17 30" src="https://user-images.githubusercontent.com/24955201/87066770-643b5a80-c213-11ea-929b-58b8c36ed4fd.png">

## TECH Notes
- Add `publicationSelectionMap` in `authors` state (`recordId` to `selected`)
  - `setPublicationSelecion([...record ids...], selected)` and `clearPublicationSelection()` actions to manage selection state
- Need to add a prop for `LiteratureSearchContainer` (or wrap it for Context) to indicate if `assignView` is enabled
  - it should render select boxes for each `LiteratureItem` whose `checked` prop is set from the state (selectionMap)
       - `onChange` should call 
    ```js
    setPublicationSelecion(
          [result.get("control_number")],
          checked,
     )
  - should display number of selected records `publicationSelectionMap.size`, if > 0
  - should display select all box
    - `onChange` should call 
    ```js
    setPublicationSelection(
          [state.search.getIn(["namespaces", namespace, "results"]).map(resultToControlNumber)],
          checked,
     )
    ```
- `assignView` prop should be only set for `AuthorsPublications` if logged in user is cataloger

harunurhan:
This is separated from #1095

PR #1122, opened 2020-07-16
ui: add publication selection

harunurhan:
Ref: #1100


harunurhan:
The reason why selection of search items is not implement in a generic way, is because for now we only need it for `AuthorPublications`

ui/src/authors/containers/AuthorPublicationsContainer.jsx
```diff
@@ -40,6 +43,7 @@ const stateToProps = state => ({
     'metadata',
     'facet_author_name',
   ]),
+  assignView: isCataloger(state.user.getIn(['data', 'roles'])),
```
drjova, line 46: are we using only `cataloger` check in general?

harunurhan: yes, for now only catalogers (and superusers) will see the UI.

Issue #1072, opened 2020-07-01
Add ("nested") classification facet for Experiments search


drjova:
The only facet for search is a new facet with the inspire_classification. You can check the values here https://old.inspirehep.net/info/Experiments/list. 

The schema has the number of the entry but not the name, so we need to decide if we modify the schema so that it has the text value instead. Otherwise, we'll need to create a dict with the translations and, in experiments/es.py, add a facet_inspire_classification with the classification name and the names of the parent classes too so that it works if we filter by a more general class.

For nesting, as Micha proposes, we could simply pass some kind of separator in the field that the UI reads and adds a padding before the corresponding checkbox.

This can be done with pathhierarchy-tokenizer in Elasticsearch
https://www.elastic.co/guide/en/elasticsearch/reference/current/analysis-pathhierarchy-tokenizer.html

 

TECH NOTES:

    We should use ES pathhierarchy-tokenizer. We need to add a new tokenizer of type "path_hierarchy" and delimiter "|" to experiment's mappings and add it to a new analyzer that should then be used in the inspire_classification field.
    After this, for the previous example, we would have ["Neutrino (flavor) experiments", "Neutrino (flavor) experiments|Reactor", "Neutrino (flavor) experiments|Reactor|Longer baselines"] as inspire_classification.
    In the frontend, we should add a TreeAggregation component using ant's tree. We should add a meta parameter in the facet indicating the character that we're splitting by ("|"), and the aggregation should be of a new type "tree". We will have to map the results returned by the backend to an object with the desired structure for the tree. So the previous example should be mapped to something like
```js
    tree = [
      {
        title: 'Neutrino (flavor) experiments',
        children: [
          {
            title: 'Reactor',
            children: [
              {
                title: 'Longer baselines'
              }
            ]
          }
        ]
      }
    ]
```
    By default only the root categories should be displayed (the children should be compressed).

PR #1123, opened 2020-07-16
facets: tree classification facet for experiments

miguelgrc:
Ref: #1072

### DIFF
--- a/ui/package.json
+++ b/ui/package.json
@@ -36,7 +36,7 @@
     "craco-antd": "1.11.0",
     "formik": "^2.0.0",
     "history": "^4.7.2",
-    "immutable": "^3.8.2",
+    "immutable": "4.0.0-rc.10",
     "inspire-schemas": "^61.2.6-2-g90a8aaf",
     "is-valid-orcid": "^1.2.0",
     "jsondiffpatch": "^0.3.11",
--- a/ui/src/common/components/TreeAggregation.jsx
+++ b/ui/src/common/components/TreeAggregation.jsx
@@ -1,0 +1,91 @@
+import React, { useState, useMemo } from 'react';
+import PropTypes from 'prop-types';
+import { List, setIn } from 'immutable';
+import { Col, Row, Tree } from 'antd';
+
+import UnclickableTag from './UnclickableTag';
+import AggregationBox from './AggregationBox';
+import { forceArray } from '../utils';
+import FormattedNumber from './FormattedNumber';
+
+function renderTitle(name, docCount) {
+  return (
+    <Row className="mv1" justify="space-between" key={name}>
+      <Col>{name}</Col>
+      <Col>
+        <UnclickableTag>
+          <FormattedNumber>{docCount}</FormattedNumber>
+        </UnclickableTag>
+      </Col>
+    </Row>
+  );
+}
+
+function buildTreeData(buckets, splitTreeBy) {
+  let tree = {};
+  buckets.forEach(bucket => {
+    const docCount = bucket.get('doc_count');
+    const key = bucket.get('key');
+    // convert to list just for `flatMap`, since Array.prototype.flatMap needs to be polyfilled
+    // a|b|c => turns into [children, a, children, b, children, c] (the path of the node)
+    const path = List(key.split(splitTreeBy)).flatMap(node => [
+      'children',
+      node,
+    ]);
+
+    tree = setIn(tree, path.push('title'), renderTitle(path.last(), docCount));
+    tree = setIn(tree, path.push('key'), key);
+  });
+
+  function convertChildrenToArray(node) {
+    if (node.children) {
+      node.children = Object.values(node.children);
+      node.children.forEach(convertChildrenToArray);
+    }
+  }
+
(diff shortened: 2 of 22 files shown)


ui/src/experiments/containers/SearchPageContainer.jsx
```diff
@@ -19,7 +23,31 @@ function renderExperimentItem(result) {
...
+  const renderAggregations = useCallback(
+    () => (
+      <LoadingOrChildren loading={loadingAggregations}>
+        <AggregationFiltersContainer namespace={EXPERIMENTS_NS} />
+      </LoadingOrChildren>
+    ),
+    [loadingAggregations]
+  );
+
+  useEffect(
```
harunurhan, line 41: unnecessary

ui/src/common/components/TreeAggregation.jsx
```diff
@@ -0,0 +1,164 @@
...
+      .toArray();
+    onChange(selections);
+  }
+
+  getCheckedKeys(buckets) {
+    const { selectionMap } = this.state;
+    const checkedKeys = buckets
+      .map(bucket => bucket.get('key'))
+      .filter(bucketKey => selectionMap.get(bucketKey));
+    return checkedKeys.toJS();
```
harunurhan, line 68: `toArray`

ui/src/common/components/TreeAggregation.jsx
```diff
@@ -0,0 +1,164 @@
...
+import { Col, Row, Tree } from 'antd';
+
+import UnclickableTag from './UnclickableTag';
+import AggregationBox from './AggregationBox';
+import { forceArray } from '../utils';
+import FormattedNumber from './FormattedNumber';
+
+export const BUCKET_NAME_SPLITTER = '_';
+
+class TreeAggregation extends Component {
```
harunurhan, line 13: we don't write new components this way anymore, instead prefer function components with hooks.

harunurhan: you can start writing the tests, then refactor which should make it easier.

ui/src/common/components/TreeAggregation.jsx
```diff
@@ -0,0 +1,164 @@
...
+      docCount,
+      splitTreeBy
+    );
+    return subTree;
+  }
+
+  render() {
+    const { name, buckets, splitTreeBy } = this.props;
+    let tree = [];
+    for (let i = 0; i < buckets.size; i += 1) {
```
harunurhan, line 117: `render` called many times so in general doing a loop like this, could lead performance problems. and that should be memoized

ui/src/search/config.js
```diff
@@ -79,7 +79,7 @@ const searchConfig = {
   },
   [EXPERIMENTS_NS]: {
     persistedQueryParamsDuringNewSearch: defaultPersistedQueryParamsDuringNewSearch,
-    onQueryChange: onAggregationlessCollectionQueryChange,
+    onQueryChange: onEventsQuerychange,
```
harunurhan, line 82: maybe rename the `onEventsQuerychange` to something that fits to `seminars`, `conferences`, `experiments`

ui/src/experiments/containers/SearchPageContainer.jsx
```diff
@@ -53,11 +94,25 @@ function ExperimentSearchPage({ loading }) {
...
   loading: state.search.getIn(['namespaces', EXPERIMENTS_NS, 'loading']),
+  loadingAggregations: state.search.getIn([
+    'namespaces',
+    EXPERIMENTS_NS,
+    'loadingAggregations',
+  ]),
+});
+
+const dispatchToProps = dispatch => ({
+  onBaseQueryChange(namespace, baseQuery) {
```
harunurhan, line 113: also this., baseQueries only change for embedded searches.

ui/src/common/components/TreeAggregation.jsx
harunurhan: also this logic should be part of `buildTree` function

miguelgrc: In which situations can that happen? The page is re-rendered when the filter changes and in that case we need to recreate the tree so it seems okay like this.

ui/src/search/config.js
miguelgrc: Maybe `onCollectionQueryChange`?

ui/src/common/components/TreeAggregation.jsx
harunurhan: you need to rebuild the tree only when `buckets` change, but this component's render method can be called for many different reasons, such other props change, internal state change, parent rerender etc.

ui/package.json
```diff
@@ -36,7 +36,7 @@
     "craco-antd": "1.11.0",
     "formik": "^2.0.0",
     "history": "^4.7.2",
-    "immutable": "^3.8.2",
+    "immutable": "4.0.0-rc.10",
```
drjova, line 39: rc?

harunurhan: yeah, it's fine it has been `rc` for so long that I think it's stable :)

Issue #1103, opened 2020-07-09
Modify behavior for assigning to stub author in the Assign API


harunurhan:
Assign API should return `stub_author_id` if `to_author_recid` is null and should set the stub author's name as being the longest name of this author on the papers
* Return ```stub_author_id``` when ```to_author_recid``` is null
* In ```assign_to_new_stub_author```, After creating the stub author and assigning the papers to it. Set the name of the stub author as the longest name of the author on the assigned papers

PR #1127, opened 2020-07-20
assign: update stub author name and return id

monaawi:
ref #1103

### DIFF
--- a/backend/inspirehep/assign/views.py
+++ b/backend/inspirehep/assign/views.py
@@ -11,7 +11,7 @@ from invenio_db import db
 
 from inspirehep.accounts.decorators import login_required_with_roles
 from inspirehep.accounts.roles import Roles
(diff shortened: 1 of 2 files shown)


backend/inspirehep/assign/views.py
```diff
@@ -65,8 +72,13 @@ def assign_view():
...
+                from_author_recid, literature_recids
+            )
         else:
             assign_to_author(from_author_recid, to_author_recid, literature_recids)
     db.session.commit()
-    return jsonify({"message": "Success"}), 200
+    if to_author_recid is None:
+        return jsonify({"stub_author_id": stub_author_id}), 200
+    else:
+        return jsonify({"message": "Success"}), 200
```
drjova, line 84:
```suggestion
    if to_author_recid is None:
        return jsonify({"stub_author_id": stub_author_id}), 200
    return jsonify({"message": "Success"}), 200
```

Issue #1120, opened 2020-07-16
Remove hidden flag for arxiv pdf from `inspire-dojson` and remove external links from `citation_pdf_url`


drjova:
- Currently we're marking them as `hidden` in the inspire-dojson, we should remove this flag and remigrate all records. 
- Filter out the internal arxiv document from fulltext_links
- Add a new `citation_pdf_urls` which contains all documents that are not `hidden:true`

Dojson hidden flag:
https://github.com/inspirehep/inspire-dojson/blob/abb782981e112e6b0c76744bfb0081d445170924/inspire_dojson/hep/rules/bdFFT.py#L46

### Acceptance

- The arxiv file should be included in the `citation_pdf_url` meta tag for example:

```html
<meta name="citation_pdf_url" content="https://inspirehep.net/files/184f3661c7f67bfb45082b015648189a" data-react-helmet="true">
```
- `citation_pdf_url` should not have any external hosted links

### TECH Notes
https://inspirehep.zulipchat.com/#narrow/stream/122202-devel/topic/files.20for.20google.20scholar/near/204519795
- add `citation_pdf_urls` field to [ui schema](https://github.com/inspirehep/inspirehep/blob/0fa1202867a78ad66b14f4808c068ecc960fe70c/backend/inspirehep/records/marshmallow/literature/ui.py), logic should be identical to https://github.com/inspirehep/inspirehep/blob/0fa1202867a78ad66b14f4808c068ecc960fe70c/backend/inspirehep/records/marshmallow/literature/ui.py#L137 but plain urls instead of `{ description, value }` objects
- `citation_pdf_urls` should contain only locally served files.
- use `citation_pdf_urls` field here in [LiteratureDocumentHead](https://github.com/inspirehep/inspirehep/blob/0fa1202867a78ad66b14f4808c068ecc960fe70c/ui/src/literature/components/LiteratureDocumentHead.jsx#L79) instead of `fulltext_links`

PR #1128, opened 2020-07-20
serializers: remove hidden flag from arxiv pdfs

pazembrz:
add new field citation_pdf_urls for arxiv pdfs

ref #1120


backend/inspirehep/records/marshmallow/literature/ui.py
```diff
@@ -187,6 +192,26 @@ def add_ads_links_for_arxiv_papers(self, data):
...
+        return document.get("hidden", False)
+
+    def get_citation_pdf_urls(self, data):
+        urls = []
+        for document in data.get("documents", []):
+            url = document.get("url")
+            if (
+                url
+                and not self.is_document_hidden(document)
+                and self.is_document_arxiv(document)
```
drjova, line 208: this logic is wrong, we should return all documents that are not `hidden` (we don't need the `is_document_arxiv` check)

backend/inspirehep/records/marshmallow/literature/ui.py
```diff
@@ -134,12 +135,16 @@ def get_arxiv_fulltext_link(self, data):
...
             return missing
-        description = data.get("description") or "fulltext"
-        url = data.get("url")
-        if url and not data.get("hidden", False):
+        description = document.get("description") or "fulltext"
+        url = document.get("url")
+        if (
+            url
+            and not self.is_document_hidden(document)
+            and not self.is_document_arxiv(document)
```
drjova, line 146: I'm not sure this check is enough, do we keep ``source`` in the ``documents``?

backend/inspirehep/records/marshmallow/literature/ui.py
pazembrz: Ah ok I misunderstood requirement. :+1:

backend/inspirehep/records/marshmallow/literature/ui.py
pazembrz: We should keep source in documents, but it's not required field. However this is how we were checking if doc is from arxiv in dojson if I understand this correctly 
https://github.com/inspirehep/inspire-dojson/commit/dec6e94df73165b06a5b15796c64fba81c408803#diff-d3b380993699eaa3534ec6fac751e842

https://github.com/inspirehep/inspire-dojson/blob/dec6e94df73165b06a5b15796c64fba81c408803/inspire_dojson/hep/rules/bdFFT.py#L67

And source is also populated from _get_source() https://github.com/inspirehep/inspire-dojson/blob/dec6e94df73165b06a5b15796c64fba81c408803/inspire_dojson/hep/rules/bdFFT.py#L99

drjova: I don't see we keep such informaton https://github.com/inspirehep/inspirehep/blob/master/backend/inspirehep/records/api/literature.py#L374 though, do I miss something?

pazembrz: Ok, When I'm analyzing this we are not able to figure out if document is from arxiv or not. Previously we were not downloading arxiv documents so all of them had source preserved, but when we download them we forget the source. I will add proper changes to store the source as this is the only way to find out if file is arxiv after downloading it. Otherwise We can just skip downloading of ArXiv files. We are missing stuff in refinement...

PR #1129, opened 2020-07-21
serializer: add new field citation_pdf_urls which will contain only local documents

pazembrz:
ref #1120

### DIFF
--- a/backend/inspirehep/records/marshmallow/literature/ui.py
+++ b/backend/inspirehep/records/marshmallow/literature/ui.py
@@ -13,6 +13,7 @@ from inspire_utils.record import get_value, get_values_for_schema
 from marshmallow import fields, missing, pre_dump
 
 from inspirehep.accounts.api import is_superuser_or_cataloger_logged_in
+from inspirehep.files.api import current_s3_instance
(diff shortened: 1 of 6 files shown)


backend/inspirehep/records/marshmallow/literature/ui.py
```diff
@@ -187,6 +189,26 @@ def add_ads_links_for_arxiv_papers(self, data):
 
         return data
 
+    @staticmethod
+    def is_document_hidden(document):
+        return document.get("hidden", False)
+
+    @staticmethod
+    def is_document_local(document):
+        return current_s3_instance.is_public_url(document["url"])
```
drjova, line 198: not sure about it, now we're using ``/files/...` will it work? Also I think it's unecessary for `get_citation_pdf_urls` since `documents` with `hidden:false` are always processed by us right?

pazembrz: Yes we are using `/files/` and `is_public_url` knows about this and checks if url is ours and is served with public url. 
Files with hidden:False are always processed by us, but if a record had an error on downloading file then this data won't be updated. So there is a chance that there will be a public file with url which is not local.

Issue #1052, opened 2020-06-26
Hide successful claims


drjova:
In the embed search in author profile, create a facet called `Claims` with the options: `Claimed papers` and `Unclaimed papers` . This should be the last facet

`authors.curated_relation:true`

<img width="153" alt="Screenshot 2020-07-09 at 17 21 58" src="https://user-images.githubusercontent.com/24955201/87058783-c80c5600-c208-11ea-99eb-4d099715cf62.png">

**TECH NOTES**

- Add a new nested aggregation Claims on `authors.curated_relation` with two options: Claimed papers and Unclaimed papers (true/false). The author whose `curated_relation` is checked should be the current one and should be sent in the request. It would be similar to `hep_self_author_affiliations_aggregation` but should have two filters in the inner aggregation.
- Don't forget to configure a filter too
- The facet should be the last one in the embedded search in author profile
- The facet should be visible only to curators

PR #1137, opened 2020-07-22
facets: add facet for claimed papers

MJedr:
Ref: #1052

### DIFF
--- a/backend/inspirehep/search/aggregations.py
+++ b/backend/inspirehep/search/aggregations.py
@@ -252,6 +252,53 @@ def hep_rpp(order, title="Exclude RPP", agg_type="checkbox"):
     }
 
 
+def hep_self_author_claimed_papers_aggregation(
+    order, author_recid, title="Claims", agg_type="checkbox"
+):
+    return {
+        "self_curated_relation": {
+            "filters": {
+                "filters": {
+                    "Claimed papers": {
+                        "nested": {
+                            "path": "authors",
+                            "query": {
+                                "bool": {
+                                    "must": [
+                                        {"term": {"authors.curated_relation": True}},
+                                        {"term": {"authors.record.$ref": author_recid}},
+                                    ]
+                                }
+                            },
+                        }
+                    },
+                    "Unclaimed papers": {
+                        "nested": {
+                            "path": "authors",
+                            "query": {
+                                "bool": {
+                                    "must_not": [
+                                        {"term": {"authors.curated_relation": True}}
+                                    ],
+                                    "must": [
+                                        {"term": {"authors.record.$ref": author_recid}}
+                                    ],
+                                }
+                            },
+                        }
+                    },
+                },
+                "meta": {
+                    "title": title,
+                    "order": order,
+                    "type": agg_type,
+                    "is_filter_aggregation": True,
+                },
+            }
+        }
+    }
+
+
 def seminar_subject_aggregation(order, title="Subject", agg_type="checkbox"):
     return {
         "subject": {
--- a/backend/inspirehep/search/facets.py
+++ b/backend/inspirehep/search/facets.py
@@ -23,6 +23,7 @@ from inspirehep.search.aggregations import (
     hep_earliest_date_aggregation,
     hep_rpp,
     hep_self_author_affiliations_aggregation,
+    hep_self_author_claimed_papers_aggregation,
     hep_self_author_names_aggregation,
     hep_subject_aggregation,
     jobs_field_of_interest_aggregation,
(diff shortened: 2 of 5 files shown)


backend/inspirehep/search/aggregations.py
```diff
@@ -252,6 +252,37 @@ def hep_rpp(order, title="Exclude RPP", agg_type="checkbox"):
...
+                            "query": {"term": {"authors.curated_relation": False}},
+                        }
+                    },
+                }
+            },
+            "meta": {
+                "title": title,
+                "order": order,
+                "type": agg_type,
+                "is_filter_aggregation": True,
```
MJedr, line 280: After adding this only one checkbox shows (for Claimed papers) and without it there are two checkboxes but when you click on one the other is checked too.

drjova: what's the request url when you click one of the two?

backend/inspirehep/search/aggregations.py
```diff
@@ -252,6 +252,37 @@ def hep_rpp(order, title="Exclude RPP", agg_type="checkbox"):
...
+                            "path": "authors",
+                            "query": {"term": {"authors.curated_relation": True}},
+                        }
+                    },
+                    "Unclaimed papers": {
+                        "nested": {
+                            "path": "authors",
+                            "query": {"term": {"authors.curated_relation": False}},
+                        }
+                    },
```
michamos, line 273: not sure this solves anything, but why not use an `other` bucket for this part? https://www.elastic.co/guide/en/elasticsearch/reference/current/search-aggregations-bucket-filters-aggregation.html#other-bucket

backend/inspirehep/search/aggregations.py
michamos: this looks very similar to the seminars thing, so I don't see why it doesn't work here.

backend/inspirehep/search/facets.py
```diff
@@ -177,6 +178,29 @@ def inner(values):
...
+                "query": {"term": {"authors.curated_relation": True}},
+            }
+        },
+        "Unclaimed papers": {
+            "nested": {
+                "path": "authors",
+                "query": {"term": {"authors.curated_relation": False}},
+            }
+        },
+    }
```
michamos, line 195: * not sure this part is correct, as you need to produce a `nested` selecting the author and the claim status together (if you have two separate `nested`, you're doing a selection based on claims for any authors here).
* also, because `curated_relation` is not mandatory, it might simply be absent, so better do a search for                                        `not True` instead of `False`.

backend/inspirehep/search/aggregations.py
MJedr: Without `"is_filter_aggregation": True` the request doesn't contain this filter. With it it seems to be correct (but only for Claimed papers)

backend/inspirehep/search/aggregations.py
michamos: also I don't see how you're restricting to the relevant author here. Doesn't this look for claims in all authors?

backend/inspirehep/search/facets.py
```diff
@@ -177,6 +178,41 @@ def inner(values):
...
+                        "must": [{"term": {"authors.record.$ref": author_recid}}],
+                    }
+                },
+            }
+        },
+    }
+
+    def inner(values):
+        filters = [filter_for_option[value] for value in values]
+        return Q("bool", must=filters)
```
michamos, line 211:
```suggestion
        return Q("bool", filter=filters)
```
as filters are more performant but don't contribute to the score, see https://www.elastic.co/guide/en/elasticsearch/reference/7.1/query-filter-context.html for a more detailed explanation.

backend/inspirehep/search/facets.py
```diff
@@ -177,6 +178,41 @@ def inner(values):
     return inner
 
 
+def curation_filter(author_recid):
```
michamos, line 181:
```suggestion
def self_author_claimed_papers_filter(author_recid):
```
to be consistent with the aggregation as it's not generic at all?

backend/tests/integration/search/test_facets.py
```diff
@@ -361,11 +363,19 @@ def test_hep_author_publications_cataloger_facets(inspire_app):
...
-        filters = current_app.config["CATALOGER_RECORDS_REST_FACETS"][
-            "hep-author-publication"
-        ]()["filters"].keys()
+        filters = set(
+            current_app.config["CATALOGER_RECORDS_REST_FACETS"][
+                "hep-author-publication"
+            ]()["filters"].keys()
+        )
+        filters.update(
+            {"self_curated_relation": curation_filter(author_recid=author_recid)}.keys()
```
michamos, line 377: why do you need to do this manually?

Issue #1131, opened 2020-07-21
date-added/date-updated queries should use range queries


michamos:
Searches such as `da 2009` or `du 2009` don't work as they don't use `range`: https://inspirehep.net/search/query-parser?q=da%202009. This should be fixed, by using `range` for all date search operators.

PR #1138, opened 2020-07-24
mappings: support nested queries for _updated and _created fields

MJedr:
Ref: #1131

### DIFF
--- a/backend/inspirehep/search/mappings/v7/records/hep.json
+++ b/backend/inspirehep/search/mappings/v7/records/hep.json
@@ -10,9 +10,11 @@
         "type": "text"
       },
       "_created": {
+        "format": "date_optional_time||yyyy-MM-dd||yyyy-MM||yyyy",
         "type": "date"
       },
       "_updated": {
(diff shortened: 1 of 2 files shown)


backend/inspirehep/search/mappings/v7/records/hep.json
```diff
@@ -10,9 +10,11 @@
         "type": "text"
       },
       "_created": {
+        "format": "yyyy-MM-dd'T'HH:mm:ss.SSSSSS||yyyy-MM-dd||yyyy-MM||yyyy",
```
MJedr, line 13: `yyyy-MM-dd'T'HH:mm:ss.SSSSSS` that is the current format, but do we need a timestamp for that field?

michamos: it's handled by `invenio-records`, so I'm not sure we can easily change the format here

MJedr: So we can't use other formats too?

Issue #1136, opened 2020-07-22
Wrong `/api/facet` response, instead of `aggs` it returns `hits`


drjova:
This example https://inspirehep.net/api/literature/facets?sort=mostrecent&size=25&page=1&q=Particles%201%20%282018%29%20%20138 returns `hits` instead of `aggs` which is wrong. The query is `Particles 1 (2018)  138`

MJedr:
Facets are not showing for this query https://inspirehep.net/literature?sort=mostrecent&size=25&page=1&q=Particles%201%20%282018%29%20138

michamos:
Yes, because the `/literature/facets endpoints doesn't return `aggregations` but full `hits`.

MJedr:
I think that for some searches it still doesn't work https://inspirehep.net/literature?sort=mostrecent&size=25&page=1&q=Phys.%20Rev.%20D95%20%282017%29%20123002

drjova:
We don't return facets for pubnote search

monaawi:
Yes, this is what we decided to do for now

PR #1161, opened 2020-07-31
search: facets search should not trigger reference match

monaawi:
* ref #1136
* Note: we will still not have facets for pub note searches

### DIFF
--- a/backend/inspirehep/search/api.py
+++ b/backend/inspirehep/search/api.py
@@ -277,6 +277,11 @@ class LiteratureSearch(InspireSearch):
         return results.execute().hits
 
 
+class LiteratureAggregationsSearch(LiteratureSearch):
+    def execute(self, *args, **kwargs):
+        return super(LiteratureSearch, self).execute(*args, **kwargs)
+
+
 class AuthorsSearch(InspireSearch):
     """Elasticsearch-dsl specialized class to search in Authors database."""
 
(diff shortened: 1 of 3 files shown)


backend/inspirehep/search/api.py
```diff
@@ -277,6 +277,11 @@ def get_records_by_pids(pids, source=None, size=10000):
         return results.execute().hits
 
 
+class LiteratureAggregationsSearch(LiteratureSearch):
+    def execute(self, *args, **kwargs):
+        return super(LiteratureSearch, self).execute(*args, **kwargs)
```
drjova, line 282:
```suggestion
        return super().execute(*args, **kwargs)
```

backend/tests/integration/search/test_api.py
```diff
@@ -81,6 +81,40 @@ def test_return_record_for_publication_info_search_with_journal_title_without_do
     assert 200 == response.status_code
 
 
+def test_get_facets_for_publication_info_search(inspire_app):
```
drjova, line 84:
```suggestion
def test_facets_for_publication_info_search(inspire_app):
```

backend/inspirehep/search/api.py
monaawi: But I need to call the execute from InspireSearch, not LiteratureSearch.

drjova: true my bad

PR #1169, opened 2020-08-03
Dispatch events to github.

benjamin-bergia:


build-and-deploy.sh
```diff
@@ -108,10 +108,25 @@ maybeDeploySmokeTestsQA() {
   fi
 }
 
+dispatch_github_event() {
```
drjova, line 111: I'd just change the name to camelCase, since is used for the other function.

benjamin-bergia: Yep sorry. I will

Issue #1162, opened 2020-07-31
Add institution facet in experiments


StellaCh:
As a user, I want to be able to filter experiments based on their institution.

In the search experiment page https://inspirehep.net/experiments add a facet called 'Institution' that contains the institutions.value of an experiment.

**Tech Notes** : 
* Add a terms aggregation on ```institutions.value```  (it is already keyword)
* Add a filter in ```EXPERIMENTS_FILTERS```
* Add the new filter and aggregation to ```records_experiments``` facets

PR #1177, opened 2020-08-06
experiments: add institution facet

harunurhan:
Ref: #1162


backend/inspirehep/records/config.py
```diff
@@ -535,7 +535,8 @@
 }
 
 EXPERIMENTS_FILTERS = {
-    "classification": must_match_all_filter("inspire_classification")
+    "classification": must_match_all_filter("inspire_classification"),
+    "institution": terms_filter("institutions.value"),
```
monaawi, line 539: This should be ```must_match_all_filter```

Issue #1102, opened 2020-07-09
Add the curated_relation field to hits for assign view


monaawi:
### SS
<img width="382" alt="Screenshot 2020-07-15 at 12 57 22" src="https://user-images.githubusercontent.com/24955201/87537463-ec4db400-c69a-11ea-8bcd-1337c6583e3b.png">

### Tech Notes
In #1100 we already set assignView for the ```AuthorPublications```. 
1. Create a class that derives from ```JSONSERIALIZER``` that will be used for ```literature_json_list_response```. It should override the ```serialize_search```. Check if there is an author param and user is superuser or curator. If yes,it should do an ES query using the ```author``` from the request params and the control numbers of the current ```search_results```, the query should be looking for literature records with these records ids that have ```curated_relation : True``` for the given author (nested query). Then, populate ```curated_relation``` within these search results. And call super. (query is similar to https://github.com/inspirehep/inspirehep/blob/1e9c1df61ff75a0b715cf6bb5a051c83431447a9/backend/inspirehep/search/aggregations.py#L258-L300)
2. In the ```LiteratureListWrappedSchema``` in ```get_ui_display```, set the ```curated_relation``` from the metadata.
3. Create the component with the ```< FileExclamationTwoTone / >``` icon (add a tooltip when you hover over the icon: *Unclaimed paper. Click on the Claim button to claim or remove it from the profile*) and dropdown menu. 
4. In LiteratureItem check if assignView is True and if the value of the curated_relation for the current user (author) is False in metadata. If the conditions are met display created component.

monaawi:
Separated from #1095

StellaCh:
I need to update the icon. Since we changed the color of the titles to black, the gray icon won't stand out anymore

StellaCh:
I updated the screenshot. Let's use <FileExclamationTwoTone />

pazembrz:
`curated_relation` flag is separate for every author in every paper. Currently UI have access to only first 10 authors, to get all authors for every paper will be to expensive as response size will be too big and filtering through 5k authors to get proper one will be slow. 
There might be few solution to this:
1. On ui serializer we can put list of all curated_relations for each paper and then on UI check if current author is on this list
2. On LiteratureListWrappedSchema  serializer query ES/DB for author and current paper to check if relation is curated.
Maybe there is better solution. 
This requires more refinement.

PR #1183, opened 2020-08-07
assign: add curated_relation to lit search hits for auth publications 

monaawi:
* ref #1102


backend/inspirehep/records/marshmallow/literature/ui.py
```diff
@@ -223,6 +223,11 @@ def get_ui_display(self, data):
             ui_display = json.loads(get_value(data, "metadata._ui_display", ""))
             if is_superuser_or_cataloger_logged_in():
                 ui_display["can_edit"] = True
+                search_type = request.values.get("search_type", "", type=str)
+                if search_type == "hep-author-publication":
+                    ui_display["curated_relation"] = get_value(
+                        data, "metadata.curated_relation", False
+                    )
```
drjova, line 230: The permission check will change in the future, it will be better if we isolate it from `is_superuser_or_cataloger_logged_in` check here

backend/inspirehep/records/marshmallow/literature/ui.py
```diff
@@ -223,6 +223,11 @@ def get_ui_display(self, data):
             ui_display = json.loads(get_value(data, "metadata._ui_display", ""))
             if is_superuser_or_cataloger_logged_in():
                 ui_display["can_edit"] = True
+                search_type = request.values.get("search_type", "", type=str)
+                if search_type == "hep-author-publication":
```
drjova, line 227: do we really need the `search_type`, content type is not enough to distinguish the request we could then have the serialize just for this `content-type` without checking in both pleases for `search_type == "hep-author-publication"`?

harunurhan:
Maybe it's good to change the commit message, since no icon now.

backend/inspirehep/records/marshmallow/literature/ui.py
monaawi: so you are saying to use a different ```content-type``` for the lit search request for author publications? I don't think a new ```content-type``` is needed, I think it is enough to use ```search_type``` in this case

drjova:
@monaawi I think we could make it simpler.

Why we need to add the `curated_relation` under any check? It's just a field and we could have it in general in the `ui_display` https://github.com/inspirehep/inspirehep/pull/1183/files#diff-19d3d162ce2a08a6469fc1ae8afae85fR228. Doing that we make the serializer cleaner, we already have `author_recid` param so we don't need to pass around the `search_type` and all the permission checks are only in the serializer https://github.com/inspirehep/inspirehep/pull/1183/files#diff-d168c2343e3473ca98acd837bd33e8fcR163

My point is, we already have undocumented params passing around and we shouldn't add more if it's possible.

WDYT?

monaawi:
I don't think that's a good idea. We need to have it at serializing time, we can't put it in ```ui_display``` from beginning because we need to know the author for the request. The idea why we added the checks was to avoid doing extra ES calls to find the ```curated_relation```. We want to do it only when needed. We pass the ```author_recid``` also in cases when we don't have the assign view, so we can't rely on that. (we pass it for example for the author facet in normal lit search). If we don't add the check, we would be doing useless ES queries in this case.

Issue #1115, opened 2020-07-15
Release experiments


StellaCh:
- In literature detailed page, link the experiments to the corresponding page. 
Example: https://inspirehep.net/literature/1809147 link CERN-LHC-ATLAS to https://inspirehep.net/experiments/1108541

-  Also, link experiments in Authors Search page and Author detail page. Example author: https://inspirehep.net/authors/1032472

- Add Experiments dropdown in search menu and extra Experiments tab after Institutions menu. 

- Hide Experiments and Institutions in the tab under a `More...` menu.
<img width="505" alt="Screenshot 2020-07-30 at 10 52 52" src="https://user-images.githubusercontent.com/24955201/88902944-27cbaf00-d253-11ea-9f16-81f9038a389f.png">

- Submission form should be visible only to curators.

**TECH NOTES** : 
* ```ExperimentList``` and ```AcceleratorExperimentsList``` do the same thing. There should be only one component, get label as prop. Change this component to have a Link to the experiment in ```renderExperiment```. 
* In the ```SearchScopeSelect``` remove the check for experiments so that they appear in the namespace dropdown. Make sure they appear in the correct order.
* In the ```HeaderMenu```, add a new menu item for Experiments. See how things fit in smaller screens. If needed, add the `More..`

MJedr:
I think that also linking experiments in Authors Search page and Author detail page is necessary.

StellaCh:
You're right, I missed this!

monaawi:
@StellaCh What about in the Job detail page? we have experiments there as well

monaawi:
@StellaCh  also we don't have the submissions experiments yet

StellaCh:
Good catch yes, everywhere we have experiments.
It's ok, we can release like we do for Institutions

MJedr:
@StellaCh for other links when they are active we hover them. Should we do it also for the links in the dropdown menu (experiments and institutions)?

StellaCh:
yes thanks!

StellaCh:
I forgot to add that next to the experiments dropdown in the tab menu we should display the 'New' green tag. No need to add it in the search dropdown

PR #1184, opened 2020-08-10
experiments: prepare release

MJedr:
Ref: #1115


ui/src/common/components/__tests__/__snapshots__/ExperimentList.test.jsx.snap
```diff
@@ -12,9 +12,13 @@ exports[`ExperimentList renders arxiv categories 1`] = `
   </span>
   <ul>
     <li
-      key="Map { \\"name\\": \\"CERN-LHC-CMS\\" }"
+      key="Map { \\"name\\": \\"CERN-LHC-CMS\\", \\"record\\": Map { \\"$ref\\": \\"http://labs.inspirehep.net/api/experiments/1110623\\" } }"
```
harunurhan, line 15: `key` should be `name` not the whole record.

ui/src/common/layouts/CollectionsMenu/CollectionsMenu.jsx
```diff
@@ -68,12 +70,35 @@ function CollectionsMenu({ currentPathname }) {
         </CollectionLink>
       </Col>
       <Col>
-        <CollectionLink
-          active={activeCollection === INSTITUTIONS_PID_TYPE}
-          to={INSTITUTIONS}
+        <Menu
+          className="__CollectionsDropdownMenu__ mh4 m-mh2"
```
harunurhan, line 74: this kind of class names are given to root tag of a components

ui/src/common/layouts/CollectionsMenu/CollectionsMenu.jsx
```diff
@@ -68,12 +70,35 @@ function CollectionsMenu({ currentPathname }) {
...
+          theme="dark"
+          mode="horizontal"
+          selectable={false}
+          inlineIndent={0}
         >
-          Institutions
-        </CollectionLink>
+          <Menu.SubMenu key="More" title="More...">
```
harunurhan, line 80: You should use `DropdownMenu` instead, since `SubMenu` has some features we don't need here.

ui/src/common/layouts/CollectionsMenu/CollectionsMenu.scss
```diff
@@ -5,4 +5,15 @@
   @media (max-width: $screen-sm-max) {
     display: none;
   }
+
+  .__CollectionsDropdownMenu__ {
+    line-height: normal;
```
harunurhan, line 10: ?

ui/src/common/layouts/CollectionsMenu/CollectionsMenu.jsx
harunurhan: does `m-mh2` exist?

ui/src/common/layouts/CollectionsMenu/CollectionsMenu.jsx
```diff
@@ -68,12 +70,35 @@ function CollectionsMenu({ currentPathname }) {
...
+              <CollectionLink
+                active={activeCollection === INSTITUTIONS_PID_TYPE}
+                to={INSTITUTIONS}
+                isDropdownLink
+              >
+                Institutions
+              </CollectionLink>
+            </Menu.Item>
+            <Menu.Item key="more.experiments">
+              <CollectionLink
```
harunurhan, line 91: These don't need to use `CollectionLink`, just a normal `Link` would be enough since we don't need any logic or customization of `CollectionLink` here.

ui/src/common/layouts/CollectionsMenu/CollectionsMenu.jsx
```diff
@@ -68,12 +70,35 @@ function CollectionsMenu({ currentPathname }) {
...
+                active={activeCollection === EXPERIMENTS_PID_TYPE}
+                to={EXPERIMENTS}
+                isDropdownLink
+                newCollection
```
harunurhan, line 95: And I would put new tag for `More...` since it's not visible enough inside the dropdown.

backend/inspirehep/records/marshmallow/common/accelerator_experiment.py
```diff
@@ -50,3 +57,6 @@ def get_name(self, item):
         if institution and accelerator and experiment:
             return "{}-{}-{}".format(institution, accelerator, experiment)
         return item.get("legacy_name")
+
+    def get_reference(self, item):
+        return get_value(item, "record")
```
harunurhan, line 62: you could just have `record = fields.Raw()` instead of this.

backend/inspirehep/records/marshmallow/common/accelerator_experiment.py
```diff
@@ -29,6 +30,12 @@ def resolve_experiment_records(self, data, many):
             for experiment in data
         ]
 
+    @post_dump
+    def remove_skip_values(self, data):
+        if not data.get("record"):
```
harunurhan, line 35: is this really necessary? why would `record` ever be an empty `dict`?

backend/tests/integration/records/serializers/json/test_jobs.py
```diff
@@ -252,3 +252,29 @@ def test_jobs_detail_json_link_alias_format(inspire_app):
...
+    expected_status_code = 200
+    expected_accelerator_experiments = [
+        {
+            "name": "LIGO",
+            "record": {"$ref": "http://labs.inspirehep.net/api/experiments/1110623"},
+        }
+    ]
+    headers = {"Accept": "application/vnd+inspire.record.ui+json"}
+    record = create_record("job")
+    record["accelerator_experiments"] = [
```
harunurhan, line 267: I don't get how does this even work, but can you put this data while creating the record `create_record("job", data=...)`

ui/src/common/layouts/CollectionsMenu/CollectionsMenu.jsx
MJedr: I asked Stella and they need similar customisation as the collection link (eg. the highlight when they are active or new tag) so that's why I used collection link. If I change to link I'll have to duplicate this code.

backend/tests/integration/records/serializers/json/test_jobs.py
MJedr: Sure, but we have a lot of tests where we update the record created by ```create_record()``` .

ui/src/common/layouts/CollectionsMenu/CollectionsMenu.scss
```diff
@@ -6,3 +6,37 @@
     display: none;
   }
 }
+
+.ant-dropdown-menu {
```
MJedr, line 10: I know that it's ugly but there was no way to override antd styling (if we don't want to use submenu).

ui/src/common/layouts/CollectionsMenu/CollectionsMenu.jsx
```diff
@@ -68,12 +72,30 @@ function CollectionsMenu({ currentPathname }) {
         </CollectionLink>
       </Col>
       <Col>
-        <CollectionLink
-          active={activeCollection === INSTITUTIONS_PID_TYPE}
-          to={INSTITUTIONS}
+        <DropdownMenu
+          className="mh4 m-mh2"
+          title={
+            <Link
```
harunurhan, line 78: you can have a `<Button>` without border or something instead of this.

ui/src/common/layouts/CollectionsMenu/CollectionsMenu.scss
harunurhan: that's not only ugly but also affect all `ant-dropdown-menu`, since whatever you leave out of `.__CollectionsMenu__` is a global style.

Whatever is rendered under `CollectionsMenu`, should have their style in `__CollectionsMenu__` class, such as the `link` class.

Since dropdown is rendered outside of the `CollectionsMenu` DOM-wise. Only way to scope the style is to use `overlayClassName` of `Dropdown` that is rendered by `DropdownMenu`


```jsx
<DropdownMenu overlayClassName="more-collections-menu"...
```

```scss
.more-collections-menu {
  // styles, classes
}
```

ui/src/common/layouts/CollectionsMenu/CollectionsMenu.scss
```diff
@@ -5,4 +5,55 @@
   @media (max-width: $screen-sm-max) {
     display: none;
   }
+  .button-no-background {
+    top: -15%;
+    padding-left: 0;
+    background-color: $layout-header-background;
+    border-color: $layout-header-background;
```
harunurhan, line 12: you can just put `0` border instead.

ui/src/common/layouts/CollectionsMenu/CollectionsMenu.scss
```diff
@@ -5,4 +5,55 @@
...
+}
+
+.more-collections-menu {
+  .ant-dropdown-menu {
+    background-color: $layout-header-background !important;
+    min-width: 160px;
+  }
+
+  .dropdown-menu-item {
+    padding-top: 5px;
```
harunurhan, line 43: no `px` in general you should use `rem` or `em` if it makes sense.

Issue #1130, opened 2020-07-21
Create a command to delete indexes with a given prefix


drjova:
Create a command which deletes indexes with a given prefix. For example:

```bash
$ inspirehep index delete inspire-v1.1.1
```

It should relete all indexes which are starting with `inspire-v1.1.1-*` including:
* `inspire-v1.1.1-records-hep-*`
* `inspire-v1.1.1-records-authors-*`
* `inspire-v1.1.1-records-experiments-*`
* ...

TECH:

You can get all indexes and their aliases by calling `current_search.client.indices.get_alias()`
Then you can filter out all indexes starting with specified prefix.
Before deleting index, delete it's aliases: https://elasticsearch-py.readthedocs.io/en/master/api.html#elasticsearch.client.IndicesClient.delete_alias
You can easily delete all indexes starting with a prefix by calling:
`indices.delete(index=prefix+"*", allow_no_indices=True, expand_wildcards="all")`

PR #1186, opened 2020-08-10
indexer cli: delete indexes with given prefix

MJedr:
Ref: #1130

### DIFF
--- a/backend/inspirehep/indexer/cli.py
+++ b/backend/inspirehep/indexer/cli.py
@@ -334,3 +334,30 @@ def create_aliases(ctx, yes_i_know, prefix_alias):
             current_search.client.indices.delete_alias("*", alias_name)
         click.echo(f"Creating alias '{alias_name}' -> '{index_name}'")
         current_search.client.indices.put_alias(index=index_name, name=alias_name)
+
+
+@index.command(
+    "delete-indexes", help="Removes indexes with specified prefix",
+)
+@click.option("--yes-i-know", is_flag=True)
+@click.option("--prefix", default="", type=str)
+@with_appcontext
+@click.pass_context
+def delete_aliases(ctx, yes_i_know, prefix):
+    indices = current_search.client.indices.get_alias()
+    prefix_regex = re.compile(f"""{prefix}.*""")
+    indices_to_delete = list(filter(prefix_regex.match, indices))
+    indices_to_delete = {k: indices[k] for k in indices_to_delete}
+
+    if not indices_to_delete:
+        click.echo("No indices matching given prefix found.")
+        return 1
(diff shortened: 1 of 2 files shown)


backend/inspirehep/indexer/cli.py
```diff
@@ -334,3 +334,32 @@ def create_aliases(ctx, yes_i_know, prefix_alias):
...
+        return 1
+    click.echo(f"""Found {len(indices_to_delete)} indices to delete""")
+    for index in indices_to_delete:
+        if not yes_i_know and not click.confirm(
+            f"This operation will remove '{index}' index. Are you sure you want to continue?"
+        ):
+            continue
+        aliases_to_delete = indices[index]["aliases"].keys()
+        for alias in aliases_to_delete:
+            current_search.client.indices.delete_alias(index=index, name=alias)
```
pazembrz, line 363: Instead getting all aliases for index and then providing them here you can just use ```"*"``` as alias name and this will remove all aliases for provided index.

pazembrz: Otherwise you can just provide all aliases as a string containing aliases separated by comma https://www.elastic.co/guide/en/elasticsearch/reference/current/indices-delete-alias.html
As this loop will call ES separately for each alias.

backend/inspirehep/indexer/cli.py
```diff
@@ -334,3 +334,32 @@ def create_aliases(ctx, yes_i_know, prefix_alias):
...
+
+
+@index.command(
+    "delete-indexes", help="Removes indexes with specified prefix",
+)
+@click.option("--yes-i-know", is_flag=True)
+@click.option("--prefix", default="", type=str)
+@with_appcontext
+@click.pass_context
+def create_aliases(ctx, yes_i_know, prefix):
```
pazembrz, line 346: This name should be different I believe :)

backend/inspirehep/indexer/cli.py
```diff
@@ -334,3 +334,30 @@ def create_aliases(ctx, yes_i_know, prefix_alias):
...
+)
+@click.option("--yes-i-know", is_flag=True)
+@click.option("--prefix", default="", type=str)
+@with_appcontext
+@click.pass_context
+def delete_aliases(ctx, yes_i_know, prefix):
+    indices = current_search.client.indices.get_alias()
+    prefix_regex = re.compile(f"""{prefix}.*""")
+    indices_to_delete = list(filter(prefix_regex.match, indices))
+    indices_to_delete = dict((k, indices[k]) for k in indices_to_delete)
```
pazembrz, line 350: `indices_to_delete = {k:indices[k] for k in indices_to_delete}` 
so you won't have to build it twice (first as tuple and then dict from this tuple)

PR #1199, opened 2020-08-13
pubnote search: journal_name together with journal volume

MJedr:
* #Ref:1056
* search for pubnotes where journal_name is with journal volume returns now up to 5 results

### DIFF
--- a/backend/inspirehep/matcher/api.py
+++ b/backend/inspirehep/matcher/api.py
@@ -58,7 +58,7 @@ def match_reference_with_config(reference, config, previous_matched_recid=None):
     same_as_previous = any(
         matched_recid == previous_matched_recid for matched_recid in matched_recids
     )
-    if 0 < len(matched_recids) <= 5:
+    if len(matched_recids) == 1:
         _add_match_to_reference(reference, matched_recids[0], config["index"])
     elif same_as_previous:
         _add_match_to_reference(reference, previous_matched_recid, config["index"])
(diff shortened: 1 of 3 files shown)


backend/inspirehep/matcher/api.py
```diff
@@ -74,19 +74,14 @@ def match_reference_with_config(reference, config, previous_matched_recid=None):
     return reference
 
 
-def match_reference(reference, previous_matched_recid=None):
-    """Match a reference using inspire-matcher.
+def get_reference_matcher_configs(reference):
```
drjova, line 77:
```suggestion
def match_reference_config(reference):
```

backend/inspirehep/matcher/api.py
```diff
@@ -129,16 +142,40 @@ def match_reference(reference, previous_matched_recid=None):
...
+    configs = get_reference_matcher_configs(reference)
+
+    matches = []
+    for config in configs:
+        matched_recids = [
+            matched_record["_source"]["control_number"]
+            for matched_record in match(reference, config)
+        ]
+        matches.extend(matched_recids)
+    matches = dedupe_list(matches)[0:5]
```
drjova, line 168: we could use a `set` instead of iterating again to remove dups and just get the first 5

MJedr: Yes, I was thinking about it but I thought that we want to keep the order of matches

PR #1202, opened 2020-08-17
collections menu: change more dropdown spacing

MJedr:
Ref: #1115

### DIFF
--- a/ui/src/common/layouts/CollectionsMenu/CollectionsMenu.scss
+++ b/ui/src/common/layouts/CollectionsMenu/CollectionsMenu.scss
@@ -40,20 +40,28 @@
   }
 
   .dropdown-menu-item {
-    padding-top: 0.3em;
-    paddig-bottom: 0.3em;
+    line-height: 2.5rem;
+    margin-top: 0.25rem;
+    padding: 0 1rem;
     &:hover {
       background-color: $layout-header-background !important;
     }
   }
 
+  .ant-dropdown-menu-item > a {
+    display: unset;
+    padding: 0;
+  }
+
   .dropdown-link {
     font-size: 1rem;
+    margin: 0;
     text-align: left;
-    color: white !important;
+    color: white;
     &:hover {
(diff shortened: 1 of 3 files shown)


ui/src/common/layouts/CollectionsMenu/CollectionsMenu.scss
```diff
@@ -35,25 +35,35 @@
 
 .more-collections-menu {
   .ant-dropdown-menu {
+    position: absolute;
```
harunurhan, line 38: why is this needed?, position seems already OK.

ui/src/common/layouts/CollectionsMenu/CollectionsMenu.scss
```diff
@@ -35,25 +35,35 @@
...
     background-color: $layout-header-background !important;
     min-width: 160px;
   }
 
   .dropdown-menu-item {
-    padding-top: 0.3em;
-    paddig-bottom: 0.3em;
+    height: 2.5rem;
```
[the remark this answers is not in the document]
MJedr: Yes, because Stella asked for the same spacing as in other dropdown's (and without it it is quite small)

ui/src/common/layouts/CollectionsMenu/CollectionsMenu.scss
```diff
@@ -40,20 +40,28 @@
...
+    line-height: 2.5rem;
+    margin-top: 0.25rem;
+    padding: 0 1rem;
     &:hover {
       background-color: $layout-header-background !important;
     }
   }
 
+  .ant-dropdown-menu-item > a {
+    display: unset;
```
harunurhan, line 52: why is that necessary?

MJedr: Because without ```display: unset``` it has ```display: block``` by default and the underline is much lower than the link

harunurhan (approved):

Issue #1194, opened 2020-08-12
add demo records to hep


drjova:
There is no easy way to prod records locally.

- [ ] Import record from url 

Change [importer](https://github.com/inspirehep/inspirehep/blob/master/backend/inspirehep/records/cli.py#L88) to use API tokens to get the full record from the records admin API. Then each developer could have it's own personal token, export it in her/his local env and retrieve records from `QA` or `PROD`.

Note: we could also rename it from `importer` to `loader` or something better.
- [ ] move the records from next demosite

Ideally we should store them as `JSON` or move https://github.com/inspirehep/inspire-next/tree/master/inspirehep/demosite/data  We could use migrator for that without many changes, for example:

```shell
$ inspirehep migrate file --force --wait /data/demo-records.xml.gz
```

TECH:
* Add possibility to provide token as another parameter for the command, also read it from env variable
* add ``` "Authorization": "Bearer {token}".format(
            token=current_app.config['AUTHENTICATION_TOKEN']
        ),``` to HEADERS in request sent https://github.com/inspirehep/inspirehep/blob/41e91c73c58a8d7e49f93410c0a9706b4e6ff417/backend/inspirehep/records/cli.py#L50

* Copy https://github.com/inspirehep/inspire-next/tree/master/inspirehep/demosite/data  to inspirehep/backend/data
* We can create separate CLI command which will just run (To discuss as running migrator is good enough) 
* Make sure the records from url could imported without validation errors
```shell
$ inspirehep migrate file --force --wait /data/demo-records.xml.gz
```

PR #1211, opened 2020-08-18
demo records

harunurhan:
Ref: #1194


backend/inspirehep/config.py
```diff
@@ -34,7 +34,7 @@
 
 # Web services and APIs
 # =====================
-AUTHENTICATION_TOKEN = "CHANGE_ME"
+AUTHENTICATION_TOKEN = "Yh2LqaWL8Ge1HBgat9H8HR3a6HXNSYtjDpLKDqno3KU5WC74j8L3al6f71up"
```
pazembrz, line 37: What token is this? Should it be here?

pazembrz (approved):
Only one comment about AUTHENTICATION_TOKEN

backend/inspirehep/config.py
harunurhan: it's a leftover, but token is invalid anyhow. I'll change it.

Issue #1191, opened 2020-08-12
rt entpoints to hep


drjova:
Move `rt` [endpoints](https://github.com/inspirehep/inspire-next/blob/master/inspirehep/modules/editor/api.py) and change the ingress to point to hep.

**Tech notes** : 
* Move [these](https://github.com/inspirehep/inspire-next/blob/92c9f08aba093421f343e39ba936f8df0c6168eb/inspirehep/modules/editor/api.py#L111-L165) endpoints to hep to ```editor/views``` as they are
* Change the permission from ```@editor_permission``` to ```@login_required_with_roles([Roles.cataloger.value])```
* All the methods needed from the rt module are already in hep so you can just use them. (Note: these methods are not tested)
* Tests for this on next are simply mocking RT and checking the correct methods are called, we could do the same
* Remove the endpoints from next
* Ask Ben to change the ingress to hep

drjova:
@MJedr did we change the ingress to point to hep?

MJedr:
No, the PR is still waiting to be merged https://github.com/inspirehep/inspire-next/pull/3959

PR #1213, opened 2020-08-18
editor views: move endpoints for rt tickets from next

MJedr:
Ref: #1191

### DIFF
--- a/backend/inspirehep/editor/views.py
+++ b/backend/inspirehep/editor/views.py
@@ -6,6 +6,7 @@
 # the terms of the MIT License; see LICENSE file for more details.
 
 from flask import Blueprint, jsonify, request
+from flask_login import current_user
 from invenio_db import db
 from invenio_records.models import RecordMetadata
 from sqlalchemy_continuum import transaction_class, version_class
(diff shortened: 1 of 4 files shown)


backend/inspirehep/editor/views.py
```diff
@@ -86,3 +91,76 @@ def get_revision(transaction_id, rec_uuid):
...
+@blueprint.route("/rt/users", methods=["GET"])
+@editor_use_api_permission.require(http_exception=403)
+def get_rt_users():
+    """View to get all rt users"""
+
+    return jsonify(tickets.get_users())
+
+
+@blueprint.route("/rt/queues", methods=["GET"])
+@editor_use_api_permission.require(http_exception=403)
```
harunurhan, line 151: shouldn't everything just be `@login_required_with_roles([Roles.cataloger.value])` ?

MJedr: In next there were two different kinds of required permissions:
```@editor_permission``` and ```@editor_use_api_permission```. The first one corresponds to  ```@login_required_with_roles``` whereas the second is set up here 
https://github.com/inspirehep/inspirehep/pull/1213/files#diff-0d4f912028d824fde064f50b4008099dR51-R55
To me it looks that it can be replaced with ```@login_required_with_roles``` but I wasn't sure. If that's true, can you confirm please?

harunurhan: I don't know why it is like that, but to me it is OK to just require the cataloger role, AFAIK this endpoints are used only for editor, which requires cataloger role anyway.

harunurhan (approved):

PR #1235, opened 2020-08-21
run ui-tests on cypress without API recordings

harunurhan:


cypress-tests-data.sh
```diff
@@ -0,0 +1,20 @@
...
+
```
drjova, line 6: could we delete the existing ones and just keep the new ones? Then we could run just ```inspirehep importer records -d data/records/literature``` , ```inspirehep importer records -d data/records/authors``` etc

.circleci/config.yml
```diff
@@ -8,8 +8,22 @@ jobs:
       - checkout
       - run: sudo chown -R 999 e2e
       - run: sh run-e2e.sh
+  cypress:
+    machine:
+      image: ubuntu-1604:202004-01
```
pazembrz, line 13: Do they have more recent image? Ubuntu 16.04 will be supported until April 2021 so it's less than a year.

README.md
```diff
@@ -209,55 +209,39 @@ Run `isort` and `flake8` checks.
...
-```bash
-docker-compose run --rm node bash # in /ui/ui-tests folder
-cd /opt/app/ui-tests
-yarn test --testPathPattern=... # to run selected tests
-```
+Visual tests are run only on `headless` mode. So `yarn test:dev` which uses the headed browser will ignore them.
+Running existing visual tests and updating/creating snapshots requires `run-cypress-tests.sh` script.
 
-NOTE: code changes in `ui/src` doesn't reflect without building the whole ui (`yarn build`)
+For continous runs (when local DB is running and has required records etc.), the script can be reduced to only the last part `sh cypress-tests-run.sh`
```
drjova, line 237:
```suggestion
For continuous runs (when local DB is running and has required records etc.), the script can be reduced to only the last part `sh cypress-tests-run.sh`
```

backend/inspirehep/records/cli.py
```diff
@@ -114,9 +125,24 @@ def importer():
     type=click.File("rb"),
     help="Path to a JSON file, example: ``data/records/literature/999108.json``.",
 )
+@click.option(
```
pazembrz, line 128: There should be 2 empty lines before this.

cypress-tests-data.sh
```diff
@@ -0,0 +1,20 @@
...
+docker-compose $files exec web inspirehep importer records -f data/records/authors/1004662.json -f data/records/authors/1060898.json -f data/records/authors/1013725.json -f data/records/authors/1078577.json -f data/records/authors/1064002.json -f data/records/authors/1306569.json --save
+
+#job search
+docker-compose $files exec web inspirehep importer records -f data/records/jobs/1811684.json -f data/records/jobs/1812904.json -f data/records/jobs/1813119.json -f data/records/jobs/1811836.json -f data/records/jobs/1812529.json
+
+#seminar search
+docker-compose $files exec web inspirehep importer records -f data/records/seminars/1811573.json -f data/records/seminars/1811750.json -f data/records/seminars/1811657.json -f data/records/seminars/1807692.json -f data/records/seminars/1807690.json
+
+#conference search
+docker-compose $files exec web inspirehep importer records -f data/records/conferences/1769332.json -f data/records/conferences/1794610.json -f data/records/conferences/1809034.json -f data/records/conferences/1776122.json -f data/records/conferences/1622944.json
```
pazembrz, line 20: Please notice that every execution of `inspirehep` takes additional ~12s to load all libraries. So here We have additional ~60s (~70s instead of ~12s from only one call). Isn't it better to load all files in one `inspirehep importer records` call?

cypress-tests-run.sh
```diff
@@ -0,0 +1,7 @@
+rc=0
+files="-f docker-compose.yml -f ./smoke-tests/docker-compose.cypress.yml"
+
+docker wait inspirehep_ui-build_1
```
pazembrz, line 4: I'm happy that it works as intended :+1:

cypress-tests-data.sh
harunurhan: well I think it's better to know, which ids explicitly we need for the screenshots. That's why I didn't bother a lot.

drjova:
@harunurhan well done 🥇

.circleci/config.yml
harunurhan: By then, we might move this to github actions, but I am not sure.

cypress-tests-data.sh
harunurhan: yeah I have been thinking that as well, I could convert to one

backend/inspirehep/records/cli.py
harunurhan: why?

pazembrz: PEP8. https://www.python.org/dev/peps/pep-0008/#blank-lines 2 blank lines should be between top level functions. Or do I interpret code wrong here?

pazembrz: Ah sorry... My mistake. There are more click options. They were hidden here ;)

.circleci/config.yml
drjova: Indeed the plan is to move to github actions

Issue #1173, opened 2020-08-05
export CV format txt/html


StellaCh:
Users would like to get a list of their papers to copy paste them in word documents.

Example in legacy: http://old.inspirehep.net/search?ln=en&ln=en&p=a+witten&of=hcv&action_search=Search&sf=earliestdate&so=d&rm=&rg=25&sc=0

 

Add another format in the dropdown menu under LaTeX(US) called 'CV format (html)' in 'Cite' and 'Cite all' that exports a paper in the following format:

For https://inspirehep.net/literature/1781327

Geodesics of the hyperbolically symmetric black hole
L. Herrera, A. Di Prisco, J. Ospino, L. Witten
arXiv:2002.07586 [gr-qc]
10.1103/PhysRevD.101.064071
Phys.Rev. D101 (2020) no.6, 064071

where title, authors, arxiv, doi are links.

The fields should be:
Title <linked>
Authors as they appear in the search results with et al. <linked>
arxiv id <linked> [arxiv category]
doi <linked>
pub note

(jira issue: [INSPIR-3635](https://its.cern.ch/jira/browse/INSPIR-3635))

StellaCh:
In legacy, we have different options for txt and html formats, let's brainstorm on ways we could have them both in 1 option.

harunurhan:
- We need to add new formats `text/html`, `text/plain`, for literature detail and list endpoints, similar to https://github.com/inspirehep/inspirehep/blob/276bd2e2eaca0f576c0cf7398daa67236c2f6c4b/backend/inspirehep/records/config.py#L103-L116
 For search endpoints, `_source` should be specified for the new formats so that only required fields retrieved from `elasticsearch`.
  - (Not sure if we want to generate also these formats on index time)
- We can use `jinja` templates serialize records to these formats.
- New formats should be added [here](https://github.com/inspirehep/inspirehep/blob/master/ui/src/literature/constants.js), so that they are visible on the UI.
- [This](https://github.com/inspirehep/inspirehep/blob/276bd2e2eaca0f576c0cf7398daa67236c2f6c4b/ui/src/literature/citeArticle.js#L6) shouldn't prefix `application` anymore but receive the full `mime-type` 
[...]

PR #1242, opened 2020-08-25
cv format: Add support for citing cv html format

monaawi:
* Ref: #1173

### DIFF
--- a/backend/inspirehep/records/config.py
+++ b/backend/inspirehep/records/config.py
@@ -107,11 +107,13 @@ LITERATURE.update(
             "application/x-bibtex": f"{INSPIRE_SERIALIZERS}:literature_bibtex_response",
             "application/vnd+inspire.latex.eu+x-latex": f"{INSPIRE_SERIALIZERS}:latex_response_eu",
             "application/vnd+inspire.latex.us+x-latex": f"{INSPIRE_SERIALIZERS}:latex_response_us",
+            "text/vnd+inspire.html+html": f"{INSPIRE_SERIALIZERS}:literature_cv_html_response",
         },
         "search_serializers": {
             "application/json": f"{INSPIRE_SERIALIZERS}:literature_json_response_search",
             "application/vnd+inspire.record.ui+json": f"{INSPIRE_SERIALIZERS}:literature_json_list_response",
             "application/x-bibtex": f"{INSPIRE_SERIALIZERS}:literature_bibtex_response_search",
+            "text/vnd+inspire.html+html": f"{INSPIRE_SERIALIZERS}:literature_cv_html_response_search",
             "application/vnd+inspire.latex.eu+x-latex": f"{INSPIRE_SERIALIZERS}:literature_latex_eu_response_search",
             "application/vnd+inspire.latex.us+x-latex": f"{INSPIRE_SERIALIZERS}:literature_latex_us_response_search",
         },
@@ -632,6 +634,15 @@ LITERATURE_SOURCE_INCLUDES_BY_CONTENT_TYPE = {
     "application/vnd+inspire.latex.us+x-latex": ["_latex_us_display"],
     "application/vnd+inspire.latex.eu+x-latex": ["_latex_eu_display"],
     "application/x-bibtex": ["_bibtex_display"],
+    "text/vnd+inspire.html+html": [
+        "titles",
+        "control_number",
+        "collaborations",
+        "authors",
+        "dois",
+        "arxiv_eprints",
+        "publication_info",
+    ],
 }
 LITERATURE_SOURCE_EXCLUDES_BY_CONTENT_TYPE = {
     "application/json": [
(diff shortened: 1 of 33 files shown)


backend/inspirehep/records/config.py
```diff
@@ -107,11 +107,13 @@
             "application/x-bibtex": f"{INSPIRE_SERIALIZERS}:literature_bibtex_response",
             "application/vnd+inspire.latex.eu+x-latex": f"{INSPIRE_SERIALIZERS}:latex_response_eu",
             "application/vnd+inspire.latex.us+x-latex": f"{INSPIRE_SERIALIZERS}:latex_response_us",
+            "text/html": f"{INSPIRE_SERIALIZERS}:literature_cv_html_response",
```
michamos, line 110: won't this be served by default if you open the API in the browser? maybe we should have a custom one.

ui/src/setupProxy.js
```diff
@@ -0,0 +1,14 @@
+const proxy = require('http-proxy-middleware');
+
+// eslint-disable-next-line func-names
```
harunurhan, line 3: leftover

ui/src/setupProxy.js
```diff
@@ -0,0 +1,14 @@
...
+
+const localProxy = proxy({
+  target: 'http://localhost:8000',
+  secure: false,
+  changeOrigin: true,
+});
+
+// eslint-disable-next-line func-names
```
harunurhan, line 11: you can just do `(app) => {...`  here, instead.

backend/inspirehep/records/marshmallow/literature/cv.py
```diff
@@ -0,0 +1,54 @@
...
+    CollaborationWithSuffixSchemaV1,
+)
+from inspirehep.records.marshmallow.literature.common.doi import DOISchemaV1
+from inspirehep.records.marshmallow.literature.common.publication_info_item import (
+    PublicationInfoItemSchemaV1,
+)
+
+
+class CVSchema(Schema):
+    title_display = fields.Method("get_title")
```
harunurhan, line 28: [optional] this can be called just `title`

backend/inspirehep/records/marshmallow/literature/cv.py
```diff
@@ -0,0 +1,54 @@
...
+    @staticmethod
+    def get_title(data):
+        titles = data.get("titles")
+
+        if not titles:
+            return missing
+
+        title = titles[0]["title"]
+        if "subtitle" in titles[0]:
+            title += f" : {titles[0]['subtitle']}"
```
harunurhan, line 53: what do you think of sth like this, instead of string concat? I know in this case it's still readable, but I feel like below is a bit more clear.

```py
first_title = titles[0]

if "subtitle" in first_title:
  return f"{first_title['title']} : {first_title['subtitle']}"

return first_title['title']
```

backend/tests/integration/records/serializers/data/cv_with_author_with_affiliations.html
```diff
@@ -0,0 +1,12 @@
+<!DOCTYPE html>
+<html>
+<body>
+  <p><b>
+    <a href="https://localhost:5000/literature/637275237">
+      This is a title.
+    </a>
+  </b></p>
+      <p>John6 Doe (<a href="https://localhost:5000/institutions/637275238">Gent U.</a>)</p>
+          <br>
```
harunurhan, line 10: in general `<p>` should have enough spacing why do we need `br` ?

ui/src/literature/components/CiteModalAction.jsx
```diff
@@ -31,6 +34,7 @@ class CiteModalAction extends Component {
     this.state = {
       modalVisible: false,
       errorMessage: null,
+      modalLoading: false,
```
harunurhan, line 37: `loadingCiteContent` or even `loading`

ui/src/literature/components/CiteModalAction.jsx
```diff
@@ -70,15 +74,17 @@ class CiteModalAction extends Component {
     let citeContent = this.citeContentCacheByFormat[format];
```
harunurhan, line 74: I think we can remove this `cache` and rely on the browser http caching only, could you at least drop a `TODO:` here.

ui/src/setupProxy.js
```diff
@@ -0,0 +1,11 @@
+const proxy = require('http-proxy-middleware');
+
+const localProxy = proxy({
+  target: 'http://localhost:8000',
```
harunurhan, line 4: now we can remove the `proxy` from `package.json` and change the docs (if any) that is referring to `proxy` prop in `package.json` to refer to this one.

backend/tests/integration/records/serializers/data/cv_with_author_with_affiliations.html
monaawi: I wanted to have extra space between literature records. Should I remove it?

harunurhan: oh right, in that case I do not know. it's up to @StellaCh

harunurhan (approved):

drjova:
@monaawi can we merge it now?

Issue #1253, opened 2020-08-27
Add FEATURE_FLAG for claim UI


harunurhan:
We shouldn't display it on prod yet.

harunurhan:
- need to set `assignView: true`, only if `ASSIGN_UI_FEATURE_FLAG` is enabled and isCataloger`  or `isSuperUser` (https://github.com/inspirehep/inspirehep/blob/22f728bead6b016a5b9326e3a5d2e45f91e0e845/ui/src/authors/containers/AuthorPublicationsContainer.jsx#L50)
- Use `config.js/getConfigFor(flag)` to read the state.
- flag should enabled locally and on QA but disabled on PROD.

PR #1262, opened 2020-08-31
Add feature flag for claims UI

miguelgrc:
* Ref: #1253


harunurhan (approved):
Do not forget to set it `true` also for QA

Issue #1273, opened 2020-09-02
Replace latexcodec with pylatexenc


michamos:
`pylatexenc` works much better than `latexcodec` (which has bugs in many corner cases and nows about fewer characters), so we should use that instead to encode/decode LaTeX-encoded text.

PR #1274, opened 2020-09-02
bibtex: improve LaTeX encoding

michamos:
* Ref #1273

### DIFF
--- a/backend/inspirehep/records/serializers/bibtex.py
+++ b/backend/inspirehep/records/serializers/bibtex.py
@@ -4,14 +4,11 @@
 #
 # inspirehep is free software; you can redistribute it and/or modify it under
 # the terms of the MIT License; see LICENSE file for more details.
-import codecs
-import re
-
-import latexcodec  # noqa: F401
 import structlog
 from invenio_records_rest.serializers.response import search_responsify
 from pybtex.database import BibliographyData, Entry, Person
 from pybtex.database.output.bibtex import Writer
+from pylatexenc.latexencode import UnicodeToLatexEncoder
 
 from ..marshmallow.literature.bibtex import BibTexCommonSchema
 from .response import record_responsify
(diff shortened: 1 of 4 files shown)


backend/inspirehep/records/serializers/bibtex.py
```diff
@@ -21,19 +21,12 @@
 
 class BibtexWriter(Writer):
     math_expression_regex = re.compile(r"(?<!\\)\$.*?\$")
```
MJedr, line 23: Is that still needed?

MJedr (approved):

tsgit:
I think going for pylatexenc is a good idea.
I note however that `\ensuremath` will break plain `TeX`
I would suggest to prepare a FAQ or help page and possibly an example macro to declare `\ensuremath{}` to just emit its argument unchanged or enclosed in `$...$` when plain `TeX` is used.
This could be declared once at the beginning of the user's `.bib` file

tsgit:
thinking about e.g. Asian names in native spelling (Hanzi, Kanji, etc.)

would it make sense to specify `unknown_char_policy` to one of

`keep` ?
this would leave the output human readable, however it would fail bibtex or be dropped by bibtex

`unihex` ? 
would preserve the content/meaning but not in a human readable form

`replace` ?
might give the best chance for authors to notice, but then what should they do?

or `fail` or `ignore` ?

tsgit:
the `ensuremath{}` part will certainly make titles less readable. simple example

```python
>>> name = u'\N{GREEK CAPITAL LETTER XI}\N{GREEK SMALL LETTER ALPHA}\N{GREEK SMALL LETTER BETA}'
>>> print(name)
Ξαβ
>>> unicode_to_latex(name, replacement_latex_protection='braces-after-macro')
'\\ensuremath{\\Xi}\\ensuremath{\\alpha}\\ensuremath{\\beta}'

Maybe that can be refined to become `\\ensuremath{\\Xi\\alpha\\beta}`  ?

Maybe wait and see what actual user feedback we get.
```

michamos:
Thanks for the review @tsgit 

> would it make sense to specify `unknown_char_policy` to one of
> 
> `keep`

This is the default, so that's what we're using. It keeps things human readable as you said, and people will notice as unknown characters will cause compilation errors.

> I note however that `\ensuremath` will break plain `TeX`
> I would suggest to prepare a FAQ or help page and possibly an example macro to declare `\ensuremath{}` to just emit its argument unchanged or enclosed in `$...$` when plain `TeX` is used.

I think the set of people who use plain TeX together with BibTeX and don't know how to define a macro is vanishingly small. I would wait until someone asks about it before adding it to the FAQ.

Issue #1217, opened 2020-08-19
Advisor name order of names


StellaCh:
In the author profile, the Advisor name is currently Last Name, First Name. It should be displayed as: First Name Last Name as it's written in the author name. This is because we received some complains from users that if there're more than 1 advisor names, it becomes confusing, eg: https://inspirehep.net/authors/1007931?ui-citation-summary=true

**TECH NOTES:**
In https://github.com/inspirehep/inspirehep/blob/beb564068dfc9e0457cf5d662fd1360bac60a441/backend/inspirehep/records/marshmallow/authors/ui.py#L22 advisor name's should go through `get_first_name` and `get_last_name` from 
https://github.com/inspirehep/inspirehep/blob/beb564068dfc9e0457cf5d662fd1360bac60a441/backend/inspirehep/records/marshmallow/utils.py#L52

In https://github.com/inspirehep/inspirehep/blob/beb564068dfc9e0457cf5d662fd1360bac60a441/ui/src/authors/components/Advisors/Advisor.jsx#L8 use `first_name` whitespace `last_name`.

PR #1298, opened 2020-09-08
authors: add first_name and last_name to advisors

monaawi:
Ref: #1217


drjova (approved):
LGTM, do we need to reindex `authors`? If yes, please add it to the commit message 😃

monaawi:
There is no need for reindexing. The change is at detail page serializer level.

Issue #1271, opened 2020-09-02
Limit authors to 10 in latex formats


michamos:
Reported by Ya'ou. Currently, we add all authors to the LaTeX formats. If a paper has more than 10 authors, e.g. https://inspirehep.net/literature/1710056, we should display `first_author, \emph{et al.}` instead of all the authors.

harunurhan:
Change https://github.com/inspirehep/inspirehep/blob/beb564068dfc9e0457cf5d662fd1360bac60a441/backend/inspirehep/records/marshmallow/literature/latex.py#L30 to limit the number of authors.

PR #1300, opened 2020-09-09
latex schema: limit rendered authors to 10

MJedr:
Ref: #1271

### DIFF
--- a/backend/inspirehep/records/serializers/templates/latex_template.tex
+++ b/backend/inspirehep/records/serializers/templates/latex_template.tex
@@ -39,7 +39,8 @@
     \VAR{- format_list(data.collaborations) }
     \VAR{- '],' }
 {%- elif 'authors' in data -%}
-    \VAR{- format_list(data.authors) ~ ',' }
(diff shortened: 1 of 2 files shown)


backend/inspirehep/records/serializers/templates/latex_template.tex
```diff
@@ -9,14 +9,15 @@
 \VAR{- '\n' if 'authors' in data or 'collaborations' in data }
 {%- if 'collaborations' in data -%}
     {%- if 'authors' in data -%}
-        \VAR{ data.authors.0 }
-        \VAR{- ' \\textit{et al.}' if data.authors|length > 1 }
+        \VAR{ data.authors[:10] ~ ',' }
+        \VAR{- ' \\textit{et al.}' if data.authors|length > 10 }
```
michamos, line 13: this can stay like it was, it's the non-collaboration case that should be fixed.

MJedr: Ok I'll change it, in the refinement it was written to fix both

michamos (approved):

PR #1319, opened 2020-09-16
Update generation of new BAI to avoid concurrency problems

pazembrz:


backend/inspirehep/pidstore/providers/bai.py
```diff
@@ -80,21 +84,38 @@ def create(
...
+                    last_exception = e
+            elif pid_from_db.object_uuid != object_uuid:
+                last_exception = PIDAlreadyExistsError(
+                    pid_value=pid_value, pid_type="bai"
+                )
+            time.sleep(current_app.config.get("PIDSTORE_TEXKEY_RETRY_DELAY", 5))
+        if last_exception:
+            raise last_exception
+
+        return provider_object
```
drjova, line 118: could we use `backoff` for example, instead of manually handle that?

pazembrz: WDYM? Minters do not have separate task.

pazembrz: Do you mean installing another package for retrying functions which will be used only in here?

drjova: don't we have [backoff](https://pypi.org/project/backoff/) already?

pazembrz: no https://github.com/inspirehep/inspirehep/blob/master/backend/poetry.lock

pazembrz: I was thinking about it but new package to be used only in one place... It's too much I believe.

drjova: :+1: weird, I thought that we're using it. At least we should add tests for this additional functionality.

backend/tests/unit/pidstore/test_providers/test_bai_provider.py
```diff
@@ -0,0 +1,85 @@
...
+import mock
+import pytest
+from invenio_pidstore.models import PersistentIdentifier, PIDStatus
+from sqlalchemy.exc import IntegrityError
+
+from inspirehep.pidstore.errors import PIDAlreadyExistsError
+from inspirehep.pidstore.providers.bai import InspireBAIProvider
+
+
+def test_bai_create_fails_after_retrying_when_bai_changes():
```
drjova, line 16: instead of unit tests we could create an integration test with existing bai and mock the provider with the same value, on the first time, and see if it raises the proper errors etc.

pazembrz: Why? What's the difference? This will test exactly same code in the same way, but additionally it will add another ~5s to integration tests (setup before and after test, cleaning ES, adding entry to db). I don't see the point of doing this.

drjova: we're mocking so many things from the app to providers, if someone for example changes the provider significantly these tests will continue to pass. 

With an integration we'll test end-to-end the functionality and in some cases (i.e. the test with `PIDAlreadyExistsError`) we will not even need to mock anything. This is a core feature and I believe we should test

pazembrz: But we already have a lot of tests for BAI minter/providers https://github.com/inspirehep/inspirehep/blob/e3d4a17b6133d967fb4d0660243e42cfa5958a94/backend/tests/integration/pidstore/minters/test_minters_bai.py#L99-L265
here I'm testing only the retry.

drjova: it is not end-to-end tested, you could integrate some of these assertions already in the other tests and just add one that checks the retry functionality.

drjova (approved):

Issue #1117, opened 2020-07-02
Create CI for author-disambiguation


MJedr:
Currently, we don't have a CI set for the repository, so there is no way to verify if the changes are correct. We need to set it up (also linters must be configured).

harunurhan:
We can use github actions, which would be very similar to [unit test setup on inspirehep ](https://github.com/inspirehep/inspirehep/blob/master/.github/workflows/backend-unit.yml), that runs https://github.com/inspirehep/inspire-disambiguation/blob/master/run-tests.sh

harunurhan:
Actually let's take this opportunity to merge `inspire-disambugation` repo into this one, under `disambiguation` folder.

PR #1322, opened 2020-09-16
Move inspire_disambiguation into hep

miguelgrc:
* Ref: #1117


michamos:
You seem to have tried to preserve the history, which is good, but then it looks like the history was completely nuked at the end anyway. Can you try to preserve it?

michamos (approved):
After checking with @miguelgrc, it seems to be a github display issue, locally blaming works.

harunurhan:
We should maybe remove the unnecessary files after this like `drone.yml`

Issue #1279, opened 2020-09-04
Author search is broken


michamos:
#1267 broke the author search by using chronological sorting whereas author search relies on TF-IDF relevance scoring so should use the default ES sorting.

A temp fix was made in https://github.com/inspirehep/kubernetes/commit/ef8f90739d1b8bf02de3926a89fbc59fec834711 which should be reverted once the issue is solved.

With an empty query, relevance scoring causes unstable sorting as there's not search query to compute the score from, so all documents have the same score, and the order probably depends on random parameters (like UUIDs assigned to documents or the session key).

There are several solutions:
* add several sorting options, so the index is sorted first by `_score` and then by `name.value` or modification timestamp
* avoid empty searches in the tests

PR #1325, opened 2020-09-17
authors: fix sort

harunurhan:
* Adds bestmatch sort option and makes it default for author search.
* Also adds docs for efficiently working with visual e2e tests and docker.

Ref: #1279


backend/inspirehep/search/mappings/v7/records/authors.json
```diff
@@ -274,7 +277,8 @@
             "copy_to": [
               "authorautocomplete",
               "names_analyzed_initials",
-              "names_analyzed"
+              "names_analyzed",
+              "name_keyword"
```
michamos, line 281: I don't think that's a correct use of `copy_to`. It's meant when you want to combine several fields into one. Better define a `value.raw` of type `keyword` as a variant, like we do for example here: https://github.com/inspirehep/inspirehep/blob/95af58d4b474a36dc9b0a1f37560d831fe793106/backend/inspirehep/search/mappings/v7/records/hep.json#L200-L204

harunurhan: I guess it's also better to have `lowercase_normalizer` just incase?

michamos (approved):

harunurhan:
We should remember to remove the hotfix on prod, when this goes in @michamos

michamos:
You can already do it on master so it goes to QA for testing, and it will be deployed on prod when we do the next release of hep.

PR #1326, opened 2020-09-17
Improve latex escaping

michamos:
* Moves the encoding from the serializers to the marshmallow schemas,
  which allows to selectively encode only some fields and avoid encoding
  others (e.g. DOI).
* Also adds encoding to LaTeX US/EU formats, previously only BibTeX was
  handled.
* Pins `pylatexenc` to get the bugfix for https://github.com/phfaist/pylatexenc/issues/44 reported by @tsgit (not released yet)

### DIFF
--- a/backend/inspirehep/records/marshmallow/literature/utils.py
+++ b/backend/inspirehep/records/marshmallow/literature/utils.py
@@ -5,7 +5,15 @@
 # inspirehep is free software; you can redistribute it and/or modify it under
 # the terms of the MIT License; see LICENSE file for more details.
 
+import re
+
+from pylatexenc.latexencode import UnicodeToLatexEncoder
+
 from inspirehep.records.api import InspireRecord
+
+# The regex selects math delimited by ``$...$`` or ``\(...\)``
+# where the delimiters are not escaped
+MATH_EXPRESSION_REGEX = re.compile(r"((?<!\\)\$.*?(?<!\\)\$|(?<!\\)\\\(.*?(?<!\\)\\\))")
 
 
 def get_parent_record(data):
@@ -20,3 +28,29 @@ def get_parent_record(data):
         data, "publication_info.parent_record"
     )
     return next(book_records, {})
+
+
+def latex_encode(text, contains_math=False):
+    """Encode a string for use in a LaTeX format.
+
+    Args:
+        contains_math (bool): when True, math environments delimited by $...$
+        or \\(...\\) are preserved to avoid double escaping. Note that $$...$$
+        is not handled.
+    """
+    if text is None:
+        return None
+
+    encode = UnicodeToLatexEncoder(
+        replacement_latex_protection="braces-after-macro"
+    ).unicode_to_latex
+
+    if not (contains_math and ("$" in text or r"\(" in text)):
+        return encode(text)
+
+    parts = MATH_EXPRESSION_REGEX.split(text)
+    encoded_text = "".join(
+        encode(part) if i % 2 == 0 else part for i, part in enumerate(parts)
+    )
+
+    return encoded_text
--- a/backend/tests/unit/records/marshmallow/literature/test_literature_marshmallow_utils.py
+++ b/backend/tests/unit/records/marshmallow/literature/test_literature_marshmallow_utils.py
@@ -1,0 +1,63 @@
+# -*- coding: utf-8 -*-
+#
+# Copyright (C) 2019 CERN.
+#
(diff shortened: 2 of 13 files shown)


backend/inspirehep/records/marshmallow/literature/utils.py
```diff
@@ -20,3 +28,33 @@ def get_parent_record(data):
...
+    math_expressions = MATH_EXPRESSION_REGEX.findall(text)
+
+    encoded_math_expressions = {
+        encode(expression): expression for expression in math_expressions
+    }
+    encoded_text = encode(text)
+
+    for encoded_math, math in encoded_math_expressions.items():
+        encoded_text = encoded_text.replace(encoded_math, math)
+
```
tsgit, line 59: interresting approach to encode everything and then substitute parts of the original back

how about avoiding encoding math parts by partitioning the text string and then recombining?

```python

parts = re.split(MATH_EXPRESSION_REGEX, title)                                                                                                   

newtitle = ''.join([encode(t) if i % 2 == 0 else t for i, t in enumerate(parts)])

```

backend/tests/unit/records/marshmallow/literature/test_literature_marshmallow_utils.py
```diff
@@ -0,0 +1,54 @@
...
+
+from inspirehep.records.marshmallow.literature.utils import latex_encode
+
+
+def test_latex_encode_returns_none_when_input_is_none():
+    assert None == latex_encode(None)
+
+
+def test_latex_encode_escapes_special_chars():
+    text = r"\ # <>"
```
tsgit, line 16: why not test all reserved LaTeX chars? 
` # $ % & \ { } _ ~ ^`

backend/inspirehep/records/marshmallow/literature/utils.py
```diff
@@ -5,8 +5,16 @@
...
 
+import re
+
+from pylatexenc.latexencode import UnicodeToLatexEncoder
+
 from inspirehep.records.api import InspireRecord
 
+# The regex selects math delimited by ``$...$`` or ``\(...\)``
+# where the delimiters are not escaped
+MATH_EXPRESSION_REGEX = re.compile(r"((?<!\\)\$.*?(?<!\\)\$|(?<!\\)\\\(.*?(?<!\\)\\\))")
```
tsgit, line 16: this will basically do the reverse of what is intended for displaymath
`some text $$a_b$$ more text`
```
MATH_EXPRESSION_REGEX.findall('some text $$a_b$$ more text')
['$$', '$$']
```

tsgit (commented):
- I can't comment on the split between /inspirehept/records/marshmallow/ and /inspirehep/records/serializers/
- The MATH_EXPRESSION_REGEX needs refinement to handle $$...$$ or we really need to sanitize title metadata
- latex_encode() could be simplified - although it's possibly clearer than my suggested approach which relies on capturing regexp

backend/inspirehep/records/marshmallow/literature/utils.py
michamos: We shouldn't have displaymath anywhere except maybe in abstracts, but we don't expose those in BibTeX/LaTeX. So handling `$$` would make it more complex for no real benefit.

backend/inspirehep/records/marshmallow/literature/utils.py
michamos: I thought it wouldn't work so easily because you don't know whether the string starts with math mode, but after testing it seems you get a leading empty string in this case:
```
In [5]: MATH_EXPRESSION_REGEX.split("$math$ non-math $math$")                                                                                                                                                        
Out[5]: ['', '$math$', ' non-math ', '$math$', '']
```
So good suggestion :+1:

tsgit (approved):
looks good now

backend/inspirehep/records/marshmallow/literature/utils.py
```diff
@@ -20,3 +28,29 @@ def get_parent_record(data):
...
+
+def latex_encode(text, contains_math=False):
+    """Encode a string for use in a LaTeX format.
+
+    Args:
+        contains_math (bool): when True, math environments delimited by $...$
+        or \\(...\\) are preserved to avoid double escaping. Note that $$...$$
+        is not handled.
+    """
+    if text is None:
```
drjova, line 41:
```suggestion
    if not text:
```

drjova (approved):
lgtm just the comment

Issue #1256, opened 2020-08-27
allow filtering of API response


michamos:
A user reports:
> In the old Marc-XML interface, I was limiting the information I obtain from Inspire by using "&ot=001,970,100,..." etc. Is there a way to do something similar in the new API? I don't want to waste the bandwidth, as the default output is very detailed.

This request to be able to get only parts of the record in the response has come up several times and would be useful to do. We could do it only for search results (similar to legacy) very easily by having a query parameter that gets converted in the search classes to the ES `_source` parameter.

harunurhan:
We can modify [search factory](https://github.com/inspirehep/inspirehep/blob/65a2f645ebcc5916659baa449b4820ebd150e760/backend/inspirehep/search/factories/search.py#L19) to override `_source` if it is passed as a request param.

API could look like this:
`?fields=authors,references,acquisition_source`


Fields should be validated against top level schema fields, and return `400` with a meaningful error message to API users.

PR #1330, opened 2020-09-18
search: add filtering of api response

monaawi:
* ref: #1256

### DIFF
--- a/backend/inspirehep/search/api.py
+++ b/backend/inspirehep/search/api.py
@@ -107,6 +107,11 @@ class InspireSearch(RecordsSearch, SearchMixin):
     def source_for_content_type(self, content_type):
         return self
 
+    def source_for_requested_fields(self, fields):
+        includes = fields.split(",")
+        includes.append("control_number")
+        return self.source(includes=includes)
+
     def execute(self, *args, **kwargs):
         if request:
             size = request.args.get("size", default=25, type=int)
--- a/backend/inspirehep/search/errors.py
+++ b/backend/inspirehep/search/errors.py
@@ -18,3 +18,10 @@ class MaximumSearchPageSizeExceeded(BaseRestError):
             )
         else:
             self.description = f"Maximum search page size exceeded."
+
+
+class FieldsParamForbidden(BaseRestError):
+    code = 400
+    description = (
+        "'fields' parameter cannot be used with the requested format or MIME type."
+    )
--- a/backend/inspirehep/search/factories/search.py
+++ b/backend/inspirehep/search/factories/search.py
@@ -10,13 +10,39 @@ from flask import request
 from invenio_records_rest.errors import InvalidQueryRESTError
 from invenio_records_rest.sorter import default_sorter_factory
 
+from ..errors import FieldsParamForbidden
 from .facet import inspire_facets_factory
 from .filter import inspire_filter_factory
 
 LOGGER = structlog.getLogger()
 
 
+def can_request_fields():
+    allowed_mimetypes = [
+        "application/json",
+        "application/vnd+inspire.record.ui+json",
+        "application/vnd+inspire.record.raw+json",
+    ]
+    has_accept_mimetypes = len(request.accept_mimetypes) > 0
+    if (
+        has_accept_mimetypes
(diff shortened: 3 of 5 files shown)


backend/inspirehep/search/factories/search.py
```diff
@@ -17,6 +17,12 @@
 
 
 def get_search_with_source(search):
+    request_fields = request.values.get("fields", "", type=str)
+    if request_fields:
+        request_fields = request_fields.split(",")
+        request_fields.append("control_number")
+        search = search.source(includes=request_fields)
+        return search
```
harunurhan, line 25: to be consistent this can be part of the `InspireSearch` already, method like `source_with_requested_fields` and you use it similarly like below.

backend/inspirehep/search/factories/search.py
```diff
@@ -17,6 +17,12 @@
 
 
 def get_search_with_source(search):
+    request_fields = request.values.get("fields", "", type=str)
+    if request_fields:
+        request_fields = request_fields.split(",")
+        request_fields.append("control_number")
```
harunurhan, line 23: is this a requirement? I would put something like `self` which is more API friendly.

backend/inspirehep/search/factories/search.py
```diff
@@ -17,6 +17,12 @@
 
 
 def get_search_with_source(search):
+    request_fields = request.values.get("fields", "", type=str)
+    if request_fields:
```
michamos, line 21: * what happens if you provide an alternative mime-type, or a `format` arg, together with `fields`? I think it should be a 400 (as it might screw up the other formats) if you request something different from JSON.
* what happens if you try to include a field that the user is not allowed to see? I suppose this will work correctly (due to filtering happening in the search response serializer at a later stage).

harunurhan: I agree with the `400`, we better handle that.

backend/inspirehep/search/factories/search.py
monaawi: I have issues in invenio record fetcher when control_number is missing :/ so I just added it

backend/inspirehep/search/factories/search.py
monaawi: Good point, didn't think that the other formats would fail. 
About the hidden fields, it is handled at the serializer level so there isn't any issue.

backend/inspirehep/search/api.py
```diff
@@ -107,6 +107,11 @@ def get_record_data_from_es(record):
     def source_for_content_type(self, content_type):
         return self
 
+    def source_for_requested_fields(self, fields):
+        includes = fields.split(",")
+        includes.append("control_number")
```
harunurhan, line 112: @michamos any fields you can think of that we should include by default?

backend/inspirehep/search/errors.py
```diff
@@ -18,3 +18,10 @@ def __init__(self, max_size=None, *args, **kwargs):
             )
         else:
             self.description = f"Maximum search page size exceeded."
+
+
+class FieldsParamForbidden(BaseRestError):
+    code = 400
+    description = (
+        "Fields parameter cannot be used with the requested format or mimetype."
```
michamos, line 26:
```suggestion
        "'fields' parameter cannot be used with the requested format or MIME type."
```

harunurhan (approved):

backend/inspirehep/search/factories/search.py
```diff
@@ -10,13 +10,39 @@
...
+        "application/vnd+inspire.record.raw+json",
+    ]
+    has_accept_mimetypes = len(request.accept_mimetypes) > 0
+    if (
+        has_accept_mimetypes
+        and next(request.accept_mimetypes.values()) not in allowed_mimetypes
+    ):
+        return False
+    requested_format = request.values.get("format", "", type=str)
+    if requested_format and requested_format != "json":
```
michamos, line 33:
```suggestion
    requested_format = request.values.get("format", "json", type=str)
    if requested_format != "json":
```

michamos (approved):
minor comments but looks good :+1:

Issue #1295, opened 2020-09-08
Changing conference name is not updating Literature "contribution to" display on search page


pazembrz:
https://github.com/inspirehep/inspirehep/blob/7da1890137a3fa2ea91a803c7c5b499d83fe5ce4/backend/inspirehep/indexer/tasks.py#L62 Should re-index all literature records which are linking to currently changed Conference.

PR #1334, opened 2020-09-21
search: reindex conference papers when conference name changed

harunurhan:
Ref: #1295

### DIFF
--- a/backend/tests/integration-async/records/indexer/test_conferences.py
+++ b/backend/tests/integration-async/records/indexer/test_conferences.py
@@ -5,6 +5,9 @@
 # inspirehep is free software; you can redistribute it and/or modify it under
 # the terms of the MIT License; see LICENSE file for more details.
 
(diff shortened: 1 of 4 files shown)


backend/tests/integration-async/records/indexer/test_conferences.py
```diff
@@ -90,3 +92,68 @@ def test_conference_record_updates_in_es_when_lit_rec_reffers_to_it(
     ]
 
     retry_until_matched(steps)
+
+
+def test_indexer_updates_conference_papers_when_name_changes(
+    inspire_app, celery_app_with_context, celery_session_worker
+):
+    conference_data = faker.record(
+        "conf", data={"titles": [{"title": "Initial Title"}]}
```
pazembrz, line 101: It should be "con" not "conf"

pazembrz (approved):

PR #1341, opened 2020-09-21
search: correct the allowed mimetypes for the api filtering

monaawi:


backend/inspirehep/search/factories/search.py
```diff
@@ -18,15 +18,17 @@
...
-    allowed_mimetypes = [
-        "application/json",
+    disallowed_mimetypes = [
         "application/vnd+inspire.record.ui+json",
-        "application/vnd+inspire.record.raw+json",
+        "application/x-bibtex",
+        "text/vnd+inspire.html+html",
+        "application/vnd+inspire.latex.eu+x-latex",
+        "application/vnd+inspire.latex.us+x-latex",
     ]
```
michamos, line 27: maybe put this somewhere in the config instead?

michamos (approved):

backend/inspirehep/search/factories/search.py
```diff
@@ -18,16 +18,10 @@
...
-        "application/vnd+inspire.record.raw+json",
-    ]
     has_accept_mimetypes = len(request.accept_mimetypes) > 0
-    if (
-        has_accept_mimetypes
-        and next(request.accept_mimetypes.values()) not in allowed_mimetypes
-    ):
+    if has_accept_mimetypes and next(
+        request.accept_mimetypes.values()
+    ) in current_app.config.get("FORBIDDEN_MIMETYPES_FOR_API_FILTERING"):
```
michamos, line 24: this is not going to work if the key is not present, as it will be `... in None`. So either pass a default arg to `get`, or drop the `get` and use `[...]`.

PR #1368, opened 2020-09-28
WIP: Build dependencies docker image

benjamin-bergia:


.github/workflows/backend-build-and-test.yml
```diff
@@ -0,0 +1,129 @@
...
+          docker run
+          --entrypoint poetry
+          --network="host"
+          inspirehep/hep:${{ github.sha }}
+          run py.test tests/unit
+
+  Deploy:
+    runs-on: ubuntu-latest
+    needs: [Build, Integration, Unit]
+    if: ${{ github.event_name == 'push' }}
```
drjova, line 96: I think we need this on `release` as well

benjamin-bergia: I will run a different WF for the releases since it's just a matter of re-tagging an existing image.

Issue #1363, opened 2020-09-24
Put the `fields` argument in `links` for search responses


michamos:
We now have the `fields` argument on search endpoints, but the search `links` don't include it. We should update https://github.com/inspirehep/inspirehep/blob/d5541f2f85f8a4d003be24031f6600278393b9a0/backend/inspirehep/records/links.py#L38-L54 to add this argument to the `next`, `self` and `json` links.

PR #1379, opened 2020-09-30
search: add the fields argument to the search links

monaawi:
* ref: #1363

### DIFF
--- a/backend/inspirehep/records/links.py
+++ b/backend/inspirehep/records/links.py
@@ -47,8 +47,24 @@ def inspire_search_links(links):
         .get("search_serializers_aliases", {})
         .keys()
     )
+    request_fields = request.values.get("fields", "", type=str)
+    url_fields = ""
+    if request_fields:
+        url_fields = f"&fields={request_fields}"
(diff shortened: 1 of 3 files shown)


backend/inspirehep/records/links.py
```diff
@@ -47,6 +47,15 @@ def inspire_search_links(links):
...
     )
+    request_fields = request.values.get("fields", "", type=str)
+    if request_fields:
+        url_fields = f"&fields={request_fields}"
+        self_url += url_fields
+        links["self"] = self_url
+        next_url = links.get("next")
+        if next_url:
+            next_url += url_fields
+            links["next"] = next_url
```
drjova, line 58: It should be only for `self` and `json`, for the other formats is useless AFAIR

monaawi: You are right, removed it for other formats. It should also be for ```next``` though.

drjova: @monaawi and `previous`

drjova (approved):

Issue #1219, opened 2020-08-19
add key to submission forms to disambiguate people with same name


michamos:
(Moved from [JIRA](https://its.cern.ch/jira/browse/INSPIR-3560))

We have a bug in the seminar submission forms (and probably others), causing the author selection not to work correctly when there are multiple authors with the same name.

PR #1381, opened 2020-09-30
ui: refactor core suggester and the form field

harunurhan:
* Moves core logic from SuggesterField to Suggester
which now provides a simpler API.
* Fixes bug #1219, also for all suggester fields
* Adds id for suggester requests so that unncessary requests are
cancelled

### DIFF
--- a/ui/src/submissions/common/components/SuggesterField.jsx
+++ b/ui/src/submissions/common/components/SuggesterField.jsx
@@ -3,6 +3,10 @@ import Suggester from '../../../common/components/Suggester';
 
 import withFormItem from '../withFormItem';
 
+function getSuggestionControlNumber(suggestion) {
+  return String(suggestion._source.control_number);
+}
+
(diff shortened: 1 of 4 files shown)


ui/src/submissions/common/components/SuggesterField.jsx
```diff
@@ -12,6 +16,10 @@ class SuggesterField extends Component {
     this.recordFieldPopulated = false;
   }
 
+  static getSuggestionControlNumber(suggestion) {
```
drjova, line 19: is this on purpose?  https://github.com/inspirehep/inspirehep/pull/1381/files#diff-3d8676fe3ea027dff622d0b717aaa7ccR6

harunurhan: no leftover.

drjova (approved):

PR #1386, opened 2020-10-02
Simple async test retry

harunurhan:


backend/tests/helpers/utils.py
```diff
@@ -156,7 +157,24 @@ def orcid_app_cli_runner():
     return runner
 
 
-# Integration-async helpers
+def retry_until_pass(assert_function, timeout=30, retry_interval=0.3):
+    last_raised_assertion_error = None
+    timeout_delta = timedelta(seconds=timeout)
+    start = datetime.now()
```
michamos, line 163: better use https://docs.python.org/3/library/time.html#time.monotonic to avoid the corner cases with leap seconds, daylight saving times, etc.

drjova (approved):

monaawi (approved):

PR #1393, opened 2020-10-05
UI bug: deleted alert in institution details.

oguzdemirbasci:


ui/src/institutions/containers/DetailPageContainer.jsx
```diff
@@ -68,6 +70,9 @@ function DetailPage({ record }) {
               </>
             }
           >
+            <Row>
+              <Col span={24}>{deleted && <DeletedAlert />}</Col>
```
harunurhan, line 74: This would render `Row` with a `Col` unnecessarily, when record is not deleted.

Do `deleted && <Row>...` instead.

I think is what is causing visual tests to fail.

harunurhan (changes requested):

harunurhan (approved):

PR #1399, opened 2020-10-05
matcher: add reference matcher cli

harunurhan:


harunurhan:
This is still WIP.

backend/inspirehep/matcher/api.py
```diff
@@ -178,10 +178,17 @@ def match_references(references):
...
 
-    return matched_references
+        if not any_reference_changed and current_record_ref != new_record_ref:
+            any_reference_changed = True
+
+        matched_references.append(reference)
+        if "record" in reference:
+            previous_matched_recid = get_recid_from_ref(reference["record"])
+
+    return matched_references if any_reference_changed else references
```
michamos, line 194: this is subtle, maybe you should say something about this in the docstring?

backend/inspirehep/matcher/cli.py
```diff
@@ -0,0 +1,61 @@
...
+@with_appcontext
+def references(batch_size, db_batch_size):
+    # TODO: logs and maybe progress?
+    literature_uuids_query = PersistentIdentifier.query.filter(
+        PersistentIdentifier.pid_type == "lit",
+        PersistentIdentifier.status == PIDStatus.REGISTERED,
+    ).with_entities(PersistentIdentifier.object_uuid)
+    uuid_chunk = []
+    matcher_tasks = []
+    for result in literature_uuids_query.yield_per(db_batch_size):
```
michamos, line 48: or more simply,
```suggestion
    for (uuid,) in literature_uuids_query.yield_per(db_batch_size):
```

backend/inspirehep/matcher/cli.py
```diff
@@ -0,0 +1,61 @@
...
+        uuid = result[0]
+        uuid_chunk.append(uuid)
+        if len(uuid_chunk) == batch_size:
```
michamos, line 51: I think this has a bug if the number of results is not an integer multiple of `batch_size`. You can use https://github.com/inspirehep/inspirehep/blob/cea05cab04a1f3d94bee7097a40402c707cebd6c/backend/inspirehep/utils.py#L23

backend/inspirehep/matcher/tasks.py
```diff
@@ -0,0 +1,35 @@
...
+from sqlalchemy.dialects.postgresql import JSONB
+
+from inspirehep.records.api import LiteratureRecord
+
+from .api import match_references
+
+
+@shared_task(
+    ignore_result=False, queue="matcher",
+)
```
michamos, line 21: you should add `acks_late=True` and retry options on possible errors (look at  migrator).

backend/inspirehep/matcher/tasks.py
```diff
@@ -0,0 +1,35 @@
...
+def match_references_by_uuids(literature_uuids):
+    # TODO: do not crash all if it fails all record
+    record_json = type_coerce(RecordMetadata.json, JSONB)
+    has_references = record_json.has_key("references")  # noqa: W601
+    selected_uuids = RecordMetadata.id.in_(literature_uuids)
+    with_references_query = RecordMetadata.query.filter(selected_uuids, has_references)
+    for record_metadata in with_references_query.all():
+        references = record_metadata.json.get("references")
+        matched_references = match_references(references)
+        if references != matched_references:
```
michamos, line 31: I think you want
```suggestion
        if references is not matched_references:
```
to avoid doing the deep equality check.

backend/tests/integration/matcher/test_tasks.py
```diff
@@ -0,0 +1,53 @@
...
+# the terms of the MIT License; see LICENSE file for more details.
+
+from helpers.utils import create_record
+from mock import patch
+
+from inspirehep.matcher.tasks import match_references_by_uuids
+from inspirehep.records.api import LiteratureRecord
+
+
+@patch("inspirehep.matcher.tasks.match_references")
```
michamos, line 15: why mock this? can't you have a real test that does the matching?

backend/inspirehep/matcher/cli.py
harunurhan: oh yeah you're right about the bug, I don't like `chunker` but I'll have a look if it works in this case.

backend/inspirehep/matcher/tasks.py
harunurhan: I guess it will be only DB related errors.

backend/tests/integration/matcher/test_tasks.py
harunurhan: wouldn't that be hard to setup? and depend on internal implementation of `reference_matcher` which is already tests and not the concern of this task.

michamos: why hard to set up? all you need is the cited record to be available, give it some easily cited info (e.g. a DOI), then cite it by the DOI. And isn't the point of an integration test to check that everything is well integrated? I would avoid mocking stuff if it's not strictly needed.

harunurhan: I would prefer to have an e2e that tests from cli instead of doing this actually.

Well the way we use integration tests sometimes, is as unit tests with DB, ES etc.

michamos (approved):

backend/tests/integration-async/matcher/test_cli.py
```diff
@@ -0,0 +1,52 @@
...
+    citer_record_2 = create_record_async("lit", data=citer_data)
+    citer_record_3 = create_record_async("lit", data=citer_data)
+    create_record_async("dat")
+    result = cli.invoke(["match", "references", "-bs", 2])
+
+    assert result.exit_code == 0
+
+    updated_citer_record_1 = LiteratureRecord.get_record(citer_record_1.id)
+    updated_citer_record_2 = LiteratureRecord.get_record(citer_record_2.id)
+    updated_citer_record_3 = LiteratureRecord.get_record(citer_record_3.id)
```
drjova, line 39:
```suggestion
    citer_record_id_1 = citer_record_1.id
    citer_record_id_2 = citer_record_2.id
    citer_record_id_3 = citer_record_3.id
    result = cli.invoke(["match", "references", "-bs", 2])

    assert result.exit_code == 0

    updated_citer_record_1 = LiteratureRecord.get_record(citer_record_id_1)
    updated_citer_record_2 = LiteratureRecord.get_record(citer_record_id_2)
    updated_citer_record_3 = LiteratureRecord.get_record(citer_record_id_3)
```
This should work

Issue #1268, opened 2020-09-02
Run integration tests under different DB name


harunurhan:
So that it won't interfere with the local dev setup and e2e tests.

TECH: `SQLALCHEMY_DATABASE_URI` env variable is used to determine DB to which Inspire is connecting.  If before test it will be set to different db name: for example `INVENIO_SQLALCHEMY_DATABASE_URI=postgresql+psycopg2://inspirehep:inspirehep@localhost:5432/inspire_2` it will use it for the tests.

This variable can be set on executing test command:
```bash
INVENIO_SQLALCHEMY_DATABASE_URI=postgresql+psycopg2://inspirehep:inspirehep@localhost:5432/inspire_2 poetry run pytest/tests/integration
```
in https://github.com/inspirehep/inspirehep/blob/d524f1fe523b61da6d3e47ea678eccd56b91a5a4/backend/run-tests.sh#L10
Or it can be hard-coded in https://github.com/inspirehep/inspirehep/blob/d524f1fe523b61da6d3e47ea678eccd56b91a5a4/backend/tests/integration/conftest.py#L37 
Please note that setting INVENIO_SQLALCHEMY_DATABASE_URI will still override value set in conftest.

BTW: Do we need [this](https://github.com/inspirehep/inspirehep/blob/d524f1fe523b61da6d3e47ea678eccd56b91a5a4/backend/tests/integration/conftest.py#L47-L57) ? Looks like it's value is overwritten if we use `INVENIO_SQLALCHEMY_DATABASE_URI`

Don't forget to also change ES index prefix by setting `SEARCH_INDEX_PREFIX` env, otherwise running tests will delete ES indexes used by local env.

Same changes should work for integration-async

MJedr:
- Some services (matcher, oai, migrator) have hardcoded index names in the config, so we have to find a proper method for updating them. One idea is to create a function which will update them based on the prefix, eg. 
```
def get_prefixed_index(index_name, index_prefix=None):
    prefix = index_prefix or current_app.config.get("SEARCH_INDEX_PREFIX")
    if prefix not in index_name:
          return prefix + index_name
    return index_name
```
but when used directly in config it gives the ``` RuntimeError: Working outside of application context.``` error.
- Some tests [like this one](https://github.com/inspirehep/inspirehep/blob/08ae47665713aa0cce90572b72e7f2eb710cd0d6/backend/tests/integration/indexer/test_cli.py#L206) remove aliases from the ES.
[...]

PR #1401, opened 2020-10-06
tests: use different db uri for tests

MJedr:
Ref: #1268


backend/inspirehep/matcher/api.py
```diff
@@ -23,11 +25,19 @@ def get_reference_from_grobid(query):
     return GrobidReferenceParser(response.text).parse()
 
 
+def _update_search_index_in_configs(configs):
```
drjova, line 28: I would avoid `deepcopy` etc. We could pull out the logic of https://github.com/inspirehep/inspirehep/blob/08ae47665713aa0cce90572b72e7f2eb710cd0d6/backend/inspirehep/indexer/cli.py#L325 and create a utility function something like:

```python
def get_prefixed_index(index_name, index_prefix=None):
    prefix = index_prefix or current_app.config.get("SEARCH_INDEX_PREFIX")
    if prefix not in index_name:
          return prefix + index_name
    return index_name
```

then we can reuse it in the `mathcer/config.py` for example:

```python
CONFIG = {    
  "algorithm": [
        {
            "queries": []
           
    ],
    "index": get_prefixed_index("records-hep"),
}
```

backend/tests/integration/indexer/test_cli.py
```diff
@@ -203,8 +205,31 @@ def _test_alias_to_index(alias_name, expected_index_name):
     assert expected_index_name in indexes[0]
 
 
+def _save_orginal_indices_aliases():
```
drjova, line 208: why we need to do that?

PR #1413, opened 2020-10-09
ui: add typescript

harunurhan:
* enables typescript during prod build and dev.
* migrates some files (utility, class component, function component etc.) to
use typescript

Ref #1375


drjova:
Looks good! Could we add the `ts` and `tsx` in something like `resolve.extensions` and omit adding the extension everywhere?

harunurhan:
I don't wanna overwrite webpack config just for this. If you import `ts(x?)` in typescript files, it works without the extension anyway.

drjova (approved):

PR #1428, opened 2020-10-13
Build UI docker image in github WF. Run the tests on the image

benjamin-bergia:


ui/package.json
```diff
@@ -12,8 +12,8 @@
     "start": "craco start",
     "build": "craco build",
     "test:unit": "craco test",
-    "test": "yarn lint && yarn test:unit && yarn test:size",
     "test:size": "yarn build && bundlesize",
+    "test": "yarn lint && yarn test:unit --no-watch && yarn test:size",
```
drjova, line 16: why do we need ``--no-watch``?

benjamin-bergia: There is a limit on the number of files you can watch simultaneously and inside the docker container it seems to be lower. So the tests fail.

drjova: @harunurhan will this affect local testing?

harunurhan: no it won't

drjova (approved):

PR #1431, opened 2020-10-14
records: remove `indexer_class` from configuration

drjova:


backend/inspirehep/records/config.py
```diff
@@ -71,7 +71,7 @@
     "default_endpoint_prefix": True,
     # XXX decide about the links
     "links_factory_imp": "inspirehep.records.links:inspire_detail_links_factory",
-    "indexer_class": RecordIndexer,
+    "indexer_class": None,
```
michamos, line 74: should we add a test for this (e.g. mocking the indexer task to do nothing, then checking that the record is not in ES)?

drjova: IMHO it's invenio configuraiton, If we start testing this then we should add tests for many others

michamos: ok then, except for the now (hopefully) unused import on L18 :)

michamos (approved):

backend/inspirehep/records/config.py
drjova: 🤦

PR #1443, opened 2020-10-15
poetry: bump inspire-dojson to v63.0.14

michamos:
I'm using poetry 1.1.3 now (I had problems with poetry 1.0.0), which has a somewhat different ordering of keys in the lockfile. Sorry for the noise. If it's an issue, maybe we could switch to the latest poetry everywhere, or someone else could bump it?


drjova:
@michamos let's bump the version everywhere :)

drjova (approved):

Issue #1374, opened 2020-09-29
Remove email addresses from search results


michamos:
For privacy reasons and to make it easier to comply with our terms of use, we should remove email addresses from the search results, for all collections for the public API.

Tech:
Literature:
On UI response we have 
 [FirstAuthorSchemaV1](https://github.com/inspirehep/inspirehep/blob/a7d56913dc8d634067d7067c59d3615b562887e4/backend/inspirehep/records/marshmallow/literature/common/author.py#L25) with email field.
There is also https://github.com/inspirehep/inspirehep/blob/183ee02c49f99518cb5c937b66db5db39aeb9917/backend/inspirehep/records/marshmallow/literature/ui.py#L40 Which contains `acquisition_source.email`. This field should be excluded. But we don't want to exclude it from Detail page, so probably we have to create separate schema for literature ui field.
https://github.com/inspirehep/inspirehep/blob/183ee02c49f99518cb5c937b66db5db39aeb9917/backend/inspirehep/records/marshmallow/literature/es.py#L39 This one will also contain `acquisition_source.email` and it's displayed instead of ui_field when content_type is json.

Authors ([AuthorsListSchema](https://github.com/inspirehep/inspirehep/blob/a7d56913dc8d634067d7067c59d3615b562887e4/backend/inspirehep/records/marshmallow/authors/ui.py#L84)): extend field `Meta.exclude` with `email_addresses` field.

Jobs: [JobsListSchema](https://github.com/inspirehep/inspirehep/blob/a7d56913dc8d634067d7067c59d3615b562887e4/backend/inspirehep/records/marshmallow/jobs/ui.py#L56) Here we have emails in `acquisition_source.email`, `contact_details[].email`, `reference_letters.emails[]`. All of those should be excluded from schema.

Seminars:
[SeminarsListSchema](https://github.com/inspirehep/inspirehep/blob/a7d56913dc8d634067d7067c59d3615b562887e4/backend/inspirehep/records/marshmallow/seminars/ui.py#L32) Here emails are stored in  `acquisition_source.email`, `contact_details[].email`. All of those should be excluded from schema.

Conferences:
[ConferencesListSchema](https://github.com/inspirehep/inspirehep/blob/183ee02c49f99518cb5c937b66db5db39aeb9917/backend/inspirehep/records/marshmallow/conferences/ui.py#L19) Here we have only `contact_details` list which should be excluded from schema.

Institutions: Nothing in here

Experiments: Nothing in here

[...]

PR #1444, opened 2020-10-15
serializers: remove emails from api response

MJedr:
Ref: #1374

### DIFF
--- a/backend/inspirehep/records/marshmallow/authors/base.py
+++ b/backend/inspirehep/records/marshmallow/authors/base.py
@@ -9,6 +9,7 @@ from marshmallow import Schema, fields
 
 from ..base import RecordBaseSchema
 from ..fields import NonHiddenRaw
+from ..utils import get_acquisition_source_without_email
 
 
 class AuthorsRawSchema(RecordBaseSchema):
@@ -24,6 +25,17 @@ class AuthorsPublicSchema(AuthorsRawSchema):
         exclude = FIELDS_TO_EXCLUDE + ["_private_notes", "_collections"]
 
     email_addresses = NonHiddenRaw(dump_only=True)
+
+
+class AuthorsPublicListSchema(AuthorsRawSchema):
+    class Meta:
+        exclude = AuthorsPublicSchema.Meta.exclude + ["email_addresses"]
+
+    acquisition_source = fields.Method("get_acquisition_source")
+
+    @staticmethod
+    def get_acquisition_source(data):
+        return get_acquisition_source_without_email(data)
 
 
 class AuthorsAdminSchema(AuthorsRawSchema):
--- a/backend/inspirehep/records/marshmallow/jobs/base.py
+++ b/backend/inspirehep/records/marshmallow/jobs/base.py
@@ -9,7 +9,13 @@
 from marshmallow import fields
 
 from inspirehep.records.marshmallow.base import RecordBaseSchema
-from inspirehep.records.marshmallow.common import AcceleratorExperimentSchemaV1
+from inspirehep.records.marshmallow.common import (
+    AcceleratorExperimentSchemaV1,
+    ContactDetailsItemWithoutEmail,
+)
+
+from ..utils import get_acquisition_source_without_email
+from .utils import get_reference_letters_without_email
 
 
 class JobsRawSchema(RecordBaseSchema):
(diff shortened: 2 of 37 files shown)


backend/inspirehep/records/marshmallow/authors/base.py
```diff
@@ -26,6 +27,17 @@ class Meta:
     email_addresses = NonHiddenRaw(dump_only=True)
 
 
+class AuthorsApiSchema(AuthorsRawSchema):
+    class Meta:
+        exclude = AuthorsPublicSchema.Meta.exclude + ["email_addresses"]
+
+    acquisition_source = fields.Method("get_acquisition_source")
```
drjova, line 34: You could add these in ``AuthorsPublicSchema``

backend/inspirehep/records/marshmallow/jobs/base.py
```diff
@@ -21,6 +27,20 @@ def __init__(self, *args, **kwargs):
...
+    reference_letters = fields.Method("get_reference_letters")
+    acquisition_source = fields.Method("get_acquisition_source")
+
+    @staticmethod
+    def get_acquisition_source(data):
+        return get_acquisition_source_without_email(data)
+
+    @staticmethod
+    def get_reference_letters(data):
+        return get_reference_letters_without_email(data)
```
drjova, line 41: You could add these in `JobsPublicSchema`

backend/inspirehep/records/marshmallow/literature/base.py
```diff
@@ -49,6 +51,19 @@ class Meta:
     report_numbers = NonHiddenRaw(dump_only=True)
 
 
+class LiteratureApiSchema(LiteraturePublicSchema):
```
drjova, line 54: ditto

backend/inspirehep/records/serializers/json/jobs.py
```diff
@@ -17,9 +17,14 @@
     wrap_schema_class_with_metadata(JobsPublicSchema), index_name="records-jobs"
 )
 
+jobs_json_2 = JSONSerializer(
```
drjova, line 20: if we add everything on `JobsPublicSchema` don't need to change anything here :)

backend/inspirehep/records/marshmallow/authors/base.py
MJedr: The problem is that `AuthorsDetailSchema` is inheriting from `AuthorsPublicSchema` and if I add it I'll exclude emails also from the detailed page and we don't want to do it

backend/inspirehep/records/marshmallow/literature/base.py
MJedr: It's the same, I can't add it to PublicSchema because I'll change detailed page

drjova:
@MJedr I see, what if we do it in ListSchema for all collections instead of adding a new Api schema? This will modify only search results.

MJedr:
But we use ListSchemas only for `"application/vnd+inspire.record.ui+json"` and we should remove emails from the response for `"application/json"`. Now we use for that PublicSchemas which have more fields than the BaseSchemas from which inherits ListSchemas so I am nor sure if we can use them for `"application/json"` response.

backend/inspirehep/records/marshmallow/authors/base.py
drjova: I see your point, so let's do the following:

We keep it as you did but rename `CollectionApiSchema` to `CollectionPublicListSchema` (i.e. `AuthorPublicListSchema`)

backend/inspirehep/records/serializers/json/literature.py
```diff
@@ -31,6 +32,16 @@
...
+    [
+        (
+            lambda _: is_superuser_or_cataloger_logged_in(),
+            wrap_schema_class_with_metadata(LiteratureAdminSchema),
+        ),
+        (None, wrap_schema_class_with_metadata(LiteraturePublicListSchema)),
+    ]
+)
+
+literature_json_detailed = ConditionalMultiSchemaJSONSerializer(
```
drjova, line 44: in some places it's `_detailed` in some others is `_record`, let unify it 

```collection_json``` -> for record response
```collection_json_search``` -> for search response

WDYT?

drjova (approved):

Issue #1430, opened 2020-10-14
Populate `_private_notes` with proceeding and conference info from submissions


harunurhan:
- add a private note with `proceeding_info` field data
- add a private note with `conference_info` field data only if `conference_record` is not present in the form data
- all private notes should have `source="submitter"`

harunurhan:
Similar to https://github.com/inspirehep/inspirehep/blob/4c5deb9e41beaef682514cce22bb64b6211177dc/backend/inspirehep/submissions/marshmallow/literature.py#L244

we shouldn't forget put pass `source="submitter"` to the builder for all `add_private_note` calls.

PR #1448, opened 2020-10-16
submission: save proceeding and conference info in private_notes

oguzdemirbasci:
Save proceeding and conference info in `_private_notes` field for literature submission.

Ref: #1430

### DIFF
--- a/smoke-tests/cypress/integration/literature.test.js
+++ b/smoke-tests/cypress/integration/literature.test.js
@@ -111,6 +111,8 @@ describe('Literature Submission', () => {
       issue: '20',
       year: '2014',
       comments: 'very private thing',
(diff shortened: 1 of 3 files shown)


smoke-tests/cypress/integration/literature.test.js
```diff
@@ -160,6 +169,8 @@ describe('Literature Submission', () => {
     };
     cy.visit('/submissions/literature');
     cy.selectLiteratureDocType('article');
+    cy.contains('Conference Info').click();
+    cy.contains('Proceedings Info (if not published in a journal)').click();
```
harunurhan, line 173: This can be `Proceedings Info` only without the full title.

harunurhan (approved):

Issue #1367, opened 2020-09-25
Fetch record's resources (metadata and schema) all at once (record-editor)


harunurhan:
`record-editor` is currently fetching record `{ metadata }` and the `$schema` sequentially. This was implemented as a POC, quickly, and stayed like that for 3 years.

First idea would be, add an endpoint that returns whatever record-editor needs at once. This wasn't done at the beginning due to `invenio` limitations.

Alternatively we can just kick-off `$schema` request at the same time, with record metadata request. Since we know the `$schema` url, from the `type` in editor url (`/editor/record/<type>/<id>`). However this puts some assumption that might cause problems later, such as "record of same type must have same schema", "all schemas have to be served under fixed url (or at least relative to the window.location)". 

**TECH**
1. Create a new endpoint  which will take as an argument the record/workflow object url. Get a record and its pid type from the provided url. Use `load_schema` from inspire_schemas to load the schema for the given pid type and update the `schema` field in the record (when processing data from workflow object url it record is under [metadata](https://github.com/inspirehep/inspirehep/blob/fda4474393b78d1f4a4223ccc9c5b60d655b70a0/record-editor/src/app/shared/interfaces/workflow-object.ts#L11)). Return data with replaced `schema` field.
2. Modify [`fetchWorkflowObject`](https://github.com/inspirehep/inspirehep/blob/fda4474393b78d1f4a4223ccc9c5b60d655b70a0/record-editor/src/app/core/services/record-api.service.ts#L53) and [`fetchRecord`](https://github.com/inspirehep/inspirehep/blob/fda4474393b78d1f4a4223ccc9c5b60d655b70a0/record-editor/src/app/core/services/holdingpen-api.service.ts#L41) methods. In the end fetch the data from a new endpoint which will return both record and the schema.
3. Remove unnecessary fetch (and fetched data processing) from [here](https://github.com/inspirehep/inspirehep/blob/1a4633bbddb06687ce6e4ad44d0df7c2362901df/record-editor/src/app/record-editor/json-editor-wrapper/json-editor-wrapper.component.ts#L162-L174) and [here](https://github.com/inspirehep/inspirehep/blob/1a4633bbddb06687ce6e4ad44d0df7c2362901df/record-editor/src/app/holdingpen-editor/holdingpen-editor.component.ts#L79-L95).

harunurhan:
[...]

PR #1451, opened 2020-10-16
record-editor: add and use GET endpoints that returns schema too

harunurhan:
Ref #1367

### DIFF
--- a/backend/inspirehep/editor/views.py
+++ b/backend/inspirehep/editor/views.py
@@ -9,6 +9,7 @@ from os.path import splitext
 
 from flask import Blueprint, current_app, jsonify, request
 from flask_login import current_user
+from inspire_schemas.api import load_schema
 from invenio_db import db
 from invenio_records.models import RecordMetadata
 from refextract import extract_references_from_string, extract_references_from_url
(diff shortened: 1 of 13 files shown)


backend/tests/integration/editor/test_editor_views.py
```diff
@@ -24,6 +24,24 @@
...
+        response = client.get(f"api/editor/conferences/{conference['control_number']}")
+
+    assert response.status_code == 200
+
+    response_data = json.loads(response.data)
+    record_metadata = response_data["record"]["metadata"]
+    schema = response_data["schema"]
+
+    assert record_metadata == dict(conference)
+    assert schema == load_schema("conferences")
```
drjova, line 42: could we add an unhappy path test (no permissions) as well?

backend/inspirehep/editor/views.py
```diff
@@ -44,6 +45,16 @@ def revert_to_revision(endpoint, pid_value):
         raise EditorRevertToRevisionError
 
 
+@blueprint.route("/<endpoint>/<int:pid_value>", methods=["GET"])
+@login_required_with_roles([Roles.cataloger.value])
+def get_record_and_schema(endpoint, pid_value):
+    pid_type = PidStoreBase.get_pid_type_from_endpoint(endpoint)
+    record = InspireRecord.get_record_by_pid_value(pid_value, pid_type)
+    return jsonify(
+        {"record": {"metadata": record}, "schema": load_schema(record["$schema"])}
```
drjova, line 54: we could also drop `metadata` right? Since we're changing the editor as well

harunurhan: I just kepth the this if we need to add more info for the record later, like updated etc.

backend/tests/integration/editor/test_editor_views.py
```diff
@@ -24,6 +24,35 @@
...
+
+
+def test_get_record_and_schema_requires_cataloger_logged_in(inspire_app):
+    user = create_user()
+    conference = create_record("con")
+
+    with inspire_app.test_client() as client:
+        login_user_via_session(user, email=cataloger.email)
```
drjova, line 50:
```suggestion
        login_user_via_session(user, email=user)
```

drjova (approved):
LGTM, just the typo in the test!!

Issue #1412, opened 2020-10-08
inspire-matcher should use `SEARCH_INDEX_PREFIX`


drjova:
Currently `inspire-matcher` https://github.com/inspirehep/inspire-matcher/blob/master/inspire_matcher/api.py#L50 doesn't use the `SEARCH_INDEX_PREFIX`, and you have to pass the exact index in the configuration (example: https://github.com/inspirehep/inspirehep/blob/master/backend/inspirehep/matcher/config.py#L37). 

**Acceptance:**
if `SEARCH_INDEX_PREFIX="preifx-abc-"` given in the configuration the index is `record-literature`, matcher should search in `preifx-abc-record-literature`.

PR #1452, opened 2020-10-16
matcher: use search index prefix

MJedr:
Ref: #1412

### DIFF
--- a/backend/inspirehep/matcher/api.py
+++ b/backend/inspirehep/matcher/api.py
@@ -12,6 +12,8 @@ from inspire_matcher import match
 from inspire_utils.dedupers import dedupe_list
 from inspire_utils.record import get_value
 
+from inspirehep.utils import get_prefixed_index
+
 from .parsers import GrobidReferenceParser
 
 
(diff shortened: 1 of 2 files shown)


backend/inspirehep/matcher/config.py
```diff
@@ -5,6 +5,9 @@
 # inspirehep is free software; you can redistribute it and/or modify it under
 # the terms of the MIT License; see LICENSE file for more details.
 
+from inspirehep.config import SEARCH_INDEX_PREFIX
```
drjova, line 8: This will take take the hardcoded value of the prefix and will not take in consideration the env changes or the overrides (i.e. ``current_app``)

backend/inspirehep/matcher/api.py
```diff
@@ -113,6 +115,9 @@ def match_reference_config(reference):
         config_data,
     ]
 
+    for config in configs:
+        config.update({"index": get_prefixed_index(config.get("index"))})
```
michamos, line 119: maybe this was discussed and I'm missing something, but why not move this to the `inspire-matcher` library, to make this transparent to the caller? Note that it already depends on Flask due to using `invenio-search`, so it can read the config.

MJedr: Yes, seems good to me. What do you think @drjova?

drjova: Totally agree, the change in the refinement it was mentioning the change to be in the `inspire-matcher`. We should close this PR and do it directly there.

Issue #1387, opened 2020-10-02
Handle versions for optimistic concurrency control in editor/REST endpoints


michamos:
When making changes to a record, it might happen that in the meantime, a concurrent transaction has made a different change to the same record. 

The way things are done currently, those modifications will be silently overwritten by the most recent transaction to be committed to the database (nothing catastrophic is happening though due to legacy version checks). The `RecordsMetadata` model uses versioning for optimistic concurrency control: SQLAlchemy maintains a version counter, which is forced to monotonically increase. In case a transaction tries to lower the version counter (as would happen if the record to be modified was read at an earlier version), an exception is raised.

We should expose this feature to the REST APIs and editor to make sure we don't silently overwrite data.

harunurhan:
Also for submissions that does immediate updates (without approvals)

harunurhan:
A cataloger did "important" editing and meanwhile a new update came in, so SAVE would fail for the cataloger. They could refresh to get the new version but then they would lose all the work. I think we have to address this issue.

@michamos

michamos:
At some point we had the idea of using the merger to combine the two update and the edit, but that would be a lot of work to implement. As a first step, we can just fail the save with a clear message of what to do, then improve the UX if it happens often.

harunurhan:
How about doing a simple diff on top level fields if different fields modified just accept the change instead of failing? Is there problem with that?

harunurhan:
Invenio is "supposed" to implement this anyway via https://github.com/inveniosoftware/invenio-records-rest/blob/896a34e47f3a1eab036c89d6348f59c2679fba86/invenio_records_rest/views.py#L883.

Etag check logic is here: https://github.com/inveniosoftware/invenio-rest/blob/0322ce1d63a6f1736bcc4dc0928c9336b96ae0e1/invenio_rest/views.py#L250-L285

This doesn't work because:
1. Invenio is returning `record.version_id` for GET but passing `record.revision_id`. Maybe this is how it is supposed work, but I have some doubts, because it seems like `revision_id = version_id + 1`, and they are maintained differently I guess.
2. Nginx servers compressed json with weak `eTag` even if it is strong, however invenio is checking eTag with `weak=False`
[...]

PR #1463, opened 2020-10-22
records: do not allow stale records updates over HTTP

harunurhan:
* Installs a fork of `invenio-records-rest` which does weak etag checks
* Modifies record-editor to send `If-Match`
* Uses `vary` header instead of `Content-Type` aware `Etag`
which fixes `Etag` and `If-None-Match` based HTTP caching

Ref #1387


backend/inspirehep/editor/views.py
```diff
@@ -50,9 +50,12 @@ def revert_to_revision(endpoint, pid_value):
 def get_record_and_schema(endpoint, pid_value):
     pid_type = PidStoreBase.get_pid_type_from_endpoint(endpoint)
     record = InspireRecord.get_record_by_pid_value(pid_value, pid_type)
-    return jsonify(
-        {"record": {"metadata": record}, "schema": load_schema(record["$schema"])}
-    )
+    json = {"record": {"metadata": record}, "schema": load_schema(record["$schema"])}
+
+    response = make_response(json)
+    set_headers_for_record_caching_and_concurrency(response, record)
```
harunurhan, line 56: this could potentially cause a state cache, when schema is changed but not the record.

backend/inspirehep/utils.py
```diff
@@ -45,3 +45,8 @@ def hash_data(data):
     if data:
         return hashlib.md5(data).hexdigest()
     raise ValueError("Data for hashing cannot be empty")
+
+
+def set_headers_for_record_caching_and_concurrency(response, record):
+    response.set_etag(str(record.revision_id), weak=True)
+    response.vary = "Content-Type,Accept,Accept-Encoding,Accept-Language"
```
michamos, line 52: I don't think you need the `content-type`, it seems vary is used to add additional *request* headers to the cache key.

backend/tests/integration/records/views/test_views.py
```diff
@@ -67,3 +67,36 @@ def test_returns_deleted_pid_error_if_not_cataloger(inspire_app):
...
+        stale_put_response = client.put(
+            f"/conferences/{record_control_number}",
+            content_type="application/json",
+            data=json.dumps(
+                faker.record("con", data={"control_number": record_control_number})
+            ),
+            headers={"If-Match": etag},
+        )
+
+    assert stale_put_response.status_code == 412
```
michamos, line 102: maybe also add a test for a 304 on a `GET` with matching `If-None-Modified`?

drjova (approved):
👏

Issue #1419, opened 2020-10-12
Disambiguation II - exact match authors from paper


pazembrz:
Take a record id from disambiguation queue and match every author (unmatched - without $ref). For each of those, search its IDs (similar to the exact match in the workflow).
If there will be exact match then store $ref to this author in literature record.

*TECH*:

* Create a celery task which will be triggered on [record commit](https://github.com/inspirehep/inspirehep/blob/master/backend/inspirehep/records/receivers.py#L18)
* For each authors, search its IDs and emails in author records (similar to the exact match in the workflow). If there's only one result, assign the signature to that author by adding `record.$ref`

michamos:
I'm reading this only now and wasn't there at the refinement, but what's the reason for not simply using celery tasks for each record like we do for indexing?

drjova:
The problem with the celery task is when to trigger it, we're indexing the record and adding a task for disambiguation (which will add  `$ref`) will mean that we might have a race condition and we end up indexing an earlier version. Adding to a queue, we make sure we're processing the records independently. I am happy to discuss this further, since this is not the final refinement.

PR #1471, opened 2020-10-23
disambiguation: add task to assign authors reference

MJedr:
Ref: #1419


backend/inspirehep/disambiguation/tasks.py
```diff
@@ -64,3 +67,21 @@ def disambiguate_signatures(self, clusters):
...
+
+
+@shared_task(ignore_result=False, bind=True)
+def disambiguate_authors(self, record_uuid):
+    record = LiteratureRecord.get_record(record_uuid)
+    if "Literature" not in record["_collections"]:
+        return
+    authors = record.get("authors", [])
+    updated_authors = []
+    for author in authors:
```
michamos, line 79: this is expensive for records with many authors, so I guess you want to restrict to the cases where relevant metadata has changed: either an author has been added/removed (check the `uuid`), or `ids` of an existing author have been changed. Any other modification of the record won't change the linked author (for this MVP, later we will add more metadata). You can do that by comparing with the previous version of the record, similarly to what we're doing for references: https://github.com/inspirehep/inspirehep/blob/7d97abcdd4f2b97d4b5a651d3c776a92d6b0533c/backend/inspirehep/records/api/literature.py#L174-L222

backend/inspirehep/disambiguation/tasks.py
```diff
@@ -64,3 +67,21 @@ def disambiguate_signatures(self, clusters):
...
+        authors_found = AuthorsSearch().disambiguate_author(author).execute().hits
```
michamos, line 80: I think the idea was to use `inspire-matcher` here, as it should make things easier in the future when more requirements are added. Not sure though, might be worth getting other opinions.

backend/inspirehep/search/api.py
```diff
@@ -324,6 +324,27 @@ def get_author_papers(author, source=None, size=10000):
...
+        emails_queries = [
+            (Q("match", email_addresses__value=email)) for email in emails
+        ]
+        id_queries = [
+            Q(
+                "bool",
+                must=[
+                    Q("match", ids__value=id_value.get("value")),
+                    Q("match", ids__schema=id_value.get("schema")),
+                ],
```
michamos, line 339: this is incorrect, because `value` and `schema` don't necessarily go together as the field is not `nested`. That's why we use a validator for matching records: https://github.com/inspirehep/inspire-matcher/blob/0ca79e7d758dea2f6ee273dcbf7b9fa15f3c4dbb/inspire_matcher/config.py#L52 (one extra reason to use `inspire-matcher`).

backend/inspirehep/records/api/literature.py
```diff
@@ -171,6 +172,36 @@ def update(self, data, disable_orcid_push=False, *args, **kwargs):
...
+                changed_authors.append(author)
+        author_ids_latest = set(
+            chain.from_iterable(get_value(self, "authors.ids.value", []))
+        )
+        author_ids_oldest = set(
+            chain.from_iterable(
+                get_value(self._previous_version, "authors.ids.value", [])
+            )
+        )
+        ids_changed = set.symmetric_difference(author_ids_latest, author_ids_oldest)
```
michamos, line 194: the general idea is the same as for references, but I think you're copy/pasting too much here. From a high-level perspective, you want to find:
* new UUIDs (you don't really care about deletions as they're gone, so there's nothing to do for them)
* UUIDs for which the metadata has been changed (as an optimization you could check the relevant metadata only, but that's error prone, as can be seen here where you forgot emails, so let's not do that right now).

So it seems like you need some way to efficiently obtain an author metadata by UUID (at least for one of the copies, for the other one you can simply iterate on `authors`), that is a have `dict` mapping `uuid` -> author `dict`, let's call this `previous_authors_by_uuid`. You could then iterate on the current authors and for each of them:
* check if it exists in `previous_authors_by_uuid` or `previous_authors_by_uuid[uuid]` is different from the current one
* if so, `yield` it for disambiguation (it's better to use generators for this kind of stuff)

backend/inspirehep/records/receivers.py
```diff
@@ -34,3 +35,4 @@ def index_after_commit(sender, changes):
                 InspireRecord(model_instance.json, model=model_instance).index(
                     force_delete=force_delete
                 )
+                disambiguate_authors.delay(str(model_instance.id))
```
michamos, line 38: I think this needs to be behind a feature flag, as it shouldn't run on prod yet.

backend/tests/integration-async/migrator/test_migrator_tasks.py
```diff
@@ -662,3 +664,151 @@ def test_update_relations_with_modified_experiments(
...
+    author_data.update(
+        {
+            "name": {"value": "Brian Gross"},
+            "ids": [{"schema": "INSPIRE BAI", "value": "J.M.Maldacena.1"},],
+            "email_addresses": [{"current": True, "value": "test@test.com"}],
+        }
+    )
+    author_record = InspireRecord.create(author_data)
+    db.session.commit()
+    time.sleep(5)
```
michamos, line 682: you shouldn't use a `sleep` here, but look at @harunurhan's way (I don't remember the details now, ask him if you don't find it).

backend/inspirehep/matcher/config.py
```diff
@@ -271,3 +271,20 @@
...
+                {
+                    "path": "emails",
+                    "search_path": "email_addresses.value",
+                    "type": "exact",
+                },
+            ]
+        }
+    ],
+    "index": "records-authors",
+    "source": ["control_number"],
```
michamos, line 289: you should also add a validator, to make sure you're not matching an ID with one `schema` with an ID with a different `schema`, same idea as in https://github.com/inspirehep/inspire-matcher/blob/0ca79e7d758dea2f6ee273dcbf7b9fa15f3c4dbb/inspire_matcher/validators.py#L78

michamos:
also I don't remember if it's part of the story, but you shouldn't touch authors that have a `curated_relation=True` (as those have been claimed, so we don't want to reassign them automatically).

MJedr:
It wasn't but I think it's a good point

backend/inspirehep/matcher/config.py
MJedr: Wouldn't it be too expensive? We'll have to query pidstore every time to check the record type

michamos: no, it's part of the record (`ids` contains `schema` and `value`).

backend/inspirehep/matcher/validators.py
```diff
@@ -0,0 +1,33 @@
...
+    identifiers = [
+        identifier.get("schema") in allowed_id_schema_values
+        for identifier in record_author_identifier
+    ]
+    result_authors_identifier = [
+        identifier in allowed_id_schema_values
+        for identifier in get_value(result, "ids.schema")
+    ]
+
+    return all(identifiers) and all(result_authors_identifier)
```
michamos, line 33: no, this is not the issue, let's discuss it.

backend/inspirehep/records/api/literature.py
```diff
@@ -171,6 +172,26 @@ def update(self, data, disable_orcid_push=False, *args, **kwargs):
...
+
+        for author_uuid in current_authors_by_uuid.keys():
+            if not previous_authors_by_uuid.get(author_uuid):
+                yield current_authors_by_uuid[author_uuid]
+            else:
+                if DeepDiff(
+                    current_authors_by_uuid[author_uuid],
+                    previous_authors_by_uuid[author_uuid],
+                    ignore_order=True,
+                ):
```
michamos, line 192: why the `DeepDiff`? why can't you just do `if current_author != previous_author`?

backend/inspirehep/records/api/literature.py
```diff
@@ -171,6 +172,26 @@ def update(self, data, disable_orcid_push=False, *args, **kwargs):
...
+    def get_modified_authors(self):
+        previous_authors = get_value(self._previous_version, "authors", [])
+        previous_authors_by_uuid = {
+            author["uuid"]: author for author in previous_authors
+        }
+
+        current_authors = get_value(self, "authors", [])
+        current_authors_by_uuid = {author["uuid"]: author for author in current_authors}
+
+        for author_uuid in current_authors_by_uuid.keys():
```
michamos, line 184: can't you just do
```suggestion
        current_authors = get_value(self, "authors", [])
        for current_author in current_authors:
            author_uuid = current_author["uuid"]
```

backend/tests/integration-async/conftest.py
```diff
@@ -31,6 +31,7 @@ def app():
     app_config["CELERY_TASK_ALWAYS_EAGER"] = False
     app_config["CELERY_TASK_EAGER_PROPAGATES"] = False
     app_config["TESTING"] = True
+    app_config["FEATURE_FLAG_ENABLE_AUTHOR_DISAMBIGUATION"] = True
```
michamos, line 34: I think it's better to do it in the testcases, otherwise it's hard to follow. Please also test the feature flag itself (nothing happens when feature is disabled).

backend/inspirehep/disambiguation/tasks.py
```diff
@@ -64,3 +68,35 @@ def disambiguate_signatures(self, clusters):
...
+            if len(set(matched_authors)) == 1:
+                author_control_number = matched_authors[0]
+                author["record"] = get_ref_from_pid("aut", author_control_number)
+                updated_authors.append(str(author_control_number))
+    if updated_authors:
+        LOGGER.info(
+            f"Updated references for authors",
+            {
+                "record_uuid": record.id,
+                "record_control_number": record["control_number"],
```
michamos, line 97: can you check which names we use in other places, so it's consistent?

backend/tests/integration-async/conftest.py
MJedr: But then how we can make sure that it is no breaking for some reason other tests?

michamos: I don't think I understand the question. If the feature flag is false in the `config.py`, you should explicitly opt-in to enable the feature flags in the tests where you need to test behavior of the feature. Then it's clear what depends on the feature flag (you can look at the other tests where we need to change some feature flags).

backend/tests/integration/search/test_api.py
```diff
@@ -901,7 +901,7 @@ def test_journal_title_normalization_without_match(inspire_app):
 
 
 def test_public_api_generates_correct_links_in_literature_search_with_fields(
-    inspire_app
+    inspire_app,
```
drjova, line 904: are these changes because of your editor settings?

backend/inspirehep/matcher/validators.py
```diff
@@ -0,0 +1,16 @@
+from inspire_utils.record import get_value
+
+
+def authors_validator(author, result):
+    record_author_identifiers = author.get("ids")
```
drjova, line 5:
```suggestion
    record_author_identifiers = author.get("ids", [])
```
otherwise it will fail in the for loop

backend/inspirehep/matcher/validators.py
```diff
@@ -0,0 +1,16 @@
+from inspire_utils.record import get_value
```
drjova, line 1: we should add the license header in the new files :)

backend/tests/integration-async/migrator/test_migrator_tasks.py
```diff
@@ -662,3 +661,219 @@ def test_update_relations_with_modified_experiments(
     ).one()
 
     assert experiment_literature_relation.literature_uuid == record.id
+
```
drjova, line 664: you could add app_config fixture and add `config["FEATURE_FLAG_ENABLE_AUTHOR_DISAMBIGUATION"] = True`

backend/inspirehep/records/api/literature.py
```diff
@@ -14,6 +14,7 @@
 import magic
 import requests
 import structlog
+from deepdiff import DeepDiff
```
drjova, line 17: I believe is not used

backend/inspirehep/records/api/literature.py
```diff
@@ -171,6 +172,23 @@ def update(self, data, disable_orcid_push=False, *args, **kwargs):
...
             self.push_authors_phonetic_blocks_to_redis()
 
+    def get_modified_authors(self):
+        previous_authors = get_value(self._previous_version, "authors", [])
+        previous_authors_by_uuid = {
+            author["uuid"]: author for author in previous_authors
+        }
+
+        current_authors = get_value(self, "authors", [])
+        current_authors_by_uuid = {author["uuid"]: author for author in current_authors}
```
michamos, line 182: you're not using these any longer

backend/inspirehep/matcher/validators.py
```diff
@@ -0,0 +1,16 @@
...
+def authors_validator(author, result):
+    record_author_identifiers = author.get("ids")
+    result_author_identifiers = get_value(result, "ids")
+    for author_id in record_author_identifiers:
+        result_identifiers = {
+            author_identifier["value"]
+            for author_identifier in result_author_identifiers
+            if author_identifier["schema"] == author_id["schema"]
+        }
+        author_identifiers = {author_id["value"]}
```
michamos, line 13: this is works but is a bit clumsy. A better way would be to build two `set`s of `(schema, value)`s, then check that the set intersection is not empty.

backend/tests/integration-async/migrator/test_migrator_tasks.py
```diff
@@ -662,3 +661,219 @@ def test_update_relations_with_modified_experiments(
...
+    def assert_disambiguation_task():
+        literature_record_from_es = InspireSearch.get_record_data_from_es(
+            literature_record
+        )
+        assert (
+            str(author_data["control_number"])
+            in literature_record_from_es["authors"][0]["record"]["$ref"]
+        )
+
+    retry_until_pass(assert_disambiguation_task, retry_interval=5)
```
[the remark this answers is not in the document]
MJedr: The default is 0.3 and the tests are flaky :/ I can reduce it.

backend/inspirehep/disambiguation/tasks.py
```diff
@@ -64,3 +68,35 @@ def disambiguate_signatures(self, clusters):
...
+
+@shared_task(ignore_result=False, bind=True)
+def disambiguate_authors(self, record_uuid):
+    record = InspireRecord.get_record(record_uuid)
+    if "Literature" not in record["_collections"]:
+        return
+    authors = list(record.get_modified_authors())
+    updated_authors = []
+    for author in authors:
+        if not author.get("curated_relation"):
```
michamos, line 81: minor: it's almost always better to write this kind of stuff as
```suggestion
        if author.get("curated_relation"):
            continue
        # do the matching
```
to avoid having to indent so much, and make it less error-prone when extending the function.

backend/tests/integration/search/test_api.py
MJedr: It's pre-commit, but I've just fixed it ;)

backend/inspirehep/records/api/literature.py
```diff
@@ -171,6 +171,22 @@ def update(self, data, disable_orcid_push=False, *args, **kwargs):
                 push_to_orcid(self)
             self.push_authors_phonetic_blocks_to_redis()
 
+    def get_modified_authors(self):
+        previous_authors = get_value(self._previous_version, "authors", [])
```
michamos, line 175: if you're not using any "advanced" features of `get_value` (there is no `.` or `[...]` in the second argument), better use the `get` method which is faster.

backend/inspirehep/records/api/literature.py
```diff
@@ -171,6 +171,22 @@ def update(self, data, disable_orcid_push=False, *args, **kwargs):
...
+        previous_authors_by_uuid = {
+            author["uuid"]: author for author in previous_authors
+        }
+
+        current_authors = get_value(self, "authors", [])
```
michamos, line 180: same

backend/inspirehep/records/api/literature.py
```diff
@@ -171,6 +171,22 @@ def update(self, data, disable_orcid_push=False, *args, **kwargs):
...
+
+        for author in current_authors:
+            author_uuid = author["uuid"]
+            if not previous_authors_by_uuid.get(author_uuid):
```
michamos, line 184:
```suggestion
            if author_uuid not in previous_authors_by_uuid:
```

backend/inspirehep/records/api/literature.py
```diff
@@ -171,6 +171,22 @@ def update(self, data, disable_orcid_push=False, *args, **kwargs):
...
+                yield author
+                continue
+            if author != previous_authors_by_uuid.get(author_uuid):
```
michamos, line 187:
```suggestion
            if author != previous_authors_by_uuid[author_uuid]:
```

backend/inspirehep/disambiguation/tasks.py
```diff
@@ -64,3 +68,36 @@ def disambiguate_signatures(self, clusters):
             disambiguation_assigned_clusters.labels("2+").inc()
             LOGGER.debug("Received cluster with more than 1 author.")
     db.session.commit()
+
+
+@shared_task(ignore_result=False, bind=True)
+def disambiguate_authors(self, record_uuid):
```
michamos, line 74: if there are changes to the linked authors in this task, then the record will be modified, so the task will fire again (and hopefully there will be no changes the second time around). Maybe we want to prevent this second task from being triggered somehow (this might be also be done in a future story, as it's not part of this one)? @drjova

drjova: good point! Yes, let's have this merged so we can plan for the next steps

backend/inspirehep/disambiguation/tasks.py
```diff
@@ -64,3 +68,36 @@ def disambiguate_signatures(self, clusters):
...
+    record = InspireRecord.get_record(record_uuid)
+    if "Literature" not in record["_collections"]:
+        return
+    authors = list(record.get_modified_authors())
```
michamos, line 78: The `list` is not needed, as the `for` loop can iterate on any iterable, which includes generators. It might be needed if you want to reuse it later (because iterating on a generator consumes it), but that doesn't seem to be the case here.
```suggestion
    authors = record.get_modified_authors()
```

backend/tests/integration/records/api/test_api_literature.py
```diff
@@ -2263,3 +2263,67 @@ def test_arxiv_url_also_supports_format_alias(inspire_app):
...
+                    "ids": [{"schema": "INSPIRE BAI", "value": "D.Matthews.1"}],
+                    "emails": ["test1@test.pl", "test1.1@test.pl"],
+                    "uuid": "ec485f53-0d75-403a-a11c-e2a6d45dd327",
+                },
+            ]
+        }
+    )
+
+    record.update(data)
+    assert len(list(record.get_modified_authors())) == 2
```
michamos, line 2329: there are two cases causing an author to count as modified:
1. new uuid
2. changed metadata

I think you're only testing 1. here (with very high probability, because UUIDs are random). If you want to test 2. (which would be useful), you need to mutate an author without touching its UUID.

michamos (approved):
a couple of small comments, but looks good overall :+1:

drjova (approved):
Agree with Micha's comments

Issue #1460, opened 2020-10-21
OAI Server should use the `SEARCH_INDEX_PREFIX`


drjova:
OAI server, https://github.com/inspirehep/invenio-oaiserver/blob/master/invenio_oaiserver/config.py#L27 uses the index directly, instead it should use `SEARCH_INDEX_PREFIX`

Acceptance:
if SEARCH_INDEX_PREFIX="preifx-abc-" given in the configuration the index is record-literature, matcher should search in preifx-abc-record-literature.

**TECH**
In [here](https://github.com/inspirehep/invenio-oaiserver/blob/5d796cf354467773a7e00b75bc382ba2bf2efef7/invenio_oaiserver/query.py#L72), [here](https://github.com/inspirehep/invenio-oaiserver/blob/5d796cf354467773a7e00b75bc382ba2bf2efef7/invenio_oaiserver/query.py#L88) and [here](https://github.com/inspirehep/invenio-oaiserver/blob/5d796cf354467773a7e00b75bc382ba2bf2efef7/invenio_oaiserver/response.py#L123) pass the index wrapped in invenio_search.utils `prefix_index` function.

PR #1488, opened 2020-10-28
setup: bump oaiserver

MJedr:
Ref:#1460


drjova:
@MJedr it has a conflict :/

drjova (approved):

PR #1491, opened 2020-10-28
add fixture to override config in app and api

MJedr:


MJedr:
[Here](https://github.com/inspirehep/inspirehep/blob/a025880bd23b6afded6d17530fba77166977a60e/backend/tests/integration-async/oai/test_views.py#L41) as expected `SEARCH_INDEX_PREFIX` will be `test-` but [here](https://github.com/inspirehep/invenio-oaiserver/blob/f96fa977eb6c443a4dd1943895d864a0735c370b/invenio_oaiserver/query.py#L89) it will be None, as is set in default config.

backend/tests/integration/conftest.py
```diff
@@ -159,6 +160,28 @@ def inspire_app(base_app, db, es_clear, vcr_config):
...
+
+            with override_config(
+                MY_FEATURE_FLAG_ACTIVE=True,
+                MY_USERNAME='username',
+            ):
+                ...
+        """
+        with mock.patch.dict(inspire_app.config, kwargs):
+            yield mock.patch.dict(inspire_app.wsgi_app.mounts["/api"].config, kwargs)
+
```
drjova, line 181:
```suggestion
        with mock.patch.dict(inspire_app.config, kwargs), mock.patch.dict(inspire_app.wsgi_app.mounts["/api"].config, kwargs):
            yield 

```

drjova (approved):
LGTM just the comment! 👍

pazembrz:
override_config was a fixture before, then team wanted to have only one fixture for setting-up tests so it was changed and we are making it fixture again? I'm lost here ;)

drjova:
@pazembrz there is a deeper problem with the `override_config`, because we have two flask apps (`app` and `api_app`) `override_config` doesn't work in all cases because it's overriding just one of the two apps.

pazembrz (approved):

Issue #1303, opened 2020-09-11
handle manual merge records on hep


drjova:
On record update we should check if the `deleted_records` field has changed, if yes we should handle the case as follows:

* Get the `pid_value` from `deleted_record.$ref`, delete the record and mark its pid_value as `REDIRECTED` to the `pid_value` of the record we are updating.

We don't want to use invenio redirect as it's not working with deleted records. We also don't want to manually reassign pids as after reassignment we still want to have access to old record by its pid (somehow).
There are also additional requirements, all of them are mentioned in zulip thread: https://inspirehep.zulipchat.com/#narrow/stream/122202-devel/topic/handle.20manual.20merge/near/211614611

We probably have to create our own redirection system which will be aware of old (redirected) records, it should also automatically redirect all pids which were redirected by record. For example:
We have records:
Rec A: {'control_number':1234}
Rec B: {control_number': 2345, 'deleted_records':[{'$ref':'.../literature/1234'}]}
Rec C:{control_number': 3456, 'deleted_records':[{'$ref':'.../literature/2345'}]}

Record C should automatically redirect also pid 1234 to itself as record B which is redirecting record A is now redirected to record C. 
Still we should be access Record A and B by their pids (while they are still redirected to record C)

We also shouldn't allow to redirect redirected pid to another record:
Rec A: {'control_number':1234}
Rec B: {control_number': 2345, 'deleted_records':[{'$ref':'.../literature/1234'}]}
Rec C:{control_number': 3456, 'deleted_records':[{'$ref':'.../literature/1234'}]}
This is not allowed.

Looks like old record should also have an info which record replaces it like it was done on NEXT: https://github.com/inspirehep/inspire-next/blob/d237fe351bdfc2d04f5276a1d10e4198cfdd3165/inspirehep/modules/workflows/tasks/manual_merging.py#L181

pazembrz:
https://github.com/pazembrz/inspirehep/tree/%231378_handle_manual_merge Here is my almost ready implementation with reassigning pids approach which we found out is not what we want.

pazembrz:
Current:

`Records`:
[get_uuid_from_pid_value](https://github.com/inspirehep/inspirehep/blob/4db5dfce77feee19850a0d839ebcb8268b7013ef/backend/inspirehep/records/api/base.py#L62-L67)
[...]

PR #1506, opened 2020-11-05
pidstore: add support for PID redirection

pazembrz:
pidstore: Change way redirection is done by re-implementing redirection
table so it will allow to acces original record after redirection
editor: Allow to edit original record after it was redirected
api: Add flag `original_record`in `InspireRecord.get_record_by_pid_value` which when set to `True` queries for original record not for redirected one.
data: Fixed records data in data folder
docker-compose: small fix for docker-compose in local env

* Table migration is required!
* After enabling FEATURE FLAG data re-migration is required!

#1303


backend/inspirehep/orcid/utils.py
```diff
@@ -160,15 +161,20 @@ def get_literature_recids_for_orcid(orcid):
...
             PersistentIdentifier.pid_type == "aut",
         )
         .one()
-        .pid_value
+    )
+
+    author_recid = (
+        author_record.pid_value
+        if not author_record.is_redirected()
+        else InspireRedirect.get_redirect(author_record).pid_value
```
michamos, line 177: so the approach here is to follow redirects rather than rewrite recids in records?

backend/inspirehep/pidstore/models.py
```diff
@@ -0,0 +1,130 @@
...
+        LOGGER.info("PID redirected successfully", old_pid=old_pid, new_pid=new_pid)
+        return True
+
+    @classmethod
+    def get_redirect(cls, pid):
+        while pid.status == PIDStatus.REDIRECTED:
+            pid = cls.get(
+                original_pid_type=pid.pid_type, original_pid_value=pid.pid_value
+            ).new_pid
+        return pid
```
michamos, line 113: Good you thought of chains of redirects! A more performant approach would be to deal with this at the DB level with a recursive CTE to avoid the N+1 query problem here, but that's much more complex, and long chains of redirects should be extremely rare, so it's probably not worth it in practice.

backend/inspirehep/pidstore/models.py
```diff
@@ -0,0 +1,130 @@
...
+
+    original_pid_type = db.Column(db.String(6), nullable=False, primary_key=True)
+    original_pid_value = db.Column(db.String(255), nullable=False, primary_key=True)
+
+    new_pid_type = db.Column(db.String(6), nullable=False)
+    new_pid_value = db.Column(db.String(255), nullable=False)
+
+    @property
+    def new_pid(self):
+        return PersistentIdentifier.get(self.new_pid_type, self.new_pid_value)
```
michamos, line 42: if we want to be able to use `InspireRedirect.new_pid` inside filters (to do stuff like `filter(InspireRedirect.new_pid == PersistentIdentifier), this should be a [`hybrid_property`](https://docs.sqlalchemy.org/en/13/orm/extensions/hybrid.html), but we probably don't need it.

backend/inspirehep/pidstore/models.py
```diff
@@ -0,0 +1,130 @@
...
+        db.Index(
+            "ix_inspire_pidstore_redirect_new_record", "new_pid_type", "new_pid_value"
+        ),
+    )
+
+    original_pid_type = db.Column(db.String(6), nullable=False, primary_key=True)
+    original_pid_value = db.Column(db.String(255), nullable=False, primary_key=True)
+
+    new_pid_type = db.Column(db.String(6), nullable=False)
+    new_pid_value = db.Column(db.String(255), nullable=False)
```
michamos, line 38: why not make these foreign keys on the pidstore instead? you'd always need to join on the pidstore to get the pid metadata, but you need to do it anyway if you want to get the `object_uuid`.

backend/inspirehep/pidstore/models.py
```diff
@@ -0,0 +1,130 @@
...
+            raise WrongRedirectionPidStatus(old_pid, new_pid)
+
+        current_redirection = cls.query.filter_by(
+            original_pid_type=old_pid.pid_type, original_pid_value=old_pid.pid_value
+        ).one_or_none()
+
+        if current_redirection:
+            if old_pid.status != PIDStatus.REDIRECTED:
+                LOGGER.warning(
+                    "There is redirection for PID while status is not correct. Trying to fix.",
```
michamos, line 64: should this ever happen? If not, I think it's ok to raise an exception here, you shouldn't try to fix stuff "silently" that is not expected to be broken.

backend/inspirehep/pidstore/models.py
```diff
@@ -0,0 +1,130 @@
...
+                    db.session.add(old_pid)
+            except IntegrityError as e:
+                raise PIDInvalidAction(e)
+            except SQLAlchemyError as e:
+                LOGGER.exception(
+                    "Failed to redirect record", old_pid=old_pid, new_pid=new_pid
+                )
+                return False
+        LOGGER.info("PID redirected successfully", old_pid=old_pid, new_pid=new_pid)
+        return True
```
michamos, line 105: wouldn't it be more useful to return the `InspireRedirect` object you just created/updated?

backend/inspirehep/pidstore/models.py
```diff
@@ -0,0 +1,130 @@
...
+                        original_pid_value=old_pid.pid_value,
+                        new_pid_type=new_pid.pid_type,
+                        new_pid_value=new_pid.pid_value,
+                    )
+                    db.session.add(redirect)
+                    old_pid.status = PIDStatus.REDIRECTED
+                    db.session.add(old_pid)
+            except IntegrityError as e:
+                raise PIDInvalidAction(e)
+            except SQLAlchemyError as e:
```
michamos, line 99: I don't think you should handle this here, as it will cause issues because you'll have inconsistent behavior (redirect not created, but the rest of the transaction still committed).

backend/inspirehep/records/api/base.py
```diff
@@ -60,17 +65,41 @@ def validate(self):
...
+            original_record(bool): if record was redirected this flag decides what this method should return
+              redirected record uuid (if False) or original record uuid (if True)
+
+        Returns: requested record uuid
+
+        """
         if not pid_type:
             pid_type = cls.pid_type
         pid = PersistentIdentifier.get(pid_type, pid_value)
+        while pid.is_redirected() and not original_record:
```
michamos, line 83: do you need this `while` here? I think you already handle chains of redirects in the `get_redirect` method.

backend/inspirehep/records/api/base.py
```diff
@@ -60,17 +65,41 @@ def validate(self):
...
     @classmethod
-    def get_uuid_from_pid_value(cls, pid_value, pid_type=None):
+    def get_uuid_from_pid_value(cls, pid_value, pid_type=None, original_record=False):
+        """Get uuid for provided PID value
+
+        Args:
+            pid_value(str): pid value to query
+            pid_type(str): pid_type to query
+            original_record(bool): if record was redirected this flag decides what this method should return
+              redirected record uuid (if False) or original record uuid (if True)
```
michamos, line 75:
```suggestion
            original_record(bool): if record was redirected this flag determines whether this method should return
              redirected record uuid (if False) or original record uuid (if True)
```

backend/inspirehep/records/api/base.py
```diff
@@ -60,17 +65,41 @@ def validate(self):
...
     @classmethod
-    def get_record_by_pid_value(cls, pid_value, pid_type=None):
+    def get_record_by_pid_value(cls, pid_value, pid_type=None, original_record=False):
+        """Get record by provided PID value
+
+        Args:
+            pid_value(str): pid value to query
+            pid_type(str): pid_type to query
+            original_record(bool): if record was redirected this flag decides what this method should return
+              redirected record (if False) or original record (if True)
```
michamos, line 95:
```suggestion
            original_record(bool): if record was redirected this flag determines whether this method should return
              redirected record uuid (if False) or original record uuid (if True)
```

backend/inspirehep/records/api/base.py
```diff
@@ -159,6 +188,9 @@ def create(cls, data, id_=None, *args, **kwargs):
             kwargs.pop("disable_relations_update", None)
             data["self"] = get_ref_from_pid(cls.pid_type, data["control_number"])
             record = super().create(data, id_=id_, **kwargs)
+
+            if not record.get("deleted") and record.get("deleted_records"):
```
michamos, line 192: are you sure the first clause is needed? what happens if you set a record to `deleted` and simultaneously add `deleted_records`?

backend/inspirehep/records/api/base.py
```diff
@@ -477,3 +537,27 @@ def revert(self, revision_id):
 
         after_record_revert.send(current_app._get_current_object(), record=self)
         return self
+
+    @property
+    def control_number(self, pid=False):
```
michamos, line 542: what is the `pid` flag used for?

backend/inspirehep/records/api/base.py
```diff
@@ -477,3 +537,27 @@ def revert(self, revision_id):
...
+        return self.get("control_number")
+
+    @property
+    def control_number_pid(self):
+        return PersistentIdentifier.get(self.pid_type, self.control_number)
+
+    @property
+    def deleted_by(self):
```
michamos, line 550:
```suggestion
    def redirected_record_ref(self):
```
and reuse `redirected_pid` property?

backend/inspirehep/records/marshmallow/base.py
```diff
@@ -8,10 +8,15 @@
...
 from inspire_dojson.utils import strip_empty_values
+from invenio_pidstore.models import PersistentIdentifier, PIDStatus
 from invenio_records_rest.schemas.json import RecordSchemaJSONV1
 from marshmallow import fields, post_dump, pre_dump
 from marshmallow.schema import Schema
 
+from inspirehep.pidstore.api import PidStoreBase
+from inspirehep.pidstore.models import InspireRedirect
+from inspirehep.records.utils import get_ref_from_pid
+
```
michamos, line 19: why do you suddenly need all these imports?

backend/inspirehep/pidstore/models.py
```diff
@@ -0,0 +1,130 @@
...
+
+    new_pid_type = db.Column(db.String(6), nullable=False)
+    new_pid_value = db.Column(db.String(255), nullable=False)
+
+    @property
+    def new_pid(self):
+        return PersistentIdentifier.get(self.new_pid_type, self.new_pid_value)
+
+    @classmethod
+    def redirect(cls, old_pid, new_pid):
```
drjova, line 45: could we separate the logic, the function is too complex imo

backend/inspirehep/pidstore/models.py
```diff
@@ -0,0 +1,130 @@
...
+                original_pid_type=pid.pid_type, original_pid_value=pid.pid_value
+            ).new_pid
+        return pid
+
+    def delete(self):
+        pid = PersistentIdentifier.get(self.original_pid_type, self.original_pid_value)
+        pid.delete()
+        db.session.delete(self)
+
+        return True
```
drjova, line 120: Why we need the bool here? 🤔

backend/inspirehep/orcid/utils.py
pazembrz: Yes, this is safer with "our" datamodel, It simple and do not require to fix too much in orcid module. It will also work when we add rewriting of recids in record in there future (as this is the plan as I remember). Also it works almost instantly and do not require to wait for a task to rewrite data in all records so at the end with this small change we can be sure that we will get here always proper data for orcid at least.

backend/inspirehep/pidstore/models.py
pazembrz: That's exactly what I thought ;)

backend/inspirehep/pidstore/models.py
pazembrz: Yeah, I also thought about this but I didn't found an use-case for it. So I left it like it is. Changing it to hybrid-property is simple.

backend/inspirehep/pidstore/models.py
pazembrz: :+1:

backend/inspirehep/pidstore/models.py
pazembrz: :+1:

backend/inspirehep/pidstore/models.py
pazembrz: :+1:  I was trying to replicate original Invenio API here, but yeah, this was stupid ;)

backend/inspirehep/pidstore/models.py
pazembrz: Ah, yes. I should re-raise SQLAlchemyError my mistake.

backend/inspirehep/records/api/base.py
pazembrz: Yes this should be an if. It was left from previous iteration of the code :/

backend/inspirehep/records/api/base.py
pazembrz: :+1:

backend/inspirehep/records/api/base.py
pazembrz: :+1:

backend/inspirehep/records/api/base.py
pazembrz: Question is do we want to redirect to deleted record? Currently this wont' be possible. You can redirect to a record then delete it, but you cannot redirect to deleted record or to record you are just deleting or to a new record you are creating in deleted state already.

backend/inspirehep/records/api/base.py
pazembrz: Leftover from time when this was not a property ;) . Sorry.

backend/inspirehep/records/api/base.py
pazembrz: The idea here was that this returns None when there was no redirection so this is why I'm not reusing `redirected_pid` here.

pazembrz: Actually now I can just remove `redirected_pid` as it is not used anymore.

backend/inspirehep/records/marshmallow/base.py
pazembrz: This is weird... As I was not even edit this file

backend/inspirehep/alembic/8f72ce9b9c3b_add_inspire_redirection_table.py
```diff
@@ -0,0 +1,49 @@
...
+branch_labels = ()
+depends_on = None
+
+
+def upgrade():
+    """Upgrade database."""
+    op.create_table(
+        "inspire_pidstore_redirect",
+        sa.Column("created", sa.DateTime(), nullable=False),
+        sa.Column("updated", sa.DateTime(), nullable=False),
```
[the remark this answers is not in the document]
pazembrz: This is already covered in Timestamp model class from which I'm inheriting. I don't think it has to be set in alembic migration right? We never set it in any alembic migration before.

backend/inspirehep/pidstore/models.py
pazembrz: :+1:

backend/inspirehep/pidstore/models.py
pazembrz: Compatibility with Invenio API? But yeah, I agree. It's unnecessary. And now when I check code, it's not called by Invenio so I can remove it.

backend/inspirehep/pidstore/models.py
pazembrz: Actually now I added proper relations and I updated this model. It should be better more readable and more useful now.

backend/inspirehep/pidstore/models.py
```diff
@@ -0,0 +1,139 @@
...
+
+    @classmethod
+    def get(cls, original_pid, reload=False):
+        redirection = None
+        if reload:
+            redirection = cls.query.filter_by(original_pid=original_pid).one_or_none()
+        elif len(original_pid.redirection) == 1:
+            redirection = original_pid.redirection[0]
+        if not redirection:
+            raise PidRedirectionMissing(original_pid)
```
michamos, line 134: why not simply return `None`?

backend/inspirehep/pidstore/models.py
```diff
@@ -0,0 +1,139 @@
...
+    new_pid = db.relationship(
+        "PersistentIdentifier",
+        backref="redirected_pids",
+        foreign_keys=new_pid_id,
+        innerjoin=True,
+        uselist=False,
+    )
+
+    @classmethod
+    def _fix_existing_redirection(cls, redirection, old_pid, new_pid):
```
michamos, line 65:
```suggestion
    def _change_existing_redirection(cls, redirection, old_pid, new_pid):
```
I was confused, but IIUC, this is not fixing the redirection so much as changing which record it should redirect to.

backend/inspirehep/pidstore/models.py
```diff
@@ -0,0 +1,139 @@
...
+            pid = cls.get(pid).new_pid
+        return pid
+
+    @classmethod
+    def get(cls, original_pid, reload=False):
+        redirection = None
+        if reload:
+            redirection = cls.query.filter_by(original_pid=original_pid).one_or_none()
+        elif len(original_pid.redirection) == 1:
+            redirection = original_pid.redirection[0]
```
michamos, line 132: could you explain to me how these two cases are different? AFAIU, in the first case you do the query explicitly, in the second case, you use the ORM to do the query for you. (BTW, how come `redirection` is a list? you're setting `uselist=False` in the `backref`, and the unique constraint on the redirect table guarantees that there can be at most one redirection for a given pid).

backend/inspirehep/records/api/base.py
```diff
@@ -306,27 +334,60 @@ def commit(self, *args, **kwargs):
         This method does nothing, instead all the work is done in ``update``.
         """
 
+    def redirect_pid(self, pid_type, pid_value):
+        try:
+            old_pid = PersistentIdentifier.get(pid_type, pid_value)
+        except PIDDoesNotExistError:
+            LOGGER.warning(
```
michamos, line 341: why would this scenario happen? it should probably be a more serious thing than just a warning.

backend/tests/integration/pidstore/test_models.py
```diff
@@ -0,0 +1,107 @@
...
+    pid_2 = PersistentIdentifier.create(
+        pid_type="a", pid_value="2", status=PIDStatus.REGISTERED
+    )
+    pid_3 = PersistentIdentifier.create(
+        pid_type="a", pid_value="3", status=PIDStatus.REGISTERED
+    )
+
+    InspireRedirect.redirect(pid_1, pid_2)
+    InspireRedirect.redirect(pid_1, pid_3)
+    assert InspireRedirect.get_redirect(pid_1) == pid_3
```
michamos, line 107: maybe also add a test for chains of redirects?

backend/tests/integration/records/api/test_api_conferences.py
```diff
@@ -311,3 +311,19 @@ def test_number_of_contributions_query_after_hard_delete(inspire_app):
 
     expected_contributions_number = 0
     assert expected_contributions_number == conference.number_of_contributions
+
+
+def test_cn_redirection_works_for_conferences(inspire_app):
```
michamos, line 316: what is `cn` (appears in the test case many collections)?

backend/inspirehep/pidstore/models.py
pazembrz: It's done like that because Invenio would not know what to do with `None` when it's trying to resolve a PID.

backend/inspirehep/pidstore/models.py
pazembrz: :+1:

backend/inspirehep/pidstore/models.py
michamos: But what does it have to do with that? AFAICT, it's used only in your code, not in the pid provider.

backend/inspirehep/pidstore/models.py
pazembrz: You can actually check in `integration/pidstore/test_models.py` there was a situation when after removing redirection in session we are trying to get it from pidstore its reusing one from cache. So looks like if we modify redirection and do not commit yet (usually in tests, I cannot find use-case in running service) it's not doing again query but it's using cached value from pid.redirection.

For the second question, I'm not setting `uselist=False` on backref (`PersistantIdentifier` model), but on original_pid and new_pid in this model (InspireRedirect). To set it on PersistentIdentifier I would have to do this in PersistantIdentifier model :/ https://docs.sqlalchemy.org/en/13/orm/relationship_api.html#sqlalchemy.orm.relationship.params.uselist

backend/inspirehep/records/api/base.py
pazembrz: This will happen now for every record which redirects a record we do not have in our DB. As you said some time before we have records which are redirecting records which were deleted and are not migrated.
```
Micha Moskovic10:56
As I said, we have deleted records that don't migrate
```

backend/tests/integration/pidstore/test_models.py
pazembrz: Its there in test_api_base.py
`test_chain_of_redirection_properly_redirects`

backend/tests/integration/records/api/test_api_conferences.py
pazembrz: cn - control number

backend/inspirehep/pidstore/models.py
pazembrz: https://github.com/inspirehep/invenio-records-rest/blob/f398bab4f4f96607045e753d25aa09bb404e9e42/invenio_records_rest/utils.py#L154 is calling resolver which is throwing Exception if record is redirected with new pid as a parameter to this exception https://github.com/inspirehep/inspirehep/pull/1506/files#diff-bab823ce6da991443e3f70c6c8dd83b71bd9cf306f65979d2f4278c4234330e9R45 So, if get() would not throw exception when redirection is missing then I would have to manually trigger this exception in get_redirect or in resolve, otherwise Invenio would not know what to do with `None` pid.
Also it's expected usually that if method `get` didn't found an object then it's throwing some exception.

backend/inspirehep/records/api/base.py
michamos: :+1: I hadn't thought of this

backend/inspirehep/pidstore/models.py
michamos: But why does it matter if you get it from the session or from the DB again? that should be an implementation detail.

michamos (approved):

backend/inspirehep/pidstore/models.py
pazembrz: Problem is that it's not taking it from the session but from cache, so it's not updated.

pazembrz: I added a small update.

michamos (approved):
great job!

Issue #1124, opened 2020-07-16
Hide acquisition_source from the public API responses


drjova:
Authors API exposes the `acquisition_source`, we should hide it from the public api for all the collections, i.e. `application/json`.

We don't need this in the UI

### TECH NOTES:

- Add `acquisition_source` in `exclude to all collections Public Schema ([example](https://github.com/inspirehep/inspirehep/blob/master/backend/inspirehep/records/marshmallow/literature/base.py#L38))

PR #1512, opened 2020-11-05
serializers: remove acquisition_source from public serializers where it can be removed

pazembrz:
* #1124


backend/inspirehep/records/marshmallow/authors/base.py
```diff
@@ -22,14 +22,21 @@ class AuthorsRawSchema(RecordBaseSchema):
...
 
 class AuthorsPublicListSchema(AuthorsRawSchema):
     class Meta:
-        exclude = AuthorsPublicSchema.Meta.exclude + ["email_addresses"]
+        exclude = AuthorsPublicSchema.Meta.exclude + [
+            "email_addresses",
+            "acquisition_source",
+        ]
 
     acquisition_source = fields.Method("get_acquisition_source")
```
MJedr, line 41: Do we still need this (and the method below to remove emails from acquisition source)?

pazembrz: You are right. It's not needed anymore. But then why we are actually taking exclude parameter from AuthorsPublicSchema? Maybe we should just inherit from it...

MJedr: It was done because we wanted to hide emails in search but not in the detailed page. But now, since we hide the whole acquisition source I think that we can inherit from AuthorsPublicSchema.

pazembrz: Ah, ok I understand the problem here...

drjova:
@pazembrz are we sure on raw and admin we get the `acquisition_source `? Are tests for those?

drjova:
@pazembrz I would also update the commit message :)

pazembrz:
> @pazembrz are we sure on raw and admin we get the `acquisition_source `? Are tests for those?

Yes, there are tests (in this commit) which are checking `admin` serializers. For `raw` it's unecessary IMHO as `admin` is inheriting from `raw`.

pazembrz:
> @pazembrz I would also update the commit message :)

:+1:

drjova (approved):

drjova:
@pazembrz it needs rebase :)

drjova (approved):

PR #1516, opened 2020-11-05
mappings: change email_adresses in authors to keyword

MJedr:
* requires remapping and reindexing of the authors


michamos:
I think it requires remapping `authors`, not `literature`.

backend/tests/integration-async/migrator/test_migrator_tasks.py
```diff
@@ -875,3 +875,125 @@ def assert_disambiguation_task():
         assert not literature_record_from_es["authors"][0].get("record")
 
     retry_until_pass(assert_disambiguation_task)
+
+
+def test_disambiguate_authors_doesnt_match_when_author_is_ambiguous(
```
michamos, line 880: you already have this test higher in this file. Additionall, why is all of this in `test_migrator_tasks.py`? it has nothing to do with the `migrator` module.

MJedr: Actually, I've added the test because:
1. It is testing disambiguation after literature record update, which wasn't tested before.
2. I have added to the test case authors with emails in same domaine what wasn't working before the change in the mappings.
I've added them to `test_migrator_tasks.py` because disambiguation task is triggered by `index_after_commit` which was testet there. I can move these tests to disambiguation module if that would be better.

drjova: @MJedr yes please these tests should be part of disambiguation

pazembrz (approved):

PR #1523, opened 2020-11-09
pidstore: remove texkeys from pidstore which are missing from record metadata

pazembrz:


backend/inspirehep/pidstore/minters/texkey.py
```diff
@@ -59,8 +60,24 @@ def prepare_and_mint(cls, object_uuid, data):
...
         # Run it without pid provided to check if texkey should be re-generated
         minter.add_texkey(None)
+        # Delete from pidstore texkeys which are not in metadata
+        minter.purge_missing()
         return minter
 
+    def purge_missing(self):
+        texkeys = (
+            PersistentIdentifier.query.filter(
+                PersistentIdentifier.object_uuid == self.object_uuid
```
michamos, line 70:
```suggestion
                PersistentIdentifier.object_uuid == self.object_uuid,
                PersistentIdentifier.object_type == "rec",
```
to benefit from the index.

backend/tests/integration/pidstore/minters/test_minters_texkey.py
```diff
@@ -252,3 +252,25 @@ def test_minter_undeleting_record(inspire_app, override_config):
...
+        record = create_record("lit", data=data)
+    generated_texkeys = record["texkeys"]
+
+    assert len(generated_texkeys) == 1
+
+    data = dict(record)
+    del data["texkeys"]
+    with override_config(FEATURE_FLAG_ENABLE_TEXKEY_MINTER=True):
+        record.update(data)
+    assert record["texkeys"] != generated_texkeys
```
michamos, line 272: this might randomly fail (rarely) if the same texkey is assigned again. Maybe assign 2 texkeys at the beginning, then remove one and check that it is actually removed instead?

pazembrz: It will be easier to just do second update record with different year in publication_info, Then it will be impossible to generate identical texkey on this update.

backend/inspirehep/pidstore/minters/texkey.py
pazembrz: :+1:

michamos (approved):

Issue #1347, opened 2020-09-22
Update author on labs


StellaCh:
Currently, when we update an author, a bibedit holdingpen entry is created. This should be handled by labs.
Requirements are in:
https://docs.google.com/document/d/1OdeisaNXd1XSvKo_GCBxS0nIUE9gmryViwmANoMP1Bk/edit#

**Tech Notes**:
* Modify the author submission [views put](https://github.com/inspirehep/inspirehep/blob/d524f1fe523b61da6d3e47ea678eccd56b91a5a4/backend/inspirehep/submissions/views.py#L107). Currently it is calling next to start a workflow. It will need to look similar to Jobs put views.
    * Add a check to make sure the user has the right permissions to edit this author. (Similar to [can_user_edit](https://github.com/inspirehep/inspirehep/blob/d524f1fe523b61da6d3e47ea678eccd56b91a5a4/backend/inspirehep/accounts/api.py#L28)) (same orcid). Need to verify: should we compare current author orcid with ids->orcid or acquisition_source->orcid? We think it is ids->orcid
    * Then, just call record update and create an rt ticket as specified in the google doc
* We need to add a ```can_edit``` field to the ```AuthorsBaseSchema```. It should return true for catalogers or user with orcid same as in ```acquisition_source```. 
* In the authors ```DetailPageContainer```, render the ```EditRecordAction``` normally only if ```can_edit``` is true. Otherwise, we will need to have a popup - discuss with Stella how exactly it should look.
* Rename the ```Comments``` [field label](https://github.com/inspirehep/inspirehep/blob/d524f1fe523b61da6d3e47ea678eccd56b91a5a4/ui/src/submissions/authors/components/AuthorForm.jsx#L308) in the ```AuthorForm``` to ```Comments to the INSPIRE team```
* We need to create an ```AuthorSubmissionPage``` with the message from the google doc. The backend put endpoint will also need to be modified to send the author control_number. This should only be used for the [redirect](https://github.com/inspirehep/inspirehep/blob/d524f1fe523b61da6d3e47ea678eccd56b91a5a4/ui/src/submissions/index.jsx#L78) with  from={`${SUBMISSIONS_AUTHOR}/:id/success`} as it is only for updates.

* We need to check for fields like institutions, experiments and advisors as it is now allowed for them to be deleted. A similar check is done in ```get_updated_record_data```
[...]

PR #1527, opened 2020-11-10
authors: implement workflowless author update

harunurhan:
Ref: #1347


backend/inspirehep/records/marshmallow/authors/base.py
```diff
@@ -43,6 +45,9 @@ class Meta:
         exclude = FIELDS_TO_EXCLUDE
 
     email_addresses = fields.Raw(dump_only=True)
+    positions = fields.Raw(dump_only=True)
+    advisors = fields.Raw(dump_only=True)
+    project_membership = fields.Raw(dump_only=True)
```
drjova, line 50: Just to be sure, these are defined again because on the Raw schema we have `NonHiddenRaw` right?

e2e/cypress/integration/authors.test.js
```diff
@@ -129,6 +138,30 @@ describe('Author Submission', () => {
...
+      formData: {
+        display_name: ': Updated',
+      },
+      expectedMetadata: {
+        name: { preferred_name: 'John Richard Ellis: Updated' },
+      },
+    });
+  });
+
+  it.only('does not show update form if user is not the owner of the author record', () => {
```
drjova, line 157: why only?

ui/src/authors/containers/DetailPageContainer/DetailPageContainer.jsx
```diff
@@ -116,8 +119,14 @@ function DetailPage({
...
                     {urls && <AuthorWebsitesAction websites={urls} />}
-                    <EditRecordAction pidType="authors" pidValue={recordId} />
                     {orcid && orcid === userOrcid && <UserSettingsAction />}
+                    {canEdit ? (
+                      <EditRecordAction pidType="authors" pidValue={recordId} />
+                    ) : (
+                      <DisabledEditRecordAction
+                        message={CAN_NOT_AUTHOR_EDIT_MESSAGE}
+                      />
+                    )}
```
drjova, line 129: could we separate this in a component? we have it in other places too

ui/src/submissions/index.jsx
```diff
@@ -75,15 +75,15 @@ class Submissions extends Component {
...
-            <Redirect
-              exact
-              from={`${SUBMISSIONS_AUTHOR}/:id/success`}
-              to={SUBMISSION_SUCCESS}
-            />
             <Redirect
               exact
               from={`${SUBMISSIONS_AUTHOR}/new/success`}
               to={SUBMISSION_SUCCESS}
+            />u
```
drjova, line 82:
```suggestion
            />
```

michamos:
AFAIU, this completely modifies author updates not to use next any longer. So it's incompatible with sending updates to legacy like we're doing now. Is there a way to put this behind a feature flag to keep existing behavior for now?

backend/inspirehep/records/marshmallow/authors/base.py
harunurhan: yep!

e2e/cypress/integration/authors.test.js
harunurhan: leftover 😄

harunurhan:
@michamos as you can see.. code is all over, so we can't really put everything under feature flag.

I guess the easiest would be to have feature flag on `PUT /submissions/author/<id>` and keep rest of the logic such as users can only edit their profile, hidden fields in the form etc.

PR #1531, opened 2020-11-11
serializers: json raw use inspire envelope schema

drjova:


backend/tests/integration/records/serializers/json/test_literature.py
```diff
@@ -1214,9 +1214,14 @@ def test_literature_raw_json_with_logged_in_cataloger(inspire_app):
         response = client.get(f"/literature/{record_control_number}", headers=headers)
 
     response_status_code = response.status_code
-
+    result_keys = response.json().keys()
```
pazembrz, line 1217: json here is not a method but a property.

drjova: You are right, I mixed up `requests` with flask requests :)

pazembrz (approved):

Issue #1499, opened 2020-11-02
Disambiguation part II - Implement the rest of the cases


drjova:
Implement the 3-6 cases on disambiguation as described in https://github.com/inspirehep/inspirehep/issues/1390

**TECH**
1. Normalize name by using `ParsedName` from inspire_schemas. You’ll be able to get the first name, last name and initials from that.
2. Create the matcher config which will query literature authors and will use extracted first name and last name as a path and literature authors `full_name` as a search path. Match author using the config. If there’s only one match (one for first name and one for last name),  assign $ref to author.
3. Create another matcher config containing `first_name`, `last_name` and `initials`. Match author using this config. If there is only one match, assign $ref to him.
4. Create a matcher config, which will be the same as the previous one, but will use a validator for collaborations. If there is only one match, assign $ref to author.
5. Create same config as previous one but add the affiliations to the paths. Match authors using it and if there is only one match, assign $ref to author.
6. Create a new record for the author if no results were found in previous points. In `_private_notes` place info about the record from which the author comes from (TBD: ask Micha what should be in `_private_notes`).

Remember, that when searching in the Literature matcher will return all the authors, not only the matched one, so you’ll have to extract the desired author from the authors list

PR #1535, opened 2020-11-16
disambiguation: disambiguate author on name, collaboration, affiliation

MJedr:
Ref: #1499
* bumps inspire-matcher, inspire-utils and inspire-schemas

### DIFF
--- a/backend/inspirehep/disambiguation/tasks.py
+++ b/backend/inspirehep/disambiguation/tasks.py
@@ -2,6 +2,8 @@ import structlog
 from celery import shared_task
 from flask import current_app
 from inspire_matcher.api import match
+from inspire_utils.name import ParsedName
+from inspire_utils.record import get_value
 from invenio_db import db
 from prometheus_client import Counter
 
@@ -10,8 +12,11 @@ from inspirehep.disambiguation.utils import (
     link_signatures_to_author,
     update_author_names,
 )
+from inspirehep.matcher.validators import (
+    affiliations_validator,
+    collaboration_validator,
+)
 from inspirehep.records.api import InspireRecord
-from inspirehep.records.utils import get_ref_from_pid
 
 LOGGER = structlog.getLogger()
 
@@ -70,6 +75,109 @@ def disambiguate_signatures(self, clusters):
     db.session.commit()
 
 
+def match_literature_author_with_config(author_data, matcher_config):
+    matched_records = [
+        matched_record
+        for matched_record in match(author_data, matcher_config)
+        if matched_record.get("inner_hits")
+    ]
+    return matched_records
+
+
+def get_record_refs_for_matched_literature_authors(matched_records):
+    matched_authors_refs = set()
+    for record_authors in matched_records:
+        matched_authors_refs.update(
+            get_value(
+                record_authors, "inner_hits.authors.hits.hits._source.record.$ref", None
+            )
+        )
+    return matched_authors_refs
+
+
+def assign_reference_to_author_if_unambiguous_match(matched_authors_refs, author):
+    if matched_authors_refs and len(matched_authors_refs) == 1:
+        author_reference_url = matched_authors_refs.pop()
+        author["record"] = {"$ref": author_reference_url}
+        return author_reference_url.split("/")[-1]
+
+
+def assign_reference_to_author_if_unambiguous_literature_author_match(
+    matched_records, author
+):
+    matched_authors_refs = get_record_refs_for_matched_literature_authors(
+        matched_records
+    )
+    matched_ref = assign_reference_to_author_if_unambiguous_match(
+        matched_authors_refs, author
+    )
+    if matched_ref:
+        return matched_ref
+
+
+def match_literature_author(author, updated_authors, record):
+    configs = [
+        current_app.config["AUTHOR_MATCHER_NAME_CONFIG"],
+        current_app.config["AUTHOR_MATCHER_NAME_INITIALS_CONFIG"],
+    ]
+
+    validators = [(collaboration_validator, affiliations_validator), None]
+
+    parsed_name = ParsedName.loads(author.get("full_name"))
+    author_matcher_data = {
+        "first_name": parsed_name.first,
+        "last_name": parsed_name.last,
+        "full_name": author.get("full_name"),
+        "collaborations": get_value(record, "collaborations.value", []),
+        "affiliations": get_value(author, "affiliations.value", []),
+    }
+
+    for config, validator in zip(configs, validators):
+        matched_records = match_literature_author_with_config(
+            author_matcher_data, config
+        )
+        matched_reference = assign_reference_to_author_if_unambiguous_literature_author_match(
+            matched_records, author
+        )
+        if not matched_reference and validator:
+            for validator_function in validator:
+                valid_matches = (
+                    match
+                    for match in matched_records
+                    if validator_function(author_matcher_data, match)
+                )
+                matched_reference = assign_reference_to_author_if_unambiguous_literature_author_match(
+                    valid_matches, author
+                )
+                if matched_reference:
+                    break
+        if matched_reference:
+            updated_authors.append(matched_reference)
+            return author
+
+
+def create_new_author(full_name, from_recid):
+    new_author_data = {
+        "name": {"value": full_name},
+        "_private_notes": [
+            {
+                "source": "INSPIRE-disambiguation",
+                "value": f"Created from literature record {from_recid}",
+            }
+        ],
+    }
+
+    new_author = create_new_empty_author()
+    new_author.update({**new_author, **new_author_data})
+    LOGGER.info(
+        "Created new author record",
+        {
+            "control_number": str(new_author.get("control_number")),
+            "full_name": full_name,
+        },
+    )
+
+
 @shared_task(ignore_result=False, bind=True)
 def disambiguate_authors(self, record_uuid):
     record = InspireRecord.get_record(record_uuid)
@@ -77,19 +185,26 @@ def disambiguate_authors(self, record_uuid):
         return
     authors = record.get_modified_authors()
     updated_authors = []
+
     for author in authors:
         if author.get("curated_relation"):
             continue
-        matched_authors = [
-            matched_record["_source"]["control_number"]
+        matched_authors = {
+            matched_record["_source"]["self"]["$ref"]
             for matched_record in match(
                 author, current_app.config["AUTHOR_MATCHER_EXACT_CONFIG"]
             )
-        ]
-        if len(set(matched_authors)) == 1:
-            author_control_number = matched_authors[0]
-            author["record"] = get_ref_from_pid("aut", author_control_number)
-            updated_authors.append(str(author_control_number))
+        }
+        if len(matched_authors) == 1:
+            updated_author_recid = assign_reference_to_author_if_unambiguous_match(
+                matched_authors, author
+            )
+            updated_authors.append(updated_author_recid)
+        if "record" not in author:
+            match_literature_author(author, updated_authors, record)
+        if "record" not in author:
+            create_new_author(author["full_name"], record["control_number"])
+
     if updated_authors:
         LOGGER.info(
             f"Updated references for authors",
--- a/backend/inspirehep/matcher/config.py
+++ b/backend/inspirehep/matcher/config.py
@@ -4,6 +4,11 @@
 #
 # inspirehep is free software; you can redistribute it and/or modify it under
 # the terms of the MIT License; see LICENSE file for more details.
+from .validators import (
+    affiliations_validator,
+    authors_validator,
+    collaboration_validator,
+)
 
 GROBID_URL = "https://grobid.inspirebeta.net"
 
@@ -277,15 +282,55 @@ AUTHOR_MATCHER_EXACT_CONFIG = {
         {
             "queries": [
                 {"path": "ids.value", "search_path": "ids.value", "type": "exact"},
+            ],
+            "validator": authors_validator,
+        },
+        {
+            "queries": [
                 {
                     "path": "emails",
                     "search_path": "email_addresses.value",
                     "type": "exact",
                 },
-            ]
-        }
+            ],
+        },
     ],
     "index": "records-authors",
-    "source": ["control_number"],
+    "source": ["self.$ref"],
     "validator": "inspirehep.matcher.validators:authors_validator",
+}
+
+AUTHOR_MATCHER_NAME_CONFIG = {
+    "algorithm": [
+        {
+            "queries": [
+                {
+                    "paths": ["first_name", "last_name"],
+                    "search_paths": ["authors.first_name", "authors.last_name"],
+                    "type": "nested",
+                    "inner_hits": {
+                        "_source": ["authors.record.$ref", "authors.affiliations.value"]
+                    },
+                },
+            ],
+        },
+    ],
+    "index": "records-hep",
+    "source": ["control_number", "collaborations.value"],
+}
+
+
+AUTHOR_MATCHER_NAME_INITIALS_CONFIG = {
+    "algorithm": [
+        {
+            "queries": [
+                {
+                    "type": "author-names",
+                    "inner_hits": {"_source": ["authors.record"]},
+                },
+            ]
+        },
+    ],
+    "index": "records-hep",
+    "source": ["control_number", "authors.full_name", "authors.record.$ref"],
 }
--- a/backend/inspirehep/matcher/validators.py
+++ b/backend/inspirehep/matcher/validators.py
@@ -10,7 +10,7 @@ from inspire_utils.record import get_value
 
 def authors_validator(author, result):
     record_author_identifiers = author.get("ids", [])
-    result_author_identifiers = get_value(result, "ids")
+    result_author_identifiers = get_value(result, "ids", [])
     record_author_id_schema_value_pairs = {
         (record_author_id["schema"], record_author_id["value"])
         for record_author_id in record_author_identifiers
@@ -22,3 +22,20 @@ def authors_validator(author, result):
     return bool(
         record_author_id_schema_value_pairs & result_author_id_schema_value_pairs
     )
+
+
+def collaboration_validator(author, result):
+    return author.get("collaborations") == get_value(
+        result, "_source.collaborations.value", []
+    )
+
+
+def affiliations_validator(author, result):
+    authors_affiliations = set(author.get("affiliations", []))
+    result_authors_affiliations = get_value(
+        result, "inner_hits.authors.hits.hits._source.affiliations.value", []
+    )
+    for affiliation_list in result_authors_affiliations:
+        if set(affiliation_list) & authors_affiliations:
+            return True
+    return False
--- a/backend/poetry.lock
+++ b/backend/poetry.lock
@@ -163,10 +163,10 @@ description = "Tools to handle automatic semantic versioning in python"
 name = "autosemver"
 optional = false
 python-versions = "*"
-version = "0.5.3"
+version = "0.5.5"
 
 [package.dependencies]
-dulwich = "*"
+dulwich = ">=0.19.6,<0.20"
 
 [[package]]
 category = "dev"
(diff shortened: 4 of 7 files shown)


backend/inspirehep/disambiguation/tasks.py
```diff
@@ -70,13 +72,38 @@ def disambiguate_signatures(self, clusters):
     db.session.commit()
 
 
+def match_literature_author(author_data, matcher_config):
+    matched_records = [
+        matched_record["inner_hits"]["authors"]["hits"]["hits"]
+        for matched_record in match(author_data, matcher_config)
+    ]
+    matched_authors_refs = []
```
drjova, line 80: you could use `get_value()` https://github.com/inspirehep/inspire-utils/blob/bf94a7ad05fac5925b625bfc2491c54b4cfb5044/inspire_utils/record.py#L33

backend/inspirehep/disambiguation/tasks.py
```diff
@@ -90,6 +117,58 @@ def disambiguate_authors(self, record_uuid):
             author_control_number = matched_authors[0]
             author["record"] = get_ref_from_pid("aut", author_control_number)
             updated_authors.append(str(author_control_number))
+            continue
```
drjova, line 120: we should separate the logic in smaller functions, it became too long.

michamos: I fully agree. If you look at what this function does from a high-level perspective, it:
*  does some initialization
* loops on all authors in the record
* inside he loop, it does a series of matches, if there's only one result, it assigns it, if not it goes to the next one, if no match, creates a new author
* does some final stuff

You could 
1. split the loop body into a `disambiguate_literature_author` which takes an element of `authors` and mutates it to add the ref or do nothing, and also return the ref/recid of the matched author (so you could still do the `updated_authors` logic).
2. refactor this `disambiguate_literature_author` to make the logic more apparent. There's lots of repetition because you call `match_literature_author` with different configs. I'd change this to take a `list` of configs, and return the first match (look at the reference matcher, it's a very similar idea).
3. split the author record creation in a separate function.

backend/inspirehep/disambiguation/tasks.py
```diff
@@ -70,26 +72,98 @@ def disambiguate_signatures(self, clusters):
...
+    for config in configs:
+        matched_recid = match_literature_author(
+            author_matcher_data, current_app.config[config]
+        )
+        if matched_recid:
+            assign_reference_to_author(author, updated_authors, matched_recid)
+            return author
+
+
+def create_new_author(author, record):
```
drjova, line 123: Instead of creating yet another create author function, we could refactor https://github.com/inspirehep/inspirehep/blob/50a36d3029a96ce9a41852486c1d81a9b6e7c9e3/backend/inspirehep/disambiguation/utils.py#L75 to take `data` param and rename it to `create_author`.

backend/inspirehep/disambiguation/tasks.py
```diff
@@ -70,26 +72,98 @@ def disambiguate_signatures(self, clusters):
...
+    author["record"] = reference
+    updated_authors.append(reference["$ref"].split("/")[-1])
+
+
+def match_literature_author_with_configs(author, updated_authors, record):
+    configs = [
+        "AUTHOR_MATCHER_NAME_CONFIG",
+        "AUTHOR_MATCHER_NAME_INITIALS_CONFIG",
+        "AUTHOR_MATCHER_COLLABORATION_CONFIG",
+        "AUTHOR_MATCHER_AFFILIATION_CONFIG",
```
drjova, line 102: I would just `[current_app.config[], current_app.config[]]` directly, instead of a list which is used later

drjova:
@MJedr great work! 😃

backend/inspirehep/disambiguation/tasks.py
```diff
@@ -70,26 +72,97 @@ def disambiguate_signatures(self, clusters):
     db.session.commit()
 
 
+def match_literature_author(author_data, matcher_config):
+    matched_records = (
+        matched_record["inner_hits"]["authors"]["hits"]["hits"]
+        for matched_record in match(author_data, matcher_config)
+    )
+    matched_authors_refs = []
```
michamos, line 80: why not make this a `set` from the start?

backend/inspirehep/disambiguation/tasks.py
```diff
@@ -70,26 +72,97 @@ def disambiguate_signatures(self, clusters):
     db.session.commit()
 
 
+def match_literature_author(author_data, matcher_config):
```
michamos, line 75:
```suggestion
def match_literature_author_with_config(author_data, matcher_config):
```
as this one is taking a config...

backend/inspirehep/disambiguation/tasks.py
```diff
@@ -70,26 +72,97 @@ def disambiguate_signatures(self, clusters):
...
+        author_reference_url = matched_authors_refs[0]
+        return {"$ref": author_reference_url}
+
+
+def assign_reference_to_author(author, updated_authors, reference):
+    author["record"] = reference
+    updated_authors.append(reference["$ref"].split("/")[-1])
+
+
+def match_literature_author_with_configs(author, updated_authors, record):
```
michamos, line 97:
```suggestion
def match_literature_author(author, updated_authors, record):
```
... as this one does not :)

backend/inspirehep/disambiguation/tasks.py
```diff
@@ -70,26 +72,97 @@ def disambiguate_signatures(self, clusters):
...
+            if "record" in author["_source"]
+        )
+    if len(set(matched_authors_refs)) == 1:
+        author_reference_url = matched_authors_refs[0]
+        return {"$ref": author_reference_url}
+
+
+def assign_reference_to_author(author, updated_authors, reference):
+    author["record"] = reference
+    updated_authors.append(reference["$ref"].split("/")[-1])
```
michamos, line 94: too much mutation makes things hard to understand. I think it's better to
```suggestion
    return reference["$ref"].split("/")[-1]
```
and accumulate stuff outside. This also avoids having to pass `updated_authors` down to functions which don't really need it to do their job.

backend/inspirehep/disambiguation/tasks.py
```diff
@@ -70,26 +72,97 @@ def disambiguate_signatures(self, clusters):
...
             matched_record["_source"]["control_number"]
             for matched_record in match(
                 author, current_app.config["AUTHOR_MATCHER_EXACT_CONFIG"]
             )
-        ]
-        if len(set(matched_authors)) == 1:
-            author_control_number = matched_authors[0]
-            author["record"] = get_ref_from_pid("aut", author_control_number)
-            updated_authors.append(str(author_control_number))
+        }
```
michamos, line 137: any reason this couldn't be written as part of `match_literature_author`? it looks like it has the same structure as the other cases (if you retrieve `self` instead of `control_number` from the author record).

backend/inspirehep/disambiguation/tasks.py
```diff
@@ -70,26 +72,97 @@ def disambiguate_signatures(self, clusters):
...
+        if len(matched_authors) == 1:
+            author_control_number = matched_authors.pop()
+            assign_reference_to_author(
+                author,
+                updated_authors,
+                get_record_ref(author_control_number, endpoint="authors"),
+            )
+            continue
+        match_literature_author_with_configs(author, updated_authors, record)
+        if not author.get("record"):
```
michamos, line 147:
```suggestion
        if "record" not in author:
```

backend/inspirehep/disambiguation/tasks.py
```diff
@@ -70,26 +72,97 @@ def disambiguate_signatures(self, clusters):
...
+            new_author_data = {
+                "name": {"value": author["full_name"]},
+                "_private_notes": [
+                    {
+                        "source": "literature_record",
+                        "value": str(record.get("control_number")),
```
michamos, line 153:
```suggestion
                        "source": "INSPIRE-disambiguation",
                        "value": f"Created from literature record {record['control_number']}",
```

backend/inspirehep/disambiguation/tasks.py
```diff
@@ -70,26 +72,97 @@ def disambiguate_signatures(self, clusters):
...
+                    }
+                ],
+            }
+            new_author = create_new_empty_author(**new_author_data)
```
michamos, line 157: the name is a misnomer now. Plus I don't think you need to do this, you could do
```suggestion
            new_author = create_new_empty_author()
            new_author.update(new_author_data)
```

backend/inspirehep/disambiguation/tasks.py
MJedr: Yes, I wanted to do it in this way but I am not sure if it would be better. In this case I have to add two conditions in `match_literature_author_config`- first one because I don't get data from `inner_hits` but from `_source` and secondly because I will take reference from `self`, not `record` as in other cases.

michamos: ok, if it's very different let's keep it separate. Maybe split it out to a different function though to avoid different "levels" of logic in the function?

backend/inspirehep/records/api/literature.py
```diff
@@ -174,7 +174,7 @@ def update(self, data, disable_orcid_push=False, *args, **kwargs):
     def get_modified_authors(self):
         previous_authors = self._previous_version.get("authors", [])
         previous_authors_by_uuid = {
-            author["uuid"]: author for author in previous_authors
+            author["uuid"]: author for author in previous_authors if "uuid" in author
```
michamos, line 177: I don't think we should do that as it hides issues. We need to fix the underlying problem.

backend/inspirehep/disambiguation/tasks.py
```diff
@@ -70,26 +71,97 @@ def disambiguate_signatures(self, clusters):
...
+    for record_authors in matched_records:
+        matched_authors_refs.update(
+            author["_source"]["record"]["$ref"]
+            for author in record_authors
+            if "record" in author["_source"]
+        )
+    return matched_authors_refs
+
+
+def assign_reference_to_author(matched_authors_refs, author):
```
michamos, line 90:
```suggestion
def assign_reference_to_author_if_unambiguous_match(matched_authors_refs, author):
```

backend/inspirehep/disambiguation/tasks.py
```diff
@@ -70,26 +71,97 @@ def disambiguate_signatures(self, clusters):
...
+    if matched_authors_refs and len(matched_authors_refs) == 1:
```
michamos, line 91: are you ever passing `None`? otherwise this works
```suggestion
    if len(matched_authors_refs) == 1:
```

backend/inspirehep/disambiguation/tasks.py
```diff
@@ -70,26 +71,97 @@ def disambiguate_signatures(self, clusters):
...
                 author, current_app.config["AUTHOR_MATCHER_EXACT_CONFIG"]
             )
-        ]
-        if len(set(matched_authors)) == 1:
-            author_control_number = matched_authors[0]
-            author["record"] = get_ref_from_pid("aut", author_control_number)
-            updated_authors.append(str(author_control_number))
+        }
+        updated_author_recid = assign_reference_to_author(matched_authors, author)
+        updated_authors.append(updated_author_recid)
```
michamos, line 142: I still think it's not nice that you're mixing different levels of concern in this function (high-level: calling `match_literature_author`, low-level: the stuff you're doing here with calling `match` with an explicit config, getting the `$ref` out of it, etc.). So I think it's better to split this part into a `match_literature_author_ids` function.

backend/inspirehep/disambiguation/tasks.py
```diff
@@ -70,26 +71,97 @@ def disambiguate_signatures(self, clusters):
...
+            new_author = create_new_empty_author()
+            new_author.update({**new_author, **new_author_data})
+            LOGGER.info(
+                "Created new author record",
+                {
+                    "control_number": str(new_author.get("control_number")),
+                    "full_name": author.get("name"),
+                },
+            )
+
```
michamos, line 164: And also split this into a  `create_new_author(full_name, from_recid)`.

backend/inspirehep/matcher/config.py
```diff
@@ -277,15 +282,93 @@
...
+                {
+                    "paths": ["first_name", "last_name"],
+                    "search_paths": ["authors.first_name", "authors.last_name"],
+                    "type": "nested",
+                    "inner_hits": {
+                        "_source": ["authors.record", "authors.affiliations.value"]
+                    },
+                },
+            ],
+            "validator": affiliations_validator,
```
michamos, line 369: why do this with a validator, can't you pass the affiliations in the search?

MJedr: That will require another changes in matcher, because if I'll add affiliations to nested query it will pass affiliations as the value to the query (as an array) and that's wrong. If I'll pass affiliations to exact match I can't use `inner_hits`, so I'll have to go through all the authors in the result to find matched one.

michamos: why is it wrong to pass affiliations to the nested query?

MJedr: Currently what's happening is:
```
{
        'query': {
            'nested': {
                'path': common_path,
                'query': {
                    'bool': {
                        'must': [
                         {"match": {
                        "authors.affiliations.value": ["CERN", "Warsaw U.]
                        }}
                       ],
                    },
                },
            },
        },
    }
```
So that would fail because of the array as the value. We can modify matcher once again to add a match query for each value if an array is passed if we can't leave the validator here.

backend/inspirehep/disambiguation/tasks.py
```diff
@@ -70,26 +75,127 @@ def disambiguate_signatures(self, clusters):
     db.session.commit()
 
 
+def match_literature_author_with_config(author_data, matcher_config):
+    matched_records = [
+        matched_record
+        for matched_record in match(author_data, matcher_config)
+        if matched_record.get("inner_hits")
+    ]
```
MJedr, line 83: I did a list comprehension here and not generator because I reuse `matched_records` later with validators.

backend/inspirehep/disambiguation/tasks.py
```diff
@@ -70,26 +75,127 @@ def disambiguate_signatures(self, clusters):
...
+    for validator in validators:
+        valid_matches = (
+            match for match in matched_references if validator(author_data, match)
+        )
+        matched_valid_authors_references = match_literature_author_ids(valid_matches)
+        if matched_valid_authors_references:
+            return matched_valid_authors_references
+
+
+def match_literature_author_ids(matched_records):
```
michamos, line 98: naming is confusing, as this doesn't do any match, it postprocesses matches and returns the matching refs, maybe something like
```suggestion
def get_record_ref_if_unambiguous_match(matched_records):
```

backend/inspirehep/disambiguation/tasks.py
```diff
@@ -70,26 +75,127 @@ def disambiguate_signatures(self, clusters):
...
+    matched_authors_refs = match_literature_author_ids(matched_records)
+    return matched_authors_refs, matched_records
+
+
+def match_literature_author_with_validator(validators, author_data, matched_references):
+    for validator in validators:
+        valid_matches = (
+            match for match in matched_references if validator(author_data, match)
+        )
+        matched_valid_authors_references = match_literature_author_ids(valid_matches)
```
michamos, line 93: same comment

backend/inspirehep/disambiguation/tasks.py
```diff
@@ -70,26 +75,127 @@ def disambiguate_signatures(self, clusters):
...
+        matched_record
+        for matched_record in match(author_data, matcher_config)
+        if matched_record.get("inner_hits")
+    ]
+    matched_authors_refs = match_literature_author_ids(matched_records)
+    return matched_authors_refs, matched_records
+
+
+def match_literature_author_with_validator(validators, author_data, matched_references):
+    for validator in validators:
```
michamos, line 89: I would move this to the outer function for the same reason

backend/inspirehep/disambiguation/tasks.py
```diff
@@ -70,26 +75,127 @@ def disambiguate_signatures(self, clusters):
...
+    matched_records = [
+        matched_record
+        for matched_record in match(author_data, matcher_config)
+        if matched_record.get("inner_hits")
+    ]
+    matched_authors_refs = match_literature_author_ids(matched_records)
+    return matched_authors_refs, matched_records
+
+
+def match_literature_author_with_validator(validators, author_data, matched_references):
```
michamos, line 88: this doesn't call the matcher (doesn't do any query), so maybe
```suggestion
def validate_literature_author_matches(validators, author_data, matched_references):
```

backend/inspirehep/disambiguation/tasks.py
```diff
@@ -70,26 +75,127 @@ def disambiguate_signatures(self, clusters):
...
 
 
+def match_literature_author_with_config(author_data, matcher_config):
+    matched_records = [
+        matched_record
+        for matched_record in match(author_data, matcher_config)
+        if matched_record.get("inner_hits")
+    ]
+    matched_authors_refs = match_literature_author_ids(matched_records)
+    return matched_authors_refs, matched_records
```
michamos, line 85: return type is a bit awkward, I don't think the first line needs to be here. Can't you put it in something like `assign_reference_to_author_if_unambiguous_literature_author_match`, which would call this then the existing `assign_reference_to_author_if_unambiguous_match`?

backend/inspirehep/disambiguation/tasks.py
```diff
@@ -70,26 +75,127 @@ def disambiguate_signatures(self, clusters):
...
+        matched_authors_refs.update(
+            get_value(
+                record_authors, "inner_hits.authors.hits.hits._source.record.$ref", None
+            )
+        )
+    if matched_authors_refs and len(matched_authors_refs) == 1:
+        return matched_authors_refs
+
+
+def assign_reference_to_author_if_unambiguous_match(matched_authors_refs, author):
```
michamos, line 110: this is a misnomer now, I think it would be better to leave the length check here (which would work together with my other suggestions).

backend/inspirehep/matcher/validators.py
```diff
@@ -22,3 +22,20 @@ def authors_validator(author, result):
...
+    )
+
+
+def affiliations_validator(author, result):
+    authors_affiliations = author.get("affiliations", [])
+    result_authors_affiliations = get_value(
+        result, "inner_hits.authors.hits.hits._source.affiliations.value", []
+    )
+    for affiliation_list in result_authors_affiliations:
+        if set(affiliation_list) == set(authors_affiliations):
```
michamos, line 39: Not all affiliations need to match, it's enough if there's one in common
```suggestion
        if set(affiliation_list) & set(authors_affiliations)
```

backend/inspirehep/matcher/validators.py
```diff
@@ -22,3 +22,20 @@ def authors_validator(author, result):
...
+
+
+def collaboration_validator(author, result):
+    return author.get("collaborations") == get_value(
+        result, "_source.collaborations.value", []
+    )
+
+
+def affiliations_validator(author, result):
+    authors_affiliations = author.get("affiliations", [])
```
michamos, line 34: better make this a `set` from the start, instead of rebuilding it inside the loop.

michamos (commented):
some small comments about code structure and logic, almost there :+1:

backend/inspirehep/disambiguation/tasks.py
```diff
@@ -70,26 +75,134 @@ def disambiguate_signatures(self, clusters):
...
+def match_literature_author_with_config(author_data, matcher_config):
+    matched_records = [
+        matched_record
+        for matched_record in match(author_data, matcher_config)
+        if matched_record.get("inner_hits")
+    ]
+    return matched_records
+
+
+def get_record_ref_if_unambiguous_match(matched_records):
```
michamos, line 87:
```suggestion
def get_record_refs_for_matched_literature_authors(matched_records):
```

michamos (approved):
Minor comment remaining, but looks very good otherwise :+1:

Issue #1504, opened 2020-11-04
Update all $ref links in all other records when merging a record


pazembrz:
When record is merged it's PID is redirected to another PID. 
When this happens make sure that all old `$ref` from other records which were pointing to merged record (merged record is deleted after merge) are pointing now to the new record.

This should work on all pid types.

**TECH**
1. Check why we call `update_records_relations` only on migration
2. When we receive a record, on `update_records_relations` check the difference between previous and a new version of the record. If there is a difference in `deleted_records` field create a new celery task which will search for the references to the deleted record. Ask Micha for the exact information about collections linking in order to search only in proper fields.
3. For every result update the reference to point to the new record.

MJedr:
It seems that we were using `update_records_relations` also in [cli which was removed](https://github.com/inspirehep/inspirehep/commit/edb9e2c3fc338945af9c93dd687e2edbb67f3e20#diff-ecb61cba1d7f51eb4d4ff616c0e1301a1f76df77a180b620b122d3bdb43b310b). We were never calling it on record update.

PR #1555, opened 2020-11-23
reference linking: update refs after record merge

MJedr:
Ref: inspirehep/inspirehep#1504

### DIFF
--- a/backend/inspirehep/records/api/base.py
+++ b/backend/inspirehep/records/api/base.py
@@ -56,6 +56,7 @@ class InspireRecord(Record):
     ]
 
     experiment_field = "accelerator_experiments.record"
+    nested_record_fields = []
 
     @staticmethod
     def strip_empty_values(data):
@@ -351,6 +352,7 @@ class InspireRecord(Record):
         InspireRedirect.redirect(old_pid, new_pid)
 
         old_record = self.get_record(old_pid_object_uuid)
+        old_record["new_record"] = get_ref_from_pid(self.pid_type, self.control_number)
         if not old_record.get("deleted"):
             old_record.delete()
 
--- a/backend/inspirehep/records/receivers.py
+++ b/backend/inspirehep/records/receivers.py
@@ -12,6 +12,7 @@ from invenio_records.models import RecordMetadata
 
 from inspirehep.disambiguation.tasks import disambiguate_authors
 from inspirehep.records.api import InspireRecord
+from inspirehep.records.tasks import redirect_references_to_merged_record
 
 LOGGER = structlog.getLogger()
 
@@ -41,3 +42,5 @@ def index_after_commit(sender, changes):
                     and current_app.config["FEATURE_FLAG_ENABLE_AUTHOR_DISAMBIGUATION"]
                 ):
                     disambiguate_authors.delay(str(model_instance.id))
+                if "new_record" in model_instance.json:
+                    redirect_references_to_merged_record.delay(str(model_instance.id))
--- a/backend/inspirehep/records/tasks.py
+++ b/backend/inspirehep/records/tasks.py
@@ -5,12 +5,23 @@
 # inspirehep is free software; you can redistribute it and/or modify it under
 # the terms of the MIT License; see LICENSE file for more details.
 
-
 import structlog
+from celery import shared_task
+from elasticsearch import TransportError
+from flask import current_app
+from inspire_matcher.api import match
+from inspire_schemas.utils import get_refs_to_schemas
+from inspire_utils.record import get_value
 from invenio_db import db
 from sqlalchemy.exc import OperationalError
 
+from inspirehep.matcher.utils import (
+    generate_matcher_config_for_nested_reference_field,
+    generate_matcher_config_for_reference_field,
+)
+from inspirehep.pidstore.api import PidStoreBase
 from inspirehep.records.api import InspireRecord, LiteratureRecord
+from inspirehep.utils import flatten_list
 
 LOGGER = structlog.getLogger()
 
@@ -43,3 +54,64 @@ def update_records_relations(uuids):
 
     db.session.commit()
     return uuids
+
+
+@shared_task(
+    bind=True,
+    queue="indexer_task",
+    acks_late=True,
+    retry_backoff=True,
+    retry_kwargs={"max_retries": 3},
+    autoretry_for=(OperationalError, TransportError),
+)
+def redirect_references_to_merged_record(self, uuid):
+    record = InspireRecord.get_record(uuid)
+    new_record_ref = record["new_record"]["$ref"]
+    deleted_record_ref = record["self"]["$ref"]
+    record_schema = PidStoreBase.get_schema_name_from_uri(record["$schema"])
+    possible_refs_to_record = get_refs_to_schemas()[record_schema]
+    update_references_pointing_to_merged_record(
+        possible_refs_to_record, deleted_record_ref, new_record_ref
+    )
+
+
+def update_references_pointing_to_merged_record(
+    refs_to_schema, merged_record_uri, new_record_uri
+):
+    for index, path in refs_to_schema:
+        config = get_config_for_given_path(index, path)
+        matched_records = match({"$ref": merged_record_uri}, config)
+        for matched_record in matched_records:
+            matched_inspire_record = InspireRecord.get_record(matched_record["_id"])
+            referenced_records_in_path = flatten_list(
+                get_value(matched_inspire_record, path[: -len(".$ref")], [])
+            )
+            for referenced_record in referenced_records_in_path:
+                update_reference_if_reference_uri_matches(
+                    referenced_record, merged_record_uri, new_record_uri
+                )
+            matched_inspire_record.update(dict(matched_inspire_record))
+            LOGGER.info(
+                f"Updated reference for record", uuid=str(matched_inspire_record.id)
+            )
+    db.session.commit()
+
+
+def get_config_for_given_path(index, path):
+    record_with_reference_pid = current_app.config["SCHEMA_TO_PID_TYPES"][index]
+    nested_fields = InspireRecord.get_subclasses()[
+        record_with_reference_pid
+    ].nested_record_fields
+    config = (
+        generate_matcher_config_for_nested_reference_field(index, path)
+        if path.split(".")[0] in nested_fields
+        else generate_matcher_config_for_reference_field(index, path)
+    )
+    return config
+
+
+def update_reference_if_reference_uri_matches(
+    reference_record, merged_record_uri, new_record_uri
+):
+    if reference_record["$ref"] == merged_record_uri:
+        reference_record.update({"$ref": new_record_uri})
--- a/backend/inspirehep/matcher/utils.py
+++ b/backend/inspirehep/matcher/utils.py
@@ -116,3 +116,25 @@ def normalize_title(raw_title):
     normalized_title = normalized_title.upper()
 
     return normalized_title
+
+
+def generate_matcher_config_for_reference_field(schema_name, path):
+    return {
+        "algorithm": [
+            {"queries": [{"path": "$ref", "search_path": path, "type": "exact",},]}
+        ],
+        "index": f"records-{schema_name}",
+    }
+
+
+def generate_matcher_config_for_nested_reference_field(schema_name, path):
+    return {
+        "algorithm": [
+            {
+                "queries": [
+                    {"paths": ["$ref"], "search_paths": [path], "type": "nested",},
+                ]
+            }
+        ],
+        "index": f"records-{schema_name}",
+    }
--- a/backend/inspirehep/pidstore/api/base.py
+++ b/backend/inspirehep/pidstore/api/base.py
@@ -99,6 +99,18 @@ class PidStoreBase(object):
         return current_app.config["SCHEMA_TO_PID_TYPES"]
 
     @staticmethod
+    def get_schema_name_from_uri(schema):
+        try:
+            schema_name = (
+                six.moves.urllib.parse.urlsplit(schema)
+                .path.split("/")[-1]
+                .split(".")[0]
+            )
+        except (TypeError, IndexError):
+            schema_name = None
+        return schema_name
+
+    @staticmethod
     def get_pid_type_from_schema(schema):
         """Return the ``pid_type`` for a given ``schema``.
 
@@ -109,14 +121,7 @@ class PidStoreBase(object):
         Returns:
             None if not found, the ``pid_type``.
         """
-        try:
-            schema_name = (
-                six.moves.urllib.parse.urlsplit(schema)
-                .path.split("/")[-1]
-                .split(".")[0]
-            )
-        except (TypeError, IndexError):
-            schema_name = None
+        schema_name = PidStoreBase.get_schema_name_from_uri(schema)
         pids_to_schema = PidStoreBase._get_config_pid_types_to_schema()
         return pids_to_schema.get(schema_name)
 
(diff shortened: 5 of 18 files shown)


backend/inspirehep/matcher/config.py
```diff
@@ -334,3 +334,232 @@
...
+                {
+                    "path": "$ref",
+                    "search_path": "institutions.record.$ref",
+                    "type": "exact",
+                },
+            ]
+        }
+    ],
+    "index": "records-jobs",
+}
```
michamos, line 565: this seems like a terrible idea to me, yet another place to update when the schema changes. Maybe we could move this information to the schemas somehow and generate these searches based on this info?

backend/inspirehep/records/tasks.py
```diff
@@ -33,6 +40,12 @@ def update_records_relations(uuids):
                     record.update_conference_paper_and_proccedings()
                     record.update_institution_relations()
                     record.update_experiment_relations()
+                deleted_record_uri = record.check_if_record_was_deleted_by_merge()
+                if deleted_record_uri:
+                    record_ref = get_value(record, "self.$ref")
+                    recalculate_references_for_merged_record.delay(
+                        deleted_record_uri, record_ref
+                    )
```
michamos, line 48: you shouldn't add this here:
* it has nothing to do with this function, which updates relations inside auxiliary database tables
* it has a race condition, because the task might be executed before the changes have been persisted to the database

I understand why you did this, because you can't easily perform the check in the `models_committed` receiver as the SQLAlchemy session is in a state that doesn't allow you to execute any DB query, which would be needed to get the previous version. But it's definitely not the right place. I don't know where to put this, maybe in the task that does the reindexing (ask the other devs).

backend/inspirehep/records/tasks.py
```diff
@@ -43,3 +56,49 @@ def update_records_relations(uuids):
 
     db.session.commit()
     return uuids
+
+
+@shared_task(ignore_result=False, bind=True, queue="migrator")
```
michamos, line 61: why these random arguments? did you copy/paste this? all of them are wrong (you should read about what they do in the celery docs).

backend/inspirehep/records/tasks.py
```diff
@@ -43,3 +56,49 @@ def update_records_relations(uuids):
...
+def recalculate_references_for_merged_record(self, merged_record_uri, record_uri):
```
michamos, line 62:
```suggestion
def redirect_references_to_merged_record(self, merged_record_uri, new_record_uri):
```

backend/inspirehep/records/api/base.py
```diff
@@ -553,3 +553,10 @@ def redirected_record_ref(self):
         if self.get("deleted") and pid.status == PIDStatus.REDIRECTED:
             redirection = InspireRedirect.get_redirect(pid)
             return get_ref_from_pid(redirection.pid_type, redirection.pid_value)
+
+    def check_if_record_was_deleted_by_merge(self):
```
michamos, line 557: don't we already have a similar function that is used to decide whether to perform the merge?

backend/inspirehep/records/api/base.py
```diff
@@ -553,3 +553,10 @@ def redirected_record_ref(self):
...
+        current_deleted_records = get_value(self, "deleted_records.$ref", [])
+        if current_deleted_records != get_value(
+            self._previous_version, "deleted_records.$ref", []
+        ):
+            return current_deleted_records.pop()
```
michamos, line 562: :warning: you might be mutating the record here (I don't believe `get_value` makes a copy, I'm not entirely sure)

backend/inspirehep/records/tasks.py
```diff
@@ -43,3 +56,49 @@ def update_records_relations(uuids):
...
+                    )
+            matched_inspire_record.update(dict(matched_inspire_record))
+            LOGGER.info(f"Updated reference for record", {"uuid": str(uuid)})
+    db.session.commit()
+
+
+def update_reference_if_reference_uri_matches(
+    reference_record, merged_record_uri, new_record_uri
+):
+    if reference_record.get("$ref") == merged_record_uri:
```
michamos, line 103: don't use `get` if you know the field is there, it's useful only in cases where the field might be absent
```suggestion
    if reference_record["$ref"] == merged_record_uri:
```

backend/inspirehep/records/tasks.py
```diff
@@ -43,3 +56,49 @@ def update_records_relations(uuids):
...
+    match_reference_with_config(
+        referenced_records_indices_and_paths, merged_record_uri, record_uri
+    )
+
+
+def match_reference_with_config(config_list, merged_record_uri, new_record_uri):
+    for index, path in config_list:
+        config = (
+            generate_matcher_config_for_nested_reference_field(index, path)
+            if path in ["authors", "publication_info", "first_author", "supervisors"]
```
michamos, line 76: * Don't hide the list of `nested_paths` here. It would be better to add a `nested_record_fields` `property` to `InspireRecord` (empty by default) and override it for `LiteratureRecord`. That would also prevent name collisions between nested and non-nested fields in different indices (we don't have those currently).
* You don't need to search in `first_author`, as it's a subset of `authors` used only for searching. It's not present in the record, so there's nothing to update.

backend/inspirehep/records/tasks.py
```diff
@@ -43,3 +56,49 @@ def update_records_relations(uuids):
 
     db.session.commit()
     return uuids
+
+
+@shared_task(ignore_result=False, bind=True, queue="migrator")
+def recalculate_references_for_merged_record(self, merged_record_uri, record_uri):
+    merged_record_schema = merged_record_uri.rsplit("/", 2)[1]
+    if merged_record_schema == "literature":
+        merged_record_schema = "hep"
```
michamos, line 65: There should be a util for this already somewhere.

backend/inspirehep/records/tasks.py
```diff
@@ -43,3 +56,49 @@ def update_records_relations(uuids):
...
+            generate_matcher_config_for_nested_reference_field(index, path)
+            if path in ["authors", "publication_info", "first_author", "supervisors"]
+            else generate_matcher_config_for_reference_field(index, path)
+        )
+        matched_records = match({"$ref": merged_record_uri}, config,)
+        for matched_record in matched_records:
+            uuid = get_value(matched_record, "_id")
+            matched_inspire_record = InspireRecord.get_record(uuid)
+            for referenced_record in get_value(
+                matched_inspire_record, path.rstrip(".$ref"), []
```
michamos, line 84: You shouldn't use `rstrip` for this kind of stuff (it removes individual matching *characters* at the end of the string). In Python 3.9 there's `removesuffix`, but in the meantime you can slice the string
```suggestion
                matched_inspire_record, path[:-len(".$ref")], []
```

backend/inspirehep/records/tasks.py
```diff
@@ -43,3 +56,49 @@ def update_records_relations(uuids):
...
+    merged_record_schema = merged_record_uri.rsplit("/", 2)[1]
+    if merged_record_schema == "literature":
+        merged_record_schema = "hep"
+    referenced_records_indices_and_paths = get_refs_to_schemas()[merged_record_schema]
+    match_reference_with_config(
+        referenced_records_indices_and_paths, merged_record_uri, record_uri
+    )
+
+
+def match_reference_with_config(config_list, merged_record_uri, new_record_uri):
```
michamos, line 72: this is a misnomer, as it does much more than matching. I think you need to step back and think about what you need to do from a high-level perspective, create functions for the different steps, then write an implementation for those functions (recursively). We can discuss this further if needed.

backend/inspirehep/records/tasks.py
```diff
@@ -43,3 +56,49 @@ def update_records_relations(uuids):
...
+            else generate_matcher_config_for_reference_field(index, path)
+        )
+        matched_records = match({"$ref": merged_record_uri}, config,)
+        for matched_record in matched_records:
+            uuid = get_value(matched_record, "_id")
+            matched_inspire_record = InspireRecord.get_record(uuid)
+            for referenced_record in get_value(
+                matched_inspire_record, path.rstrip(".$ref"), []
+            ):
+                if isinstance(referenced_record, list):
```
michamos, line 86: use `force_list` from `inspire_utils.helpers` instead

backend/inspirehep/records/api/base.py
```diff
@@ -553,3 +554,10 @@ def redirected_record_ref(self):
         if self.get("deleted") and pid.status == PIDStatus.REDIRECTED:
             redirection = InspireRedirect.get_redirect(pid)
             return get_ref_from_pid(redirection.pid_type, redirection.pid_value)
+
+    def get_new_deleted_records(self):
+        current_deleted_records = get_value(self, "deleted_records.$ref", [])
+        if current_deleted_records != get_value(
+            self._previous_version, "deleted_records.$ref", []
+        ):
+            return current_deleted_records[-1]
```
michamos, line 563: no, you don't know whether the new one is the last one. You need to compare the two sets.

backend/inspirehep/records/receivers.py
```diff
@@ -41,3 +42,4 @@ def index_after_commit(sender, changes):
                     and current_app.config["FEATURE_FLAG_ENABLE_AUTHOR_DISAMBIGUATION"]
                 ):
                     disambiguate_authors.delay(str(model_instance.id))
+                redirect_references_to_merged_record.delay(str(model_instance.id))
```
michamos, line 45: so one extra very rarely needed task per record update. Is that really how we want this to be done? (cc @drjova @pazembrz). Maybe we should have a single task instead that does both indexing and this?

backend/inspirehep/records/tasks.py
```diff
@@ -43,3 +54,64 @@ def update_records_relations(uuids):
...
     db.session.commit()
     return uuids
+
+
+@shared_task(
+    bind=True,
+    acks_late=True,
+    retry_backoff=True,
+    retry_kwargs={"max_retries": 3},
+    autoretry_for=(OperationalError, RequestError, TransportError),
```
michamos, line 64: * I would put this in the `indexer_task` queue, as it's the same kind of stuff.
* `RequestError` is a subclass of `TransportError` if I'm not mistaken.

backend/inspirehep/records/receivers.py
pazembrz: IMHO as this will be a rare task so optimization is not really important here, so it's better to keep it separate. We don't want bloated celery tasks, and it will be easier to keep logic of indexing and post-redirection data handling separate.

michamos: It's not a rare task, it will be scheduled on every record update, but the task itself will do nothing most of the time.

backend/inspirehep/records/tasks.py
```diff
@@ -43,3 +54,66 @@ def update_records_relations(uuids):
...
+):
+    for index, path in refs_to_schema:
+        config = get_config_for_given_path(index, path)
+        matched_records = list(match({"$ref": merged_record_uri}, config,))
+        for matched_record in matched_records:
+            matched_inspire_record = InspireRecord.get_record(matched_record["_id"])
+            referenced_records_in_path = get_value(
+                matched_inspire_record, path[: -len(".$ref")], []
+            )
+            for referenced_record in force_list(*referenced_records_in_path):
```
michamos, line 91: are you sure that this shouldn't be
```suggestion
            for referenced_record in force_list(referenced_records_in_path):
```
please add a test for the case where `referenced_records_in_path` is a list.

backend/inspirehep/records/tasks.py
```diff
@@ -43,3 +54,66 @@ def update_records_relations(uuids):
...
+    queue="indexer_task",
+    acks_late=True,
+    retry_backoff=True,
+    retry_kwargs={"max_retries": 3},
+    autoretry_for=(OperationalError, TransportError),
+)
+def redirect_references_to_merged_record(self, uuid):
+    record = InspireRecord.get_record(uuid)
+    deleted_record_refs = record.get_new_deleted_records()
+    if deleted_record_refs:
```
michamos, line 70: better avoid unneeded indentation
```suggestion
    if not deleted_record_refs:
        return
    # the rest
```

backend/inspirehep/records/tasks.py
```diff
@@ -43,3 +54,66 @@ def update_records_relations(uuids):
...
+        record_ref = get_value(record, "self.$ref")
```
michamos, line 71: you need this to be available, otherwise things will fail, and it should always be there, so better do
```suggestion
        record_ref = record["self"][$ref"]
```
which is 1) slightly more performant and 2) fails early and loudly in case it's not there.

backend/inspirehep/records/receivers.py
MJedr: Maybe let's run the task only if the `deleted_records` from `model_instance` is not empty? I don't know how often it happens that records have `deleted_records`, but it will reduce somehow the number of calls of this task

backend/inspirehep/records/tasks.py
```diff
@@ -43,3 +54,66 @@ def update_records_relations(uuids):
...
+                possible_refs_to_record, deleted_record_ref, record_ref
+            )
+
+
+def update_references_pointing_to_merged_record(
+    refs_to_schema, merged_record_uri, new_record_uri
+):
+    for index, path in refs_to_schema:
+        config = get_config_for_given_path(index, path)
+        matched_records = list(match({"$ref": merged_record_uri}, config,))
```
michamos, line 85: You can iterate on the iterator returned by `match` directly:
```suggestion
        matched_records = match({"$ref": merged_record_uri}, config,)
```

backend/inspirehep/records/receivers.py
michamos: that's a great idea :+1: We have ~10k Literature records that have data in `deleted_records`: https://inspirehep.net/literature?sort=mostrecent&size=25&page=1&q=deleted_records%3A%2A, so it will be a big difference.

backend/inspirehep/records/tasks.py
MJedr: The problem is with `authors.affiliations.record` since it returns a list for every author, so I have to flatten this list somewhere

michamos: https://github.com/inspirehep/inspirehep/blob/9ad059f74b0467b546728ceeb6227415c89f1663/backend/inspirehep/utils.py#L36

backend/inspirehep/records/tasks.py
```diff
@@ -43,3 +55,64 @@ def update_records_relations(uuids):
...
+):
+    for index, path in refs_to_schema:
+        config = get_config_for_given_path(index, path)
+        matched_records = match({"$ref": merged_record_uri}, config)
+        for matched_record in matched_records:
+            matched_inspire_record = InspireRecord.get_record(matched_record["_id"])
+            referenced_records_in_path = flatten_list(
+                get_value(matched_inspire_record, path[: -len(".$ref")], [])
+            )
+            for referenced_record in force_list(referenced_records_in_path):
```
michamos, line 90: I don't think you need the `force_list` here, as `flatten_list` already returns a `list`.

michamos (approved):
a couple minor comments, looks good otherwise :+1:

backend/inspirehep/records/tasks.py
MJedr: For authors.affiliations it doesn't work, probably because it's an array in nested fields

michamos: that's why you need `flatten_list`, but `force_list` on top doesn't add anything (unless there's a bug in the former).

PR #1556, opened 2020-11-23
v1.2.23

drjova:


michamos (approved):
My change on dojson requires remigration

pazembrz (approved):
My changes for pidstore requires alembic migration!
If we want to use redirection it requires `FEATURE_FLAG_ENABLE_REDIRECTION_OF_PIDS` to be turned on and data re-migrated.

MJedr (approved):

PR #1561, opened 2020-11-24
pidstore: register pid if it is missing during pid deletion

pazembrz:


backend/inspirehep/pidstore/minters/base.py
```diff
@@ -145,7 +145,17 @@ def update(cls, object_uuid, data):
     @classmethod
     def delete(cls, object_uuid, data):
```
drjova, line 146: is migrator calling `delete`?

backend/inspirehep/pidstore/minters/base.py
```diff
@@ -145,7 +145,17 @@ def update(cls, object_uuid, data):
...
         if "control_number" in data:
-            pid_provider = cls.provider.get(data["control_number"], cls.pid_type)
+            try:
+                pid_provider = cls.provider.get(data["control_number"], cls.pid_type)
+            except PIDDoesNotExistError:
+                if data.get("deleted", False) == True:
+                    cls.mint(object_uuid, data)
```
michamos, line 152: does this fail if there's a texkey/doi/arXiv ID/ORCID/BAI collision with a non-deleted record?

backend/inspirehep/pidstore/minters/base.py
pazembrz: Yes it is. Twice actually: https://github.com/inspirehep/inspirehep/blob/c02b88d7a64420854f13770e082be14945a79cd6/backend/inspirehep/migrator/tasks.py#L470-L471 here directly, and in record.create_or_update (as update and create are also calling minter.delete)

backend/inspirehep/pidstore/minters/base.py
pazembrz: This code is in `ControlNumberMinter`. So this will be called only for control_numbers as we only support redirecting control_numbers (texkeys are not redirected but re-claimed)

michamos (approved):

PR #1574, opened 2020-12-01
disambiguation: pass authors data directly to create_new_author

MJedr:
* solves bug with assigning already existing pid to newly created authors

### DIFF
--- a/backend/tests/integration-async/disambiguation/test_disambiguation_tasks.py
+++ b/backend/tests/integration-async/disambiguation/test_disambiguation_tasks.py
@@ -630,3 +630,28 @@ def test_disambiguate_authors_create_new_author(
         )
 
     retry_until_pass(assert_disambiguation_task)
+
+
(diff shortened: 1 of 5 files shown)


backend/tests/integration-async/disambiguation/test_disambiguation_tasks.py
```diff
@@ -630,3 +630,63 @@ def assert_disambiguation_task():
         )
 
     retry_until_pass(assert_disambiguation_task)
+
+
+def test_disambiguate_authors_create_two_author_with_same_name(
```
michamos, line 635: I actually meant doing this through the disambiguation, to have a test of everything that's going on. So create a paper with two authors with same names (with no matching author record), and check that works correctly.

MJedr: So I did it, in test I created a record with two authors with same names for which new authors are created (what is tested in assertion)

michamos (approved):

PR #1579, opened 2020-12-03
hep: fix failing tests

MJedr:


backend/tests/integration/indexer/test_indexer_literature.py
```diff
@@ -86,8 +86,7 @@ def test_indexer_deletes_record_from_es(inspire_app, datadir):
 def test_indexer_creates_proper_fulltext_links_in_ui_display_files_enabled(
     inspire_app, s3
 ):
-    create_s3_bucket("1")
-    create_s3_bucket("f")
+    create_s3_bucket("8")
```
drjova, line 89: we should add a test for not hidden files as well

drjova (approved):

PR #1599, opened 2020-12-14
poetry: use pyproject.toml instead of setup.py

michamos:
* Newer versions of poetry don't use the info of `setup.py`, which is a
  problem as we rely heavily on entry-points. The solution is to move
  everything to `pyproject.toml` instead, and drop the `setup.py`.
* I also looked into dropping `setup.cfg`. This is used for `isort`,
  `flake8` and `coverage` config. New versions of `isort` and `coverage`
  can use `pyproject.toml` instead, so I updated them and adapted their
  config. For now the `setup.cfg` file is still needed for `flake8 (see
  https://gitlab.com/pycqa/flake8/-/issues/428).
* To make this commit, I needed to setup the `prettier` pre-commit hook.
  The repo has changed, so I changed the `pre-commit` config, and also
  had to extend the `prettier` ignore list.


drjova (approved):
Thank you @michamos LGTM

PR #1600, opened 2020-12-14
poetry: bump inspire-query-parser to v6.0.33

michamos:


MJedr (approved):

backend/poetry.lock
```diff
@@ -4640,7 +4640,8 @@ inspire-matcher = [
     {file = "inspire-matcher-9.0.8.tar.gz", hash = "sha256:2e10c71c3e576f0984adabe53f2c8292757a849927387aa8160bb62b92765285"},
 ]
 inspire-query-parser = [
-    {file = "inspire-query-parser-6.0.31.tar.gz", hash = "sha256:b0556a950012f4432458823afbf3b6d7a1294ebf0449a1d1db22bcfaf6cb3857"},
+    {file = "inspire-query-parser-6.0.33.tar.gz", hash = "sha256:6c448af5f9fb58cf636114bc2c238e4f3aa3029d5334625c09460a7a0323a141"},
+    {file = "inspire_query_parser-6.0.33-py2-none-any.whl", hash = "sha256:d7cb7c9cb67803a588ea50946dbe7330410ca6543fd27eb3817be0fdbe9d6342"},
```
michamos, line 4644: I don't think we need to build a `python2` wheel for the query parser @drjova. Actually we could update it to be python3-only as `next` is not using it AFAIK.

drjova: Indeed we can drop it!

PR #1609, opened 2021-01-05
submissions: fix author update

drjova:


backend/inspirehep/submissions/views.py
```diff
@@ -160,6 +162,12 @@ def get_updated_record_data(self, update_data, record):
             "advisors",
         ]
 
+        updated_schemas = get_value(update_data, "ids.schema", default=[])
+        if updated_schemas:
+            for _id in record.get("ids", []):
+                if _id["schema"] not in updated_schemas:
+                    update_data["ids"].append(_id)
```
michamos, line 169: I don't think this is correct, as we want to keep outdated identifiers on the record too (some records have two ORCIDs for example, one of which is deprecated).

drjova: Hmm, in that case indeed it will not work. Do we want the same for `BAI` and `LINKEDIN` which are part of the form?

drjova: @michamos Also do we need to keep the order with the updated first?

michamos: I don't think we ever have duplicated BAIs or LINKEDIN, but for others it might happen (ORCID in particular). Order is indeed important (first to appear is the current one).

drjova: Thanks, still even if we don't have more than one BAI or LINKED so far, should I handle that case to?

drjova: It's a more complex, but I will update the PR

michamos (approved):

Issue #1601, opened 2020-12-14
Datasets icon for papers


StellaCh:
Users complain that they can't find hepdata links to papers. The HEPData link is hidden in the detailed page and you can view HEPData links in `View in: HEPData`, eg. https://inspirehep.net/literature/1810348 . At the moment, users go to legacy to view these links because they don't know they exist on labs.

We need to the HEPData info more visible:
Add a new icon <DatabaseOutlined /> that links to HEPData.
The icon should appear only if there're links to datasets in the paper. The icon should be visible both to search and detailed page
<img width="642" alt="Screenshot 2020-12-14 at 15 37 41" src="https://user-images.githubusercontent.com/24955201/102105039-2defb000-3e2f-11eb-8a9c-385edf1e38dd.png">

The icon 'datasets' should be the last item of the icons list.

drjova:
TECH NOTES:

* Remove `hepdata` from https://github.com/inspirehep/inspirehep/blob/master/backend/inspirehep/records/marshmallow/literature/common/external_system_identifier.py#L12
* Create a new field `hepdata_link` in the serializer https://github.com/inspirehep/inspirehep/blob/a325629aea4e8717997a26ec0f37b5d2e62ad313/backend/inspirehep/records/marshmallow/literature/ui.py#L40
* Add `UrlsAction` her https://github.com/inspirehep/inspirehep/blob/bf5025906d19daf5dc018407a280321d6eb849fa/ui/src/literature/containers/DetailPageContainer/DetailPageContainer.jsx#L132 for the `hep_data_link`

PR #1612, opened 2021-01-05
literature: add dataset link field and icon

MJedr:
Ref: #1601


ui/src/literature/containers/DetailPageContainer/DetailPageContainer.jsx
```diff
@@ -130,6 +131,13 @@ function DetailPage({
                         pidValue={controlNumber}
                       />
                     )}
+                    {datasetLinks &&
+                    <UrlsAction
+                    urls={datasetLinks}
+                    icon={<DatabaseOutlined />}
+                    text="datasets"
+                        />
+                        }
```
drjova, line 140: I believe you will have to update tests and e2e tests

MJedr: They pass, but I can update the snapshot in e2e

backend/inspirehep/records/marshmallow/literature/ui.py
```diff
@@ -210,6 +211,25 @@ def get_citation_pdf_urls(self, data):
                 urls.append(url)
         return urls or missing
 
+    def get_datasets(self, data):
+        dataset_links = []
+        dataset_schema_to_url_link_prefix_map = {
+            "hepdata": "https://www.hepdata.net/record/",
+        }
+        dataset_schema_to_description_map = {"hepdata": "HEPData"}
+        all_links = get_value(data, "external_system_identifiers", [])
```
drjova, line 220: You could use `get_values_for_schema`

i.e.

```python
get_values_for_schema(data['external_system_identifiers'], 'HEPData')
```

MJedr: It will work for now, but what if we'll add other dataset sources?

ui/src/literature/containers/DetailPageContainer/DetailPageContainer.jsx
drjova: It's important to add the link to the record, so we make sure it displayed correctly :)

drjova: We could add the link to test the snapshot :)

backend/inspirehep/records/marshmallow/literature/ui.py
```diff
@@ -210,6 +211,25 @@ def get_citation_pdf_urls(self, data):
                 urls.append(url)
         return urls or missing
 
+    def get_datasets(self, data):
+        dataset_links = []
+        dataset_schema_to_url_link_prefix_map = {
```
drjova, line 216: `dataset_schema_to_url_link_prefix_map` and `dataset_schema_to_description_map` should be moved on the top of the file (uppercase).

drjova (approved):

Issue #1565, opened 2020-11-26
Reuse the Jinja environments for caching


michamos:
From https://inspirehep.zulipchat.com/#narrow/stream/122202-devel/topic/migration.20profiling/near/217984830:

Compiling Jinja templates to bytecodes is expensive, but Jinja has built-in caching mechanisms. Unfortunately, they are not used in use in our case, because we recreate the `Environment` object on every call:
 https://github.com/inspirehep/inspirehep/blob/e4fd6b57fe2ccf78a315f4e0c0b33dd0e3a0d263/backend/inspirehep/records/serializers/latex.py#L44-L48 and
https://github.com/inspirehep/inspirehep/blob/e4fd6b57fe2ccf78a315f4e0c0b33dd0e3a0d263/backend/inspirehep/records/serializers/cv.py#L28-L30
This should be refactored to have global environments that can be reused during the whole application lifetime, as explained in https://jinja.palletsprojects.com/en/2.11.x/api/#basics.

drjova:
TECH NOTES:

* Create an environment in `records/serializers/jinja.py` with `loader` the `template` folder of `inspirehep.records.serializers.templates`.
* Change https://github.com/inspirehep/inspirehep/blob/e4fd6b57fe2ccf78a315f4e0c0b33dd0e3a0d263/backend/inspirehep/records/serializers/latex.py#L44-L48 and https://github.com/inspirehep/inspirehep/blob/e4fd6b57fe2ccf78a315f4e0c0b33dd0e3a0d263/backend/inspirehep/records/serializers/cv.py#L28-L30 to use the new environment instead.

PR #1618, opened 2021-01-06
jinja: reuse environments for caching

MJedr:
Ref #1565

### DIFF
--- a/backend/inspirehep/records/serializers/jinja.py
+++ b/backend/inspirehep/records/serializers/jinja.py
@@ -1,0 +1,21 @@
+# -*- coding: utf-8 -*-
+#
+# Copyright (C) 2020 CERN.
+#
+# inspirehep is free software; you can redistribute it and/or modify it under
+# the terms of the MIT License; see LICENSE file for more details.
+
+from pathlib import Path
+
+import jinja2
+
+templates_path = str(Path(__file__).parent.resolve() / "templates")
+
+jinja_cv_env = jinja2.Environment(
+    loader=jinja2.FileSystemLoader(templates_path),
+)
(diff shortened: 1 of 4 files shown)


backend/inspirehep/records/serializers/jinja.py
```diff
@@ -0,0 +1,23 @@
...
+# inspirehep is free software; you can redistribute it and/or modify it under
+# the terms of the MIT License; see LICENSE file for more details.
+
+import os
+
+import jinja2
+
+current_abs_path = os.path.abspath(__file__)
+directory_path = os.path.dirname(current_abs_path)
+templates_path = os.path.join(directory_path, "templates")
```
michamos, line 14: I know you just copy/pasted this, but it's better to use `Pathlib`:
```python
from pathlib import Path
templates_path = Path(__file__).parent / "templates"
```

drjova (approved):

backend/inspirehep/records/serializers/jinja.py
```diff
@@ -0,0 +1,22 @@
...
+
+import os
+from pathlib import Path
+
+import jinja2
+
+templates_path = Path(__file__).parent / "templates"
+
+jinja_cv_env = jinja2.Environment(
+    loader=jinja2.FileSystemLoader(os.path.abspath(templates_path)),
```
michamos, line 16: sorry if I wasn't clear, I meant replacing all `os.path` by `pathlib` equivalents: https://docs.python.org/3/library/pathlib.html#correspondence-to-tools-in-the-os-module, so also this.

michamos (approved):

PR #1619, opened 2021-01-06
global: addition of flake8 step in pre-commit

drjova:


.pre-commit-config.yaml
```diff
@@ -11,3 +11,8 @@ repos:
     rev: '5.6.4'
     hooks:
       - id: isort
+- repo: https://gitlab.com/pycqa/flake8
+  rev: 3.8.4
+  hooks:
+  - id: flake8
+    additional_dependencies: [flake8-typing-imports==1.6.0]
```
MJedr, line 18: why not the a higher version (the newest is `1.10.0`)?

drjova: Actually is completely wrong, thanks

MJedr (approved):

Issue #1564, opened 2020-11-26
Fix idx_object index


pazembrz:
Create alembic migration which will modify index `idx_object` from `pidstore_pid` tablein a way that it will allow to querying on only `object_uuid` or `object_uuid` and `object_type`. The index should be on columns (`object_uuid`, `object_type`) in this particular order, currently is in opposite order.

PR #1620, opened 2021-01-06
alembic: add migration to fix idx_object in pidstore_pid table

MJedr:
Ref: #1564
* requires alembic migration


pazembrz (approved):

pazembrz:
Remember that this alembic migration has to be run manually after deployment :)

PR #1637, opened 2021-01-11
literature: add dataset_links to search serializer

MJedr:


backend/inspirehep/records/marshmallow/literature/base.py
```diff
@@ -50,6 +56,22 @@ class Meta:
     documents = NonHiddenRaw(dump_only=True)
     publication_info = NonHiddenRaw(dump_only=True)
     report_numbers = NonHiddenRaw(dump_only=True)
+    dataset_links = fields.Method("get_datasets")
```
drjova, line 59: I wouldn't add this in the raw schema, since the record will be invalid.

MJedr: It's in public schema, not raw :)

drjova: Sorry github diff :/

Issue #1293, opened 2020-09-08
Check mimetype before we attach the file


drjova:
On `next` check the mimetype before attaching a file. We have examples which  the files attached were html error pages i.e.
https://inspirehep.net/api/files/a8e7a2f6-fd5d-4cd9-b228-c75364d8918a/fulltext.pdf

**TECH NOTES**
Files are downloaded in https://github.com/inspirehep/inspire-next/blob/11dcad6d3ecf6f5b5457b1af014121c73c1dcc3e/inspirehep/modules/workflows/tasks/actions.py#L382, we should check there if the mimetype is correct. (https://github.com/inspirehep/inspire-next/blob/11dcad6d3ecf6f5b5457b1af014121c73c1dcc3e/inspirehep/modules/workflows/utils/__init__.py#L309)

PR #1638, opened 2021-01-11
literature: check mime type before attaching a file

drjova:
* Ref: #1293

### DIFF
--- a/.prettierignore
+++ b/.prettierignore
@@ -1,3 +1,4 @@
 *.html
+*.yaml
 backend
 .prettierignore
--- a/backend/inspirehep/records/api/literature.py
+++ b/backend/inspirehep/records/api/literature.py
@@ -9,6 +9,7 @@ import threading
 import uuid
 from concurrent.futures import ThreadPoolExecutor, TimeoutError, as_completed
 from io import BytesIO
+from mimetypes import guess_type
 
 import magic
 import orjson
@@ -41,6 +42,7 @@ from inspirehep.records.errors import (
     ImportArticleError,
     ImportConnectionError,
     ImportParsingError,
+    InvalidMimeTypeError,
     UnknownImportIdentifierError,
 )
 from inspirehep.records.marshmallow.literature import LiteratureElasticSearchSchema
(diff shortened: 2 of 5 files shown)


.prettierignore
```diff
@@ -1,3 +1,4 @@
 *.html
+*.yaml
```
michamos, line 2: why?

backend/inspirehep/records/api/literature.py
```diff
@@ -421,7 +423,16 @@ def add_file(
                 return result
             file_data = download_file_from_url(url)
             new_key = hash_data(file_data)
+
             mimetype = magic.from_buffer(file_data, mime=True)
+            mimetype_guess, _ = guess_type(url)
```
michamos, line 428: I don't think this will be very effective for the main use-case (avoiding invalid PDFs). If the `url` doesn't end in `.pdf` (which I expect happens pretty often, e.g. on INSPIRE), this won't do anything. Maybe we should additionally maintain a black/whitelist?

drjova: it's not at all, but after the discussion with Thorsten there is no way to have a blacklist, we should accept any kind of files actually. Another solution will be to verify the response `content-type` but still if it supposes to be a `pdf` and the url is `text/html` and we check the content of the url and it's `html` then it's exactly the same issue.

tsgit: what I meant to convey is that you can't trust the content-type, extension, mime-type or any number of things. you have to verify the format of the file once you have a local copy.
for PDF, PS, png, jpg, gif there are tools to not only check that the file has the proper "magic sequence" in the first few dozen bytes of the file (python-magic),
but that the entire file is syntactically correct. think pylint or xmllint for other file types.
Specifically for PDF, you can run `gs -q -DNODISPLAY <file> -c quit` or PyPDF2 open or some such (e.g. mutool). Similarly the most commonly used figure formats have a variety of tools, for example PIL's `im.verify()`

tsgit: never trust external sources, always verify

Issue #1623, opened 2021-01-07
metadata is lost on author update (form) if not part of it


drjova:

michamos:
Not only a few fields are affected. It seems to be a structural issue with the author update, that it completely discards all information not present in the form (@hoc3426 reports wikipedia links also disappear). The update mechanism should probably be revisited to update the existing record rather than overwriting everything.

PR #1651, opened 2021-01-15
author submission: don't override _private_notes

MJedr:
Ref: #1623

### DIFF
--- a/backend/inspirehep/submissions/views.py
+++ b/backend/inspirehep/submissions/views.py
@@ -14,7 +14,7 @@ from flask import Blueprint, abort, current_app, request, url_for
 from flask.views import MethodView
 from flask_login import current_user
 from inspire_schemas.builders import JobBuilder
-from inspire_utils.record import get_value, get_values_for_schema
+from inspire_utils.record import get_value
 from invenio_db import db
 from invenio_pidstore.errors import PIDDoesNotExistError
 from jsonschema import SchemaError, ValidationError
@@ -135,11 +135,8 @@ class AuthorSubmissionsResource(BaseSubmissionsResource):
                 )
         except PIDDoesNotExistError:
             abort(404)
-
         data = self.load_data_from_request()
-        updated_record_data = self.get_updated_record_data(data, record)
-
-        record.update(updated_record_data, data)
+        self.update_author_record(data, record)
         db.session.commit()
 
         if not is_superuser_or_cataloger_logged_in():
@@ -150,7 +147,7 @@ class AuthorSubmissionsResource(BaseSubmissionsResource):
 
         return jsonify({"pid_value": record["control_number"]})
 
-    def get_updated_record_data(self, update_data, record):
+    def update_author_record(self, update_data, record):
         optional_fields = [
             "email_addresses",
             "public_notes",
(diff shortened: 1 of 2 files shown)


backend/inspirehep/submissions/views.py
```diff
@@ -54,13 +54,25 @@
...
         if field not in update_form_data and field in record_data:
             del record_data[field]
+    fields_to_extend = {
+        field: value
+        for (field, value) in update_form_data.items()
+        if field in fields_to_extend
+    }
+    extended_fields = {
+        field: value + record[field] for field, value in fields_to_extend.items()
+    }
```
drjova, line 73: unfortunately, this will cause duplicates.

drjova:
@MJedr we have to do it for all the fields in the form, it's not just `_private_notes`

backend/inspirehep/submissions/views.py
MJedr: Yes, but in `_private_notes` do we want to replace the note `{"value": "a value"}` or do we want to append it to the list of `_private_notes`?

backend/inspirehep/submissions/views.py
```diff
@@ -160,21 +157,29 @@ def get_updated_record_data(self, update_data, record):
...
+        for field in optional_fields:
+            if field not in set(update_data) and field in set(dict(record)):
+                del record[field]
+
+        for field in update_data:
+            if field == "_private_notes" and record.get(field):
+                record[field].extend(update_data[field])
+            elif field == "ids":
+                record_ids = record.get("ids", [])
+                record_ids_schemas = set(get_value(record, "ids.schema", []))
```
drjova, line 170: hmm we should calculate this once, not in the loop

backend/inspirehep/submissions/views.py
```diff
@@ -160,21 +157,29 @@ def get_updated_record_data(self, update_data, record):
...
+                for id_dict in update_data["ids"]:
```
drjova, line 171: we don't have to put it inside `for field in update_data:`. What I would recommend is the following:

Check `_private_notes`
Check `ids`
For loop for fields that are not in `['_private_notes', 'ids']`

By doing that you would avoid the 3 nested loops

backend/inspirehep/submissions/views.py
```diff
@@ -160,21 +157,32 @@ def get_updated_record_data(self, update_data, record):
...
-                    and record_id["value"] != updated_orcid[0]
-                ):
-                    update_data["ids"].append(record_id)
-        except KeyError:
-            pass
 
-        return get_updated_record_data(record, update_data, optional_fields)
+        for field in optional_fields:
+            if field not in set(update_data) and field in set(dict(record)):
+                del record[field]
```
drjova, line 163: why to duplicate the code of `get_updated_record_data`? Do I miss something?

backend/inspirehep/submissions/views.py
```diff
@@ -160,21 +157,32 @@ def get_updated_record_data(self, update_data, record):
...
+        if "_private_notes" in update_data and "_private_notes" in record:
+            record["_private_notes"].extend(update_data["_private_notes"])
+            del update_data["_private_notes"]
+
+        if "ids" in update_data:
+            record_ids = record.get("ids", [])
+            record_ids_schemas = set(get_value(record, "ids.schema", []))
+            for id_dict in update_data["ids"]:
+                if (id_dict["schema"] not in record_ids_schemas) or (
+                    id_dict["schema"] == "ORCID" and id_dict not in record_ids
```
drjova, line 174: Shouldn't this be like the following?

```suggestion
                    id_dict["schema"] == "ORCID" and id_dict['value'] not in record_ids
```

drjova: if it's true then we need to add tests for this case too.

backend/inspirehep/submissions/views.py
```diff
@@ -160,21 +157,32 @@ def get_updated_record_data(self, update_data, record):
...
+                ):
+                    record["ids"].append(id_dict)
+                elif id_dict["schema"] != "ORCID":
+                    for record_id in record_ids:
+                        if record_id["schema"] == id_dict["schema"]:
+                            record_id["value"] = id_dict["value"]
+            del update_data["ids"]
```
drjova, line 181:
```suggestion
       
            record_ids = record.get("ids", [])
            record_ids_schemas = set(get_value(record, "ids.schema", []))
            for id_dict in update_data.get("ids", []):
                if (id_dict["schema"] not in record_ids_schemas) or (
                    id_dict["schema"] == "ORCID" and id_dict not in record_ids
                ):
                    record["ids"].append(id_dict)
                elif id_dict["schema"] != "ORCID":
                    for record_id in record_ids:
                        if record_id["schema"] == id_dict["schema"]:
                            record_id["value"] = id_dict["value"]
            date_data.pop("ids",None)
```

backend/inspirehep/submissions/views.py
MJedr: I think that there is no difference if it will be `id_dict['value']` or `id_dict` because the schema in this case should always be `ORCID`.

We're already testing update when orcid is not in the record ids here https://github.com/inspirehep/inspirehep/blob/17a0dec6bbd9d9a8bd27b94e99afa278dde58250/backend/tests/integration/submissions/test_submissions_views.py#L312
 and in other tests we're testing the case when other ids from update data are missing in record ids for example linkedin here
https://github.com/inspirehep/inspirehep/blob/6bdf7ce1f9a0b8b977e5f13c1aa735119db387d6/backend/tests/integration/submissions/test_submissions_views.py#L283

backend/inspirehep/submissions/views.py
```diff
@@ -160,21 +157,28 @@ def get_updated_record_data(self, update_data, record):
...
-            pass
 
-        return get_updated_record_data(record, update_data, optional_fields)
+        if "_private_notes" in update_data and "_private_notes" in record:
+            record["_private_notes"].extend(update_data["_private_notes"])
+            del update_data["_private_notes"]
+
+        record_ids = record.get("ids", [])
+        record_ids_schemas = set(get_value(record, "ids.schema", []))
+        for id_dict in update_data["ids"]:
```
drjova, line 167:
```suggestion
        for id_dict in update_data.get("ids", []):
```

drjova (approved):
LGTM 👍 Just the comments. We have to test it thoroughly on QA

Issue #1648, opened 2021-01-14
Fix circular imports in serializers


drjova:
```   

...:
---------------------------------------------------------------------------
ImportError                               Traceback (most recent call last)
<ipython-input-2-82a7ec573853> in <module>
----> 1 from inspirehep.records.marshmallow.literature.common import PublicationInfoItemSchemaV1

~/src/inspirehep/backend/inspirehep/records/marshmallow/literature/__init__.py in <module>
      7 # flake8: noqa
      8
----> 9 from .authors import LiteratureAuthorsSchema
     10 from .base import (
     11     LiteratureAdminSchema,

~/src/inspirehep/backend/inspirehep/records/marshmallow/literature/authors.py in <module>
     10
     11 from ..fields import NestedWithoutEmptyObjects
---> 12 from .common import AuthorSchemaV1
     13 from .common.author import SupervisorSchema
     14

~/src/inspirehep/backend/inspirehep/records/marshmallow/literature/common/__init__.py in <module>
     11 from .collaboration import CollaborationSchemaV1
     12 from .collaboration_with_suffix import CollaborationWithSuffixSchemaV1
---> 13 from .conference_info_item import ConferenceInfoItemSchemaV1
     14 from .doi import DOISchemaV1
     15 from .external_system_identifier import ExternalSystemIdentifierSchemaV1

~/src/inspirehep/backend/inspirehep/records/marshmallow/literature/common/conference_info_item.py in <module>
      9
     10 from inspirehep.pidstore.api import PidStoreBase
---> 11 from inspirehep.records.api import InspireRecord
     12
     13

~/src/inspirehep/backend/inspirehep/records/api/__init__.py in <module>
     12 from .authors import AuthorsRecord  # noqa: F401
     13 from .base import InspireRecord  # noqa: F401
---> 14 from .conferences import ConferencesRecord  # noqa: F401
     15 from .data import DataRecord  # noqa: F401
     16 from .experiments import ExperimentsRecord  # noqa: F401

~/src/inspirehep/backend/inspirehep/records/api/conferences.py in <module>
      8 from inspirehep.pidstore.api.conferences import PidStoreConferences
      9 from inspirehep.records.api.base import InspireRecord
---> 10 from inspirehep.records.marshmallow.conferences import ConferencesElasticSearchSchema
     11 from inspirehep.records.models import (
     12     ConferenceLiterature,

[...]

PR #1652, opened 2021-01-15
pytest: use testmon for integration tests

michamos:
* Introduces and documents `testmon` to speed up integration tests
  locally, and adds a wrapper script to make it more convenient.
* Works around #1648 by not running `unit` tests with `testmon`.
* Updates `pytest` to latest version, moves its config to
  `pyproject.toml`, and removes useless `pytest-runner` and config
  options.


drjova (approved):
💯

Issue #1621, opened 2021-01-07
Move /authorlist to hep


drjova:
Move `/editor/authorlist/text` to `hep`.

Tech Notes:
* https://github.com/inspirehep/inspire-next/blob/master/inspirehep/modules/editor/api.py#L74-L82
* Change routing on k8s

MJedr:
What about this view:
https://github.com/inspirehep/inspire-next/blob/2fe0a80ba872dd91ffdfec9134b8be7a7a0cc78c/inspirehep/modules/tools/views.py#L51-L71
Should we leave it on next?

michamos:
yes, this will go away once we switch off legacy.

PR #1681, opened 2021-01-29
editor: move authorlist from next

MJedr:
Ref:#1621

### DIFF
--- a/backend/inspirehep/editor/authorlist_utils.py
+++ b/backend/inspirehep/editor/authorlist_utils.py
@@ -1,0 +1,347 @@
+import re
+
+from inspire_schemas.api import LiteratureBuilder
+from refextract.documents.pdf import replace_undesirable_characters
+
+re_emptyline = re.compile(r"\n\s*\n", re.UNICODE)
+re_hyphens = re.compile(
+    r"(\\255|\u02D7|\u0335|\u0336|\u2212|\u2013|\u002D|\uFE63|\uFF0D)", re.UNICODE
+)
+re_multiple_space = re.compile(r"\s{2,}", re.UNICODE)
+re_potential_key = re.compile(r"^(?:\d|[^\w.'-])+$", re.UNICODE)
+re_trailing_nonword = re.compile(r"((?:\d|[^\w,.'-])+ )", re.UNICODE)
+re_symbols = re.compile(r"[^\w ]", re.UNICODE)
+
+
+def split_id(word):
+    """
+    Separate potential aff-ids .
+    E.g.: '*12%$' -> ['*', '12' '%', '$']
+    """
+
+    aff_ids = []
+
+    symbols = re_symbols.findall(word)
+    if symbols:
+        aff_ids += symbols
+        for rest in re_symbols.split(word):
+            if rest:
+                aff_ids.append(rest)
+    else:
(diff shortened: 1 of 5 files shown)


backend/inspirehep/editor/utils.py
```diff
@@ -0,0 +1,345 @@
+import re
```
michamos, line 1: call this `authorlist.py`? it's not generic utils, so the file name is a bit of a misnomer.

backend/inspirehep/editor/views.py
```diff
@@ -250,3 +251,14 @@ def upload():
...
+
+@blueprint.route("/authorlist/text", methods=["POST"])
+@login_required_with_roles([Roles.cataloger.value])
+def authorlist_text():
+    """Run authorlist on a piece of text."""
+    try:
+        parsed_authors = authorlist(request.json["text"])
+        return jsonify(parsed_authors)
+    except Exception as err:
+        return jsonify(status=500, message=" / ".join(err.args)), 500
```
michamos, line 264: wouldn't this be better as a `400`, as it's not really something that blew up, but rather incorrect input that the script doesn't understand. Or does the record editor expect a `500` here?

MJedr: Yes, it's required by editor but it seems that it can be changed easily:
https://github.com/inspirehep/inspirehep/blob/f9c41f9578cd0c6aea6f4e2dc78236c8b5c34ea9/record-editor/src/app/shared/components/extract-actions/author-extract-actions.component.ts#L100-L109

record-editor/src/app/shared/components/extract-actions/author-extract-actions.component.ts
```diff
@@ -99,7 +97,7 @@ export class AuthorExtractActionsComponent {
 
   private onError(error: ApiError) {
     this.toastrService.clear(this.extractingToast.toastId);
-    if (error.status === 500 && error.message) {
+    if (error.status === 400 && error.message) {
```
michamos, line 100: do you need to do anything on `next` to take these changes into account (bump version or something like that)?

michamos (approved):
couple of small comments, but looks good!

record-editor/src/app/shared/components/extract-actions/author-extract-actions.component.ts
MJedr: I'll have to change it manually because authorlist functionality needs to stay at next to for the view that we are using (or we can leave status code 500).

backend/inspirehep/editor/authorlist_utils.py
```diff
@@ -0,0 +1,345 @@
...
+        return "digit"
+    warnings.append("CAUTION! Using symbols (# and stuff) as aff-IDs.")
+    return "symbol"
+
+
+def create_author_fullname(author_names, warnings):
+    fullname = " ".join(author_names)
+    if len(author_names) == 1:
+        warnings.append("Author without firstname: %s" % fullname)
+    return fullname
```
pazembrz, line 68: IMHO it should be cleaner to not use warnings from outside, but create here a separate warnings instance and return it together with fullname. When you can `extend` outside warnings list.

Issue #1676, opened 2021-01-28
keywording API


michamos:
DESY have a whole process to assign keywords to records. This involves sending the records to experts, which assign keywords, which then need to be added back to the record. This requires two pieces of metadata: `_desy_bookkeeping` to manage the process and the `keywords` themselves.

Suggestion by @michamos : build a simple API to modify this information on a record. 

We should create `curation` module with a REST API prefix `/curation`.

We could have something like `PUT /curation/literature/<pid_value>/keywords` with a payload of the form:
```
{ 
"_desy_bookkeeping": {...},
"keywords": [...]
}
```
which would *append* an element to the `_desy_bookkeeping` array and *replace* all `keywords` with `schema` `INSPIRE` with the provided list of keywords (each element being optional, so you can provide only one or the other).

pazembrz:
Tech:
* Create new module called `curation` in HEP
* this module should have a view which will accept only PUT
* * `PUT /curation/literature/<pid_value>/keywords`
* As authentication use tokens assigned to their accounts.
* * Only curators should be allowed to use this API.
It should accept json with one or both keys:
`_desy_bookkeeping` - dictionary and `keywords`- list

`_desy_bookkeeping` should be added to `_desy_bookkeeping` list on the record as is.

If `keywords` are provided, remove from record all keywords with `INSPIRE` schema. Then add those new keywords with `INSPIRE` schema and without `source`. Then update records and commit changes. If there will be any error return it as a response in the API with proper text and status code.
If update was successful return 200 status code

Remember to [register](https://github.com/inspirehep/inspirehep/commit/415ee4f550307776db1fec437fa9ed7f08546bbf#diff-bafe36cadd61f15c41fb61105985bd5c72b3da49f96083e515001b621fddda47) this module views in setup.py

PR #1692, opened 2021-02-02
curation: add keyword api

MJedr:
Ref: #1676

### DIFF
--- a/backend/inspirehep/curation/views.py
+++ b/backend/inspirehep/curation/views.py
@@ -1,0 +1,70 @@
+# -*- coding: utf-8 -*-
+#
+# Copyright (C) 2019 CERN.
+#
+# inspirehep is free software; you can redistribute it and/or modify it under
+# the terms of the MIT License; see LICENSE file for more details.
+
+from flask import Blueprint, abort
+from invenio_db import db
+from jsonschema.exceptions import ValidationError
+from webargs import fields
+from webargs.flaskparser import FlaskParser
+
+from inspirehep.accounts.decorators import login_required_with_roles
+from inspirehep.accounts.roles import Roles
+from inspirehep.records.api import LiteratureRecord
+from inspirehep.serializers import jsonify
+
+blueprint = Blueprint("inspirehep_curation", __name__, url_prefix="/curation")
+parser = FlaskParser()
+
+
+@parser.error_handler
+def handle_error(error, req, schema, error_status_code, error_headers):
+    message = f"Incorrect input type for fields: {''.join(error.field_names)}"
+    abort(400, message)
+
+
+@blueprint.route("/literature/<int:pid_value>/keywords", methods=["PUT"])
+@login_required_with_roles([Roles.cataloger.value])
+@parser.use_args(
+    {
+        "_desy_bookkeeping": fields.Dict(required=False),
+        "keywords": fields.List(fields.String, required=False),
+    },
+    locations=("json",),
+)
+def add_keywords(args, pid_value):
+    keywords = args.get("keywords")
+    desy_bookkeeping = args.get("_desy_bookkeeping")
+    if not keywords and not desy_bookkeeping:
+        return (
+            jsonify(
+                success=False,
+                message="None of required fields was passed",
+            ),
+            400,
+        )
+
+    record = LiteratureRecord.get_record_by_pid_value(pid_value)
+
+    if desy_bookkeeping:
+        record_desy_bookkeeping = record.get("_desy_bookkeeping", [])
+        record_desy_bookkeeping.append(desy_bookkeeping)
+        record["_desy_bookkeeping"] = record_desy_bookkeeping
+
+    if keywords:
+        record_keywords = record.get("keywords", [])
+        other_keywords = [
+            keyword for keyword in record_keywords if keyword.get("schema") != "INSPIRE"
+        ]
+        for keyword in keywords:
+            other_keywords.append({"value": keyword, "schema": "INSPIRE"})
+        record["keywords"] = other_keywords
+    try:
+        record.update(dict(record))
+        db.session.commit()
+        return jsonify(success=True)
+    except ValidationError as exception:
+        return jsonify(success=False, message=exception.message), 400
--- a/backend/pyproject.toml
+++ b/backend/pyproject.toml
@@ -157,6 +157,7 @@ inspirehep_tools = "inspirehep.tools.views:blueprint"
 inspirehep_assign = "inspirehep.assign.views:blueprint"
 inspirehep_matcher = "inspirehep.matcher.views:blueprint"
 invenio_oaiserver = "invenio_oaiserver.views.server:blueprint"
+inspirehep_curation = "inspirehep.curation.views:blueprint"
 
 [tool.poetry.plugins."invenio_config.module"]
 inspirehep = "inspirehep.config"
--- a/backend/tests/integration/curation/test_curation_views.py
+++ b/backend/tests/integration/curation/test_curation_views.py
@@ -1,0 +1,176 @@
+# -*- coding: utf-8 -*-
+#
+# Copyright (C) 2019 CERN.
+#
+# inspirehep is free software; you can redistribute it and/or modify it under
+# the terms of the MIT License; see LICENSE file for more details.
+
+import orjson
+from helpers.utils import create_record, create_user
+from invenio_accounts.testutils import login_user_via_session
+
+from inspirehep.accounts.roles import Roles
+from inspirehep.records.api import LiteratureRecord
+
+
+def test_add_keywords_replace_old_keywords_with_new(inspire_app):
+    user = create_user(role=Roles.cataloger.value)
+    record = create_record(
+        "lit", data={"keywords": [{"value": "Machine learning", "schema": "INSPIRE"}]}
+    )
+    with inspire_app.test_client() as client:
+        login_user_via_session(client, email=user.email)
+        response = client.put(
+            f""""/curation/literature/{record["control_number"]}/keywords""",
+            content_type="application/json",
+            data=orjson.dumps({"keywords": ["Deep Learning"]}),
+        )
+    updated_record = LiteratureRecord.get_record_by_pid_value(record["control_number"])
+
+    assert response.status_code == 200
+    assert updated_record["keywords"] == [
+        {"value": "Deep Learning", "schema": "INSPIRE"}
+    ]
+
+
+def test_add_keywords_updated_desy_bookkeeping(inspire_app):
+    user = create_user(role=Roles.cataloger.value)
+    record = create_record(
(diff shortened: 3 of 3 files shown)


backend/inspirehep/curation/views.py
```diff
@@ -0,0 +1,63 @@
...
+
+    datatype_warnings = check_keyword_dict_datatypes(json, required_fields)
+    if datatype_warnings:
+        return (
+            jsonify(
+                success=False,
+                message=f"""{" ".join(datatype_warnings)}""",
+            ),
+            400,
+        )
```
michamos, line 45: this is not something you should be doing manually but use `webargs` for this (validate the payload based on marshmallow schema).

backend/inspirehep/curation/views.py
```diff
@@ -0,0 +1,63 @@
...
+
+    if desy_bookkeeping:
+        record_desy_bookkeeping = record.get("_desy_bookkeeping", [])
+        record_desy_bookkeeping.append(desy_bookkeeping)
+
+    builder = LiteratureBuilder()
+    for keywords in keywords:
+        builder.add_keyword(keywords)
+    if builder.record.get("keywords"):
+        record["keywords"] = builder.record.get("keywords")
```
michamos, line 59: this is incorrect, you should remove all keywords with `schema: INSPIRE` (don't touch the others!) and add all of the provided keywords with no `source` and `schema: INSPIRE`.

backend/inspirehep/curation/views.py
```diff
@@ -0,0 +1,65 @@
...
+
+@blueprint.route("/literature/<int:pid_value>/keywords", methods=["PUT"])
+@login_required_with_roles([Roles.cataloger.value])
+@parser.use_kwargs(
+    {
+        "keywords": fields.List(fields.String, required=False),
+        "_desy_bookkeeping": fields.Dict(required=False),
+    }
+)
+def add_keywords(pid_value, **kwargs):
```
drjova, line 36:
```suggestion
def add_keywords(pid_value, keywords, _desybookkeeping):
```

backend/inspirehep/curation/views.py
```diff
@@ -0,0 +1,65 @@
...
+def handle_error(error, req, schema, error_status_code, error_headers):
+    message = f"""Incorrect input type for fields: {" ".join(error.field_names)}"""
+    abort(400, message)
+
+
+@blueprint.route("/literature/<int:pid_value>/keywords", methods=["PUT"])
+@login_required_with_roles([Roles.cataloger.value])
+@parser.use_kwargs(
+    {
+        "keywords": fields.List(fields.String, required=False),
```
drjova, line 32:
```suggestion
        "keywords": fields.List(fields.String, required=False, missing=[]),
```

backend/inspirehep/curation/views.py
```diff
@@ -0,0 +1,65 @@
...
+        "_desy_bookkeeping": fields.Dict(required=False),
```
drjova, line 33:
```suggestion
        "_desy_bookkeeping": fields.Dict(required=False, missing={}),
```

backend/inspirehep/curation/views.py
```diff
@@ -0,0 +1,65 @@
...
+    }
+)
+def add_keywords(pid_value, **kwargs):
+    keywords = kwargs.get("keywords", [])
+    desy_bookkeeping = kwargs.get("_desy_bookkeeping", {})
```
michamos, line 38: providing a default argument to `get` is useful in case you're going to process it further no matter what, but here you're doing explicit `if` checks later, so it's not really needed.

backend/inspirehep/curation/views.py
```diff
@@ -0,0 +1,65 @@
...
+from webargs import fields
+from webargs.flaskparser import FlaskParser
+
+from inspirehep.accounts.decorators import login_required_with_roles
+from inspirehep.accounts.roles import Roles
+from inspirehep.records.api import LiteratureRecord
+from inspirehep.serializers import jsonify
+
+blueprint = Blueprint("inspirehep_curation", __name__, url_prefix="/curation")
+parser = FlaskParser()
```
drjova, line 19: why do we need this? `from webargs.flaskparser import parser` should be enough no?

backend/inspirehep/curation/views.py
```diff
@@ -0,0 +1,65 @@
...
+                message="None of required fields was passed",
+            ),
+            400,
+        )
+
+    record = LiteratureRecord.get_record_by_pid_value(pid_value)
+
+    if desy_bookkeeping:
+        record_desy_bookkeeping = record.get("_desy_bookkeeping", [])
+        record_desy_bookkeeping.append(desy_bookkeeping)
```
drjova, line 53: the record is never updated with these values, or I'm missing something?

backend/inspirehep/curation/views.py
```diff
@@ -0,0 +1,65 @@
...
+from inspirehep.records.api import LiteratureRecord
+from inspirehep.serializers import jsonify
+
+blueprint = Blueprint("inspirehep_curation", __name__, url_prefix="/curation")
+parser = FlaskParser()
+
+
+@parser.error_handler
+def handle_error(error, req, schema, error_status_code, error_headers):
+    message = f"""Incorrect input type for fields: {" ".join(error.field_names)}"""
```
drjova, line 24:
```suggestion
    message = f"Incorrect input type for fields: {''.join(error.field_names)}"
```

backend/inspirehep/curation/views.py
```diff
@@ -0,0 +1,65 @@
...
+    abort(400, message)
+
+
+@blueprint.route("/literature/<int:pid_value>/keywords", methods=["PUT"])
+@login_required_with_roles([Roles.cataloger.value])
+@parser.use_kwargs(
```
michamos, line 30: Minor: I think using `use_args` is a little bit more clear in this case, as you don't have `required` fields, so you need to use `**kwargs` (instead of `args`).

backend/inspirehep/curation/views.py
michamos, line 53:
```suggestion
       record.setdefault("_desy_bookkeeping", []).append(desy_bookkeeping)
```
your version has a bug in case `_desy_bookkeeping` is not part of the record yet (you should maybe add a test for that): the first line will create an empty list, the second line will append to it, but that list is never assigned to the record.

backend/inspirehep/curation/views.py
```diff
@@ -0,0 +1,65 @@
...
+    {
+        "keywords": fields.List(fields.String, required=False),
+        "_desy_bookkeeping": fields.Dict(required=False),
+    }
+)
+def add_keywords(pid_value, **kwargs):
+    keywords = kwargs.get("keywords", [])
+    desy_bookkeeping = kwargs.get("_desy_bookkeeping", {})
+
+    if not any([keywords, desy_bookkeeping]):
```
michamos, line 40: I would go for the explicit version here, as there are only 2 variables:
```suggestion
    if not keywords and not desy_bookkeeping:
```

backend/inspirehep/curation/views.py
MJedr: I am overriding parser's `handle_error` method, it won't work without that

backend/inspirehep/curation/views.py
```diff
@@ -0,0 +1,65 @@
...
+
+    if desy_bookkeeping:
+        record_desy_bookkeeping = record.get("_desy_bookkeeping", [])
+        record_desy_bookkeeping.append(desy_bookkeeping)
+
+    if keywords:
+        record_keywords = record.get("keywords", [])
+        for record_keyword in record_keywords:
+            if record_keyword.get("schema") == "INSPIRE":
+                record_keywords.remove(record_keyword)
```
michamos, line 59: this is accidentally quadratic. Better do
```suggestion
    other_keywords = [keyword for keyword in record_keywords if keyword.get("schema") != "INSPIRE"]
    record["keywords"] = other_keywords
```
(and later append to `other_keywords` on L61).

backend/inspirehep/curation/views.py
```diff
@@ -0,0 +1,65 @@
...
+        for keyword in keywords:
+            record_keywords.append({"value": keyword, "schema": "INSPIRE"})
+
+    record.update(dict(record))
```
michamos, line 63: what happens in case of validation error here (e.g. passing an empty string as one of the keywords, repeated keywords, wrong structure in `_desy_bookkeeping`, etc.)? I think that should be handled in order to display a nice error message rather than the generic error message for 500 internal error.

backend/pyproject.toml
```diff
@@ -157,6 +157,7 @@ inspirehep_tools = "inspirehep.tools.views:blueprint"
 inspirehep_assign = "inspirehep.assign.views:blueprint"
 inspirehep_matcher = "inspirehep.matcher.views:blueprint"
 invenio_oaiserver = "invenio_oaiserver.views.server:blueprint"
+invenio_curation = "inspirehep.curation.views:blueprint"
```
michamos, line 160:
```suggestion
inspirehep_curation = "inspirehep.curation.views:blueprint"
```

backend/inspirehep/curation/views.py
michamos: why do you need to override `handle_error`?

MJedr: To return custom message and status code, without that it would be `422 Unprocessable Entity` without detailed message

backend/inspirehep/curation/views.py
```diff
@@ -0,0 +1,67 @@
...
+        ]
+        for keyword in keywords:
+            other_keywords.append({"value": keyword, "schema": "INSPIRE"})
+        record["keywords"] = other_keywords
+    try:
+        record.update(dict(record))
+        db.session.commit()
+        return jsonify(success=True)
+    except ValidationError as exception:
+        return jsonify(success=False, message=exception.message), 500
```
michamos, line 67: I would say this is a 400 for a wrong request, not a generic server internal error.

michamos (approved):
minor comment, looks good otherwise :+1:

backend/inspirehep/curation/views.py
```diff
@@ -0,0 +1,67 @@
...
+@parser.error_handler
+def handle_error(error, req, schema, error_status_code, error_headers):
+    message = f"Incorrect input type for fields: {''.join(error.field_names)}"
+    abort(400, message)
+
+
+@blueprint.route("/literature/<int:pid_value>/keywords", methods=["PUT"])
+@login_required_with_roles([Roles.cataloger.value])
+@parser.use_args(
+    {"_desy_bookkeeping": fields.Dict(), "keywords": fields.List(fields.String)},
```
michamos, line 32: don't you need the `required=False` any longer?

Issue #1666, opened 2021-01-26
Migrator deleted records


drjova:
On [https://github.com/inspirehep/inspirehep/blob/master/backend/inspirehep/migrator/tasks.py](migrator) we delete `pids` for 
`deleted_records` blindly without checking if the record even exist on hep or has `deleted:True`.  

More details:
https://inspirehep.zulipchat.com/#narrow/stream/122199-ops/topic/bugninja/near/224035593

pazembrz:
Tech Notes:
Instead of [doing this](https://github.com/inspirehep/inspirehep/blob/f9c41f9578cd0c6aea6f4e2dc78236c8b5c34ea9/backend/inspirehep/migrator/tasks.py#L470-L472), deleted record should be loaded and then method `record.delete()` should be called to properly handle deletion. 
Watch out as record which should be deleted might be missing in inspire (process this exception so it won't stop migration).

Issue #1656, opened 2021-01-20
Redirect of deleted records


drjova:
The problem seems to be that that https://github.com/inspirehep/inspirehep/blob/03d47172983da1531a657ab6ed6b798c6a9e1245/backend/inspirehep/records/api/base.py#L204 is following redirects

PR #1695, opened 2021-02-03
#1666  and #1656

pazembrz:


backend/inspirehep/records/api/base.py
```diff
@@ -363,6 +366,22 @@ def redirect_pids(self, pids):
                 self.redirect_pid(pid_type, pid_value)
             return pids
 
+    @classmethod
+    def delete_records_from_record_deleted(cls, data):
```
drjova, line 370:
```suggestion
    def delete_records_from_ deleted_records(cls, data):
``` 
:)

backend/inspirehep/records/api/base.py
```diff
@@ -387,6 +406,7 @@ def update(self, data, *args, **kwargs):
             if data.get("deleted"):
                 self.pidstore_handler.delete(self.id, self)
             else:
+                self.delete_records_from_record_deleted(data)
```
drjova, line 409: what about create? 🤔

drjova:
@pazembrz 👏

backend/inspirehep/records/api/base.py
pazembrz: :+1:

backend/inspirehep/records/api/base.py
pazembrz: Ah, you are right :+1:  . It's OK for literature (as it's calling update inside create) but not for other types.

backend/inspirehep/records/api/base.py
```diff
@@ -363,6 +366,22 @@ def redirect_pids(self, pids):
...
+        # Hack for migrator in case new record takes pids from other records
+        # which should be deleted but they are not deleted yet.
+        for pid in data.get("deleted_records", []):
+            pid_type, pid_value = PidStoreBase.get_pid_from_record_uri(pid["$ref"])
+            try:
+                record_to_delete = cls.get_record_by_pid_value(
+                    pid_value, pid_type, original_record=True
+                )
+            except PIDDoesNotExistError:
+                # Pid is not registered!
```
drjova, line 380: do we need this comment? :)

drjova (approved):
just the comment!

backend/inspirehep/records/api/base.py
pazembrz: Comment is not needed, but I added a log as this is potential error/bug so we want to have some kind of trace that this situation happened.

PR #1698, opened 2021-02-03
records: fix versioning issues

drjova:
**Please do not review and do not merge**

### DIFF
--- a/backend/inspirehep/records/api/data.py
+++ b/backend/inspirehep/records/api/data.py
@@ -17,3 +17,23 @@ class DataRecord(CitationMixin, InspireRecord):
     es_serializer = DataElasticSearchSchema
     pid_type = "dat"
     pidstore_handler = PidStoreData
+
+    @classmethod
+    def create(
+        cls,
+        data,
+        *args,
+        **kwargs,
+    ):
+        record = super().create(data, **kwargs)
+        record.update_refs_in_citation_table()
+        return record
+
+    def update(
+        self,
+        data,
+        *args,
+        **kwargs,
+    ):
+        super().update(data)
+        self.update_refs_in_citation_table()
--- a/backend/inspirehep/records/api/literature.py
+++ b/backend/inspirehep/records/api/literature.py
@@ -91,12 +91,42 @@ class LiteratureRecord(
             date = self.created.strftime("%Y-%m-%d")
         return date
 
+    def update_record_relationships(self):
+        self.update_authors_records_table()
+        self.update_refs_in_citation_table()
+        self.update_conference_paper_and_proccedings()
+        self.update_institution_relations()
+        self.update_experiment_relations()
+
     @classmethod
-    def create(cls, data, disable_orcid_push=False, *args, **kwargs):
+    def create(
+        cls,
+        data,
+        disable_orcid_push=False,
+        disable_relations_update=False,
+        *args,
+        **kwargs,
+    ):
+        LiteratureRecord.update_authors_signature_blocks_and_uuids(data)
+        LiteratureRecord.update_refs_to_conferences(data)
+        data = LiteratureRecord.add_files(data)
+
         with db.session.begin_nested():
             record = super().create(data, **kwargs)
-            record.update(dict(record), disable_orcid_push=disable_orcid_push, **kwargs)
-            return record
+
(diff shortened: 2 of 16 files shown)


michamos:
:see_no_evil:

backend/inspirehep/records/api/literature.py
```diff
@@ -288,56 +328,56 @@ def push_authors_phonetic_blocks_to_redis(self):
...
         if "deleted" in data and data["deleted"]:
-            LOGGER.info("Record is deleted, skipping files.", uuid=self.id)
+            LOGGER.info("Record is deleted, skipping files.")
             return data
 
         documents = data.pop("documents", [])
         figures = data.pop("figures", [])
 
-        added_documents = self.add_documents(documents)
+        added_documents = LiteratureRecord.add_documents(documents)
```
pazembrz, line 344: Isn't better to make this `classmethod` and call it like that?
```suggestion
        added_documents = cls.add_documents(documents)
```

backend/inspirehep/records/api/literature.py
```diff
@@ -157,23 +184,36 @@ def update_refs_to_conferences(data):
...
+
+        if disable_orcid_push:
+            LOGGER.info(
+                "Record ORCID PUSH disabled",
+                recid=self.get("control_number"),
+                uuid=str(self.id),
+            )
+        else:
+            push_to_orcid(self)
+        self.push_authors_phonetic_blocks_to_redis()
```
pazembrz, line 216: This is almost identical to create() except super().update/super().create calls, can you maybe split this to two separate methods which will be called before and after super().update/super().create calls? So we won't have to keep two identical peaces of code separate?

drjova:
@pazembrz updated

backend/inspirehep/records/api/data.py
```diff
@@ -17,3 +17,23 @@ class DataRecord(CitationMixin, InspireRecord):
...
+        record.update_refs_in_citation_table()
+        return record
+
+    def update(
+        self,
+        data,
+        *args,
+        **kwargs,
+    ):
+        self.update_refs_in_citation_table()
```
pazembrz, line 38: This should be called after `super().update()` IMHO otherwise it will update data for current record not for updated one.

backend/inspirehep/records/api/mixins.py
```diff
@@ -332,18 +328,6 @@ def update_self_citations(self):
...
-
-        if disable_relations_update:
-            LOGGER.info(
-                "Record citation update disabled",
-                recid=self.get("control_number"),
-                uuid=str(self.id),
-            )
-        else:
-            self.update_refs_in_citation_table()
-
```
pazembrz, line 346: Was this an issue? As you removed this you added separate update for DataRecord class. This code is now duplicated in LiteratureRecord and in DataRecord. And actually it's error prone (as we can see). IMHO you should add `create` modethod here which calls `super().create(data)` and then `self.update_refs_in_citation_table` acordingly. You won't duplicate identical code then in two classes. You have already Mixin, use it.

backend/inspirehep/records/api/data.py
drjova: True thanks!

backend/inspirehep/records/api/literature.py
drjova: we need the model already

backend/inspirehep/records/api/mixins.py
drjova: This is a very valid point, but I would like to have a full control in the literature class and call the relations update manually.

backend/inspirehep/records/api/literature.py
```diff
@@ -288,56 +327,56 @@ def push_authors_phonetic_blocks_to_redis(self):
...
                 )
 
-    def add_files(self, data):
+    @staticmethod
+    def add_files(data):
         if not current_app.config.get("FEATURE_FLAG_ENABLE_FILES", False):
             LOGGER.info("Files are disabled")
             return data
 
         if "deleted" in data and data["deleted"]:
```
pazembrz, line 336: If you already here, can you change this?
```suggestion
        if data.get("deleted", False):
```

pazembrz (approved):

PR #1702, opened 2021-02-05
set etc/hosts for e2e to work with proper ui nginx config

pazembrz:
TEST

### DIFF
--- a/.github/workflows/integration.yml
+++ b/.github/workflows/integration.yml
@@ -226,6 +226,15 @@ jobs:
         with:
           ref: ${{ env.REF }}
 
+      - name: Pull images
+        run: >
+          docker pull inspirehep/hep:${{ env.REF_TAG }} &
+          docker pull inspirehep/next:latest &
+          docker pull inspirehep/next-assets:latest &
+          docker pull inspirehep/editor:${{ env.REF_TAG }} &
+          docker pull inspirehep/ui:${{ env.REF_TAG }} &
+          wait
+
       - name: Start hep worker container
         uses: ./.github/actions/docker-start-container
         with:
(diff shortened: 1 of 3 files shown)


.github/workflows/integration.yml
```diff
@@ -297,6 +306,9 @@ jobs:
           name: ui
           options: --mount type=bind,source=${{ github.workspace }}/ui/docker/nginx/config/local.conf,destination=/etc/nginx/conf.d/default.conf
 
+      - name: Set hosts on ui
+        run: docker exec ui sh -c 'echo -e "127.0.0.1 hep-web\n127.0.0.1 next-web\n127.0.0.1 record-editor" >> /etc/hosts'
```
benjamin-bergia, line 310: This is not the correct way to modify the hosts file.
https://docs.docker.com/engine/reference/commandline/run/#add-entries-to-container-hosts-file---add-host

.github/workflows/integration.yml
```diff
@@ -226,6 +226,15 @@ jobs:
         with:
           ref: ${{ env.REF }}
 
+      - name: Pull images
```
benjamin-bergia, line 229: What's the difference between this and adding `&` in the action?

benjamin-bergia:
All this changes so you don't need to add 
```yaml
network_mode: "host"
```
to the compose file?

.github/workflows/integration.yml
pazembrz: This pulls all images in parallel (for all actions). In actions you download only one image (for example `hep`)and then you run it, then you download next image *for example `next` this time) and then you run it. which might be significantly slower.

pazembrz:
> All these changes so you don't need to add
> 
> ```yaml
> network_mode: "host"
> ```
> 
> to the compose file?

Yes, because on my machine it will break as I have already PostgreSQL and ES running for other service on my machine (on their native ports) And I would like not to fight with this every-time I want to run tests ;) Otherwise I could run everything naively.

.github/workflows/integration.yml
pazembrz: :+1: Thanks, I completely forgot about that.

benjamin-bergia (approved):

Issue #1691, opened 2021-02-02
Add edit button in conferences search


StellaCh:
The edit button in Conferences detailed page https://inspirehep.net/conferences/1836988 needs to be displayed also in the conferences search page: https://inspirehep.net/conferences?sort=dateasc&size=25&page=1&start_date=upcoming&q= and it should be visible only to curators

**TECH NOTES:**
https://github.com/inspirehep/inspirehep/blob/aa66423414d75800ca120fb87feb7dd291ee34f1/backend/inspirehep/records/marshmallow/conferences/ui.py#L22
This line should be updated as follows.
```Python3
class ConferencesListSchema(CatalogerCanEditMixin, ConferencesPublicListSchema):
```

PR #1704, opened 2021-02-09
ui: display edit button in conference search results

oguzdemirbasci:
Ref: #1691


drjova:
@oguzdemirbasci we should update the e2e tests as well :)

oguzdemirbasci:
> @oguzdemirbasci we should update the e2e tests as well :)

Yeah, I have been looking which e2e tests needed to be updated. But screenshots are taken before login. So shouldn't they be the same after this update?

drjova:
@oguzdemirbasci I know it's not part of the issue but could you please write an e2e to test this functionality? we should also add for all collections that have the edit button

drjova (approved):

Issue #1712, opened 2021-02-15
Increase parallelism during continuous migration


michamos:
See https://inspirehep.zulipchat.com/#narrow/stream/195298-experts/topic/Edits.20not.20taking/near/226354725

PR #1714, opened 2021-02-15
migrator: size batches according to number of workers

michamos:
* Ref #1712.


backend/inspirehep/migrator/tasks.py
```diff
@@ -101,6 +101,20 @@ def read_file(source):
                 yield line
 
 
+def count_consumers_for_queue(queue_name):
```
michamos, line 104: I left this function here as no other piece of code needs this for now.

pazembrz (approved):
Looks nice and clean :+1:

Issue #1655, opened 2021-01-20
Don't escape `{` and `}` in LaTeX/BibTeX titles


michamos:
See https://inspirehep.zulipchat.com/#narrow/stream/195298-experts/topic/bibtex.20issues/near/223258988.

We have lots of titles with spurious `{` and `}` outside of math for historical reasons. These shouldn't be escaped, and it's too much work to correct the metadata, so we need to add a hack and avoid escaping them.

TECH NOTES:
* use a [conversion rule](https://pylatexenc.readthedocs.io/en/latest/latexencode/#pylatexenc.latexencode.UnicodeToLatexEncoder.conversion_rules) with a `RULE_DICT` containing `{ord("{"): "{", ord("}"): "}"}` on top of `defaults` in https://github.com/inspirehep/inspirehep/blob/1e5197109a13d7afe5fab116ca868a95b29aba48/backend/inspirehep/records/marshmallow/literature/utils.py#L44.
* while we're at it, unpin `pylatexenc` and use the latest upstream version as the required fix has been released.

PR #1716, opened 2021-02-16
latex: do not escape curly brackets + upgrade pylatexenc

pazembrz:
#1655


michamos (approved):
:+1: Note that this requires reindexing.

Issue #1642, opened 2021-01-12
HAL push


drjova:
On record create/update we should check if the record has `_export_to.HAL: true` and if it's `True` we should send a task that pushes this record to HAL with the following logic https://github.com/inspirehep/inspire-hal/blob/master/inspire_hal/bulk_push.py#L62-L88. 

Note: We could copy the logic of `tei` and `sword` functionality directly.

Blocked by https://github.com/inspirehep/inspirehep/issues/1641

**TECH NOTES**
1. create a HAL module
2. Copy sword and tei modules from here https://github.com/inspirehep/inspire-hal/tree/master/inspire_hal/core
2. Create a celery task which will do the push to hal using the logic from [here](https://github.com/inspirehep/inspire-hal/blob/master/inspire_hal/bulk_push.py#L62-L88)
3. Call the task in [create](https://github.com/inspirehep/inspirehep/blob/03d47172983da1531a657ab6ed6b798c6a9e1245/backend/inspirehep/records/api/literature.py#L92) and [update](https://github.com/inspirehep/inspirehep/blob/03d47172983da1531a657ab6ed6b798c6a9e1245/backend/inspirehep/records/api/literature.py#L157) methods of the LiteratureRecord

michamos:
When the push fails for some reason, Mathieu should be able to see the failure. Either we log it and then we build some periodic task to send the logs to the chat/to Mathieu by email, or we send failures immediately. It's probably better to decouple this from the application and have the application just do logs.

PR #1718, opened 2021-02-17
hal-push: Hal push module is implemented.

oguzdemirbasci:
Ref: #1642


backend/inspirehep/hal/tasks.py
```diff
@@ -0,0 +1,96 @@
...
+        backoff = (4 ** (self.request.retries + 1)) * 60
+
+        LOGGER.warning(
+            f"hal_push task raised an exception. Retrying in {backoff} secs.",
+            recid=record["control_number"],
+        )
+        raise self.retry(max_retries=3, countdown=backoff, exc=exc)
+    except Exception:
+        LOGGER.warning("hal_push task failed", recid=record["control_number"])
+        raise
```
michamos, line 62: actually I don't think you need any of this, and can simply use celery auto-retry on some specific exceptions, like we're doing in https://github.com/inspirehep/inspirehep/blob/9878076d6f7389fbb4358ab553f0a7e620fd31b5/backend/inspirehep/indexer/tasks.py#L58-L60

I think the ORCID stuff was written before we were using a compatible celery version and we never changed it.

backend/inspirehep/hal/utils.py
```diff
@@ -0,0 +1,176 @@
...
+    return {el["control_number"]: _get_hal_id(el) for el in institutions}
+
+
+def _get_hal_id(record):
+    for el in record.get("external_system_identifiers", []):
+        if el.get("schema") == "HAL":
+            return el["value"]
+
+
+def get_db_records(pids):
```
michamos, line 158: this function was used when HAL push was outside of INSPIRE as there was nothing better. You should remove this and use `InspireRecord.get_records_by_pids` instead.

backend/inspirehep/hal/tasks.py
```diff
@@ -0,0 +1,75 @@
...
+def _hal_push(record):
+    if "Literature" in record["_collections"] or "HAL Hidden" in record["_collections"]:
+        tei = convert_to_tei(record)
+
+        hal_id = ""
+        ids = record.get("external_system_identifiers", [])
+
+        for id_ in ids:
+            if id_["schema"] == "HAL":
+                hal_id = id_["value"]
```
michamos, line 66: we have a util for this, `get_values_for_schema` in `inspirehep_utils.record`.

backend/inspirehep/hal/tasks.py
```diff
@@ -0,0 +1,75 @@
...
+from inspirehep.records.api import LiteratureRecord
+
+LOGGER = structlog.getLogger()
+
+
+CELERY_HAL_PUSH_RETRY_ON_SWORD_EXCEPTIONS = (
+    RequestTimeOut,
+    NotFound,
+    PackagingFormatNotAvailable,
+    NotAcceptable,
```
michamos, line 29: does it make sense to retry on all exceptions? If the result can't possibly change, it makes no sense retrying. `RequestTimeOut` probably yes, the others I'm not sure.

backend/inspirehep/records/api/literature.py
```diff
@@ -175,6 +176,8 @@ def update(self, data, disable_orcid_push=False, *args, **kwargs):
                 push_to_orcid(self)
             self.push_authors_phonetic_blocks_to_redis()
 
+            push_to_hal(self)
```
michamos, line 179: we need a boolean flag so we can disable this in the migrator (like for `disable_orcid_push`). Question @drjova: rename `disable_orcid_push` -> `disable_external_push` and use it for both, or have a separate flag? I would be in favor of the first option to have a cleaner API, but it's up to you. @oguzdemirbasci you'll also need to adapt the migrator to perform the `hal_push` separately if needed in the same way we do the `orcid_push` currently.

backend/inspirehep/hal/tasks.py
```diff
@@ -0,0 +1,60 @@
...
+from inspirehep.records.api import LiteratureRecord
+
+LOGGER = structlog.getLogger()
+
+
+@shared_task(
+    bind=True,
+    retry_backoff=2,
+    retry_kwargs={"max_retries": 6},
+    autoretry_for=(RequestTimeOut,),
```
michamos, line 25: So this goes to the general `celery` queue. We probably need a separate queue for this with separate workers, so we can tune the amount of concurrency if needed (I'll ask HAL if they have restrictions).

backend/tests/unit/hal/test_core_tei.py
```diff
@@ -0,0 +1,101 @@
+# -*- coding: utf-8 -*-
```
michamos, line 1: I'd like to see a real test (with VCR) where we push a record to HAL, at least for the happy flow. Currently we have no idea if this whole thing is working at all.

backend/inspirehep/hal/tasks.py
michamos: I asked them and they have no limits currently. If we overwhelm them, they might need to put some in place. Still, I believe it's better to have a separate queue for this I believe to better see what's going on.

backend/inspirehep/records/api/literature.py
drjova: Totally agree, let's have one flag, we already have enough of them

backend/inspirehep/hal/api.py
```diff
@@ -0,0 +1,38 @@
...
+    if not current_app.config["FEATURE_FLAG_ENABLE_HAL_PUSH"]:
+        LOGGER.info("HAL push feature flag not enabled")
+        return
+
+    # Ensure there is a control number. This is not always the case because of broken store_record.
+    if "control_number" not in record:
+        return
+
+    if is_hal_set(record):
+        _send_push_task(kwargs={"recid": record["control_number"]})
```
pazembrz, line 33: As you already provide correct queue on task definition, you can easily replace it with just `current_celery_app.send_task("inspirehep.hal.tasks.hal_push, kwargs={"recid": record["control_number"]}) `
Or even better you can call it with:
`hal_push.delay(recid=record["control_number"])` Which is much more readable.

backend/tests/integration/hal/test_core_tei.py
```diff
@@ -0,0 +1,92 @@
...
+    db.session.commit()
+
+    yield
+
+    record = InspireRecord.get_record_by_pid_value(902725, "ins")
+    del record.json["external_system_identifiers"]
+    update(RecordMetadata).where(RecordMetadata.id == record.id).values(
+        json=record.json
+    )
+    db.session.commit()
```
pazembrz, line 52: I don't see this fixture to be used anywhere

backend/tests/integration/hal/test_core_sword.py
```diff
@@ -0,0 +1,100 @@
...
+
+    # hal update
+    receipt = _hal_push(record)
+    assert receipt
+    assert receipt.parsed
+
+    institute_record = InspireRecord.get_record_by_pid_value(912490, "ins")
+    institute_record.hard_delete()
+    record.hard_delete()
+    db.session.commit()
```
pazembrz, line 100: This test should probably be moved to integration-async. Also deleting record at the end is unnecessary.

backend/tests/integration/hal/conftest.py
```diff
@@ -0,0 +1,139 @@
...
+@pytest.fixture(autouse=True, scope="function")
+def assert_all_played(request, vcr_cassette):
+    """
+    Ensure that all all episodes have been played in the current test.
+    Only if the current test has a cassette.
+    """
+    yield
+
+
+@pytest.fixture
```
pazembrz, line 102: This fixture can be replaced with something like this:
```python3
data = json.loads(_get_fixture(filename))
return InspireRecord.create(data)
```

backend/inspirehep/hal/utils.py
```diff
@@ -0,0 +1,145 @@
...
+
+    pids = [("ins", str(pid)) for pid in affiliation_recids]
+    institutions = InspireRecord.get_records_by_pids(pids)
+
+    return {el["control_number"]: _get_hal_id(el) for el in institutions}
+
+
+def _get_hal_id(record):
+    for el in record.get("external_system_identifiers", []):
+        if el.get("schema") == "HAL":
```
pazembrz, line 144: Can you use here `get_values_for_schema` form `inspire-utils`?

backend/inspirehep/hal/utils.py
```diff
@@ -0,0 +1,145 @@
...
+        ...                 '$ref': '/api/conferences/972464',
+        ...             },
+        ...         },
+        ...     ],
+        ... }
+        >>> conference_record = get_conference_record(record)
+        >>> conference_record['control_number']
+        972464
+    """
+    pub_info = get_value(record, "publication_info.conference_record[0]")
```
pazembrz, line 92: you can use here `record.get_value("publication_info.conference_record[0]")`

backend/inspirehep/hal/utils.py
```diff
@@ -0,0 +1,145 @@
...
+    Args:
+        record(InspireRecord): a record.
+    Returns:
+        list(str): the HAL domains of the record.
+    Examples:
+        >>> record = {'inspire_categories': [{'term': 'Experiment-HEP'}]}
+        >>> get_domains(record)
+        ['phys.hexp']
+    """
+    terms = get_value(record, "inspire_categories.term", default=[])
```
pazembrz, line 125: you can use here `record.get_value("inspire_categories.term", default=[])`

backend/inspirehep/hal/utils.py
```diff
@@ -0,0 +1,145 @@
...
+        972464
+    """
+    pub_info = get_value(record, "publication_info.conference_record[0]")
+    if not pub_info:
+        return default
+
+    return InspireRecord.get_record_by_pid_value(get_recid_from_ref(pub_info), "con")
+
+
+def get_divulgation(record):
```
pazembrz, line 99: Returning 1/0 instead True/False this is is so "C" way, but it's not our API so we can't do anything with this... :/

backend/tests/integration/hal/conftest.py
```diff
@@ -0,0 +1,107 @@
...
+            "mq",
+            "hep-worker",
+            "hep-web",
+            "ui",
+            "next-web",
+            "next-worker",
+            "flower",
+        ),
+        "record_mode": record_mode,
+    }
```
drjova, line 68: @oguzdemirbasci we have already [configuration](https://github.com/inspirehep/inspirehep/blob/master/backend/tests/conftest.py#L12) for that you don't need it

drjova: neither the other two vcr config

drjova:
@oguzdemirbasci looks good, just the comments!

drjova (approved):

PR #1721, opened 2021-02-22
merge the backend and python base image

benjamin-bergia:
I merged both images and reordered a bit the steps.


backend/Dockerfile
```diff
@@ -5,22 +5,32 @@
 # inspirehep is free software; you can redistribute it and/or modify it under
 # the terms of the MIT License; see LICENSE file for more details.
 
-FROM inspirehep/python:3.6-with-certs
-RUN apt update && apt install poppler-utils -y
+FROM python:3.6.13
```
michamos, line 8: why aren't we using the latest python 3.6, but pinning 3.6.13?

benjamin-bergia: I am afraid that by letting 3.6 people running compose locally get older versions during their tests on their computers.

michamos: doesn't docker check every time if there's a new version? here the major drawback is that we don't automatically benefit from bug and security fixes, which I thought was the reason to inline the certificate stuff in the first place.

benjamin-bergia: main benefit is not maintaining two images. In any case we have to be strict on which version is used by people locally and in the WF. Even if newer tag were fetched automatically you would run the risk of testing different things locally, in your PR and on master. It will also simplify troubleshooting if we know for a fact that the basei mage didn't change. I don't think it's a good idea to go with a tag like 3.6 and updating a single tag in a single Dockerfile when needed is not that much trouble.

michamos: In principle patch versions of Python are perfectly backwards compatible, so it shouldn't cause issues. @drjova WDYT? pin the base python image or automatically get patch updates?

benjamin-bergia: Don't forget that we are testing and building prod images here. There is no dev or testing images anymore.

drjova (approved):

PR #1726, opened 2021-02-24
lockfile: update inspire-json-merger to 11.0.14

pazembrz:


michamos (approved):

drjova:
@pazembrz do we need this PR?

pazembrz:
@drjova We have inspire-json-merger on HEP so it would be good to keep it in sync. Here problem is that json-merger also has to be updated ( I will do this right now)

Issue #1559, opened 2020-11-24
Handle race condition when doing multiple updates


michamos:
Some tasks (indexing, disambiguation) are done asynchronously after updating the record, and rely on the following property:

> The previous run of the task was for the previous version of the record.

This property is used implicitly when diffing versions of the record in order to detect
* which references have changed and their records need reindexing;
* which authors have changed and need disambiguating;
* etc.

However, the property as stated is **false**, due to the asynchronicity: it's possible for two (or more) updates of the record to be made before the task has a chance to run, causing the diff of the changes on the DB not to correspond to the diff of successive runs of the tasks. This should be fixed, as it will cause issues (wrong citation counts when updating references, wrongly disambiguated authors, etc.).

We focus here on indexing only. For disambiguation new task will be created.

Tech notes in comment below https://github.com/inspirehep/inspirehep/issues/1559#issuecomment-754013872
~~**Tech Notes:**~~
~~Solution to this problem will be quite complex.~~
~~* Add 2 locks (redis) for record control number. One will be for indexing and second for disambiguation.~~
~~* Locks should be acquired in [here](https://github.com/inspirehep/inspirehep/blob/a325629aea4e8717997a26ec0f37b5d2e62ad313/backend/inspirehep/records/receivers.py#L37) before indexing and disambiguation will run. And they should be released at the end of indexing and disambiguation tasks. So next indexing and disambiguation won't run until previous one will finish.~~
~~* check which version is on ES currently (based on `updated` field from db and `updated` field on ES) and index/disambiguate ONLY if `record.updated` is NOT OLDER than `es_data['_updated]`.~~
~~Before executing indexing and disambiguation add a check `if InspireSearch.get_record_data_from_es(rec)['_updated'] <= rec.updated: then reindex/disambiguate`~~
~~Remember that `_updated` is in isoformat, so it should be parsed first.~~
~~* all `get_modified_*` methods should diff between `last_indexed_version` and `current_version`. To properly get this version is should be passed from `index_after_commit` receiver to index/disambiguation task.~~

michamos:
I don't think that would work, as sometimes we index due to changes not reflected in the DB record (e.g. citation count) which wouldn't update the `_updated` field.

pazembrz:
I Updated tech notes.

pazembrz:
We are changing the idea qhat and how we want to fix here

*Tech Notes* 
[...]

PR #1727, opened 2021-02-24
Fix indexing and versioning issues

pazembrz:
#1559


backend/inspirehep/records/api/base.py
```diff
@@ -410,21 +460,29 @@ def update(self, data, *args, **kwargs):
...
+                else:
+                    self.delete_records_from_deleted_records(data)
+                    self.pidstore_handler.update(self.id, self)
+                    if self.get("deleted_records"):
+                        self.redirect_pids(self["deleted_records"])
+                self.update_model_created_with_legacy_creation_date()
+                flag_modified(self.model, "json")
+                db.session.merge(self.model)
+                db.session.flush()
+                # After flush new version should be created.
```
drjova, line 480: could we add a feature flag for this?

pazembrz: Ah, yes. Right. I forgot about It...

backend/tests/integration-async/records/indexer/test_jobs.py
```diff
@@ -0,0 +1,29 @@
...
+        current_search.flush_and_refresh("records-jobs")
+        expected_records_count = 0
+        record_lit_es = JobsSearch().get_record(str(record.id)).execute().hits
+        assert expected_records_count == len(record_lit_es)
+
+    record = JobsRecord.create(faker.record("job"))
+    db.session.commit()
+
+    record.delete()
+    record.index(delay=False)
```
drjova, line 27: why not commit as the rest of the tests :)

drjova (commented):
**@pazembrz looks good 🥇 Just the comment and please squash the commits :)

backend/tests/integration-async/records/indexer/test_jobs.py
pazembrz: This one was moved directly from integration. And in integration we don't commit. I will update it accordingly to be consistent with other async tests.

drjova (approved):

Issue #1607, opened 2021-01-04
Handle superseded ORCIDs


michamos:
@tsgit reports that ORCID are increasingly deprecating some IDs and redirecting them to others. Example: https://orcid.org/0000-0001-7942-3265. We don't handle this case currently during the ORCID push. We should change that and:

* add the new ORCID on the author record
* delete the remote account/remote token for the ORCID as we won't be able to push with it
* contact the author to let them know they need to login again and re-enable the ORCID push for their new ORCID account

Tech notes:
1. In  https://github.com/inspirehep/inspire-service-orcid/blob/master/inspire_service_orcid/exceptions.py  add the exception `MovedPernamentlyException` with status code 301 (for orcid api exceptions reference look [here](https://github.com/ORCID/ORCID-Source/blob/master/orcid-api-web/tutorial/api_errors.md))
2. In orcid utils add a function which will take the old orcid, the new orcid and author recid as the arguments and:
    * Get the author recid by orcid (from pidstore) and then get author record
    * Add the new orcid to the `ids` list 
    * Delete the push access token for the old orcid from db 
    * Send an email to the author which orcid has been deprecated that he needs to login again and re-enable the ORCID push for their new ORCID account
4. In [`orcid_push` task](https://github.com/inspirehep/inspirehep/blob/4ed3692a05fc6d061715eb3dd9b0a572ea56a155/backend/inspirehep/orcid/tasks.py#L197) raise exception on `MovedPernamentlyException`, retrieve old orcid from the exception message (it will look like that ```<developer-message>301 Moved Permanently: This account is deprecated. Please refer to account: https://qa.orcid.org/0000-0000-0000-0000. ORCID https://qa.orcid.org/0000-1111-0000-0000</developer-message>```) and call the function created in the previous step with deprecated and new orcid

tsgit:
just this past week

0000-0002-8640-2574  -> 0000-0002-4563-5806
0000-0003-0346-5586  -> 0000-0001-6135-1144
0000-0001-5044-9584  -> 0000-0002-2258-7741
0000-0001-7942-3265  -> 0000-0001-7207-7793
0000-0003-4508-1169  -> 0000-0002-3097-2324

I updated the info on legacy. You may want to delete tokens.

StellaCh:
Content of the email to user:

subject: Deprecated ORCID in INSPIRE

Hello,

Your ORCID <old ORCID> has been deprecated and has been replaced by <new ORCID>.
[...]

PR #1740, opened 2021-03-03
orcid: handle deprecated orcid accounts

MJedr:
Ref: #1607

### DIFF
--- a/backend/inspirehep/orcid/tasks.py
+++ b/backend/inspirehep/orcid/tasks.py
@@ -30,6 +30,7 @@ from . import domain_models, exceptions
 
 LOGGER = structlog.getLogger()
 USER_EMAIL_EMPTY_PATTERN = "{}@FAKEEMAILINSPIRE.FAKE"
+ORCID_REGEX = r"\d{4}-\d{4}-\d{4}-\d{3}[0-9X]"
 
 
 def legacy_orcid_arrays():
--- a/backend/inspirehep/orcid/utils.py
+++ b/backend/inspirehep/orcid/utils.py
@@ -11,10 +11,12 @@
 from contextlib import contextmanager
 from itertools import chain
 
+import structlog
 from elasticsearch_dsl import Q
 from flask import current_app
 from inspire_dojson.utils import get_recid_from_ref
-from inspire_utils.record import get_values_for_schema
+from inspire_utils.record import get_value, get_values_for_schema
+from invenio_accounts.models import User
 from invenio_db import db
 from invenio_oauthclient.models import RemoteAccount, RemoteToken, UserIdentity
 from invenio_oauthclient.utils import oauth_link_external_id
@@ -25,9 +27,13 @@ from redis_lock import Lock
 from sqlalchemy import cast, type_coerce
 from sqlalchemy.dialects.postgresql import JSONB
 
+from inspirehep.mailing.api.authors import send_orcid_push_disabled_email
+from inspirehep.orcid.push_access_tokens import delete_access_token
 from inspirehep.pidstore.models import InspireRedirect
 from inspirehep.records.api import AuthorsRecord
 from inspirehep.search.api import LiteratureSearch
+
+LOGGER = structlog.getLogger()
 
 
 def _split_lists(sequence, chunk_size):
@@ -226,3 +232,32 @@ def distributed_lock(lock_name, expire=10, auto_renewal=True, blocking=False):
 
 class DistributedLockError(Exception):
     pass
+
+
+def update_moved_orcid(old_orcid, new_orcid):
+    author_record = AuthorsRecord.get_record_by_pid_value(old_orcid, "orcid")
+    if new_orcid not in get_value(author_record, "ids.value", []):
+        new_author_ids = [
+            {"schema": "ORCID", "value": new_orcid},
+            *author_record["ids"],
+        ]
+        author_record["ids"] = new_author_ids
+        author_record.update(dict(author_record))
+    remove_access_token_for_orcid_account(old_orcid, new_orcid)
+    db.session.commit()
+    LOGGER.info("ORCID updated", new_orcid=new_orcid, old_orcid=old_orcid)
+
+
+def remove_access_token_for_orcid_account(orcid, new_orcid):
+    user_orcid_account = RemoteAccount.query.filter(
+        RemoteAccount.user_id == UserIdentity.id_user, UserIdentity.id == orcid
+    ).one()
+    user_orcid_account.extra_data["orcid"] = new_orcid
+    user_orcid_account.extra_data["allow_push"] = False
+    UserIdentity.query.filter(
+        user_orcid_account.user_id == UserIdentity.id_user,
+    ).first().id = new_orcid
+    access_token = user_orcid_account.remote_tokens[0]
+    delete_access_token(access_token.access_token, new_orcid)
+    user_email = User.query.filter(User.id == user_orcid_account.user_id).first().email
+    send_orcid_push_disabled_email(user_email, orcid)
--- a/backend/inspirehep/mailing/api/authors.py
+++ b/backend/inspirehep/mailing/api/authors.py
@@ -1,0 +1,29 @@
+# -*- coding: utf-8 -*-
+#
+# Copyright (C) 2019 CERN.
+#
+# inspirehep is free software; you can redistribute it and/or modify it under
+# the terms of the MIT License; see LICENSE file for more details.
+
+import structlog
+from flask import current_app, render_template
+
+from inspirehep.mailing.providers.flask_mail import send_email
+
+LOGGER = structlog.getLogger()
+
+
+def get_orcid_push_disabled_email_html_content(orcid):
+    return render_template("mailing/authors/orcid_push_disabled.html", orcid=orcid)
+
+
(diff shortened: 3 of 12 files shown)


backend/inspirehep/orcid/utils.py
```diff
@@ -226,3 +233,37 @@ def distributed_lock(lock_name, expire=10, auto_renewal=True, blocking=False):
 
 class DistributedLockError(Exception):
     pass
+
+
+def update_moved_orcid(old_orcid, new_orcid):
+    try:
+        author_record = AuthorsRecord.get_record_by_pid_value(old_orcid, "orcid")
+        if new_orcid not in get_value(author_record, "ids.value"):
```
michamos, line 241:
```suggestion
        if new_orcid not in author_record.get_value("ids.value", []):
```

backend/inspirehep/orcid/utils.py
```diff
@@ -226,3 +233,37 @@ def distributed_lock(lock_name, expire=10, auto_renewal=True, blocking=False):
...
+            new_author_ids = [
+                {"schema": "ORCID", "value": new_orcid},
+                *author_record["ids"],
+            ]
+            author_record["ids"] = new_author_ids
+            author_record.update(dict(author_record))
+    except PIDDoesNotExistError:
+        LOGGER.info(
```
michamos, line 249: I don't think this should be logged with such a low log level. If we're trying to push to this ORCID, we necessarily need to have an author profile and account corresponding to this ORCID. So if you can't find it, something is really wrong and we want to know about it.

backend/inspirehep/orcid/utils.py
```diff
@@ -226,3 +233,37 @@ def distributed_lock(lock_name, expire=10, auto_renewal=True, blocking=False):
...
+            "ORCID account was moved, but no profile matching ORCID number found on INSPIRE",
+            orcid=old_orcid,
+        )
+    user_orcid_account = RemoteAccount.query.filter(
+        RemoteAccount.user_id == UserIdentity.id_user, UserIdentity.id == old_orcid
+    ).one_or_none()
```
michamos, line 255:
```suggestion
    ).one()
```
same reason, this better exist.

backend/inspirehep/orcid/utils.py
```diff
@@ -226,3 +233,37 @@ def distributed_lock(lock_name, expire=10, auto_renewal=True, blocking=False):
...
+    if user_orcid_account:
+        remove_access_token_for_orcid_account(user_orcid_account, old_orcid)
+    LOGGER.info("ORCID update", new_orcid=new_orcid, old_orcid=old_orcid)
+    db.session.commit()
+
+
+def remove_access_token_for_orcid_account(user_orcid_account, orcid):
+    access_token = RemoteToken.query.filter(
+        RemoteToken.id_remote_account == user_orcid_account.id
+    ).first()
```
michamos, line 265: a relation has been set up, so you can simply do
```suggestion
    access_token = user_orcid_account.remote_tokens[0]
```

backend/inspirehep/orcid/utils.py
```diff
@@ -226,3 +233,37 @@ def distributed_lock(lock_name, expire=10, auto_renewal=True, blocking=False):
...
+            "ORCID account was moved, but no profile matching ORCID number found on INSPIRE",
+            orcid=old_orcid,
+        )
+    user_orcid_account = RemoteAccount.query.filter(
+        RemoteAccount.user_id == UserIdentity.id_user, UserIdentity.id == old_orcid
+    ).one_or_none()
+    if user_orcid_account:
+        remove_access_token_for_orcid_account(user_orcid_account, old_orcid)
+    LOGGER.info("ORCID update", new_orcid=new_orcid, old_orcid=old_orcid)
+    db.session.commit()
```
michamos, line 259: `remove_access_token_for_orcid_account` already commits, because `delete_access_token` does: https://github.com/inspirehep/inspirehep/blob/9878076d6f7389fbb4358ab553f0a7e620fd31b5/backend/inspirehep/orcid/push_access_tokens.py#L64

backend/inspirehep/orcid/utils.py
```diff
@@ -226,3 +233,37 @@ def distributed_lock(lock_name, expire=10, auto_renewal=True, blocking=False):
...
+    except PIDDoesNotExistError:
+        LOGGER.info(
+            "ORCID account was moved, but no profile matching ORCID number found on INSPIRE",
+            orcid=old_orcid,
+        )
+    user_orcid_account = RemoteAccount.query.filter(
+        RemoteAccount.user_id == UserIdentity.id_user, UserIdentity.id == old_orcid
+    ).one_or_none()
+    if user_orcid_account:
+        remove_access_token_for_orcid_account(user_orcid_account, old_orcid)
```
michamos, line 257: there's something I hadn't thought of before: if we only delete the remote token, the user will still have a user account corresponding to that ORCID but without a valid token. However, if he tries to login again with his new ORCID, a new account will be created, but he won't be able to reuse the same email address. How should we deal with that @drjova? Deleting the whole user account seems like a better solution, but it might fail if the user has made any modifications to a record (e.g. jobs, seminars, author profile), as the SQLAlchemy transaction table has a foreign key on the user table. The whole account handling is kind of a mess currently.

backend/inspirehep/orcid/utils.py
MJedr: So you mean to raise an exception there?

backend/inspirehep/orcid/utils.py
```diff
@@ -226,3 +232,38 @@ def distributed_lock(lock_name, expire=10, auto_renewal=True, blocking=False):
...
+def update_moved_orcid(old_orcid, new_orcid):
+    author_record = AuthorsRecord.get_record_by_pid_value(old_orcid, "orcid")
+    if new_orcid not in get_value(author_record, "ids.value", []):
+        new_author_ids = [
+            {"schema": "ORCID", "value": new_orcid},
+            *author_record["ids"],
+        ]
+        author_record["ids"] = new_author_ids
+        author_record.update(dict(author_record))
+        db.session.commit()
```
michamos, line 246: why are you committing here? I think it's best to run this whole thing in a single transaction as we don't want to partially apply these changes.

backend/inspirehep/orcid/utils.py
```diff
@@ -226,3 +232,38 @@ def distributed_lock(lock_name, expire=10, auto_renewal=True, blocking=False):
...
+        "ORCID updated",
+        new_orcid=new_orcid,
+        old_orcid=old_orcid,
+        orcid_account=user_orcid_account.id,
+    )
+
+
+def remove_access_token_for_orcid_account(user_orcid_account, orcid, new_orcid):
+    access_token = user_orcid_account.remote_tokens[0]
+    delete_access_token(access_token.access_token, orcid)
```
michamos, line 261: you can't use this function as it commits at the end, and again, you want to do all of this in a single transaction.

backend/inspirehep/orcid/utils.py
```diff
@@ -226,3 +232,38 @@ def distributed_lock(lock_name, expire=10, auto_renewal=True, blocking=False):
...
+    user_orcid_account.extra_data["orcid"] = new_orcid
+    user_orcid_account.extra_data["allow_push"] = False
```
michamos, line 263: interesting that this works thanks to using the correct `Mutable` SQLAlchemy pattern, unlike what they do in `invenio-records` (@pazembrz will be please :smiley: ) https://github.com/inveniosoftware/invenio-oauthclient/blob/19f0e0c73e62ac1a85285612c754e290c4faa9fa/invenio_oauthclient/models.py#L53.

backend/inspirehep/orcid/tasks.py
```diff
@@ -260,6 +262,10 @@ def orcid_push(self, orcid, rec_id, oauth_token, kwargs_to_pusher=None):
         # thus backoff power 5 is (secs): [5, 25, 125, 625]
         backoff = 5 ** (self.request.retries + 1)
         raise self.retry(max_retries=4, countdown=backoff, exc=exc)
+    except MovedPermanentlyException as exc:
+        old_orcid, new_orcid = re.findall(ORCID_REGEX, exc.args[0])
+        update_moved_orcid(old_orcid, new_orcid)
```
michamos, line 267: wouldn't it be better to move this kind of business logic to the `domain_models` `push` function?

michamos (approved):
small comment regarding code organization, looks great otherwise :+1:

backend/inspirehep/orcid/utils.py
```diff
@@ -226,3 +232,37 @@ def distributed_lock(lock_name, expire=10, auto_renewal=True, blocking=False):
...
+    if new_orcid not in get_value(author_record, "ids.value", []):
+        new_author_ids = [
+            {"schema": "ORCID", "value": new_orcid},
+            *author_record["ids"],
+        ]
+        author_record["ids"] = new_author_ids
+        author_record.update(dict(author_record))
+    user_orcid_account = RemoteAccount.query.filter(
+        RemoteAccount.user_id == UserIdentity.id_user, UserIdentity.id == old_orcid
+    ).one()
```
michamos, line 248: Also I would move this query to the `remove_access_token_for_orcid_account` function, so it can take only `old_orcid` and `new_orcid` args.

PR #1744, opened 2021-03-08
github: build images on release

drjova:


.github/workflows/integration.yml
```diff
@@ -7,6 +7,8 @@ on:
     branches: [master]
 
 env:
+  MASTER: ${{ github.event.pull_request.head.sha || github.sha }}
```
benjamin-bergia, line 10: You don't seem to be using this anywhere
```suggestion
```

.github/workflows/integration.yml
```diff
@@ -424,15 +428,31 @@ jobs:
             docker logs $i >${name}.log
           done
 
+      - name: Get images and tags
+        id: images
+        run: |
+          hep_tag=${{ needs.backend_build.result == 'skipped' && 'master' || env.REF_TAG }}
+          editor_tag=${{ needs.editor_build.result == 'skipped' && 'master' || env.REF_TAG }}
+          ui_tag=${{ needs.ui_build.result == 'skipped' && 'master' || env.REF_TAG }}
```
benjamin-bergia, line 436: I would remove this and add:
`BACKEND_IMAGE`, `UI_IMAGE` and `EDITOR_IMAGE` to the **workflow** env. so we can reuse it in the steps where we start the containers.

benjamin-bergia (changes requested):

.github/workflows/integration.yml
```diff
@@ -499,8 +485,8 @@ jobs:
         uses: ./.github/actions/docker-tag-and-push
         if: ${{ github.event_name == 'pull_request_target' }}
         with:
-          image: ${{ matrix.images }}
-          pull-tag: ${{ env.REF_TAG }}
+          image: ${{ matrix.images.image }}
+          pull-tag: ${{ matrix.images.version }}
```
benjamin-bergia, line 489: images doesn't have a version field

.github/workflows/integration.yml
```diff
@@ -424,15 +428,31 @@ jobs:
...
+          ui_tag=${{ needs.ui_build.result == 'skipped' && 'master' || env.REF_TAG }}
+
+          images=''
+          images=($images '{"image":"inspirehep/hep","tag":"'"$hep_tag"'"}' )
+          images=($images '{"image":"inspirehep/ui","tag":"'"$ui_tag"'"}' )
+          images=($images '{"image":"inspirehep/editor","tag":"'"$editor_tag"'"}' )
+          images_json=$(echo $images | jq . | jq -sc .)
+
+          json=$(echo '{"include":'"$images_json"'}' | jq . | jq -sc .)
+          echo '::set-output name=images::'$json
```
benjamin-bergia, line 445: You are probably better off hardcoding the json here

benjamin-bergia:
Just to recap because I think it's currently wrong on this PR. Here is the tags we want:
- For the image that have been built in this workflow:
  - `ref-*`
  - `master`
  - `latest`
  - `build-latest` only ig they have a build stage
- For all the images used in the workflow even if not built
  - `master-*`

drjova:
@benjamin-bergia updated

benjamin-bergia (approved):

Issue #1729, opened 2021-02-25
record_uuid is None when trying to delete deleted_records in migration


michamos:
https://sentry.inspirebeta.net/inspire-qa/hep/issues/396169/?query=is%3Aunresolved

@pazembrz is saying: 
> 😸 Pawel Zembrzuski: Ah, it's bad. I will have to investigate this. https://github.com/inveniosoftware/invenio-pidstore/blob/8f6e99787da5a07d38712f7eb16146608304cdb9/invenio_pidstore/models.py#L119 why they even allow for null in there...
> 😸 Pawel Zembrzuski: After quick look. Maybe this is a problem, https://github.com/inspirehep/inspirehep/blob/e32babc9a14b9c0edacd5a28aad38e9ac67b16e1/backend/inspirehep/records/api/base.py#L181-L188 When a new deleted record is added it might happen that we will create a new pidstore "deleted" entry for a record without ID?

PR #1746, opened 2021-03-08
records api: fix pid creation for deleted legacy records

MJedr:
Ref: #1729

### DIFF
--- a/backend/tests/integration/records/api/test_api_base.py
+++ b/backend/tests/integration/records/api/test_api_base.py
@@ -727,3 +727,34 @@ def test_create_record_which_redirects_non_existing_pid_when_redirection_is_turn
     )
     assert LiteratureRecord.get_record_by_pid_value(rec["control_number"])
 
+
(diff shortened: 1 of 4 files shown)


backend/tests/integration/records/api/test_api_base.py
```diff
@@ -727,3 +727,32 @@ def test_create_record_which_redirects_non_existing_pid_when_redirection_is_turn
...
+    }
+
+    cls = InspireRecord.get_class_for_record(json_record)
+    record = cls.create_or_update(
+        json_record,
+        disable_orcid_push=True,
+        disable_relations_update=True,
+    )
+    assert record.id
+    assert InspireRecord.get_record_by_pid_value("1775082", "exp")
```
pazembrz, line 758: You can check here if record is correctly marked as deleted (in metadata), also check if it's PID status is also DELETED

pazembrz (approved):
LGTM :+1:

Issue #1661, opened 2021-01-22
bibtex: take earliest good publication info, not most complete


michamos:
See https://inspirehep.zulipchat.com/#narrow/stream/195298-experts/topic/bibtex.20issues/near/223653296.

We should change the code in https://github.com/inspirehep/inspirehep/blob/e9280dc55bf69d38516ecbf7ab241b6a91d467e8/backend/inspirehep/records/marshmallow/literature/bibtex.py#L132-L142 to take the earliest (chronological, position in the array as a fallback) `publication_info` having all of  `journal_title`, `journal_volume`, `page_start`/`artid`.

PR #1753, opened 2021-03-09
bibtex: get_best_publication_info function updated

oguzdemirbasci:
Ref: #1661

### DIFF
--- a/backend/inspirehep/records/marshmallow/literature/bibtex.py
+++ b/backend/inspirehep/records/marshmallow/literature/bibtex.py
@@ -135,11 +135,16 @@ class BibTexCommonSchema(BaseSchema):
             entry
             for entry in publication_info
             if entry.get("material", "publication") == "publication"
+            and entry.get("journal_title")
+            and entry.get("journal_volume")
+            and (entry.get("artid") or entry.get("page_start"))
         ]
         if not only_publications:
             return {}
 
-        return sorted(only_publications, key=len, reverse=True)[0]
+        return sorted(
+            only_publications, key=lambda item: item.get("year", float("inf"))
+        )[0]
 
     def get_authors_with_role_author(self, data):
         return self.get_authors_with_role(data.get("authors", []), "author")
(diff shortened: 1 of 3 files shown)


backend/inspirehep/records/marshmallow/literature/bibtex.py
```diff
@@ -135,11 +135,16 @@ def get_best_publication_info(data):
             entry
             for entry in publication_info
             if entry.get("material", "publication") == "publication"
+            and entry.get("journal_title", "")
+            and entry.get("journal_volume", "")
+            and (entry.get("artid", "") or entry.get("page_start", ""))
```
drjova, line 140:
```suggestion
            and entry.get("journal_title")
            and entry.get("journal_volume")
            and (entry.get("artid") or entry.get("page_start"))
```

backend/inspirehep/records/marshmallow/literature/bibtex.py
```diff
@@ -135,11 +135,16 @@ def get_best_publication_info(data):
...
         ]
         if not only_publications:
             return {}
 
-        return sorted(only_publications, key=len, reverse=True)[0]
+        return sorted(
+            only_publications, key=lambda item: item.get("year", float("inf"))
+        )[0]
```
drjova, line 147: why we need different sorted?

oguzdemirbasci: It should return the earliest one.

drjova: can year be `None`? if yes I would add `[] and entry.get("year")` and we'll not need the `float(inf)`

oguzdemirbasci: In that case, it throws an error; `None` type cannot be compared with `int` type. And I thought the items with a year info should come before that the ones without year info, that's why I put `float(inf)` by default.

drjova: you can ignore the ones without `year` https://github.com/inspirehep/inspirehep/pull/1753/files#diff-33e33d63c34a6b46d8b67464e2c19087c32e16705c5589cf3aef05d824f432f0R138

oguzdemirbasci: @michamos Should we ignore the publications with no year info even if there is no publication having year data? In that case, It will return `none` while there are publications having all of `journal_title`, `journal_volume`, `page_start/artid` but not `year`.

drjova (approved):

Issue #1765, opened 2021-03-15
[UI] Bulk interface for assign conferences


drjova:
Create a new action item 'tools' in literature, visible only to curators and superusers with the options: 'assign conference'

The prototype is found here: https://www.figma.com/proto/TUAMWktXlnKLOivYGg9p2V/Other-collections?node-id=344%3A5206&viewport=1676%2C306%2C0.3636794090270996&scaling=min-zoom

Add a checkbox on the left of each search item in literature. (This exists already for authors). A checkbox next to the number of results should also be added that allows users to select all results.
Assign conference: A drawer with a conference search is displayed (similar to the claiming interface for authors). When the user selects the appropriate conference, a confirmation modal appears.

UI:
* Add a new route `curation` here https://github.com/inspirehep/inspirehep/blob/master/ui/src/App.jsx, and use the same components there (`curation/literature` etc.).
* If user is `curator` or `superuser` https://github.com/inspirehep/inspirehep/blob/master/ui/src/literature/containers/LiteratureSearchContainer.jsx#L98 add a drop down (as described in the mockup)
  * Disable drop down if search results are more than `9999` results

* If user is `curator` or `superuser` https://github.com/inspirehep/inspirehep/blob/master/ui/src/literature/containers/SearchPage.jsx#L21 create an AssignVIewContext as https://github.com/inspirehep/inspirehep/blob/master/ui/src/authors/containers/AuthorPublicationsContainer.jsx#L27-L36
 and an AssignDrawer as https://github.com/inspirehep/inspirehep/blob/master/ui/src/authors/components/AssignDrawer/AssignDrawer.jsx
  * Add the confirmation box
  * We should pass to the API the current  query for assigning conferences if user select all otherwise we should pass the recids of the selected record. Display an alert with a message from the API response.

@StellaCh what message should we display on success or error?

StellaCh:
Finally, we are going forward with the option to create a new interface for curators. What do you need to refine this solution?

StellaCh:
Regarding 
>@StellaCh what message should we display on success or error?
It should be the same messages we display for the claiming interface when we assign papers to authors.

StellaCh:
As discussed, if I have 500 results and I select the first page with 250 results we should display:

- [ ] 250/500 results

Issue #1735, opened 2021-03-02
Add BAI to Literature record in disambiguation


michamos:
Currently, during disambiguation we set `authors.record` to the ref of the author profile. We should also add the BAI of the author record, as that gets used in the search, or have a different mechanism to maintain the BAI in sync with the author record.

**TECH NOTES**
1. In matcher configs `AUTHOR_MATCHER_NAME_CONFIG`, `AUTHOR_MATCHER_NAME_INITIALS_CONFIG` and `AUTHOR_MATCHER_EXACT_CONFIG` add `ids` to `_source`. 
2. Create a new function `assign_bai_to_author` which will take matched authors as an argument and will assign bai to an author. The bai should be in `inner_hits.authors.hits.hits._source.ids`.
3. If [author reference was found](https://github.com/inspirehep/inspirehep/blob/6d3133b75b610243dce084cba3298c8cd89f4f30/backend/inspirehep/disambiguation/tasks.py#L114) call the `assign_bai_to_author` with `matched_records` as an argument.

michamos:
Current solution is incorrect: BAI should be added in all cases, not only when there is an exact match based on author identifiers, like has been done in #1772.

PR #1772, opened 2021-03-16
disambiguation: assign bai to author

MJedr:
Ref: #1735, #1765
* Fix matcher validator errors

### DIFF
--- a/backend/inspirehep/disambiguation/tasks.py
+++ b/backend/inspirehep/disambiguation/tasks.py
@@ -3,7 +3,7 @@ from celery import shared_task
 from flask import current_app
 from inspire_matcher.api import match
 from inspire_utils.name import ParsedName
-from inspire_utils.record import get_value
+from inspire_utils.record import get_value, get_values_for_schema
 from invenio_db import db
 from prometheus_client import Counter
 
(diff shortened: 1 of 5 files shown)


backend/inspirehep/disambiguation/tasks.py
```diff
@@ -105,16 +105,30 @@ def assign_reference_to_author_if_unambiguous_match(matched_authors_refs, author
...
     matched_authors_refs = get_record_refs_for_matched_literature_authors(
-        matched_records
+        matched_record_list
     )
     matched_ref = assign_reference_to_author_if_unambiguous_match(
         matched_authors_refs, author
     )
     if matched_ref:
+        author_ids = get_value(
+            matched_record_list[0], "inner_hits.authors.hits.hits._source.ids", []
```
drjova, line 117: what if `len(matched_record_list) == 0`?

MJedr: then there will be no matched_ref

drjova: It will raise an `IndexError` though

MJedr: No, because `matched_ref` is True only in case when there is one author match. 
https://github.com/inspirehep/inspirehep/blob/1aefc405c244823a29ca7b40f8555831abf52097/backend/inspirehep/disambiguation/tasks.py#L98-L102
So if there is no match reference can't be assigned (and `None` is returned from `assign_reference_to_author_if_unambiguous_match`) and this part of the code won't be executed.

drjova (approved):

Issue #1768, opened 2021-03-15
[Backend] Bulk interface for assign conferences 


drjova:
BACKEND:
Refactor https://github.com/inspirehep/inspirehep/blob/master/backend/inspirehep/assign/views.py#L67-L84 to be author specific and add 2 new endpoints for assigning conferences to literature:
1. Endpoint that accepts `recids` and create a celery task which will create a chain of update tasks  
~~2. Endpoint that accepts a search query and create a celery task which gets the `recids` from ES  and create a chain of update tasks~~

The API should return `200`, `400`  or `500`

pazembrz:
Literature record should be assigned as `conference paper` to conference unless the paper already has `proceedings` among `document_type`

PR #1774, opened 2021-03-17
backend: Bulk interface for assign conferences

pazembrz:
#1768

### DIFF
--- a/backend/inspirehep/assign/tasks.py
+++ b/backend/inspirehep/assign/tasks.py
@@ -1,0 +1,141 @@
+# -*- coding: utf-8 -*-
+#
+# Copyright (C) 2021 CERN.
+#
+# inspirehep is free software; you can redistribute it and/or modify it under
+# the terms of the MIT License; see LICENSE file for more details.
+import structlog
+from celery import shared_task
+from inspire_schemas.builders import LiteratureBuilder
+from invenio_db import db
+from invenio_pidstore.errors import PIDDoesNotExistError
+from jsonschema import ValidationError
+from sqlalchemy.exc import (
+    DisconnectionError,
+    OperationalError,
+    ResourceClosedError,
+    TimeoutError,
+    UnboundExecutionError,
+)
+from sqlalchemy.orm.exc import NoResultFound, StaleDataError
+
+from inspirehep.records.api import ConferencesRecord, LiteratureRecord
+from inspirehep.records.errors import MissingArgumentError
+
+LOGGER = structlog.getLogger()
+
+
+@shared_task(
+    ignore_results=False,
+    queue="assign",
+    acks_late=True,
+    retry_backoff=2,
+    retry_kwargs={"max_retries": 6},
+    autoretry_for=(
+        NoResultFound,
+        StaleDataError,
+        DisconnectionError,
+        TimeoutError,
+        UnboundExecutionError,
+        ResourceClosedError,
+        OperationalError,
+    ),
+)
+def assign_paper_to_conference(literature_recids, conference_recid):
+    try:
+        conference = ConferencesRecord.get_record_by_pid_value(conference_recid)
+    except PIDDoesNotExistError:
+        LOGGER.exception(
+            "Cannot assign papers to conference. Conference does not exist.",
+            conference_recid=conference_recid,
+        )
+    conference_ref = conference.get("self")
+    cnum = conference.get("cnum")
+    for recid in literature_recids:
+        try:
+            record = LiteratureRecord.get_record_by_pid_value(recid)
+        except PIDDoesNotExistError:
+            LOGGER.error(
+                "Cannot assign record to conference. Record does not exist.",
+                literature_recid=recid,
+                conference_recid=conference_recid,
+            )
+            continue
+        with db.session.begin_nested():
+            try:
+                updated_data = assign_conference(record, conference_ref, cnum)
+                record.update(updated_data)
+            except ValidationError as err:
+                LOGGER.exception(
+                    "Cannot assign conference to paper.",
+                    recid=recid,
+                    cnum=cnum,
+                    conference_recid=conference_recid,
+                    err=err,
+                )
+            except MissingArgumentError:
+                LOGGER.error(
+                    "CNUM and conference $ref are required.",
+                    cnum=cnum,
+                    conference_ref=conference_ref,
+                    record_recid=recid,
+                )
+    db.session.commit()
+
+
+def assign_conference(record, conference_ref, cnum):
+    builder = LiteratureBuilder(record=record)
+    if not cnum:
+        raise MissingArgumentError("cnum is required.")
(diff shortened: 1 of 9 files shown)


backend/inspirehep/assign/tasks.py
```diff
@@ -0,0 +1,141 @@
...
+def assign_paper_to_conference(literature_recids, conference_recid):
+    try:
+        conference = ConferencesRecord.get_record_by_pid_value(conference_recid)
+    except PIDDoesNotExistError:
+        LOGGER.exception(
+            "Cannot assign papers to conference. Conference does not exist.",
+            conference_recid=conference_recid,
+        )
+    conference_ref = conference.get("self")
+    cnum = conference.get("cnum")
```
MJedr, line 53: Aren't they always present?

backend/inspirehep/assign/tasks.py
```diff
@@ -0,0 +1,141 @@
...
+
+
+def assign_conference(record, conference_ref, cnum):
+    builder = LiteratureBuilder(record=record)
+    if not cnum:
+        raise MissingArgumentError("cnum is required.")
+    if not conference_ref:
+        raise MissingArgumentError("$ref is required.")
+
+    if not {"proceedings", "conference paper"}.intersection(
```
MJedr, line 93: Why is that? For example if `document_type` is `article` why you are changing it?

MJedr (commented):
So finally we always pass recids to the task (even if all the records were selected)?

backend/inspirehep/assign/tasks.py
pazembrz: https://github.com/inspirehep/inspire-schemas/blob/afc2553e150fc83e72aa3187926caa126a0a096d/inspire_schemas/records/conferences.yml#L98-L100 https://github.com/inspirehep/inspire-schemas/blob/afc2553e150fc83e72aa3187926caa126a0a096d/inspire_schemas/records/conferences.yml#L265-L266
They are not mandatory. It's possible that some old conference is missing those.

backend/inspirehep/assign/tasks.py
pazembrz: > So finally we always pass recids to the task (even if all the records were selected)?
Yes.

pazembrz: I'm not changing this, I'm adding another document_type. This requirement was added when I was working on the task https://github.com/inspirehep/inspirehep/issues/1768#issuecomment-800209835

pazembrz:
> So finally we always pass recids to the task (even if all the records were selected)?

Yes.

backend/inspirehep/assign/views.py
```diff
@@ -64,9 +66,9 @@ def assign_to_author(from_author_recid, to_author_recid, literature_recids):
     unstub_author_by_recid(to_author_recid)
 
 
-@blueprint.route("", methods=["POST"])
+@blueprint.route("authors", methods=["POST"])
 @login_required_with_roles([Roles.cataloger.value])
-def assign_view():
+def author_assign_view():
```
drjova, line 71:
```suggestion
def authors_assign_view():
```

backend/inspirehep/assign/views.py
```diff
@@ -82,3 +84,37 @@ def assign_view():
...
+
+
+@blueprint.route("literature_conferences", methods=["POST"])
+@login_required_with_roles([Roles.cataloger.value])
+def literature_assign_conferences_view():
+    body = request.get_json()
+    conference_recid = body.get("conference_recid")
+    literature_recids = body.get("literature_recids")
+
+    if not conference_recid or not literature_recids:
```
drjova, line 96: Instead I would use `@parser.use_args` decorator to check the args

backend/inspirehep/assign/views.py
```diff
@@ -82,3 +84,37 @@ def assign_view():
     if to_author_recid is None:
         return jsonify({"stub_author_id": stub_author_id}), 200
     return jsonify({"message": "Success"}), 200
+
+
+@blueprint.route("literature_conferences", methods=["POST"])
```
drjova, line 89:
```suggestion
@blueprint.route("conference", methods=["POST"])
```

backend/inspirehep/assign/views.py
```diff
@@ -64,9 +66,9 @@ def assign_to_author(from_author_recid, to_author_recid, literature_recids):
     unstub_author_by_recid(to_author_recid)
 
 
-@blueprint.route("", methods=["POST"])
+@blueprint.route("authors", methods=["POST"])
```
drjova, line 69:
```suggestion
@blueprint.route("author", methods=["POST"])
```

backend/inspirehep/assign/views.py
pazembrz: But this is not bulk like action, we are assigning/replacing one author for another in multiple papers. I don't think that `authors` here will be correct.

backend/inspirehep/assign/views.py
pazembrz: :+1:

backend/inspirehep/assign/views.py
pazembrz: :+1:

backend/inspirehep/assign/views.py
pazembrz: :+1:

backend/inspirehep/assign/tasks.py
```diff
@@ -0,0 +1,141 @@
...
+    retry_kwargs={"max_retries": 6},
+    autoretry_for=(
+        NoResultFound,
+        StaleDataError,
+        DisconnectionError,
+        TimeoutError,
+        UnboundExecutionError,
+        ResourceClosedError,
+        OperationalError,
+    ),
```
drjova, line 42: aren't we use this errors everywhere? can we move them and reuse the same?

backend/inspirehep/assign/views.py
```diff
@@ -82,3 +87,43 @@ def assign_view():
...
+            400,
+        )
+
+    if not isinstance(literature_recids, list):
+        return (
+            jsonify(
+                {"message": "literature_recids has to be a list of literature recids"}
+            ),
+            400,
+        )
```
drjova, line 121: we don't need these checks with the parser

pazembrz: Ah, yes. You are right ;)

backend/inspirehep/assign/tasks.py
pazembrz: We use similar ones in migrator tasks, they are not 100% identical, so I would have to create a common group of exceptions and manually add rest of them in proper places. It's not used that often to be worth additional complexity IMHO.

drjova (approved):

drjova (approved):

Issue #1779, opened 2021-03-18
Run complete author disambiguation on update


michamos:
Currently, we check for the existence of a linked author `record` in the later steps of disambiguation as a proxy for whether the first disambiguation step based on identfiers produced a match. While this is correct for new signatures, which don't have a `record` field yet, it's wrong for updates where we might need to reassign the signature to a different record, even if a `record` field is already present because disambiguation has been run previously. Instead, there should be a more direct check in
https://github.com/inspirehep/inspirehep/blob/9d3ef0776defeed49edd0892c238606d78babd15/backend/inspirehep/disambiguation/tasks.py#L226-L233.

Also, `updated_record` seems not to be updated when matching other literature records.

PR #1785, opened 2021-03-19
disambiguation: fix assigning bai and disambiguation of updated author

MJedr:
Ref: #1779

### DIFF
--- a/backend/inspirehep/disambiguation/tasks.py
+++ b/backend/inspirehep/disambiguation/tasks.py
@@ -79,57 +79,35 @@ def match_literature_author_with_config(author_data, matcher_config):
     matched_records = [
         matched_record
         for matched_record in match(author_data, matcher_config)
-        if matched_record.get("inner_hits")
+        if get_value(
+            matched_record, "inner_hits.authors.hits.hits[0]._source.record.$ref"
+        )
     ]
     return matched_records
 
 
-def get_record_refs_for_matched_literature_authors(matched_records):
-    matched_authors_refs = set()
-    for record_authors in matched_records:
-        matched_authors_refs.update(
-            get_value(
-                record_authors, "inner_hits.authors.hits.hits._source.record.$ref", None
-            )
-        )
-    return matched_authors_refs
+def get_reference_and_bai_if_unambiguous_match(matched_refs_ids):
+    if len(matched_refs_ids) == 1:
+        author_reference, author_ids = matched_refs_ids.popitem()
+        author_bais = get_values_for_schema(author_ids, "INSPIRE BAI")
+        return {
+            "author_reference": author_reference,
+            "author_bai": author_bais[0] if author_bais else None,
+        }
 
 
-def assign_reference_to_author_if_unambiguous_match(matched_authors_refs, author):
-    if matched_authors_refs and len(matched_authors_refs) == 1:
-        author_reference_url = matched_authors_refs.pop()
-        author["record"] = {"$ref": author_reference_url}
-        return author_reference_url.split("/")[-1]
+def get_reference_and_bai_if_unambiguous_literature_author_match(matched_records):
+    matched_refs_ids = {
+        get_value(
+            matched_author, "inner_hits.authors.hits.hits[0]._source.record.$ref"
+        ): get_value(matched_author, "inner_hits.authors.hits.hits[0]._source.ids", [])
+        for matched_author in matched_records
+    }
+    matched_author_data = get_reference_and_bai_if_unambiguous_match(matched_refs_ids)
+    return matched_author_data
 
 
-def assign_reference_to_author_if_unambiguous_literature_author_match(
-    matched_records, author
-):
-    matched_record_list = list(matched_records)
-    matched_authors_refs = get_record_refs_for_matched_literature_authors(
-        matched_record_list
-    )
-    matched_ref = assign_reference_to_author_if_unambiguous_match(
-        matched_authors_refs, author
-    )
-    if matched_ref:
-        author_ids = get_value(
-            matched_record_list[0], "inner_hits.authors.hits.hits._source.ids", []
-        )
-        assign_bai_to_author(author_ids, author)
-        return matched_ref
-
-
-def assign_bai_to_author(author_ids, author):
-    author_bai_list = get_values_for_schema(author_ids, "INSPIRE BAI")
-    if author_bai_list and "INSPIRE BAI" not in get_value(author, "ids.schema", []):
-        author["ids"] = [
-            *author.get("ids", []),
-            {"schema": "INSPIRE BAI", "value": author_bai_list[0]},
-        ]
-
-
-def match_literature_author(author, updated_authors, record):
+def match_literature_author(author, record):
     configs = [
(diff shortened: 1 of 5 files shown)


backend/inspirehep/disambiguation/tasks.py
```diff
@@ -114,19 +114,28 @@ def assign_reference_to_author_if_unambiguous_literature_author_match(
...
+        return
+    literature_author["ids"] = [
+        *literature_author.get("ids", []),
+        {"schema": "INSPIRE BAI", "value": matched_author_bai_list[0]},
+    ]
+
+
+def get_literature_author_bai(literature_author):
+    literature_author_ids = literature_author.get("ids", [])
+    for idx, author_id_dict in enumerate(literature_author.get("ids", [])):
```
drjova, line 135: I'm not sure I understand this code, why couldn't we use https://github.com/inspirehep/inspire-utils/blob/master/inspire_utils/record.py#L94?

```python
get_values_for_schema(literature_author, 'INSPIRE BAI')
```

MJedr: The case is that I would like to remove the whole `{"schema": "INSPIRE BAI", "value" "A.Value"}` from the id list (because it can't be used later for matching) and `get_values_for_schema` is returning only the value

backend/inspirehep/disambiguation/tasks.py
```diff
@@ -114,19 +114,32 @@ def assign_reference_to_author_if_unambiguous_literature_author_match(
...
+
+def remove_author_bai_from_id_list(literature_author):
+    literature_author_ids = get_value(literature_author, "ids", [])
+    literature_author_bai_dict = [
+        bai_dict
+        for bai_dict in literature_author_ids
+        if bai_dict["schema"] == "INSPIRE BAI"
+    ]
+    if literature_author_bai_dict:
+        literature_author_ids.remove(literature_author_bai_dict[0])
```
michamos, line 141: don't do this. You're mutating the record, which might have unintended consequences.
Unless I'm misunderstanding something, I believe this whole function can be replaced by
```python
def author_ids_without_bais(literature_author):
    literature_author_ids = get_value(literature_author, "ids", [])
    return [
        bai_dict
        for bai_dict in literature_author_ids
        if bai_dict["schema"] != "INSPIRE BAI"
    ]
```

backend/inspirehep/disambiguation/tasks.py
```diff
@@ -206,32 +239,21 @@ def disambiguate_authors(self, record_uuid):
...
-        }
-        if len(matched_authors_refs) == 1:
-            updated_author_recid = assign_reference_to_author_if_unambiguous_match(
-                matched_authors_refs, author
+        author_bai = remove_author_bai_from_id_list(author)
+        updated_author_recid = match_author(author, updated_authors)
+        if not updated_author_recid:
+            updated_author_recid = match_literature_author(
+                author, updated_authors, record
             )
```
michamos, line 247: are you populating the BAI in this case? doesn't look like it. 

Additionally, I think this function is slowly becoming a big mess. From a high-level perspective, what you need to do here is:
1. match the current record against author records, ignoring the BAI among IDs,
2.  if there is no result, match against literature records,
3. and if still no results and there was no assigned record yet, create a new record
4. in all cases, you need to populate the signature with the record ref and BAI.

Does the code match this high-level description? not really, as e.g. 1. would be `match_author`, but that function does both more and less: it doesn't take care of the BAI (you need to do that separately before in an awkward way that requires you to restore it back at the end, which is error-prone and mixing different levels of concern), but OTOH it returns results through mutation of the underlying record (for the BAI) and through a return value (for the recid). That's nasty! Instead, it should 
* internally deal with removing the BAI, without that concern being visible to the outside world  (this function) as it's a detail of how we match authors.
* return both BAI and recid (through a `tuple` or a `dict`, up to you).

Same pattern for `match_literature_author` which should only match and return both pieces of info, then `create_new_author`. Then once at the end, you can check if you have a new BAI and record ID, and put them into the signature if needed.

backend/inspirehep/disambiguation/tasks.py
```diff
@@ -196,6 +173,30 @@ def create_new_author(full_name, from_recid):
     return new_author
 
 
+def match_author(author):
+    matched_authors = match(author, current_app.config["AUTHOR_MATCHER_EXACT_CONFIG"])
```
michamos, line 177: are you avoiding the match based on existing BAI in the record? I don't see it.

backend/inspirehep/disambiguation/tasks.py
```diff
@@ -79,57 +79,35 @@ def match_literature_author_with_config(author_data, matcher_config):
...
-    if matched_ref:
-        author_ids = get_value(
-            matched_record_list[0], "inner_hits.authors.hits.hits._source.ids", []
-        )
-        assign_bai_to_author(author_ids, author)
-        return matched_ref
+def get_reference_and_bai_if_unambiguous_match(matched_refs_ids):
+    if len(matched_refs_ids.keys()) == 1:
+        author_reference, author_ids = matched_refs_ids.popitem()
+        author_bai = get_values_for_schema(author_ids, "INSPIRE BAI")
```
michamos, line 92:
```suggestion
        author_bais = get_values_for_schema(author_ids, "INSPIRE BAI")
```

backend/inspirehep/disambiguation/tasks.py
```diff
@@ -79,57 +79,35 @@ def match_literature_author_with_config(author_data, matcher_config):
...
-        matched_authors_refs, author
-    )
-    if matched_ref:
-        author_ids = get_value(
-            matched_record_list[0], "inner_hits.authors.hits.hits._source.ids", []
-        )
-        assign_bai_to_author(author_ids, author)
-        return matched_ref
+def get_reference_and_bai_if_unambiguous_match(matched_refs_ids):
+    if len(matched_refs_ids.keys()) == 1:
```
michamos, line 90: The length of a `dict` is defined by the number of keys:
```suggestion
    if len(matched_refs_ids) == 1:
```

backend/inspirehep/disambiguation/tasks.py
MJedr: Yes, I've added it to the validator.

michamos: :+1: I missed it

backend/tests/integration-async/disambiguation/test_disambiguation_tasks.py
```diff
@@ -765,9 +772,100 @@ def assert_authors_records_exist_in_es():
...
+    def assert_disambiguation_on_record_update():
+        literature_record_from_es = InspireSearch.get_record_data_from_es(
+            literature_record_3
+        )
+        assert (
+            literature_record_from_es["authors"][0]["ids"]
+            == literature_record["authors"][0]["ids"]
+        )
+
+    retry_until_pass(assert_disambiguation_on_record_update, retry_interval=2)
```
michamos, line 871: I think there is one more test case to consider: if an author record is created on record creation due to an ambiguous match, and the match is still ambiguous during update, it remains assigned to the same record instead of creating yet another one.

michamos (approved):

PR #1796, opened 2021-03-25
Fix circular imports

pazembrz:


backend/inspirehep/records/api/base.py
```diff
@@ -533,6 +534,10 @@ def get_enhanced_es_data(self, serializer=None):
                 f"{self.__class__.__name__} is missing data serializer!"
             )
         if not serializer:
+            if isinstance(self.es_serializer, str):
```
drjova, line 537: we could use this https://tedboy.github.io/flask/generated/werkzeug.import_string.html instead

drjova (commented):
Still I'm not sure about this solution, this is a work around the problem not solving it

backend/inspirehep/records/api/base.py
pazembrz: Problem is that werkzeug is actually doing a wrapper on top of `__import__` just like python is doing with `import_module`. Still if we want to use it then we will have to keep code the same and replace only `import_module` with `import_string` so I don't see what benefit it will give us (please notice that part of the code you marked is actually doing module caching so it won't be imported on every call)

PR #1799, opened 2021-03-29
hal-push: check if flag is set in migrator task

oguzdemirbasci:


MJedr (commented):
Can you add a test for it maybe?

backend/inspirehep/migrator/tasks.py
```diff
@@ -439,6 +439,10 @@ def run_orcid_push(uuids):
 
 @shared_task(ignore_results=False, queue="migrator", acks_late=True)
 def run_hal_push(uuids):
+    if not current_app.config["FEATURE_FLAG_ENABLE_HAL_PUSH"]:
```
pazembrz, line 442: I'm not sure if this is the place where we want to do this check. As It will already create a task (unnecessary). It should be IMHO checked before we call this task.

oguzdemirbasci:
> Can you add a test for it maybe?

`FEATURE_FLAG_ENABLE_HAL_PUSH` should be `true` for QA and `false` for prod. How can I write a test for it?

MJedr:
> > Can you add a test for it maybe?
> 
> `FEATURE_FLAG_ENABLE_HAL_PUSH` should be `true` for QA and `false` for prod. How can I write a test for it? @MJedr

You can do a fixture which will set feature flag True/False like this one 
https://github.com/inspirehep/inspirehep/blob/050f94c8340f3867fad913216474d8be91324193/backend/tests/integration-async/conftest.py#L138
and check if works as expected

drjova (approved):

drjova:
@oguzdemirbasci could we add a test?

drjova:
@oguzdemirbasci what's the status of that?

drjova:
@oguzdemirbasci can we merge it?

drjova:
@oguzdemirbasci what's the status of this? :)

oguzdemirbasci:
@drjova There is a strange error in test. I am on it now.

backend/tests/integration-async/migrator/test_migrator_tasks.py
```diff
@@ -596,6 +598,32 @@ def assert_migrator_task():
     retry_until_pass(assert_migrator_task)
 
 
+def test_migrate_recids_from_mirror_external_push_enabled(
```
drjova, line 601: @oguzdemirbasci if you move these tests on `integration` and mock the `run_hal_push` you will be able to `assert_called_once`. This test unfortunately doesn't test much :/

drjova (approved):
🥇

Issue #1748, opened 2021-03-08
Best match sorting for conferences


StellaCh:
Currently, when I type a few keywords in the conference search, I don't see relevant results in the 1st page, as the results are sorted chronologically. For the bulk assignment interface, this should be improved.

In the conference search, implement a new sorting option 'best match' which returns conferences that match the following fields:

-  title
- acronym
- cnum 
- start-date
- place/city

**TECH NOTES**
* Using the data from Stella check if search based on `_score` will give correct results (the record should appear in the top of result list):
  * In [record config](https://github.com/inspirehep/inspirehep/blob/6d3133b75b610243dce084cba3298c8cd89f4f30/backend/inspirehep/records/config.py#L602) add a new sorting config:
  ```
    "bestmatch": {
        "title": "Best match",
        "fields": ["-_score"],
        "order": 3,
    }
  ```

StellaCh:
Examples of searches that should return these examples on top of the search results:
- `C22-07-27` should return https://inspirehep.net/conferences?sort=dateasc&size=25&page=1&start_date=all&q=cnum%3AC22-07-27
- `CHEP 2019` should return https://inspirehep.net/conferences/1732179
- `HYP 2022` should return https://inspirehep.net/conferences/1830349
- `ICRC 1947` should return https://inspirehep.net/conferences/978213
- `ICHEP 2014` should return https://inspirehep.net/conferences/1203206
- `International Workshop on High Energy Physics 2017` should return https://inspirehep.net/conferences/1505770
- `International conference on string theory 2014` should return https://inspirehep.net/conferences/1305234

PR #1806, opened 2021-03-30
conferences: best match sorting option added to result page

oguzdemirbasci:
Ref: #1748


drjova:
@oguzdemirbasci looks good! Do we have a test case to make sure we get the expected results?

drjova (commented):
Looks good, the only thing that is missing is to test the sorting that it's correct

drjova:
@oguzdemirbasci ☝️

drjova (approved):

PR #1808, opened 2021-03-31
backend: search_engines module

drjova:
* Sitemap cli to create and upload sitemaps on S3.

* Render cli to create and upload rendered records on S3.

Ref: inspirehep/kubernetes#226


backend/inspirehep/search_engines/tasks.py
```diff
@@ -54,4 +59,28 @@ def create_sitemap():
...
+        ConflictError,
+        ConnectionError,
+        ConnectionTimeout,
+        NotFoundError,
+        RequestError,
+        HTTPError,
+    ),
+)
+def render_records():
+    collections = get_rendered_records()
```
michamos, line 81: I don't understand the strategy here. This is a single task, that will render all the records sequentially. Assuming a very conservative 3s/record for rendering+uploading and 1.6M records, this will take more than 55 days. I think it would make more sense to fire a rendering task, rendering and uploading a single record, on record update (and maybe skipping the rendering if the last render for that record is younger than N days).

Issue #1749, opened 2021-03-08
Display public notes in search


StellaCh:
Add public notes in search for the curator interface:
<img width="457" alt="Screenshot 2021-03-15 at 10 11 59" src="https://user-images.githubusercontent.com/24955201/111137189-c86ea500-857e-11eb-8aa5-06112ba99f3b.png">

These should be displayed, similarly to the display in the detailed page. As the last item in the search results

**TECH NOTES**
1. Create a new component PublicNotes with `InlineList` inside which will take publicNotes (it’s a list of maps). Render the value from `PrivateNotes`.
2. In LiteratureItem get publicNotes and then create new `InlineUL` component here https://github.com/inspirehep/inspirehep/blob/9d3ef0776defeed49edd0892c238606d78babd15/ui/src/literature/components/LiteratureItem.jsx#L151 with the new `PublicNotes` component inside. It should be rendered only if `canEdit` is `True`

PR #1811, opened 2021-04-01
ui: display public notes in search results

oguzdemirbasci:
Ref: #1749


ui/src/literature/components/LiteratureItem.jsx
```diff
@@ -148,6 +150,13 @@ function LiteratureItem({ metadata, searchRank }) {
             {eprints && <ArxivEprintList eprints={eprints} />}
           </InlineUL>
         </div>
+        {canEdit && (
```
drjova, line 153: I would explicit check the roles, in the future `canEdit` might be the submitter as well and we don't want to display the pub notes

ui/src/literature/containers/LiteratureSearchContainer.jsx
```diff
@@ -206,6 +210,7 @@ const stateToProps = (state, { namespace }) => ({
   ]),
   results: state.search.getIn(['namespaces', namespace, 'results']),
   isCitationSummaryVisible: isCitationSummaryEnabled(state),
+  isCatalogerLoggedIn: isCataloger(state.user.getIn(['data', 'roles'])),
```
drjova, line 213:
```suggestion
  isCatalogerOrAdmin: isSuperUser(state.user.getIn(['data', 'roles'])) || isCataloger(state.user.getIn(['data', 'roles']))
```

oguzdemirbasci: Yeah, it was my first thought too. But `isCataloger` actualy returns true if the role is `cataloger` or `superuser` as described here https://github.com/inspirehep/inspirehep/blob/master/ui/src/common/authorization.js#L18-L20

drjova (approved):

Issue #1769, opened 2021-03-15
Get HAL ID from hal response


drjova:
On create record, get from the HAL response the HAL ID add it `ids` `{“schema”: “HAL”}`, on this record disable push to HAL.

TECH NOTES:
* Update the record data with external identifier after `create`. https://github.com/inspirehep/inspirehep/blob/9d3ef0776defeed49edd0892c238606d78babd15/backend/inspirehep/hal/tasks.py#L59
```Python3
data = dict(record)
if "external_system_identifiers" not in data.keys():
    data["external_system_identifiers"] = []
data["external_system_identifiers"].append({"value": receipt.id, "schema": "HAL"})
```
* Call the update record function with `disable_external_push` flag set `True`.
```Python3
record.update(data, disable_external_push=True)
db.session.commit()
``` 

If hal create throws an `HTTPResponseError` saying that the record already exists in the server, update the record with `hal id` which is contained in the receipt.

* Handle `HTTPResponseError` here:
https://github.com/inspirehep/inspirehep/blob/9d3ef0776defeed49edd0892c238606d78babd15/backend/inspirehep/hal/tasks.py#L59
* Add returning `hal id` to `external_system_identifiers` and update the record.
```Python3
data = dict(record)
if "external_system_identifiers" not in data.keys():
    data["external_system_identifiers"] = []
data["external_system_identifiers"].append({"value": receipt.id, "schema": "HAL"})
record.update(data, disable_external_push=True)
db.session.commit()
```

PR #1812, opened 2021-04-01
hal: add hal id to record after create

MJedr:
Ref: #1769


backend/inspirehep/hal/tasks.py
```diff
@@ -65,12 +68,30 @@ def _hal_push(record):
...
         hal_id = hal_value[0] if hal_value else ""
-
         receipt = None
-        if hal_id:
-            receipt = update(tei, hal_id)
-            LOGGER.info("HAL updated.", recid=record["control_number"], hal_id=hal_id)
-        else:
-            receipt = create(tei)
-            LOGGER.info("HAL created.", recid=record["control_number"], receipt=receipt)
+        try:
```
drjova, line 72: these can be easily split into two functions `hal_create` and `had_update`, the logic would be easier to test and maintain without the `if/else` `try/except`  spaghetti.

drjova (approved):

Issue #1780, opened 2021-03-18
Increase HAL push timeout


oguzdemirbasci:
Currently, timeout for the HAL push task is 30 secs. It should be increased to ~2 mins to avoid `RequestTimeOut` errors.

TECH NOTES:

* Add 2 mins timeout on `requests` to HAL push.

PR #1815, opened 2021-04-06
hal: increase push timeout

MJedr:
Ref: #1780

### DIFF
--- a/backend/inspirehep/hal/config.py
+++ b/backend/inspirehep/hal/config.py
@@ -89,5 +89,5 @@ HAL_IGNORE_CERTIFICATES = False
 """Whether to check certificates when connecting to HAL."""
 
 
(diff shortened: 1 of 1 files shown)


backend/inspirehep/hal/tasks.py
```diff
@@ -38,6 +38,7 @@ def _get_error_message_from_hal_exception(exception):
     retry_backoff=2,
     retry_kwargs={"max_retries": 6},
     autoretry_for=(RequestTimeOut,),
+    time_limit=2 * 60,
```
drjova, line 41: shouldn't we do it on request timeout?

MJedr: Yes, you're right, thanks 👍

pazembrz (approved):

Issue #1747, opened 2021-03-08
Bulk assign interface CDS


StellaCh:
Add an action in literature visible only to curators and superusers 'export to CDS' 

The prototype is found here: https://www.figma.com/proto/TUAMWktXlnKLOivYGg9p2V/Other-collections?node-id=344%3A5206&viewport=1676%2C306%2C0.3636794090270996&scaling=min-zoom

Set in the record `_export_to: [{'CDS': true}]`

1. Export to CDS: A confirmation modal is displayed.
2. 
TECH
1. For backend write the task similar to this one https://github.com/inspirehep/inspirehep/blob/master/backend/inspirehep/assign/tasks.py#L44
2. Add the (https://github.com/inspirehep/inspirehep/blob/master/ui/src/literature/components/ToolAction.jsx)[ToolAction] new new option - export to cds
3. Add `onClickExportToCDS` function which will render Modal witch `onOk` will pass record ids (from `state.literature.get('literatureSelection')`) to the correct endpoint

StellaCh:
@MJedr I tested this, works well!👍
Can you please add a success message when you click on confirming the CDS action? It should be similar to the success message we get when 'assigning to conference'
<img width="323" alt="Screenshot 2021-04-08 at 11 52 10" src="https://user-images.githubusercontent.com/24955201/114006613-e1f2cc00-9860-11eb-8711-def98c59d515.png">

It should say:
Export successful!
[number of papers] selected papers exported to CDS
 
This is something that was missing from the prototype, but the user would need some kind of confirmation that their actions took place.

PR #1825, opened 2021-04-09
export to cds: add notifications

MJedr:
Ref: #1747

### DIFF
--- a/ui/src/actions/literature.js
+++ b/ui/src/actions/literature.js
@@ -21,6 +21,9 @@ import {
   assignSuccess,
   assignError,
   assigning,
+  exportToCdsSuccess,
+  exportToCdsError,
+  exporting,
 } from '../literature/assignNotification';
 
 function fetchingLiteratureReferences(query) {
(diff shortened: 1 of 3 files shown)


ui/src/actions/literature.js
```diff
@@ -154,11 +156,16 @@ export function assignPapers(conferenceId, conferenceTitle) {
...
 export function exportToCds() {
-  return (dispatch, getState, http) => {
-    const papers = getState().literature.get('literatureSelection');
-    http.post('/assign/export-to-cds', {
-      literature_recids: papers,
-    });
-    dispatch(clearLiteratureSelection());
+  return async (dispatch, getState, http) => {
+    try {
+      const papers = getState().literature.get('literatureSelection');
```
drjova, line 161: we're missing the `loading` action here, we should create a task for that.

drjova (approved):

ui/src/actions/literature.js
MJedr: @drjova I can add it, it shouldn't take long

drjova: That would be great, thanks!

Issue #1794, opened 2021-03-24
Don't allow undeleting records


michamos:
In the record editor, it's possible to undelete deleted records. This is not currently handled correctly, some PIDs (at least the primary ones like `lit`, `con`, etc.) remain deleted. It seems that undeleting records is a very rare occurrence and is not actually a required feature, so we should simply disallow undeleting (or even updating) deleted records.

TECH NOTES:
* Add `force_undelete` flag in `update` function on  [record base](https://github.com/inspirehep/inspirehep/blob/master/backend/inspirehep/records/api/base.py#L445)
* Throw a validation error in update record when `deleted: true` changed to `deleted: false`

drjova:
@michamos what we will do with the redirected records?

PR #1836, opened 2021-04-14
records api: don't allow undeleting records

MJedr:
ref: #1794

### DIFF
--- a/backend/inspirehep/records/api/base.py
+++ b/backend/inspirehep/records/api/base.py
@@ -24,6 +24,7 @@ from invenio_records.api import Record
 from invenio_records.errors import MissingModelError
 from invenio_records.models import RecordMetadata
 from invenio_records.signals import after_record_revert, before_record_revert
+from jsonschema import ValidationError
 from sqlalchemy import tuple_
 from sqlalchemy.orm.attributes import flag_modified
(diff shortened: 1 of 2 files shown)


backend/inspirehep/records/api/base.py
```diff
@@ -460,6 +461,13 @@ def update(self, data, *args, **kwargs):
             # To be sure that when someone edits redirected record by mistake tries to undelete record as this is not supported for now
             raise CannotUndeleteRedirectedRecord(self.pid_type, data["control_number"])
 
+        if (
+            not force_undelete
+            and self._previous_version.get("deleted")
+            and self.get("deleted") is False
```
drjova, line 467:
```suggestion
            and not self.get("deleted") 
```

pazembrz, line 467: Here `deleted` key can be None so this check will return False and allow to undelete record while it shouldn't . 
```suggestion
            and self.get("deleted", False) is False
```
OR 
```suggestion
            and not self.get("deleted")
```

pazembrz: Ah sorry @drjova  I was too slow ;)

drjova (approved):

Issue #1797, opened 2021-03-29
Display more info for book chapters


michamos:
For book chapters, we currently display (and link) to the book title if there's a `publication.parent_record`. We should also display the subtitle in that case, as well as the page range (like we do on regular journal publications).

So https://inspirehep.net/literature/1642228 should display:
```
Part of From My Vast Repertoire ... : Guido Altarelli's Legacy, 123-151
```

**TECH NOTES**
1. Change name of `ParentRecordLink` to `ParentRecordInfo` and display also `subtitle` from parentRecord (if present) as link
2. Pass to the component also `page_start` and `page_end` from `publicationInfo` and display them as a text

PR #1837, opened 2021-04-14
ui: display subtitle and pages for book chapters

oguzdemirbasci:
Ref: #1797


ui/src/literature/components/ParentRecordInfo.jsx
```diff
@@ -5,7 +5,11 @@ import { Link } from 'react-router-dom';
 import { getRecordIdFromRef } from '../../common/utils';
 import { LITERATURE } from '../../common/routes';
 
-function ParentRecordLink({ parentRecord }) {
+function ParentRecordInfo({ parentRecord, publicationInfo }) {
+  const pageStart = publicationInfo
+    ? publicationInfo.get(0).get('page_start')
```
drjova, line 10: can something like `.get('key', '')` work?

ui/src/literature/components/ParentRecordInfo.jsx
```diff
@@ -5,7 +5,13 @@ import { Link } from 'react-router-dom';
...
 import { LITERATURE } from '../../common/routes';
 
-function ParentRecordLink({ parentRecord }) {
+function ParentRecordInfo({ parentRecord, publicationInfo }) {
+  const pageStart = publicationInfo
+    ? publicationInfo.get(0, {}).get('page_start')
+    : null;
+  const pageEnd = publicationInfo
+    ? publicationInfo.get(0, {}).get('page_end')
+    : null;
```
drjova, line 14: you don't need the `null` imo, it will return `undefined`

oguzdemirbasci: When `publicationInfo` is `undefined`, `publicationInfo.get(...)` throws error. I have to check this before assignment. If `publicationInfo` is `undefined`, `pageStart` and `pageEnd` should also be `undefined` I think.

drjova: What I mean you could  do the following

```
const pageStart = publicationInfo.get(0, {}).get('page_start', null)
```

oguzdemirbasci: It throws an error when `publicationInfo` is `undefined` since it doesn't have a `get` function.

drjova (approved):

PR #1856, opened 2021-04-28
Revert "v1.2.48"

pazembrz:
Reverts inspirehep/inspirehep#1855


michamos (changes requested):
I approve the release, so I disapprove the revert :smile:

Issue #1830, opened 2021-04-12
mint PIDs for existing author BAIs


michamos:
We have a feature flag to enable BAI PID minting, but we can't turn it on in prod currently for two reasons:
* BAIs will start to diverge from legacy (legacy is minting more PIDs than what is exposed in author records)
* the sequential approach to BAIs (`J.Smith.3` is minted after `J.Smith.4`) makes it impossible to mint PIDs for records one by one during the migrator, as we might mint a BAI that is still to come later in the migration, causing a collision.

So we would need to have a way to have a two-step process after switching off legacy:
1. Mint PIDs for all existing BAIs in author records through some script/CLI.
2. Enable the PID minting feature flag.

The procedure should be tested on QA by turning off author disambiguation and the minting feature flag, removing all existing BAI PIDs from the pidstore, running the script then turning on minting and disambiguation again. No collisions should occur.

pazembrz:
Tech Notes:
* Enable `FEATURE_FLAG_ENABLE_BAI_PROVIDER` Flag.
* Add new Feature Flag to disable BAI minter `create` method ONLY https://github.com/inspirehep/inspirehep/blob/60155f9ddc036932bb3653e11e1696ab2b4daf7c/backend/inspirehep/pidstore/minters/bai.py#L49. 

Now all PIDS which are already present in records should be minted. Records without BAI should be ignored until second flag will be enabled.

PR #1863, opened 2021-04-29
pidstre: Add flag to not create new BAIs. Add cli to recreate bais for all authors.

pazembrz:
* #1830


backend/inspirehep/pidstore/cli.py
```diff
@@ -0,0 +1,140 @@
...
+        pid_type="bai", pid_provider="external", status="R"
+    ).count()
+
+    if not click.confirm(
+        f"There are {new_not_external_bais_count-start_not_external_bais_count} new bais with non external pid_provider added to the Pidstore."
+        f" {length_external-new_external_bais_count} BAIs with external pid_provider are deleted"
+        " Should those changes be commited to DB?"
+    ):
+        db.session.rollback()
+        return
```
michamos, line 132: you're not testing for `yes_i_know` here, and you don't need to test for the output of `click.confirm`, as aborting the script will also rollback the transaction.

michamos (approved):
Minor comment, looks like a very careful approach :+1:

backend/inspirehep/pidstore/cli.py
pazembrz: I'm not testing for `yes_i_know` here as I want this to always display and ask for confirmation. This is really delicate operation and committing wrong data here will actually require a lot of work to undone it.
I am also aware that aborting will rollback, but here I want to be as much straightforward with the code as possible. This is dangerous operation here which at the end will actually modify Pidstore and records metadata in one big commit.

Issue #1844, opened 2021-04-19
`index_record` and migrator dependent records calculation are out of sync


michamos:
In `index_record` (used when indexing individual records following a change on labs), we have logic to reindex dependent records for which some of the indexed metadata might change following the modification of the record under consideration: https://github.com/inspirehep/inspirehep/blob/6d91451cd96de4a327b8c19c7e75e15a1eba0c95/backend/inspirehep/indexer/tasks.py#L113-L133.

The migrator doesn't use that task, but uses `batch_index` instead, as it deals with several records at once. It has a separate migration step in https://github.com/inspirehep/inspirehep/blob/1210863236fd2ed32f46c53cc4d97595472670a8/backend/inspirehep/migrator/tasks.py#L384 to perform similar logic.

The problem is that there is a discrepancy between the two places. The migrator misses for example `get_linked_author_records_uuids_if_author_changed_name` which causes papers to disappear from profiles when author names change. We should make sure the migrator is kept in sync with the `index_record` task.

PR #1867, opened 2021-05-03
indexing: unify index_record and migrator indexing task

MJedr:
ref: #1844

### DIFF
--- a/backend/inspirehep/indexer/tasks.py
+++ b/backend/inspirehep/indexer/tasks.py
@@ -14,7 +14,6 @@ from elasticsearch import (
     NotFoundError,
     RequestError,
 )
-from flask import current_app
 from sqlalchemy.exc import (
     DisconnectionError,
     OperationalError,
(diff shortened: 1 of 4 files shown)


oguzdemirbasci (approved):
Looks good to me 👍

backend/inspirehep/indexer/tasks.py
```diff
@@ -109,6 +109,13 @@ def index_record(self, uuid, record_version=None, force_delete=None):
                 error=err,
             )
 
+    uuids_to_reindex = get_references_to_update(record)
+
+    if uuids_to_reindex:
+        batch_index(list(uuids_to_reindex))
+
+
+def get_references_to_update(record):
```
drjova, line 118: since is used also in other modules, it should be moved as a method to `LiteratureRecord`. Isn't used only for literature?

MJedr: No, it's also for Authors and Conferences (see line 129 and 134). If you prefer we can do the record method for those record classes.

drjova: Shall we move it to `indexer/api.py` then? it shouldn't be part of the tasks :)

drjova (approved):

PR #1870, opened 2021-05-04
backend: update to python 3.8

drjova:


michamos:
Any reason for not going to 3.9 directly?

drjova:
because it  breaks some dependencies, we can invest time a bit later imo

backend/pyproject.toml
```diff
@@ -25,7 +25,7 @@ Flask = ">=1.0.2"
 werkzeug  = ">=0.16.1,<1.0.0"
 msgpack = "==0.6.2"
 
-python = ">=3.6,<3.8"
+python = ">=3.6,<3.9"
```
michamos, line 28: don't you want to require at least 3.8 now? if we run all the CI and deployments on 3.8, we better make sure that the developers are using the same version, otherwise things might start failing locally when incompatible with older versions.

pazembrz (approved):

michamos:
I don't understand. The `Dockerfile` is not updated, so won't it still run python 3.6?

PR #1902, opened 2021-05-26
v1.2.37

drjova:


pazembrz (approved):
`pidstre:Add flag to not create new BAIs` During tests i found that there is a small bug which I didn't have time to investigate yet. If this will be disabled by feature flag then It's ok to release.

oguzdemirbasci (approved):

Issue #1739, opened 2021-03-03
Handle hidden records correctly


StellaCh:
See https://inspirehep.zulipchat.com/#narrow/stream/195298-experts/topic/Fermilab.20collection

Some records are not part of the main Literature collection so should not be shown in the default search, yet they're not really private so knowing unauthenticated users should be able to search for them somehow and records in those collections should be visible.

## Classification
We'll need to refine our current public/private simple permission distinction and have three collection types in total:
1. **public records**: these records have `Literature` in `_collections`. They are show in the search by default and any user is allowed to see the records.
2. **hidden records**: these records have at least one of the following in `_collections`, but are not public:
* `CDF Notes` 
* `CDS Hidden` 
* `D0 Preliminary Notes` 
* `Fermilab` 
* `HAL Hidden` 
* `LArSoft Notes` 
* `SLAC` 
* `ZEUS Preliminary Notes` 

They will not be visible in the search by default but will require adding a request arg, and any user is allowed to see the records.
3. **private records**: these records don't have values in `_collections` making them either public or hidden. They are never shown in the search to non-privileged users, and those users are not allowed to access the records either.

Note that behaviors 1. and 3. already exist, 2. is new

## New features
These features only concern non-privileged users (not authenticated, or authenticated by having no superadmin or curator privileges). Admins and curators should still get the current behavior (including the visible collection facet).

1. For **hidden records**, adding a `collection={collection_name}` request argument (example: https://inspirehep.net/literature?sort=mostrecent&size=25&page=1&q=&collection=fermilab) will switch the search from searching for `_collections:Literature` (this currently happens in https://github.com/inspirehep/inspirehep/blob/6f7e3aab8f4b820a9e24b431a8749a0b0dff1b30/backend/inspirehep/search/api.py#L191) to searching in that collection. Non-privileged users should never see **private records** in search results.
2. When a hidden collection is active, any search typed in the search bar (until I go back to the main page or switch to a non-Literature collection) should stay in that collection.
3. The permission check on record access needs to be changed to only disallow non-authenticated access to **private records**. So https://inspirehep.net/literature/1861814 should return a 200 for everyone, instead of 401 as is currently the case for non-privileged users.

drjova:
TECH NOTES:

[...]

PR #1904, opened 2021-05-27
search: handle hidden records correctly

MJedr:
ref: #1739

### DIFF
--- a/backend/inspirehep/access_control.py
+++ b/backend/inspirehep/access_control.py
@@ -5,6 +5,8 @@
 # inspirehep is free software; you can redistribute it and/or modify it under
 # the terms of the MIT License; see LICENSE file for more details.
 
+from flask import current_app
+
 from inspirehep.accounts.decorators import login_required_with_roles
 from inspirehep.accounts.roles import Roles
 
@@ -37,6 +39,8 @@ class SessionSuperuserPermission(InspireBasePermissionCheck):
 
 class LiteraturePermissionCheck(SessionCatalogerPermission):
     def can(self):
-        if "Literature" not in self.record.get("_collections", []):
+        if not set(current_app.config["NON_PRIVATE_LITERATURE_COLLECTIONS"]) & set(
+            self.record.get("_collections", [])
+        ):
             return self.cataloger_check()
         return True
--- a/backend/inspirehep/search/api.py
+++ b/backend/inspirehep/search/api.py
@@ -75,7 +75,7 @@ class SearchMixin(object):
                 index=self.alias,
                 doc_type=self.Meta.doc_types,
                 body={"ids": uuids},
-                **kwargs
+                **kwargs,
             )
             results = [document["_source"] for document in documents["docs"]]
         except RequestError:
@@ -187,10 +187,19 @@ class LiteratureSearch(InspireSearch):
 
     def query_for_superuser_or_users(self, query_string):
         if not is_superuser_or_cataloger_logged_in():
-            user_query = Q(
-                IQ(query_string, self) & Q("term", _collections="Literature")
-            )
-            return self.query(user_query)
+            if "_collections" in query_string:
+                user_query = Q(IQ(query_string, self))
+                return self.query(user_query).filter(
+                    "terms",
+                    _collections=current_app.config[
+                        "NON_PRIVATE_LITERATURE_COLLECTIONS"
+                    ],
+                )
+            else:
+                user_query = Q(
+                    IQ(query_string, self) & Q("term", _collections="Literature")
+                )
+                return self.query(user_query)
         return self.query(IQ(query_string, self))
 
     def query_from_iq(self, query_string):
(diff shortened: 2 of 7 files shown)


backend/inspirehep/access_control.py
```diff
@@ -37,6 +37,17 @@ def can(self):
 
 class LiteraturePermissionCheck(SessionCatalogerPermission):
     def can(self):
-        if "Literature" not in self.record.get("_collections", []):
+        collections = {
```
michamos, line 40: maybe make it clear what these collections are?
```suggestion
        non_private_collections = {
```

backend/inspirehep/search/api.py
```diff
@@ -186,10 +186,9 @@ def execute(self, *args, **kwargs):
         return results
 
     def query_for_superuser_or_users(self, query_string):
+        collection = request.args.get("collection", default="literature", type=str)
         if not is_superuser_or_cataloger_logged_in():
-            user_query = Q(
-                IQ(query_string, self) & Q("term", _collections="Literature")
-            )
+            user_query = Q(IQ(query_string, self) & Q("term", _collections=collection))
```
michamos, line 191: what prevents a non-priviledged user from searching in private collections now, by setting something like `collection=d0 internal notes`?

backend/inspirehep/search/api.py
```diff
@@ -187,9 +192,29 @@ def execute(self, *args, **kwargs):
 
     def query_for_superuser_or_users(self, query_string):
         if not is_superuser_or_cataloger_logged_in():
-            user_query = Q(
-                IQ(query_string, self) & Q("term", _collections="Literature")
+            non_private_collections = [
```
michamos, line 195: This is used in two places now. Maybe move it to the config?

backend/inspirehep/search/api.py
```diff
@@ -187,9 +192,29 @@ def execute(self, *args, **kwargs):
...
+                "Fermilab",
+                "HAL Hidden",
+                "LArSoft Notes",
+                "SLAC",
+                "ZEUS Preliminary Notes",
+            ]
+            validation_regex = f"_collections: *({'|'.join(non_private_collections)})"
+            matched_collections = re.match(
+                validation_regex, query_string.lower(), re.IGNORECASE
             )
```
michamos, line 209: 1. I'm confused, I thought the idea was to use the `collection` request arg? why are you trying to get the collection from the query string here? I thought the idea was that by default you search in `Literature`, and if a `collection` is passed you search in that collection?
2. Also this is a rather bad way of restricting to allowed collections, as you're relying on one specific spelling of the query, which doesn't necessarily match what the query parser understands. So the correct way to do this is to try to extract the collections from the query parser (which is the thing that knows how to parser a query), but I don't think you actually need to do that due to 1., as it's easy to figure out if a collection is allowed if you don't need to extract it from the query string.

MJedr: No, actually with @drjova we discussed that the first idea is too hackish (from the frontend side) and will require many changes and we decided to do it in a way it's handled on legacy.

backend/tests/integration/search/test_api.py
```diff
@@ -1052,3 +1052,69 @@ def test_public_api_generates_correct_links_in_literature_search_with_fields(
...
+            "_collections": ["D0 Preliminary Notes"],
+            "titles": [{"title": "A hidden record"}],
+        },
+    )
+    with inspire_app.test_client() as client:
+        url = "/api/literature?q=_collections%3A%20fermilab%20or%20_collections%3A%20HAL%20hidden%20or%20_collections%3A%20Literature"
+        response = client.get(url)
+
+    assert response.status_code == 200
+    assert len(response.json["hits"]["hits"]) == 3
```
michamos, line 1120: can you also add a test that makes sure we never return records from private collactions?

backend/tests/integration/records/views/test_views_literature.py
```diff
@@ -850,3 +850,16 @@ def test_literature_json_put_redirected_record(inspire_app):
...
+    record = create_record(
+        "lit",
+        data={
+            "_collections": ["Fermilab"],
+            "titles": [{"title": "A literature record from fermilab collection"}],
+        },
+    )
+    with inspire_app.test_client() as client:
+        response = client.get(f"/literature/{record.control_number}")
+    assert response.status_code == 200
```
michamos, line 865: do we have a test that makes sure users can't access private records?

michamos (approved):

backend/inspirehep/search/api.py
```diff
@@ -187,10 +187,19 @@ def execute(self, *args, **kwargs):
...
+                    "terms",
+                    _collections=current_app.config[
+                        "NON_PRIVATE_LITERATURE_COLLECTIONS"
+                    ],
+                )
+            else:
+                user_query = Q(
+                    IQ(query_string, self) & Q("term", _collections="Literature")
+                )
+                return self.query(user_query)
```
drjova, line 202:
```suggestion
            if "_collections" in query_string:
                user_query = Q(IQ(query_string, self))
                return self.query(user_query).filter(
                    "terms",
                    _collections=current_app.config[
                        "NON_PRIVATE_LITERATURE_COLLECTIONS"
                    ],
                )
            
            user_query = Q(
                IQ(query_string, self) & Q("term", _collections="Literature")
            )
            return self.query(user_query)
```

drjova (approved):
💯  just the comment :)

PR #1907, opened 2021-05-28
records_api: (fix) mint identifiers even when id is provided

pazembrz:


backend/inspirehep/pidstore/providers/recid.py
```diff
@@ -53,7 +53,16 @@ def create(cls, object_type=None, object_uuid=None, **kwargs):
                 )
         else:
             LOGGER.info("Control number provided", recid=kwargs["pid_value"])
-            RecordIdentifier.insert(kwargs["pid_value"])
+            pid = PersistentIdentifier.query.filter_by(
```
michamos, line 56: IMHO this is a bit dirty. The method is `create`, but here you're doing a workaround for the case the PID already exists after all. I would `create` to error in case the PID can't be created. I believe it would be cleaner to have a no-op minter to appease invenio, and always do the actual minting in `InspireRecord.create`. Let's see what @drjova likes best.

drjova: I agree with @michamos it would be cleaner to error if cannot be created.

pazembrz: Actually you both are right. I will move this to `ControlNumberMinter.mint` as we still need this check because when we do record create through web API Invenio is first minting control_number and then the same minting is called from inside our record.create() method. If i will leave without this check anywhere there won't be any way to create record through API anymore.

drjova (approved):

Issue #1850, opened 2021-04-22
UI thinks the user is logged in, but that's not the case


michamos:
We've received many tickets from users who are confused because the overall UI shows they're logged in, but when navigating to pages that do backend checks whether they have permissions to edit something, they can't because they are not in fact logged in.

Example:
![image](https://user-images.githubusercontent.com/2895793/115679067-76295c80-a352-11eb-9ab4-1d6b646befe3.png)

The root cause of this seems to be that [some browsers periodically clear cookies](https://www.cookiestatus.com/) set through the `document.cookie` DOM API.

Possible solutions:
* periodically refresh the cookie, to avoid the timeout
* set the cookie from the backend, not the frontend (that seems to be the case though :confused: )
* refresh the login state more often (expensive, would need to happen on every request to be 100% accurate)
* have the UI watch the status of the session cookie
* ...

michamos:
It seems to happen not only with Safari, see https://inspirehep.zulipchat.com/#narrow/stream/122202-devel/topic/inconsistent.20login/near/237466200.

PR #1911, opened 2021-05-31
ui: logout and notify user when session expires.

oguzdemirbasci:
Ref: #1850


e2e/cypress/integration/user.test.js
```diff
@@ -9,26 +11,56 @@ describe('user', () => {
...
+        url: '/api/accounts/me',
+        failOnStatusCode: false,
+      })
+        .its('status')
+        .should('equal', 401);
+
+      cy.window().trigger('mouseover', 'topRight');
+      cy.tick(1800000);
+      cy.clock().invoke('restore');
+      cy.wait(500);
```
drjova, line 62: left overs?

drjova (commented):
LGTM 👏  why on mobile they are so many notifications?

e2e/cypress/integration/user.test.js
oguzdemirbasci: Actually, there is only one notification. Since it is sticky, it was staying when scroll down the page. So cypress screenshot took it multiple times.

oguzdemirbasci: Yes, the notification appears after 1800000 millis (30 mins) of inactivity. `tick` moves the time so we don't need to wait for 30 mins.

drjova (approved):

PR #1939, opened 2021-06-10
hal: allow to configure sword2 logging level

pazembrz:


backend/inspirehep/hal/__init__.py
```diff
@@ -0,0 +1,12 @@
+import logging
+import os
+
+import structlog
+
+LOGGER = structlog.getLogger()
+
+sword_logging_level = os.environ.get("SWORD2_LOGGING_LEVEL")
+if sword_logging_level:
+    LOGGER.info(f"Setting sword2.connection logger to '{sword_logging_level}' level")
```
michamos, line 10: you're doing structured logging here, why are you interpolating vars?

pazembrz: Argh, right. My mind was somehow on normal logging ;)

michamos (approved):

PR #1968, opened 2021-06-18
schemas: add revision_id to records envelope

pazembrz:
Required for https://github.com/inspirehep/inspirehep/issues/1962

### DIFF
--- a/backend/inspirehep/records/marshmallow/base.py
+++ b/backend/inspirehep/records/marshmallow/base.py
@@ -22,8 +22,6 @@ warnings.filterwarnings(
 class EnvelopeSchema(RecordSchemaJSONV1):
     """This schema is the envelope of the API.
 
-    Invenio API returns the following structure:
-
     Examples:
         {
             'uuid': '...',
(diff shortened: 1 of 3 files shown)


backend/inspirehep/records/marshmallow/base.py
```diff
@@ -38,6 +38,7 @@ class EnvelopeSchema(RecordSchemaJSONV1):
     """
 
     uuid = fields.String(dump_only=True, attribute="pid.object_uuid")
+    revision_id = fields.Integer(dump_only=True, attribute="metadata.revision_id")
```
drjova, line 41: could you please also update the docstring?

drjova (approved):
just the comment 👍

backend/inspirehep/records/marshmallow/base.py
pazembrz: :+1:

pazembrz: Wait. From what I can see docstring is about what Invenio is returning not for what we are returning here... I'm confused.

drjova: we're extending `RecordSchema` from invenio

PR #1994, opened 2021-06-25
pidstore: add a cli command which will process only authors without BAIs to speed it up.

pazembrz:


backend/inspirehep/pidstore/cli.py
```diff
@@ -138,3 +138,50 @@ def mint_bais(yes_i_know, create):
...
+    current_app.config["FEATURE_FLAG_ENABLE_BAI_CREATION"] = True
+    all_authors = PersistentIdentifier.query.filter_by(
+        pid_type="aut", status="R"
+    ).with_entities("object_uuid")
+    bai_authors = PersistentIdentifier.query.filter_by(
+        pid_type="bai", status="R"
+    ).with_entities("object_uuid")
+    all_authors = {author[0] for author in all_authors}
+    bai_authors = {author[0] for author in bai_authors}
+    missing_bais = all_authors - bai_authors
```
michamos, line 172: this should be a DB join, similar to what we have in https://github.com/inspirehep/inspirehep/blob/7150712b6f2b4204d8e08da0fad406257901c8e7/backend/inspirehep/records/utils.py#L111-L120

backend/inspirehep/pidstore/cli.py
```diff
@@ -33,7 +33,7 @@ def inspire_pidstore():
     help="Set this flag to also create missing BAIs. Without this flag only BAIs which are already in author metadata will be minted.",
 )
 @with_appcontext
-def mint_bais(yes_i_know, create):
+def mint_bais(yes_i_know, create, fast):
```
michamos, line 36: typo?

backend/inspirehep/pidstore/cli.py
pazembrz: Discussed on zoom

backend/inspirehep/pidstore/cli.py
```diff
@@ -138,3 +138,50 @@ def mint_bais(yes_i_know, create):
...
+    all_authors = {author[0] for author in all_authors}
+    bai_authors = {author[0] for author in bai_authors}
+    missing_bais = all_authors - bai_authors
+    length = len(missing_bais)
+    with click.progressbar(all_authors, length=length) as authors_bar:
+        for author_uuid in authors_bar:
+            author_record = AuthorsRecord.get_record(str(author_uuid))
+            author_data = dict(author_record)
+            BAIMinter.mint(author_uuid, author_data)
+            if author_data.get("ids") != author_record.get("ids"):
```
michamos, line 179: I'm not sure this check is needed, as we know they have no BAI yet, and we know this is going to create one.

michamos (approved):
Minor comment, looks good!

backend/inspirehep/pidstore/cli.py
pazembrz: Actually it's quite helpful as with this we are checking if BAI was actually created or not, as if there were no BAI and now there will be one it will differ. Otherwise if BAI was not created because of something then we do not want to update record. Additionally I can add a log to display info that BAI for this author was not created.

PR #2009, opened 2021-06-29
hep: add index task to a separate queue

drjova:


michamos (approved):
As I mentioned on the chat, this can alternatively be done in the config: https://docs.celeryproject.org/en/stable/userguide/routing.html#automatic-routing.

Issue #2013, opened 2021-06-30
Cache info in the editor


StellaCh:
"when I re-open a record in the editor (I made changes and saved the record in the editor before)
it shows cached information, when I then try to save the record I get an error message.
I have to reload the web-page to get the current data."

drjova:
TECH NOTES:
* On `412` status code error we should automatically refresh the editor here https://github.com/inspirehep/inspirehep/blob/master/record-editor/src/app/holdingpen-editor/holdingpen-save-button/holdingpen-save-button.component.ts#L183-L196

oguzdemirbasci:
With this pr https://github.com/inspirehep/inspirehep/pull/2036, page is redirected to the detail view when the record successfully saved. If save is failed, user should refresh the page manually when they see the error message. I am closing this issue.

PR #2030, opened 2021-07-07
editor: page refresh when 412 error responded

oguzdemirbasci:
Ref: #2013


drjova (approved):

michamos:
as I've already discussed privately with @oguzdemirbasci on the chat, I don't think this is a good solution:
```quote
I think there's a much broader problem with caching
  13:29

like you need to refresh the page in the UI too after changes for them to be seen
  13:29

maybe there's a unique solution to both problems
  13:29

(I don't know much about browser caching)
  13:30

there's also https://github.com/inspirehep/inspirehep/issues/2012
  13:30

if the editor closes after saving, then they'll need to reopen the editor to make a further change, at which point "fresh" data should be loaded
  13:31

to summarize, I don't think it makes much sense to work on this bug in isolation, it's connected to other things
```

Issue #2008, opened 2021-06-29
HAL pushes stale metadata


michamos:
We're sending the HAL push task during record update, not after record commit: https://github.com/inspirehep/inspirehep/blob/9281776185dfc3b363e0c900fd17523c0c5a9aad/backend/inspirehep/records/api/literature.py#L128-L129

This might cause the task to fire too early (before the record is commited), and retrieve the previous version from the DB. The solution is to move this to the `models_committed` receiver, like for disambiguation: https://github.com/inspirehep/inspirehep/blob/9281776185dfc3b363e0c900fd17523c0c5a9aad/backend/inspirehep/records/receivers.py#L44.

The same applies to ORCID (but is less critical, as we check the version there).

drjova:
TECH NOTES:

* We should send to orcid and hal on `CREATE` and `UPDATE` in the receivers https://github.com/inspirehep/inspirehep/blob/master/backend/inspirehep/records/receivers.py#L20-L46

michamos:
ORCID requires pushing on `DELETE` too.

PR #2031, opened 2021-07-07
hal push: don't push stale data

MJedr:
ref: #2008

### DIFF
--- a/backend/inspirehep/hal/api.py
+++ b/backend/inspirehep/hal/api.py
@@ -25,7 +25,11 @@ def push_to_hal(record):
 
     if is_hal_set(record):
         current_celery_app.send_task(
-            "inspirehep.hal.tasks.hal_push", kwargs={"recid": record["control_number"]}
+            "inspirehep.hal.tasks.hal_push",
+            kwargs={
+                "recid": record["control_number"],
+                "record_version_id": record.model.version_id,
+            },
         )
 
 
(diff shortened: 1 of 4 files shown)


backend/inspirehep/hal/api.py
```diff
@@ -25,7 +25,11 @@ def push_to_hal(record):
 
     if is_hal_set(record):
         current_celery_app.send_task(
-            "inspirehep.hal.tasks.hal_push", kwargs={"recid": record["control_number"]}
+            "inspirehep.hal.tasks.hal_push",
+            kwargs={
+                "recid": record["control_number"],
+                "record_version_id": record.model.version_id,
+            },
```
michamos, line 32: the TECH NOTES in the issue talk about moving the trigger from the record class to the receivers. Why did you change the approach?

MJedr: It was discussed with @drjova, the issue is that we need to pass the feature flag which is disabling push in some cases (for update).

michamos: that flag was used only in the migrator, and we don't use that any longer.

MJedr: That's not true, it is also used in hal push:
https://github.com/inspirehep/inspirehep/blob/4e47de34dc2e0500772425e9fa6687722dc127ac/backend/inspirehep/hal/tasks.py#L110

michamos: ok, my bad, I hadn't realized that :+1:

drjova (approved):

Issue #2004, opened 2021-06-29
retry on BAI minting and on author disambiguation task


michamos:
BAI minting can fail if there are collisions with concurrent transactions. We should retry in that case like we do for texkey minting. The author disambiguation task should also retry in case of error. See discussion in https://inspirehep.zulipchat.com/#narrow/stream/122199-ops/topic/author.20disambiguation.

PR #2035, opened 2021-07-08
bai: change retry policy 

MJedr:
ref: #2004

### DIFF
--- a/backend/inspirehep/pidstore/providers/bai.py
+++ b/backend/inspirehep/pidstore/providers/bai.py
@@ -6,8 +6,8 @@
 # the terms of the MIT License; see LICENSE file for more details.
 import re
 import string
-import time
 
+import backoff
 import structlog
 from flask import current_app
 from inspire_utils.name import format_name
@@ -78,6 +78,12 @@ class InspireBAIProvider(InspireBaseProvider):
         return int(last_number) + 1
 
     @classmethod
+    @backoff.on_exception(
+        backoff.constant,
+        (IntegrityError, PIDAlreadyExistsError),
+        max_tries=5,
+        interval=5,
+    )
     def create(
         cls, pid_value=None, object_uuid=None, data=None, object_type=None, **kwargs
     ):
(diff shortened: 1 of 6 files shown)


backend/inspirehep/pidstore/providers/bai.py
```diff
@@ -78,6 +78,9 @@ def next_bai_number(cls, bai):
         return int(last_number) + 1
 
     @classmethod
+    @backoff.on_exception(
+        backoff.expo, (IntegrityError, PIDAlreadyExistsError), max_tries=5
```
michamos, line 82: I think the point is that you don't need exponential backoff in this case. AFAIU, the failure mode is the following: there's a concurrent transaction that has created the next BAI with the same prefix already, so the current transaction comptues the same BAI, which fails when inserting as that PID already exists. But this means that the other transaction has already been committed (otherwise there would be no issue, this transaction would win), so you don't need to wait before retrying.

MJedr: But how it's possible that it was failing with current approach? We were retrying 5 times with `time.sleep(5)` for every retry and still we were getting that error.

michamos: ok, I looked some more and I believe this is the reason: https://github.com/inspirehep/inspirehep/blob/2f0da315b970425b012467f3698343b17a21c0a6/backend/inspirehep/records/api/base.py#L472.
In principle, newly created/modified model instances should be flushed to the DB before querying so they're available to the DB. If they're not flushed, they're simply not present in the DB (they exist only in the Python interpreter), so can't be queried. See also this excelled answer https://stackoverflow.com/a/4202016.

So I guess the failure mode is not a concurrent transaction, but a different BAI created in the same transaction, which isn't flushed to the DB yet.

Question: do we really need to disable autoflush? it was done by @pazembrz in https://github.com/inspirehep/inspirehep/pull/1727 while trying to fix the versioning issues, but I'm not sure it's actually needed.

MJedr: It seems very reasonable, but this is only in update method. I think that this problem exists also on create (because we're creating same author in the concurrent transactions)

michamos: still, *exponential backoff* doesn't make a lot of sense to me. Why not have a constant backoff insteand?

MJedr: Make sense to me too, what value do you suggest in that case?

michamos: 5s like previously?

backend/inspirehep/pidstore/providers/bai.py
```diff
@@ -88,39 +94,25 @@ def create(
...
-            else:
-                provider_object = super().get(pid_value=pid_value)
-                break
-                # Correct pid already assigned to this record
-            time.sleep(current_app.config.get("PIDSTORE_BAI_RETRY_DELAY", 5))
-        if last_exception:
-            raise last_exception
+                # Pid created successfully.
+            except IntegrityError:
+                raise
```
michamos, line 110: this doesn't do anything, you can just drop the whole `try...except`.

backend/inspirehep/pidstore/providers/bai.py
```diff
@@ -88,39 +94,25 @@ def create(
...
+                    object_type=object_type,
+                    object_uuid=object_uuid,
+                    status=PIDStatus.REGISTERED,
+                    **kwargs,
                 )
-            else:
-                provider_object = super().get(pid_value=pid_value)
-                break
-                # Correct pid already assigned to this record
-            time.sleep(current_app.config.get("PIDSTORE_BAI_RETRY_DELAY", 5))
```
michamos, line 121: Is this variable defined in the config? if so you should remove it there too.

MJedr: No, it's not, I'll drop it 👍

michamos (approved):

Issue #2012, opened 2021-06-30
page after saving the editor


StellaCh:
"I'm trying to modify this record and save https://inspirehep.net/editor/record/literature/1831503 but clicking on the save button does nothing and the editor stays open"

Clicking save should go to the previous page that the curator was browsing- record detailed page or search results. If this is not possible, return to the detailed page of the record.

PR #2036, opened 2021-07-08
record-editor: go to detail view when the record successfully saved.

oguzdemirbasci:
Ref: #2012

### DIFF
--- a/record-editor/src/app/record-editor/record-toolbar/record-toolbar.component.ts
+++ b/record-editor/src/app/record-editor/record-toolbar/record-toolbar.component.ts
@@ -48,7 +48,8 @@ import { HOVER_TO_DISMISS_INDEFINITE_TOAST } from '../../shared/constants';
   ],
   changeDetection: ChangeDetectionStrategy.OnPush,
 })
-export class RecordToolbarComponent extends SubscriberComponent
+export class RecordToolbarComponent
+  extends SubscriberComponent
   implements OnInit {
   // `undefined` if there is no record being edited
   record: object;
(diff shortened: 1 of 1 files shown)


record-editor/src/app/record-editor/record-toolbar/record-toolbar.component.ts
```diff
@@ -133,19 +134,22 @@ export class RecordToolbarComponent extends SubscriberComponent
...
     this.domUtilsService.unregisterBeforeUnloadPrompt();
-
-    if (this.pidType === 'conferences') {
+    this.toastrService.success('Changes are saved', 'Success');
+    if (this.pidType === 'conferences' || this.pidType === 'literature') {
       // direct assignment to window.location  doesn't compile with this version of TS
-      window.location.href = `/conferences/${this.record['control_number']}`;
-    } else {
-      this.toastrService.success('Changes are saved', 'Success');
+      setInterval(
```
drjova, line 148: set inteval why? 🤔

record-editor/src/app/record-editor/record-toolbar/record-toolbar.component.ts
```diff
@@ -133,19 +134,22 @@ export class RecordToolbarComponent extends SubscriberComponent
...
+      (error) => this.onSaveError(error)
+    );
   }
 
   private onSaveSuccess() {
     this.domUtilsService.unregisterBeforeUnloadPrompt();
-
-    if (this.pidType === 'conferences') {
+    this.toastrService.success('Changes are saved', 'Success');
+    if (this.pidType === 'conferences' || this.pidType === 'literature') {
```
drjova, line 146: what about authors?

michamos: yeah, we want this behavior for any pid type.

record-editor/src/app/record-editor/record-toolbar/record-toolbar.component.ts
oguzdemirbasci: In order to show the success message. It will be confusing redirecting the page without displaying a feedback.

record-editor/src/app/record-editor/record-toolbar/record-toolbar.component.ts
oguzdemirbasci: Updating 👍

record-editor/src/app/record-editor/record-toolbar/record-toolbar.component.ts
michamos: I think it's ok, it only redirects on success, right, not in case of error? curators will learn that if they see the detailed view, it means that their changes have been applied.

drjova (approved):

Issue #2034, opened 2021-07-07
Garbage collect stub authors with no papers


michamos:
In the disambiguation process, we create a new stub author (having `stub: true` in the metadata) if there's no unambiguous match for a given signature. Manual intervention or a subsequent disambiguation run might remove the paper from that author profile, causing that author to have no associated papers. In the long term, these authors will pile up and cause noise in the authors collection. 

To avoid that, we should garbage collect stub authors. This can happen either periodically, or during disambiguation/claiming (but is more tricky to get right).

MJedr:
**TECH NITES**
* Write a cli in a disambiguation module which will:
    - query all authors with `stub: true`
    - for every author it will check if it has any linked papers - if no, delete record
* in kuberneres add a cronjob which will run the task (similarly to [mailing cronjob](https://github.com/inspirehep/kubernetes/tree/master/hep/overlays/mail-jobs-weekly))
* add a cronjob to the [hep kustomisation file](https://github.com/inspirehep/kubernetes/blob/c383bbf91f1dd1b7b7cabc55b697d39febfdf8c7/hep/environments/prod/kustomization.yml#L24)

PR #2053, opened 2021-07-20
disambiguation: add cli to remove stub authors

MJedr:
ref: inspirehep/inspirehep#2034

### DIFF
--- a/backend/inspirehep/alembic/2d7ea622feda_add_index_to_authors_records.py
+++ b/backend/inspirehep/alembic/2d7ea622feda_add_index_to_authors_records.py
@@ -1,0 +1,38 @@
+#
+# This file is part of Invenio.
+# Copyright (C) 2016-2018 CERN.
+#
+# Invenio is free software; you can redistribute it and/or modify it
+# under the terms of the MIT License; see LICENSE file for more details.
+
+"""add index to authors_records"""
+
+from alembic import op
+
+# revision identifiers, used by Alembic.
+revision = "2d7ea622feda"
+down_revision = "412aeb064d68"
+branch_labels = ()
+depends_on = None
+
+
+def upgrade():
+    """Upgrade database."""
+    with op.get_context().autocommit_block():
+        op.create_index(
+            "ix_records_authors_id_type_authors_id",
+            "records_authors",
+            ["id_type", "author_id"],
+            unique=False,
+            postgresql_concurrently=True,
+        )
+
+
+def downgrade():
+    """Downgrade database."""
+    with op.get_context().autocommit_block():
+        op.drop_index(
+            "ix_records_authors_id_type_authors_id",
+            table_name="records_authors",
+            postgresql_concurrently=True,
+        )
--- a/backend/inspirehep/disambiguation/cli.py
+++ b/backend/inspirehep/disambiguation/cli.py
@@ -1,0 +1,64 @@
+# -*- coding: utf-8 -*-
+#
+# Copyright (C) 2021 CERN.
+#
+# inspirehep is free software; you can redistribute it and/or modify it under
+# the terms of the MIT License; see LICENSE file for more details.
+import click
+from elasticsearch_dsl import Q
+from inspire_utils.record import get_values_for_schema
+from invenio_db import db
+
+from inspirehep.records.api import AuthorsRecord
+from inspirehep.records.models import RecordsAuthors
+from inspirehep.search.api import AuthorsSearch
+
+
+@click.group()
+def disambiguation():
+    """Commands for disambiguation"""
+
+
+@disambiguation.command()
+def clean_stub_authors():
+    """Removes all the authors created by disambiguation and having no linked papers."""
+    # We get all the stub authors (created by disambiguation) from ES and we verify
+    # in db if the returned records are stub (ES data might be outdated)
+    stub_authors_query = Q("term", stub=True)
+    stub_authors = (
+        AuthorsSearch().query(stub_authors_query).source(["control_number"]).execute()
+    )
+    stub_authors_control_numbers = [
+        ("aut", str(author["control_number"])) for author in stub_authors
+    ]
+    # We change isolation level in db to the higher one (serializable) to avoid
+    # issues with race condition
+    db.session.connection(execution_options={"isolation_level": "SERIALIZABLE"})
+    stub_authors_verified = AuthorsRecord.get_records_by_pids(
+        stub_authors_control_numbers
+    )
+    stub_authors_bais = {
+        get_values_for_schema(author["ids"], "INSPIRE BAI")[0]: author
+        for author in stub_authors_verified
+    }
+    # We verify which authors have linked papers
+    stub_authors_with_papers = set(
+        query_authors_with_linked_papers_by_bai(stub_authors_bais.keys())
+    )
+    # For every author who has not linked papers we delete record
+    for author_bai in set(stub_authors_bais.keys()).difference(
+        stub_authors_with_papers
+    ):
+        author = stub_authors_bais[author_bai]
+        author.delete()
+    db.session.commit()
+
+
+def query_authors_with_linked_papers_by_bai(authors_bais):
+    query = RecordsAuthors.query.filter(
+        RecordsAuthors.id_type == "INSPIRE BAI",
+        RecordsAuthors.author_id.in_(authors_bais),
+    )
+
+    for data in query.yield_per(100).with_entities(RecordsAuthors.author_id):
+        yield data.author_id
--- a/backend/inspirehep/records/models.py
+++ b/backend/inspirehep/records/models.py
@@ -179,6 +179,7 @@ class RecordsAuthors(db.Model):
             "id_type",
             "record_id",
         ),
+        db.Index("ix_records_authors_id_type_authors_id", "id_type", "author_id"),
     )
     id = db.Column(db.Integer, primary_key=True)
     author_id = db.Column(Text, nullable=False)
--- a/backend/inspirehep/cli.py
+++ b/backend/inspirehep/cli.py
@@ -10,6 +10,7 @@
 from invenio_base.app import create_cli
 
 from inspirehep.cds.cli import cds
+from inspirehep.disambiguation.cli import disambiguation
 from inspirehep.factory import create_app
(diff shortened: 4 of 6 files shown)


backend/inspirehep/disambiguation/cli.py
```diff
@@ -0,0 +1,37 @@
...
+    stub_authors = AuthorsSearch().query(stub_authors_query).execute()
+    for author in stub_authors:
+        author_data = author.to_dict()
+        author_bai = get_values_for_schema(author_data["ids"], "INSPIRE BAI")[0]
+        author_papers_query = Q(
+            "nested",
+            path="authors",
+            query=Q("match", authors__ids__value__raw=author_bai),
+        )
+        author_papers = LiteratureSearch().query(author_papers_query).execute()
```
michamos, line 32: ES might be inconsistent with the DB, and here you're changing the DB based on ES data, that's dangerous. Two things might happen:
1. the author record is not actually `stub` any longer, so you shouldn't delete it even if it has no papers
2. the author record has papers associated, but ES doesn't know yet.

I think the best course of action would be to use the DB only for this. That might require some work if you want it to be efficient (adding indices/auxiliary tables). Alternatively, you could get `stub` record IDs from ES, use `get_records_by_pids` to get the DB records and confirm they're still stub and get the BAI, then use a query against the `RecordsAuthors` table to check if there are no associated papers (that probably requires adding a new index on that table). I believe you also need a higher transaction isolation level than the standard `read committed` (I would guess `serializable`), because you're removing author records based on a predicate (are there associated papers), which the standard optimistic concurrency approach using a version counter on the author record doesn't protect.

backend/inspirehep/disambiguation/cli.py
```diff
@@ -0,0 +1,45 @@
...
+    stub_authors = (
+        AuthorsSearch().query(stub_authors_query).source(["control_number"]).execute()
+    )
+    stub_authors_control_numbers = [
+        ("aut", str(author["control_number"])) for author in stub_authors
+    ]
+    stub_authors_db = list(
+        AuthorsRecord.get_records_by_pids(stub_authors_control_numbers)
+    )
+    db.session.connection(execution_options={"isolation_level": "SERIALIZABLE"})
```
michamos, line 34: I think you need to move this above L31, as the reading of the author records is part of the same serializable transaction.

backend/inspirehep/disambiguation/cli.py
```diff
@@ -0,0 +1,45 @@
...
+        if not author["stub"]:
+            continue
+        author_bai = get_values_for_schema(author["ids"], "INSPIRE BAI")[0]
+        author_papers = RecordsAuthors.query.filter(
+            RecordsAuthors.id_type == "INSPIRE BAI",
+            RecordsAuthors.author_id == author_bai,
+        ).all()
+        if not author_papers:
+            author.delete()
+    db.session.commit()
```
michamos, line 45: I suppose you've thought about this, but this may now fail if the transaction is not serializable due to other concurrent transactions. That will cause the CLI command to exit with a non-zero exit code and the cronjob to retry (if setup correctly). I think that's acceptable and no extra error handling is needed, but it might be something to point out.

backend/inspirehep/disambiguation/cli.py
```diff
@@ -0,0 +1,45 @@
...
+from inspirehep.search.api import AuthorsSearch
+
+
+@click.group()
+def disambiguation():
+    """Commands for disambiguation"""
+
+
+@disambiguation.command()
+def clean_stub_authors():
```
michamos, line 23: I would add some more comments to this function to explain what's going on w.r.t. consistency, as it might look superficially like you're doing useless work (reading from DB data you've already obtained through ES) and you don't have explicit tests for concurrent scenarios (I don't think you should add those, it's probably not worth it).

backend/inspirehep/disambiguation/cli.py
```diff
@@ -0,0 +1,45 @@
...
+        AuthorsSearch().query(stub_authors_query).source(["control_number"]).execute()
+    )
+    stub_authors_control_numbers = [
+        ("aut", str(author["control_number"])) for author in stub_authors
+    ]
+    stub_authors_db = list(
+        AuthorsRecord.get_records_by_pids(stub_authors_control_numbers)
+    )
+    db.session.connection(execution_options={"isolation_level": "SERIALIZABLE"})
+    for author in stub_authors_db:
```
michamos, line 35: here you're doing lots of queries, one for each author, which is expensive. It's better to filter the stub authors first, then do a join against the `RecordsAuthors` table (potentially batched to avoid too long statements, look at how `get_records_by_pids` is implemented).

backend/inspirehep/disambiguation/cli.py
```diff
@@ -0,0 +1,45 @@
...
+        if not author["stub"]:
+            continue
+        author_bai = get_values_for_schema(author["ids"], "INSPIRE BAI")[0]
+        author_papers = RecordsAuthors.query.filter(
+            RecordsAuthors.id_type == "INSPIRE BAI",
+            RecordsAuthors.author_id == author_bai,
+        ).all()
```
michamos, line 42: * you don't need `all` results here, all you want to know if this is empty or not, so you can do `count` or `first` instead of `all`.
* I think you need to add an index for this, otherwise it will be very slow.

backend/inspirehep/records/models.py
```diff
@@ -179,6 +179,7 @@ class RecordsAuthors(db.Model):
             "id_type",
             "record_id",
         ),
+        db.Index("ix_id_type_authors_id", "id_type", "author_id"),
```
michamos, line 182: there is a global index for namespaces, so the table name should be included

backend/inspirehep/disambiguation/cli.py
```diff
@@ -0,0 +1,59 @@
...
+    )
+    # For every author who has not linked papers we delete record
+    for author_bai in stub_authors_bais.difference(stub_authors_with_papers):
+        author = AuthorsRecord.get_record_by_pid_value(author_bai, "bai")
+        author.delete()
+    db.session.commit()
+
+
+def query_authors_with_linked_papers_by_bai(authors_bais):
+    query = RecordsAuthors.query.filter(
```
michamos, line 53: if you add `with_entities(RecordsAuthors.author_id)`, this will be executed as an index-only scan. Now, the main table still needs to be accessed to retrieve the other columns you don't care about, but are returned by SQLAlchemy.

backend/inspirehep/disambiguation/cli.py
```diff
@@ -0,0 +1,59 @@
...
+        for author in AuthorsRecord.get_records_by_pids(stub_authors_control_numbers)
+    }
+    # We verify which authors have linked papers
+    stub_authors_with_papers = set(
+        query_authors_with_linked_papers_by_bai(stub_authors_bais)
+    )
+    # For every author who has not linked papers we delete record
+    for author_bai in stub_authors_bais.difference(stub_authors_with_papers):
+        author = AuthorsRecord.get_record_by_pid_value(author_bai, "bai")
+        author.delete()
```
michamos, line 48: Does this remove the BAI from the PIDstore or mark it as deleted? I guess we want to actually remove the BAI from the PIDstore, otherwise we'll get ugly things like `Z.Wang.4235` at some point.

backend/inspirehep/records/models.py
MJedr: Sorry, forgot to add the file to git, thanks :)

backend/inspirehep/disambiguation/cli.py
MJedr: Doesn't it remove bai too?
https://github.com/inspirehep/inspirehep/blob/cb15e67376042277a0550a2ae0f8b7147a8d65ae/backend/inspirehep/records/api/base.py#L501

michamos: Looks like it, I hadn't checked :+1: https://github.com/inspirehep/inspirehep/blob/cb15e67376042277a0550a2ae0f8b7147a8d65ae/backend/inspirehep/pidstore/providers/bai.py#L117-L119

backend/inspirehep/records/models.py
michamos: Not sure I understand, we can discuss it :)

backend/inspirehep/alembic/2d7ea622feda_add_index_to_authors_records.py
```diff
@@ -0,0 +1,35 @@
...
+
+def upgrade():
+    """Upgrade database."""
+    # ### commands auto generated by Alembic - please adjust! ###
+    op.create_index(
+        "ix_records_authors_id_type_authors_id",
+        "records_authors",
+        ["id_type", "author_id"],
+        unique=False,
+    )
```
michamos, line 27: you need to create the index concurrently to avoid locking the table, like in https://github.com/inspirehep/inspirehep/blob/cb15e67376042277a0550a2ae0f8b7147a8d65ae/backend/inspirehep/alembic/318758a589d5_modify_idx_object_index_in_pidstore_pid_.py#L21-L29

michamos: what you did is not sufficient, you need to go into autocommit mode (disable transactions), look at the above snippet.

backend/inspirehep/disambiguation/cli.py
```diff
@@ -0,0 +1,59 @@
...
+        get_values_for_schema(author["ids"], "INSPIRE BAI")[0]
+        for author in AuthorsRecord.get_records_by_pids(stub_authors_control_numbers)
+    }
+    # We verify which authors have linked papers
+    stub_authors_with_papers = set(
+        query_authors_with_linked_papers_by_bai(stub_authors_bais)
+    )
+    # For every author who has not linked papers we delete record
+    for author_bai in stub_authors_bais.difference(stub_authors_with_papers):
+        author = AuthorsRecord.get_record_by_pid_value(author_bai, "bai")
```
michamos, line 47: you fetched the author records already on L39. If you assign them to a variable (watch out, I believe it's an iterator, not a list), you can avoid doing these extra queries.

michamos (approved):
small comment, looks good otherwise :+1:

Issue #2045, opened 2021-07-14
Authors created through claiming have `B.Stub.*` BAI


michamos:
When manually kicking out a paper from an author profile through claiming (`Claim>This is not my paper`), a new author is created through https://github.com/inspirehep/inspirehep/blob/17a5208bbbd8580dd1e0534c44524d001f33420a/backend/inspirehep/assign/views.py#L57-L64. The problem is that a stub author record is created first, then updated to have the correct metadata, which causes the `B.Stub.*` BAI to be created in the first stage. The solution is to pass (at least) the correct author name to `create_stub_author`. Maybe we should even remove https://github.com/inspirehep/inspirehep/blob/17a5208bbbd8580dd1e0534c44524d001f33420a/backend/inspirehep/disambiguation/utils.py#L78 so the record will be invalid if an author name is not provided immediately, allowing to catch these errors more easily.

PR #2061, opened 2021-07-22
assign authors: don't create incorrect bai for stub authors

MJedr:
ref: #2045

### DIFF
--- a/backend/inspirehep/assign/views.py
+++ b/backend/inspirehep/assign/views.py
@@ -43,29 +43,42 @@ def unstub_author_by_recid(author_recid):
         author.update(dict(author))
 
 
-def assign_papers(from_author_recid, to_author_recid, literature_recids):
-    author_signatures = []
-    for record in get_literature_records_by_recid(literature_recids):
-        from_author = get_author_by_recid(record, from_author_recid)
-        from_author["record"] = get_record_ref(to_author_recid, endpoint="authors")
-        from_author["curated_relation"] = True
-        author_signatures.append(from_author)
+def assign_papers(
+    to_author_recid, author_papers, author_signatures, is_stub_author=False
+):
+    for author, record in zip(author_signatures, author_papers):
+        author["record"] = get_record_ref(to_author_recid, endpoint="authors")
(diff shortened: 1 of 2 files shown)


backend/inspirehep/assign/views.py
```diff
@@ -78,8 +78,10 @@ def author_assign_view():
     literature_recids = body["literature_recids"]
     with db.session.begin_nested():
         if to_author_recid is None:
+            from_author = AuthorsRecord.get_record_by_pid_value(from_author_recid)
+            from_author_name = from_author["name"]["value"]
```
michamos, line 82: I don't think this is the name to use. Basically you need to take the logic from `update_author_names` and instead of mutating the already constructed author, it should provide you with the author name, that you can then pass to `create_new_stub_author`. I don't think you need to change anything outside `assign_to_new_stub_author`.

MJedr: Ok, but the name used to create name vairants (as in `update_author_names`) should be the name of the existing author that we're unassigning records from?

backend/inspirehep/assign/views.py
```diff
@@ -43,29 +43,37 @@ def unstub_author_by_recid(author_recid):
...
-    to_author = create_new_stub_author()
-    author_signatures = assign_papers(
-        from_author_recid, to_author["control_number"], literature_recids
-    )
-    update_author_names(to_author, author_signatures)
+    author_papers = list(get_literature_records_by_recid(literature_recids))
+    author_signatures = get_author_signatures(from_author_recid, author_papers)
+    stub_author_data = update_author_names({"name": {}}, author_signatures)
+    to_author = create_new_stub_author(**stub_author_data)
+    assign_papers(to_author["control_number"], author_papers, author_signatures, True)
```
michamos, line 69:
```suggestion
    assign_papers(to_author["control_number"], author_papers, author_signatures, is_stub_author=True)
```
so it's more clear to the reader what the arg represents.

michamos (approved):
minor comment only

Issue #2064, opened 2021-07-22
link papers when creating new profile from BAI


michamos:
Legacy had profiles (with BAI) with no associated records. There exist in the system as Literature records with a BAI but no linked record. Users/curators can create new author profiles for those, but upon creation the existing signatures in Literature records with that BAI don't get linked to the newly created author profile. We should fix that and automatically add the author record ref to the Literature record.

MJedr:
**TECH NOTES**
1. Create a function that will take as an argument an author record. It should query db for all the papers where this author exists (by using his BAI). For every paper in the result set, assign the reference to this author.
1. On author create call the new function.

PR #2072, opened 2021-07-27
authors: link papers when creating new profile

MJedr:
ref: #2064

### DIFF
--- a/backend/inspirehep/records/api/authors.py
+++ b/backend/inspirehep/records/api/authors.py
@@ -5,11 +5,15 @@
 # inspirehep is free software; you can redistribute it and/or modify it under
 # the terms of the MIT License; see LICENSE file for more details.
 import structlog
+from inspire_utils.record import get_values_for_schema
+from invenio_db import db
 
 from inspirehep.records.marshmallow.authors import AuthorsElasticSearchSchema
+from inspirehep.records.models import RecordsAuthors
 from inspirehep.search.api import AuthorsSearch
 
 from ...pidstore.api import PidStoreAuthors
+from ..utils import get_author_by_bai
 from .base import InspireRecord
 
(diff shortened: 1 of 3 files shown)


backend/inspirehep/records/api/base.py
```diff
@@ -169,6 +169,20 @@ def get_record(cls, record_uuid, with_deleted=False, record_version=None):
             raise NoResultFound
         return record
 
+    @classmethod
+    def get_records_by_ids(cls, ids):
```
michamos, line 173: how is this different from `get_records_by_pids`?

MJedr: The major difference is that in `RecordsAuthors` there are not pids, we have only ids, so unless I don't do the join in `query_author_papers` I can't get them back from db (I haven't thought about it before, but might I change it). Then the issue is that I am getting literature records by calling `InspireRecord`, so I need to get class record class [there]( https://github.com/inspirehep/inspirehep/pull/2072/files#diff-fc03de0e68b26d840b1e069376322e3b7339601165daa83a7c4ac829b547b992R183-R184). If it's ok, I can modify `get_records_by_pids` by it there

michamos: you don't need to pass pid objects to `get_records_by_pids`, but `(pid_type, pid_value)` tuples, which is exactly what you're doing too, so I fail to see the difference. Why can't you use `LiteratureRecord` instead of `InspireRecord`, given you're already hardcoding `lit` pids?

MJedr: But id is not a pid, isn't it (and it's not in pid table)? If not, I can't pass it in the tuple because record won't be find.
No because of circular dependency issue

michamos (approved):

Issue #2047, opened 2021-07-15
Avoid creating duplicates on HAL push


michamos:
1. We often get `StaleDataError` during HAL push. This happens because the HAL push task races the author disambiguation task which is started at the same time, so it often happen that author disambiguation finishes, and commits the update, between the read of the HAL push and the write to the DB. 

We retry the task in this case: https://github.com/inspirehep/inspirehep/blob/69fa78b8d17f5000883982c1b90b9bdae785d7ed/backend/inspirehep/hal/tasks.py#L44 but the problem is that the HAL push is not idempotent, as the record might have been created successfully on the HAL side, causing a duplicate record to be created when retrying.

The only thing the HAL push writes to the record is the HAL ID, so it seems safe to deal with the exception inside the task and reload the record from DB when it happens. This can't cause missing updates on HAL as the other update will cause a new push in any case. 

2. We might also need to implement a lock system [like for ORCID](https://github.com/inspirehep/inspirehep/blob/69fa78b8d17f5000883982c1b90b9bdae785d7ed/backend/inspirehep/orcid/domain_models.py#L211) to avoid pushing concurrently from two different tasks, as this case is not handled correctly on their side (they said so by email).

MJedr:
**TECH NOTES**
1. [Here](https://github.com/inspirehep/inspirehep/blob/69fa78b8d17f5000883982c1b90b9bdae785d7ed/backend/inspirehep/hal/tasks.py#L58-L59) instead of raising an error reload record from db. Remove `StaleDataError` from autoretry in celery task.
2. For the lock system use [same context manager as in orcid](https://github.com/inspirehep/inspirehep/blob/69fa78b8d17f5000883982c1b90b9bdae785d7ed/backend/inspirehep/orcid/utils.py#L198). Use it after [this line](https://github.com/inspirehep/inspirehep/blob/69fa78b8d17f5000883982c1b90b9bdae785d7ed/backend/inspirehep/hal/tasks.py#L74) (maybe it would be reasonable to move the `distributed_lock` from orcid module to somewhere else). As the lock name use the `hal:{recid}`.

PR #2090, opened 2021-08-04
hal: handle stale data error for record update & introduce lock system

MJedr:
ref: #2047

### DIFF
--- a/backend/inspirehep/hal/tasks.py
+++ b/backend/inspirehep/hal/tasks.py
@@ -19,6 +19,7 @@ from sword2.exceptions import RequestTimeOut
 from inspirehep.hal.core.sword import create, update
 from inspirehep.hal.core.tei import convert_to_tei
 from inspirehep.records.api import LiteratureRecord
+from inspirehep.utils import distributed_lock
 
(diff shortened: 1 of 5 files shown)


backend/inspirehep/hal/tasks.py
```diff
@@ -72,13 +73,15 @@ def _hal_push(record):
...
         hal_id = hal_value[0] if hal_value else ""
-        if hal_id:
-            receipt = _hal_update(tei, hal_id, record)
-        else:
-            receipt = _hal_create(tei, record)
-        if receipt and receipt.id != hal_id:
-            _write_hal_id_to_record(record, receipt.id)
-        return receipt
+        lock_name = f"hal:{record['control_number']}"
+        with distributed_lock(lock_name, blocking=True):
```
drjova, line 77: I don't understand why we need the lock here :/

MJedr: Because we want to prevent the race condition with disambiguation as @michamos wrote in the description in #2047

drjova (approved):

Issue #2107, opened 2021-08-05
Flaky test `test_lit_record_removed_from_es_when_hard_deleted`


drjova:
```test_lit_record_removed_from_es_when_hard_deleted``` flaky test is failing because of:

```python
    def assert_hits():
        current_search.flush_and_refresh("records-hep")

        result = es_search("records-hep")

        result_total = get_value(result, "hits.total.value")

>       assert expected_hits_count == result_total

E       assert 1 == 2

```

PR #2113, opened 2021-08-06
fix flaky integration-async tests

MJedr:
ref: #2107

### DIFF
--- a/backend/tests/integration-async/records/indexer/test_literature.py
+++ b/backend/tests/integration-async/records/indexer/test_literature.py
@@ -10,8 +10,7 @@ import pytest
 from flask_sqlalchemy import models_committed
 from helpers.factories.models.user_access_token import AccessTokenFactory
 from helpers.providers.faker import faker
-from helpers.utils import es_search, retry_until_pass
-from inspire_utils.record import get_value
+from helpers.utils import retry_until_pass
 from invenio_db import db
 from invenio_search import current_search
 from sqlalchemy.orm.exc import StaleDataError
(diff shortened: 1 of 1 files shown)


backend/tests/integration-async/records/indexer/test_literature.py
```diff
@@ -31,12 +30,11 @@ def assert_record():
...
 
-def assert_es_hits_count(expected_hits_count):
+def assert_es_hits_for_record_id_count(recid, expected_hits_count=1):
     def assert_hits():
         current_search.flush_and_refresh("records-hep")
-        result = es_search("records-hep")
-        result_total = get_value(result, "hits.total.value")
-        assert expected_hits_count == result_total
+        hits = LiteratureSearch().get_record(str(recid)).execute().hits
+        assert expected_hits_count == len(hits)
```
drjova, line 37: I would just check if the control_number is on elasticsearch

backend/tests/integration-async/records/indexer/test_literature.py
```diff
@@ -31,12 +30,20 @@ def assert_record():
...
+        assert hits
+
+    retry_until_pass(assert_hits, retry_interval=5)
+
+
+def assert_record_not_in_es(recid):
+    def assert_hits():
+        current_search.flush_and_refresh("records-hep")
+        hits = LiteratureSearch().query_from_iq(f"recid:{recid}")
+        assert hits
```
drjova, line 46: how this is checking if the recid is not in the ES?

backend/tests/integration-async/records/indexer/test_literature.py
```diff
@@ -31,12 +30,20 @@ def assert_record():
...
+
+    retry_until_pass(assert_hits, retry_interval=5)
+
+
+def assert_record_not_in_es(recid):
+    def assert_hits():
+        current_search.flush_and_refresh("records-hep")
+        hits = LiteratureSearch().query_from_iq(f"recid:{recid}").execute().hits
+        assert not hits
```
drjova, line 46: this will always be `false` no? What is the structure of `hits`?

MJedr: It's a list of `hits` so if es doesn't find anything matching query it's empty

MJedr: And I think it should always be empty considering that recid is unique

backend/tests/integration-async/records/indexer/test_literature.py
```diff
@@ -270,7 +277,17 @@ def test_record_created_through_api_is_indexed(inspire_app, clean_celery_session
...
-    assert_es_hits_count(1)
+
+    def assert_record_in_es():
+        results = (
+            LiteratureSearch()
+            .query_from_iq(f"recid:{data['control_number']}")
+            .execute()
+            .hits
+        )
+        assert len(results) == 1
```
drjova, line 288: I would just assert if results

drjova: or can't we used the function you have created instead of duplicating it?

MJedr: yes, i've just done it

drjova (approved):

Issue #2125, opened 2021-08-09
Display students on advisor profile


michamos:
From https://rt.inspirehep.net/Ticket/Display.html?id=1121227:

> [...] one important feature
that got lost along the way relative to the previous version was the "students"
of any given individual. The "advisor" is there, and I assume the information
on students is still in the underlying database.

> This is particularly useful to understand potential conflicts of interest
(e.g. for review panels) and facilities understanding some aspects of career
trajectories (e.g. which experiments people were on and in what context). Would
it be possible to add the "students" feature back in the new version?

As discussed in the ticket, this would require reliably linking the students to the advisors. Currently we don't really link to an author's advisors (there is no record ref to the advisor record) but generate a search for the name in the author display. @hoc3426 mentions that many `advisors` fields contain an `INSPIRE ID`, which could be used to more reliably link back to the advisor.

This probably also requires creating a new table to hold this relation.

MJedr:
**TECH NOTES**
1. Create a new table students_advisors, which should contain: 
* student_id (record uuid),
* advisor_id (record uuid).
  Put an index on author_id column and advisor_id column.
2. Create a mixin which will: 
* prepare (generate) entries for table (by using $ref of the advisor in student record)
*  update table (put all the entries to the table).
*  delete all the entries from table. 
* hard delete all the entries from table. That mixing should be used in AuthorsRecord.
3. In AuthorsRecord, add a method to update relationships in the new table (update table method from mixin) and then call it in `create`.
4. Add a new field to ES schema for authors: `students` which will contain `name` and reference to student record.
5. In authors ES serializer add a new method that will populate `students` (for `preferred_name`, if it's not available use `name.value`).
6. Create a new method in AuthorsRecord that will take all the advisors if `full_name` or `preferred_name` changed. It should return uuids.
7. To reindex records affected by change, call the method from point 6 [here](https://github.com/inspirehep/inspirehep/blob/61348b82d55d950419daa9d5ced8436ef52e7aa2/backend/inspirehep/indexer/api.py#L28)

michamos:
it would be good to add the `degree_type` to the table and ES too, so we can display the type of student. We potentially have this info for the `advisor` (not required): https://github.com/inspirehep/inspire-schemas/blob/8b0dd65c43c4e890e99b7c766250b995526abaee/inspire_schemas/records/authors.yml#L52.

StellaCh:
[...]

PR #2146, opened 2021-08-13
authors: add student-advisor relationship

MJedr:
ref: #2125

### DIFF
--- a/backend/inspirehep/records/api/authors.py
+++ b/backend/inspirehep/records/api/authors.py
@@ -7,7 +7,10 @@
 import structlog
 from inspire_utils.record import get_values_for_schema
 from invenio_db import db
+from invenio_pidstore.models import PersistentIdentifier
 
+from inspirehep.pidstore.api.base import PidStoreBase
+from inspirehep.records.api.mixins import StudentsAdvisorMixin
 from inspirehep.records.marshmallow.authors import AuthorsElasticSearchSchema
 from inspirehep.records.models import RecordsAuthors
 from inspirehep.search.api import AuthorsSearch
@@ -19,7 +22,7 @@ from .base import InspireRecord
 LOGGER = structlog.getLogger()
 
 
-class AuthorsRecord(InspireRecord):
+class AuthorsRecord(InspireRecord, StudentsAdvisorMixin):
     """Authors Record."""
 
     es_serializer = AuthorsElasticSearchSchema
@@ -68,6 +71,7 @@ class AuthorsRecord(InspireRecord):
     def create(cls, data, id_=None, *args, **kwargs):
         record = super().create(data, id_, **kwargs)
         record.assign_author_to_papers()
+        record.update_students_advisors_table()
         return record
 
     def assign_author_to_papers(self):
@@ -97,3 +101,44 @@ class AuthorsRecord(InspireRecord):
 
         for data in query.yield_per(100).with_entities(RecordsAuthors.record_id):
             yield data.record_id
+
+    def update(self, data, *args, **kwargs):
+        with db.session.begin_nested():
+            super().update(data, *args, **kwargs)
+            self.update_students_advisors_table()
+
+    def get_linked_advisors_when_name_changes(self):
+        if not self.get("advisors"):
+            return set()
+        if (
+            self.get_value("name.preferred_name")
+            != self._previous_version.get_value("name.preferred_name")
+        ) or (
+            not self.get_value("name.preferred_name")
+            and (
+                self.get_value("name.value")
+                != self._previous_version.get_value("name.value")
+            )
+        ):
+            advisors_references = self.get_value("advisors.record.$ref")
+            advisors_pids = [
+                PidStoreBase.get_pid_from_record_uri(uri)[1]
+                for uri in advisors_references
+            ]
+            advisor_uuids = (
+                PersistentIdentifier.query.with_entities(
+                    PersistentIdentifier.object_uuid
+                )
+                .filter(
+                    PersistentIdentifier.pid_type == "aut",
+                    PersistentIdentifier.pid_value.in_(advisors_pids),
+                )
+                .all()
+            )
+            return set(str(uuid_list[0]) for uuid_list in advisor_uuids)
+
+        return set()
+
+    def hard_delete(self):
+        self.delete_students_advisors_table_entries()
+        super().hard_delete()
--- a/backend/inspirehep/records/api/mixins.py
+++ b/backend/inspirehep/records/api/mixins.py
@@ -7,9 +7,11 @@
 
 import structlog
 from flask import current_app
+from inspire_dojson.utils import get_recid_from_ref
 from inspire_utils.date import fill_missing_date_parts
 from inspire_utils.record import get_values_for_schema
 from invenio_db import db
+from invenio_pidstore.models import PersistentIdentifier
 from sqlalchemy import and_, func, not_, or_, text
 
 from inspirehep.records.models import (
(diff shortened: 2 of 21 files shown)


backend/inspirehep/records/api/mixins.py
```diff
@@ -587,3 +590,79 @@ def get_modified_experiment_uuids(self):
...
+                    uuid=str(self.id),
+                )
+                continue
+            advisor_recid = get_recid_from_ref(advisor["record"])
+            advisor_record_uuid = (
+                PersistentIdentifier.query.with_entities(
+                    PersistentIdentifier.object_uuid
+                )
+                .filter_by(pid_value=str(advisor_recid), pid_provider="recid")
+                .scalar()
```
michamos, line 614: you're doing one query per advisor here. It would be better to do one query per record instead. Why do you need to write this query BTW, don't we have helpers already to get a list of record UUIDs given a list of PIDs?

backend/inspirehep/records/api/mixins.py
```diff
@@ -587,3 +590,79 @@ def get_modified_experiment_uuids(self):
...
+        return table_entries_buffer
+
+    def update_students_advisors_table(self):
+        """Puts all advisors and student ids in students_advisors table"""
+        deleted_count = self.delete_students_advisors_table_entries()
+
+        if (
+            self.get("deleted", False)
+            or self.pid_type not in ["aut"]
+            or "Authors" not in self["_collections"]
```
michamos, line 633: how can this happen? aren't you adding the mixin to `AuthorsRecord` only?

backend/inspirehep/records/marshmallow/authors/base.py
```diff
@@ -16,6 +21,32 @@ class AuthorsRawSchema(RecordBaseSchema):
...
+            .filter(
+                StudentsAdvisors.student_id == RecordMetadata.id,
+                StudentsAdvisors.advisor_id == record.id,
+            )
+            .all()
+        )
+        student_entries = []
+        for student_json, student_degree in student_records:
+            student_data = {
+                "name": self.get_student_name(student_json),
```
michamos, line 43: wouldn't it be better to store this data in the new table too, instead of doing the join here?

MJedr: I think that this should be quite fast since we have index on `student_id`

backend/inspirehep/records/api/mixins.py
MJedr: Does it matter so much? How many advisors we might have per record? Of course I can make one query, but then I'll have to do some transformations to retrieve refs and then match query results with advisors cause an advisor might not have `$ref`.

No, we have only `get_records_by_pids` which returns all the metadata.

michamos: OK, I see how it's a little bit more work. I don't know if it's worth it. I think so, but I'll let @drjova decide. If not, you should probably use `pid_type` instead of `pid_provider` (everywhere).

backend/inspirehep/alembic/6063cc09aec9_add_students_advisors_table.py
```diff
@@ -0,0 +1,70 @@
...
+            name="fk_students_advisors_student_id",
+        ),
+        sa.PrimaryKeyConstraint(
+            "advisor_id", "student_id", name=op.f("pk_students_advisors")
+        ),
+    )
+    op.create_index(
+        "ix_students_advisors_student_id",
+        "students_advisors",
+        ["advisor_id", "student_id"],
```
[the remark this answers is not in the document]
MJedr: But won't the index on `student_id` and `advisor_id` be useful to this kind of search?
https://github.com/inspirehep/inspirehep/pull/2146/files/cee4526d5fd927cfc6764d5591191779980c6f74#diff-e35a8a45c8ecb58888ebe46c544224d041c296a28a295f6dc85be38d1423f916R34-R37

backend/inspirehep/records/marshmallow/authors/base.py
drjova: IMO it's better to use a simple join instead of duplicating data

backend/inspirehep/alembic/6063cc09aec9_add_students_advisors_table.py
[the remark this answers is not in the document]
michamos: I think there are several points you're misundertsanding.

* Here you're creating an index that's exactly the same as the one provided by the primary key (same columns, same order). That's never useful, it just creates a duplicate data structure. A primary key always creates an index, as the whole point of a primary key is to efficiently retrieve rows given the data in the primary key, which requires an index (and also to enforce the uniqueness constraint, which also requires an index to do efficiently).
* In a multi-column index, the ordering is important. Basically, an index (or at least a B-tree index, which is the default for non-composite fields) is a data structure that maintains a sorted list of rows. The sort order is specified by the columns. So on a `(column1, column2)` index, you can efficiently query a row given specific values for `column1` and `column2`, but also given specific values for `column1` only. However, you can't easily find rows given `column2` only. You can think of this in analogy to a dictionary (the book, not the data structure), where you need to find a definition. Because the dictionary is sorted, you can easily find a word or a prefix of a word, but finding a suffix requires scanning the whole dictionary. 
In your linked query, you're specifying `student_id` and want to get all `advisor_id`s that match, but that's exactly like trying to find a suffix given the ordering you've specified here. So you need either to define an index in the opposite order, or probably better, one containing only `student_id` (getting the rows requires going back to the main table then, but that's needed in any case because your index doesn't `degree_type`, and the index will be smaller so more efficient).

If you have some time to spare and want to learn more, I highly recommend reading https://use-the-index-luke.com/.

backend/inspirehep/records/api/authors.py
```diff
@@ -97,3 +101,50 @@ def query_author_papers(bai):
...
+                != self._previous_version.get_value("name.preferred_name")
+            )
+            or (
+                self.get_value("name.preferred_name")
+                != self._previous_version.get_value("name.preferred_name")
+            )
+            or (
+                self.get_value("name.value")
+                != self._previous_version.get_value("name.value")
+            )
```
michamos, line 126: * the second clause is now a superset of the first one
* it's a bit too broad (`name.value` changes only matter if there's no `preferred_name`), however I don't think that really matters. We'll be reindexing advisors a little bit too frequently in some cases.

backend/inspirehep/records/api/authors.py
```diff
@@ -97,3 +101,50 @@ def query_author_papers(bai):
...
+            advisors_pids = [
+                PidStoreBase.get_pid_from_record_uri(uri)[1]
+                for uri in advisors_references
+            ]
+            advisor_uuids = (
+                PersistentIdentifier.query.with_entities(
+                    PersistentIdentifier.object_uuid
+                )
+                .filter(
+                    PersistentIdentifier.pid_provider == "recid",
```
michamos, line 138: you don't need the `pid_provider` if also specifying `pid_type` (anywhere)

backend/inspirehep/records/api/mixins.py
```diff
@@ -587,3 +590,75 @@ def get_modified_experiment_uuids(self):
...
+                    recid=self.get("control_number"),
+                    uuid=str(self.id),
+                )
+                continue
+            advisor_recid = get_recid_from_ref(advisor["record"])
+            advisor_record_uuid = (
+                PersistentIdentifier.query.with_entities(
+                    PersistentIdentifier.object_uuid
+                )
+                .filter_by(pid_value=str(advisor_recid), pid_provider="recid")
```
michamos, line 613: better use `pid_type`

drjova (approved):

Issue #1947, opened 2021-06-14
Harvest hepdata 


StellaCh:
HEPData records are often (always?) supplementary material related to an INSPIRE Literature record. For INSPIRE records that have a corresponding HEPData record, we need to add an external link to HEPData (through `external_system_identifiers` with `schema=HEPDATA`). 

On legacy, this was done in https://github.com/inspirehep/inspire/blob/master/bibtasklets/bst_hepdata.py, by basically doing a [search for all HEPData records through their API](https://github.com/inspirehep/inspire/blob/37c905be9a569a6c25ced045eb84545ddb7ac3a5/bibtasklets/bst_hepdata.py#L84-L103), extracting the `inspire_id` from the response, comparing that to the set of all INSPIRE records having a `HEPData` link and filling in the gaps. If a record needs to be linked to HEPData, the `external_system_identifiers.value` should be set to `ins{inspire_recid}`.

Instead of querying all of HEPData every time (especially because we'll hit the ES 10000 record search limit in a couple of years), we could use the dedicated API which returns exactly the data we need: https://www.hepdata.net/search/ids?inspire_ids=true&last_updated=2021-08-10&sort_by=latest

MJedr:
**TECH NOTES**
Create a cron job that will run every day and that will query hepdata api https://www.hepdata.net/search/ids?inspire_ids=true&last_updated=2021-08-10&sort_by=latest (pass correct date). Add each returned id to the `external_identifiers` of a corresponding inspire record (entry should have `schema: HEPDATA` and value `ins{returned_recid}`)

PR #2148, opened 2021-08-17
backend: hepdata harvest

MJedr:
ref: #1947

### DIFF
--- a/backend/inspirehep/hepdata/cli.py
+++ b/backend/inspirehep/hepdata/cli.py
@@ -1,0 +1,64 @@
+# -*- coding: utf-8 -*-
+#
+# Copyright (C) 2021 CERN.
+#
+# inspirehep is free software; you can redistribute it and/or modify it under
+# the terms of the MIT License; see LICENSE file for more details.
+from datetime import datetime, timedelta
+
+import click
+import requests
+import structlog
+from inspire_utils.record import get_values_for_schema
+from invenio_db import db
+
+from inspirehep.records.api import LiteratureRecord
+
+LOGGER = structlog.getLogger()
+HEPDATA_URL = "https://www.hepdata.net/search/ids"
+
+
+@click.group()
+def hepdata():
+    """Commands for disambiguation"""
+
+
+@hepdata.command("harvest")
+@click.option(
+    "--since",
+    "-s",
+    "since",
+    default=None,
+    help="Date from when hepdata should be harvested. If not provided the date from a day before would be used.",
+)
+def harvest(since):
+    if since:
+        try:
+            since = datetime.strptime(since, "%Y-%m-%d").date()
+        except ValueError:
+            raise ValueError(
+                f"`since`: {since} is in wrong format. Should be in ISO format: YYYY-MM-DD."
+            )
+    else:
+        since = (datetime.now() - timedelta(1)).strftime("%Y-%m-%d")
+    payload = {"inspire_ids": True, "last_updated": since, "sort_by": "latest"}
+    hepdata_response = requests.get(HEPDATA_URL, params=payload)
+    hepdata_response.raise_for_status()
+    ids = hepdata_response.json()
+    LOGGER.info(
+        "Collected hepdata ids, assigning them to INSPIRE records.",
+        number_of_collected_ids=len(ids),
+    )
+    records_to_update = list(
+        LiteratureRecord.get_records_by_pids([("lit", str(recid)) for recid in ids])
+    )
+    for record in records_to_update:
+        record_external_identifiers = record.setdefault(
+            "external_system_identifiers", []
+        )
+        if not get_values_for_schema(record_external_identifiers, "HEPDATA"):
+            record_external_identifiers.append(
+                {"schema": "HEPDATA", "value": f"ins{record['control_number']}"}
+            )
+            record.update(dict(record))
+    db.session.commit()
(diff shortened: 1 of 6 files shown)


backend/inspirehep/hepdata/cli.py
```diff
@@ -0,0 +1,66 @@
...
+    """Commands for disambiguation"""
+
+
+@hepdata.command("harvest")
+@click.option(
+    "--since",
+    "-s",
+    "since",
+    default=None,
+    help="Date from when hepdata should be harvested. If not provided today's date would be used.",
```
michamos, line 29: Maybe use yesterday instead? otherwise there's a gap between the beginning of the day and the harvest time. Another approach is to store the last run info like we do for CDS, but that might not be needed for such a simple thing.

michamos: You need to update the docstring

backend/inspirehep/hepdata/cli.py
```diff
@@ -0,0 +1,66 @@
...
+    else:
+        since = (datetime.now() - timedelta(1)).strftime("%Y-%m-%d")
+    payload = {"inspire_ids": True, "last_updated": since, "sort_by": "latest"}
+    hepdata_response = requests.get(HEPDATA_URL, params=payload)
+    try:
+        hepdata_response.raise_for_status()
+    except Exception as exc:
+        message = "Hepdata harvest failed!"
+        click.echo(click.style(message, fg="red"))
+        click.echo(exc.message)
```
michamos, line 48: what's the point of this? I think you can just let it fail and the exception will go to sentry and the logs.

backend/inspirehep/hepdata/cli.py
```diff
@@ -0,0 +1,66 @@
...
+    ids = hepdata_response.json()
+    click.echo(f"Collected {len(ids)} hepdata ids, assigning them to INSPIRE records.")
```
michamos, line 50: nobody is reading these messages, it won't be an interactive command but a cronjob in most cases. Wouldn't it make more sense to use `structlog` here?

backend/inspirehep/hepdata/cli.py
```diff
@@ -0,0 +1,66 @@
...
+    records_to_update = list(
+        LiteratureRecord.get_records_by_pids([("lit", str(recid)) for recid in ids])
+    )
+    for record in records_to_update:
+        record_external_identifiers = record.get("external_system_identifiers", [])
+        record_external_identifiers_no_hepdata = [
+            identifier
+            for identifier in record_external_identifiers
+            if identifier["schema"] != "HEPDATA"
+        ]
```
michamos, line 60: I think it would make sense to distinguish different cases here:
* no `HEPDATA`: update record
*  `HEPDATA` already in record: do nothing (might have been harvested already yesterday)

drjova (approved):

backend/inspirehep/hepdata/cli.py
```diff
@@ -0,0 +1,66 @@
...
+        record_external_identifiers_no_hepdata.append(
+            {"schema": "HEPDATA", "value": f"ins{record['control_number']}"}
+        )
+        record["external_system_identifiers"] = record_external_identifiers_no_hepdata
+        record.update(dict(record))
+        db.session.commit()
```
michamos, line 66: maybe do everything in one transaction?

backend/tests/integration/hepdata/test_cli.py
```diff
@@ -0,0 +1,54 @@
...
+    rec_1 = LiteratureRecord.get_record_by_pid_value(record_1["control_number"])
+    assert rec_1["external_system_identifiers"] == [
+        {"schema": "HEPDATA", "value": "ins1882568"}
+    ]
+
+
+def test_hepdata_harvest_incorrect_date_passed(inspire_app, cli):
+    create_record("lit", data={"control_number": 1882568})
+    result = cli.invoke(["hepdata", "harvest", "--since", "2021-08"])
+    assert result.exit_code == -1
```
michamos, line 46: why `-1`? I would expect `0` for success.

MJedr: Because date is wrong, so it won't be a success

backend/inspirehep/hepdata/cli.py
```diff
@@ -0,0 +1,63 @@
...
+        LiteratureRecord.get_records_by_pids([("lit", str(recid)) for recid in ids])
+    )
+    for record in records_to_update:
+        record_external_identifiers = record.get("external_system_identifiers", [])
+        if not get_values_for_schema(record_external_identifiers, "HEPDATA"):
+            record_external_identifiers.append(
+                {"schema": "HEPDATA", "value": f"ins{record['control_number']}"}
+            )
+        record["external_system_identifiers"] = record_external_identifiers
+        record.update(dict(record))
```
michamos, line 62: this should be moved inside the `if`

backend/inspirehep/hepdata/cli.py
```diff
@@ -0,0 +1,63 @@
...
+    ids = hepdata_response.json()
+    LOGGER.info(
+        "Collected hepdata ids, assigning them to INSPIRE records.",
+        number_of_collected_ids=len(ids),
+    )
+    records_to_update = list(
+        LiteratureRecord.get_records_by_pids([("lit", str(recid)) for recid in ids])
+    )
+    for record in records_to_update:
+        record_external_identifiers = record.get("external_system_identifiers", [])
```
michamos, line 56:
```suggestion
        record_external_identifiers = record.setdefault("external_system_identifiers", [])
```
so you can get rid of L61.

michamos (approved):

Issue #2080, opened 2021-08-02
Don't allow attaching HTML/JS/CSS to records


michamos:
Currently we don't impose any restrictions on the type of files attached to a record. That causes issues when we expect a PDF but we get some publisher page instead (because we have been detected as a robot and banned) but also poses security issues as we serve the files using the `inspirehep.net` domain, so any page and script that we host has access to all associated secrets (e.g. cookies).

We should simply disallow these file types, as there is currently no real use for attaching them. This would catch problems and avoid the security issue.

MJedr:
**TECH NOTES**
1. Check the mime type in the response headers [here](https://github.com/inspirehep/inspire-next/blob/52b1795c12df0a15bb9bf2daaa5e974d16233ed0/inspirehep/modules/workflows/utils/__init__.py#L325) and if it’s `HTML/JS/CSS` don’t attach the file.

michamos:
I don't think that's a sufficient approach. We need to forbid these files from being attached to INSPIRE records, not merely to workflows, as files might be attached directly without being downloaded by workflows first (e.g. inputting the URL in the record editor).

PR #2149, opened 2021-08-17
records: disable attaching html/js files

oguzdemirbasci:
Ref: #2080


backend/inspirehep/records/api/literature.py
```diff
@@ -484,6 +487,14 @@ def add_file(
             filename = filename or key
             if not filename:
                 filename = new_key
+            if mimetype in RESTRICTED_MIMETYPES:
```
drjova, line 490: I would add this in the `config.py` to be able to configure it if needed

drjova: `FILES_RESTRICTED_MIMETYPES`

backend/inspirehep/records/api/literature.py
```diff
@@ -484,6 +487,14 @@ def add_file(
...
+                LOGGER.error(
+                    "Unsupported file type - Aborting",
+                    key=key,
+                    mimetype=mimetype,
+                    thread=threading.get_ident(),
+                )
+                raise UnsupportedFileError(mimetype)
```
michamos, line 497: This will cause a 500 and display a generic message in the editor. I think it would be good to expose this to the API (the invenio way is to subclass `invenio-rest.RESTException`, not sure that's how it should be done here). @drjova wdyt?

oguzdemirbasci: I updated `UnsupportedFileError` definition. It is a subclass of `RESTException` now. I think it should work.

drjova: Inheriting `RESTException` should be enough

backend/tests/integration/records/api/test_api_literature.py
```diff
@@ -1503,6 +1507,60 @@ def test_update_record_add_more_documents(inspire_app, s3):
...
+    expected_documents = [
+        {
+            "source": "PoS",
+            "key": expected_document_key,
+            "url": current_s3_instance.get_public_url(expected_document_key),
+            "original_url": "http://original-url.com/4",
+            "filename": "PoS(ICHEP2020)903.pdf",
+        },
+    ]
+    with pytest.raises(UnsupportedFileError):
```
drjova, line 1556: let's tests also the code and message on an API call

backend/inspirehep/records/errors.py
```diff
@@ -58,6 +58,14 @@ class DownloadFileError(Exception):
     pass
 
 
+class UnsupportedFileError(RESTException):
+    def __init__(self, message="", **kwargs):
```
drjova, line 62:
```suggestion
    def __init__(self, mimetype="", **kwargs):
```

drjova (approved):
@oguzdemirbasci just the comment, LGTM

Issue #2123, opened 2021-08-09
make permission system more flexible


michamos:
From https://inspirehep.zulipchat.com/#narrow/stream/195298-experts/topic/BaBar.20internal.20notes.20access

We currently have a 3-tiered permission system for literature-type records: public (`Literature` collection), hidden (not visisble by default), private (visible only to curators). The private collections however need to be accessible by selected groups of people. We should add for each private collection two different roles, one allowing to view/search records in that collection, and the other allowing also to edit records.

drjova:
TECH NOTES:
* Permissions should be `name-of-collection-read` and `name-of-collection-read-write`
* Update [LiteraturePermissionCheck](https://github.com/inspirehep/inspirehep/blob/eb6fa8d303b2007adf1583971b98281d4f63756a/backend/inspirehep/access_control.py#L40) to check for permissions if the user has the role for this specific collection
* Update [SessionCatalogerPermission](https://github.com/inspirehep/inspirehep/blob/master/backend/inspirehep/records/config.py#L122)  to check for permissions if the user has the role for this specific collection
* Update [query_for_superuser_or_users](https://github.com/inspirehep/inspirehep/blob/master/backend/inspirehep/search/api.py#L188) to check for permissions if the user has the role for this specific collection
* Create a new mixin similar to [CatalogerCanEditMixin](https://github.com/inspirehep/inspirehep/blob/ba5fe0fdd9aa4daebad985376068487edccf995a/backend/inspirehep/records/marshmallow/common/mixins.py#L10) to check for permissions if the user has the role for this specific collection and replace `CatalogerCanEditMixin` [here]( https://github.com/inspirehep/inspirehep/blob/ba5fe0fdd9aa4daebad985376068487edccf995a/backend/inspirehep/records/marshmallow/literature/ui.py#L45)

michamos:
I think you're missing the editor views in https://github.com/inspirehep/inspirehep/blob/master/backend/inspirehep/editor/views.py

PR #2163, opened 2021-08-20
permissions: `read` and `read_write` permissions added for collections.

oguzdemirbasci:
ref: #2123

### DIFF
--- a/backend/inspirehep/accounts/api.py
+++ b/backend/inspirehep/accounts/api.py
@@ -7,6 +7,7 @@
 
 
 import structlog
+from flask import current_app
 from flask_login import current_user
 from inspire_utils.record import get_value, get_values_for_schema
 from invenio_oauthclient.models import RemoteAccount, UserIdentity
@@ -19,10 +20,23 @@ LOGGER = structlog.getLogger()
 
 def is_superuser_or_cataloger_logged_in():
     """Check if current authenticated user is cataloger/superuser."""
-    if current_user and current_user.is_authenticated:
+    if is_user_logged_in():
         user_roles = {role.name for role in current_user.roles}
         return bool(user_roles & {Roles.cataloger.value, Roles.superuser.value})
     return False
+
+
+def is_superuser_logged_in():
+    """Check if current authenticated user is superuser."""
+    if is_user_logged_in():
+        user_roles = {role.name for role in current_user.roles}
+        return bool(user_roles & {Roles.superuser.value})
+    return False
+
+
+def is_user_logged_in():
+    """Check if current authenticated user exists"""
+    return current_user and current_user.is_authenticated
 
 
 def can_user_edit_record(record):
@@ -66,3 +80,49 @@ def get_current_user_orcid():
 
 def get_current_user_remote_orcid_account():
     return RemoteAccount.query.filter_by(user_id=current_user.get_id()).one_or_none()
+
+
+def get_allowed_roles_for_collections(collections, read_only=True):
+    private_collections = set(collections) - set(
+        current_app.config["NON_PRIVATE_LITERATURE_COLLECTIONS"]
+    )
+    roles = []
+    for col in private_collections:
+        collection_role_prefix = col.lower().replace(" ", "-")
+        roles.append(collection_role_prefix + "-read-write")
+        if read_only:
+            roles.append(collection_role_prefix + "-read")
+    return roles
+
+
+def _check_permissions_for_private_collections(collections, read_only=True):
+    """
+    Check if the user has read or write access.
+
+    Parameters:
+        collections (list): list of collections
+        read_only (bool): True, read permission check. False, read/write permission check.
+
+    Returns:
+        set: allowed roles. Access granted if not empty.
+    """
+    # superuser and cataloger always have access
+    allowed_roles = [Roles.superuser.value, Roles.cataloger.value]
+    if collections:
+        allowed_collection_roles = get_allowed_roles_for_collections(
+            collections, read_only
+        )
+        allowed_roles.extend(allowed_collection_roles)
+
+    user_roles = {role.name for role in current_user.roles}
+    has_access = user_roles & set(allowed_roles)
+
+    return has_access
+
+
+def check_permissions_for_private_collection_read(collections):
+    return _check_permissions_for_private_collections(collections, read_only=True)
+
+
+def check_permissions_for_private_collection_read_write(collections):
+    return _check_permissions_for_private_collections(collections, read_only=False)
--- a/backend/inspirehep/editor/views.py
+++ b/backend/inspirehep/editor/views.py
@@ -17,7 +17,11 @@ from invenio_records_rest.utils import set_headers_for_record_caching_and_concur
 from refextract import extract_references_from_string, extract_references_from_url
 from sqlalchemy_continuum import transaction_class, version_class
 
-from inspirehep.accounts.decorators import login_required_with_roles
+from inspirehep.accounts.api import (
+    check_permissions_for_private_collection_read,
+    check_permissions_for_private_collection_read_write,
+)
+from inspirehep.accounts.decorators import login_required, login_required_with_roles
 from inspirehep.accounts.roles import Roles
 from inspirehep.files.api import current_s3_instance
 from inspirehep.matcher.api import match_references
@@ -28,8 +32,8 @@ from inspirehep.rt import tickets
 from inspirehep.serializers import jsonify
 from inspirehep.utils import hash_data
 
-from .authorlist_utils import authorlist
 from ..rt.errors import EmptyResponseFromRT, NoUsersFound
+from .authorlist_utils import authorlist
 from .errors import EditorGetRevisionError, EditorRevertToRevisionError
 
 blueprint = Blueprint("inspirehep_editor", __name__, url_prefix="/editor")
@@ -37,7 +41,7 @@ LOGGER = structlog.getLogger()
 
 
 @blueprint.route("/<endpoint>/<int:pid_value>/revisions/revert", methods=["PUT"])
-@login_required_with_roles([Roles.cataloger.value])
+@login_required
 def revert_to_revision(endpoint, pid_value):
     """Revert given record to given revision"""
     try:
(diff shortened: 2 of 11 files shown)


backend/inspirehep/search/api.py
```diff
@@ -185,16 +187,25 @@ def execute(self, *args, **kwargs):
                 LOGGER.exception("Match reference error.", query_string=query_string)
         return results
 
-    def query_for_superuser_or_users(self, query_string):
+    def query_by_user_role(self, query_string):
         if not is_superuser_or_cataloger_logged_in():
             if "_collections" in query_string:
+                collections = current_app.config["NON_PRIVATE_LITERATURE_COLLECTIONS"]
```
MJedr, line 193: Maybe you can rename this variable to `non_private_collection` because this name might be a bit confusing.

backend/inspirehep/search/api.py
```diff
@@ -185,16 +187,25 @@ def execute(self, *args, **kwargs):
...
+                if current_user and current_user.is_authenticated:
+                    private_collection_roles = set(current_user.roles) - set(
+                        map(lambda c: c.value, Roles)
```
MJedr, line 196: Maybe ` {role.value for role in Roles}`, it would be more pythonic way I believe.

backend/inspirehep/search/api.py
```diff
@@ -185,16 +187,25 @@ def execute(self, *args, **kwargs):
                 LOGGER.exception("Match reference error.", query_string=query_string)
         return results
 
-    def query_for_superuser_or_users(self, query_string):
+    def query_by_user_role(self, query_string):
```
MJedr, line 190: This function became very complex with lots of indentation levels, I think that it might be good to split it in parts

backend/inspirehep/accounts/api.py
```diff
@@ -66,3 +75,32 @@ def get_current_user_orcid():
...
+    # superuser can do everything
+    roles.append(Roles.superuser.value)
+    user_roles = {role.name for role in current_user.roles}
+
+    # read-write roles to have read access
+    if not readwrite:
+        user_roles = {
+            role if not role.endswith("-read-write") else role[:-6]
+            for role in user_roles
+        }
```
MJedr, line 102: I am not sure if I understand it. Why can't you add `-read` for everything in line 91 (`read-write` has always access to read)?

backend/inspirehep/records/marshmallow/common/mixins.py
```diff
@@ -17,3 +20,29 @@ def set_can_edit(data, orginal_data):
...
+        self.post_dumps.append(self.set_can_edit)
+
+    @staticmethod
+    def set_can_edit(data, orginal_data):
+        if is_superuser_or_cataloger_logged_in():
+            data["can_edit"] = True
+        elif current_user and current_user.is_authenticated:
+            private_collection_roles = set(current_user.roles) - set(
+                map(lambda c: c.value, Roles)
+            )
```
michamos, line 37: This is not JS ;)
```suggestion
            private_collection_roles = set(current_user.roles) - {role.value for role in Roles}
```

backend/inspirehep/accounts/api.py
```diff
@@ -66,3 +75,32 @@ def get_current_user_orcid():
 
 def get_current_user_remote_orcid_account():
     return RemoteAccount.query.filter_by(user_id=current_user.get_id()).one_or_none()
+
+
+def check_permissions_for_private_collections(
+    roles=[], collections=None, readwrite=False
```
michamos, line 81: * :warning: [mutable default argument](https://docs.python-guide.org/writing/gotchas/#mutable-default-arguments) for `roles`, this is almost certainly not what you want. 
* Also it's not clear how this function is supposed to be used, what is the supplied `roles` supposed to be? Does it actually need to be parametric or is it always set to `[Roles.cataloger.value]`? 
* I also have the impression that this function is doing too many things and is mixing up the allowed roles for the collections and the user roles, I would split it into logical parts, which would be something like
    + define `get_allowed_roles_for_collections(collections, read_only=True)`: returns the union of all allowed roles for all collections, `read_only=True` returns `-read` and `read-write`, `False` returns only `-read-write` (this includes `superuser` and `cataloger`.
    + use that to intersect with the current user roles and give a result
    + then define two non-parametric `can_user_read_collections`/`can_user_edit_collections` that take only a `collections` arg.

backend/inspirehep/editor/views.py
```diff
@@ -28,48 +29,61 @@
...
             pid_value, pid_type, original_record=True
         )
-        revision_id = request.json["revision_id"]
-        record.revert(revision_id)
-        db.session.commit()
-        return jsonify(success=True)
+
+        if check_permissions_for_private_collections(
+            [Roles.cataloger.value], record.get("_collections", []), readwrite=True
+        ):
```
michamos, line 52: I would invert the logic for clarity:
```python
if not ...:
    return jsonify(message="Unauthorized, code=403), 403
    
# authorized here, so do the work

File: backend/inspirehep/editor/views.py [hunk 8]
```diff
@@ -28,48 +29,61 @@
...
     record = InspireRecord.get_record_by_pid_value(
         pid_value, pid_type, original_record=True
     )
-    json = {"record": {"metadata": record}, "schema": load_schema(record["$schema"])}
 
-    response = make_response(json)
-    set_headers_for_record_caching_and_concurrency(response, record)
+    if check_permissions_for_private_collections(
+        [Roles.cataloger.value], record.get("_collections", [])
+    ):
```
michamos, line 72: same

backend/inspirehep/editor/views.py
```diff
@@ -79,26 +93,34 @@ def get_revisions(endpoint, pid_value):
...
-                    "revision_id": revision.revision_id,
-                    "user_email": user_email,
-                    "transaction_id": transaction_id,
-                    "rec_uuid": record.id,
-                }
-            )
-        return jsonify(revisions)
+        if check_permissions_for_private_collections(
+            [Roles.cataloger.value], record.get("_collections", [])
+        ):
```
michamos, line 98: same

backend/inspirehep/records/marshmallow/common/mixins.py
```diff
@@ -17,3 +20,29 @@ def set_can_edit(data, orginal_data):
...
         return data
+
+
+class CanEditByCollectionPermissionMixin:
+    def __init__(self, *args, **kwargs):
+        super().__init__()
+        self.post_dumps.append(self.set_can_edit)
+
+    @staticmethod
+    def set_can_edit(data, orginal_data):
```
michamos, line 31: can you have somehow only a single permission checking logic, and reuse it here, instead of doing the same thing twice?

backend/inspirehep/search/api.py
```diff
@@ -185,16 +187,25 @@ def execute(self, *args, **kwargs):
...
+                    )
+                    for role in private_collection_roles:
+                        collection_role = role.name
+                        col = None
+                        if "-read-write" in collection_role:
+                            col = collection_role.split("-read-write")[0]
+                        elif "-read" in collection_role:
+                            col = collection_role.split("-read")[0]
+                        if col and col not in collections:
+                            collections.append(col)
```
michamos, line 206: I would split this check into a different function

backend/inspirehep/accounts/api.py
oguzdemirbasci: Yes, splitting the function is a good idea. When `read_only=True` in `get_allowed_roles_for_collections`, if it returns both `-read` and `-read-write`, users with `-read` role can have the write permission. Is it what we want? Or do you suggest to check all roles in the intersection?

michamos: we allow if there's a non-empty intersection between the user roles and the allowed roles. The `read-write` role allows to read, so `read_only=True` (corresponding to a `read` operation) should also include `read-write`, in order for a user with `read-write` to be allowed to read. Or am I misunderstanding something?

backend/inspirehep/accounts/api.py
```diff
@@ -66,3 +80,46 @@ def get_current_user_orcid():
...
     return RemoteAccount.query.filter_by(user_id=current_user.get_id()).one_or_none()
+
+
+def get_allowed_roles_for_collections(collections, read_only=True):
+    private_collections = set(collections) - set(
+        current_app.config["NON_PRIVATE_LITERATURE_COLLECTIONS"]
+    )
+    roles = []
+    for col in private_collections:
+        roles.append(col + "-read-write")
```
drjova, line 91: this will not work, for example `HEP HIDDEN-read-write` is not a valid role. We need to replace ` ` with `-` and maybe lowercase the collection

backend/inspirehep/search/api.py
```diff
@@ -185,16 +188,29 @@ def execute(self, *args, **kwargs):
...
+            roles_to_names = current_app.config["COLLECTION_ROLES_TO_COLLECTION_NAMES"]
+            return [
+                roles_to_names[role] for role in roles if role in roles_to_names.keys()
+            ]
+
+        def query_by_collections_for_users():
+            non_private_collections = copy.deepcopy(
+                current_app.config["NON_PRIVATE_LITERATURE_COLLECTIONS"]
+            )
+            if current_user and current_user.is_authenticated:
```
drjova, line 202: we can reuse `is_logged_in_user` function

backend/inspirehep/search/api.py
```diff
@@ -185,16 +188,29 @@ def execute(self, *args, **kwargs):
...
+                collection_roles = get_collection_for_user_roles(current_user.roles)
```
drjova, line 203: collections

backend/tests/integration-async/editor/test_editor_api.py
```diff
@@ -76,6 +95,45 @@ def test_get_revisions(inspire_app, clean_celery_session, record_with_two_revisi
...
+):
+    user_readwrite = create_user(role="hep-hidden-read-write")
+    with inspire_app.test_client() as client:
+        login_user_via_session(client, email=user_readwrite.email)
+        response = client.get(
+            "/api/editor/literature/111/revisions", content_type="application/json"
+        )
+        assert response.status_code == 200
+        logout(client)
+
```
drjova, line 135: I would add a test for curators as well, just to make sure that this case works as well

drjova (approved):

PR #2178, opened 2021-08-25
ui: numbers are missing from the list view

hanadikaleel:
#2164


hanadikaleel:
Screenshot 2021-08-25 at 15 54 52

drjova:
@hanadikaleel we should write a test to make sure it won't happen again

drjova (approved):

Issue #2183, opened 2021-08-30
author 


michamos:
**Describe the bug**

**To Reproduce**

**Expected behavior**

**Screenshots**

**Environment (please complete the following information):**
 - OS: [e.g. iOS8.1, Windows 7]

**Additional context**

Issue #2177, opened 2021-08-25
RecordsAuthors table is inconsistent


michamos:
See https://inspirehep.zulipchat.com/#narrow/stream/122199-ops/topic/inconsistent.20authors_records.20table

The data there doesn't seem to match what's in the records, for example `1836509` is affected. This causes disambiguation to fail sometimes due to https://sentry.inspirehep.net/inspire-prod/hep/issues/37362/events/784580/.

PR #2184, opened 2021-08-31
assign: fix assigning papers to new author

MJedr:
ref: #2183

### DIFF
--- a/backend/inspirehep/assign/views.py
+++ b/backend/inspirehep/assign/views.py
@@ -106,13 +106,10 @@ def author_assign_view():
     to_author_recid = body.get("to_author_recid")
     from_author_recid = body["from_author_recid"]
     literature_recids = body["literature_recids"]
-    with db.session.begin_nested():
-        if to_author_recid is None:
-            stub_author_id = assign_to_new_stub_author(
-                from_author_recid, literature_recids
-            )
-        else:
-            assign_to_author(from_author_recid, to_author_recid, literature_recids)
+    if to_author_recid is None:
+        stub_author_id = assign_to_new_stub_author(from_author_recid, literature_recids)
+    else:
+        assign_to_author(from_author_recid, to_author_recid, literature_recids)
     db.session.commit()
     if to_author_recid is None:
         return jsonify({"stub_author_id": stub_author_id}), 200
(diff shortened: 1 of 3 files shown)


backend/tests/integration-async/assign/test_views.py
```diff
@@ -0,0 +1,77 @@
...
+from helpers.providers.faker import faker
+from helpers.utils import create_user
+from invenio_accounts.testutils import login_user_via_session
+from invenio_db import db
+from sqlalchemy.exc import ResourceClosedError
+
+from inspirehep.records.api import AuthorsRecord, LiteratureRecord
+
+
+def test_assign_doesnt_craise_resources_closed_error(inspire_app, override_config):
```
michamos, line 11:
```suggestion
def test_assign_doesnt_raise_resource_closed_error(inspire_app, override_config):
```

backend/tests/integration-async/assign/test_views.py
```diff
@@ -0,0 +1,77 @@
...
+                                paper_2["control_number"],
+                                paper_3["control_number"],
+                            ],
+                            "from_author_recid": author_record["control_number"],
+                        }
+                    ),
+                    content_type="application/json",
+                )
+        except ResourceClosedError:
+            assert False
```
michamos, line 77: are you sure this test is failing without your change? I tried something extremely similar yesterday on a worker and I didn't get any exception.

backend/inspirehep/assign/views.py
```diff
@@ -49,7 +49,12 @@ def assign_papers(
 ):
     author_bai = get_values_for_schema(to_author_record["ids"], "INSPIRE BAI")[0]
     for record in author_papers:
-        lit_author = get_author_by_recid(record, from_author_recid)
+        try:
+            lit_author = get_author_by_recid(record, from_author_recid)
+        # it might happen that stub author BAI already existed
+        # and we already assigned papers to his profile.
+        except StopIteration:
+            continue
```
michamos, line 57: This is a workaround for #2177. I don't think we should ignore the issue but rather find and fix the root cause. How is this related to this issue?

MJedr: I am not sure if that's a workaround for another issue cause this one is about `get_author_by_recid` and the other one is about BAI. In the case that I try to fix the following problem, due to which my test for commiting was failing:   
we create a new stub author and [here](https://github.com/inspirehep/inspirehep/blob/39bbacf372b720b6cfed7ce577f74abf72256dc5/backend/inspirehep/records/api/authors.py#L91) we assign reference to this new author so if we try to get an old author by recid it's not valid anymore and we get an error.

michamos: You're right of course about the issue being different, I got confused by the `StopIteration` that made it look superficially similar. I don't understand the failure mode however. Let's discuss it.

michamos (approved):

PR #2188, opened 2021-09-02
literature: add task to fix invalid entries in RecordsAuthors table

MJedr:
ref: #2177

### DIFF
--- a/backend/inspirehep/records/api/base.py
+++ b/backend/inspirehep/records/api/base.py
@@ -649,3 +649,17 @@ class InspireRecord(Record):
         if self.get("deleted") and pid.status == PIDStatus.REDIRECTED:
             redirection = InspireRedirect.get_redirect(pid)
             return get_ref_from_pid(redirection.pid_type, redirection.pid_value)
+
+    @classmethod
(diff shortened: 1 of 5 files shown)


backend/inspirehep/records/api/base.py
```diff
@@ -649,3 +649,16 @@ def redirected_record_ref(self):
         if self.get("deleted") and pid.status == PIDStatus.REDIRECTED:
             redirection = InspireRedirect.get_redirect(pid)
             return get_ref_from_pid(redirection.pid_type, redirection.pid_value)
+
+    @classmethod
+    def get_all_ids_by_update_date(cls, date=None):
```
michamos, line 654:
```suggestion
    def get_recids_by_updated_datetime(cls, before=None, after=None):
```
to have a more consistent naming with existing methods, and slightly generalizing it? Both `before` and `after` should be optional (not added to the query if not present) instead of using today. Also `updated` is a `datetime`, so provides finer granularity that just a date.

michamos (approved):
minor comment, but looks good otherwise :+1:

PR #2189, opened 2021-09-02
stub authors cli: verify in db if author is stub

MJedr:


backend/inspirehep/disambiguation/cli.py
```diff
@@ -67,3 +62,11 @@ def query_authors_with_linked_papers_by_bai(authors_bais):
 
     for data in query.yield_per(100).with_entities(RecordsAuthors.author_id):
         yield data.author_id
+
+
+def get_stub_authors_by_pids(author_record_ids):
+    query = RecordMetadata.query.filter(
+        RecordMetadata.json.op("?")("stub"), RecordMetadata.id.in_(author_record_ids)
```
michamos, line 69: This is incorrect, as there could be `stub: False` in the record. I think it's easier if you just keep what was there before and add in the `stub_authors_bais` comprehension an `if authors.get("stub") is True` (almost all records should still be `stub`, so you don't gain much by filtering at the DB level).

If instead you want to do it this way, you could do (with appropriate imports)
```suggestion
        type_coerce(RecordMetadata.json, JSONB).contains({"stub": True}), RecordMetadata.id.in_(author_record_ids)
```

backend/inspirehep/disambiguation/cli.py
```diff
@@ -27,17 +27,12 @@ def clean_stub_authors():
     # We get all the stub authors (created by disambiguation) from ES and we verify
     # in db if the returned records are stub (ES data might be outdated)
     stub_authors_query = Q("term", stub=True)
-    stub_authors_search = (
-        AuthorsSearch().query(stub_authors_query).source(["control_number"])
-    )
-    stub_authors_control_numbers = [
-        ("aut", str(author["control_number"])) for author in stub_authors_search.scan()
-    ]
+    stub_authors_search = AuthorsSearch().query(stub_authors_query)
```
michamos, line 30: I suppose you want to set an empty source too.

drjova (dismissed):

michamos (approved):

Issue #2169, opened 2021-08-24
Use relaxed matching on `journal_title` only with GROBID


michamos:
@hoc3426 reports in https://inspirehep.zulipchat.com/#narrow/stream/122199-ops/topic/bugninja/near/250177495 incorrectly matched references. This seems to be a regression introduced by https://github.com/inspirehep/inspirehep/pull/1119: a more relaxed matching on journal title was introduced as a workaround for GROBID issues, which affects all reference matching, not only when doing searches and using GROBID. The changes in that PR (new config to match `journal_title` prefixes and taking the first of up to 5 ambiguous results) should not affect the standard reference matcher, but should only be used when matching references produced by GROBID interactively in the user search.

PR #2194, opened 2021-09-03
matcher: use relaxed title matching only with grobid

MJedr:
ref: #2169

### DIFF
--- a/backend/inspirehep/matcher/api.py
+++ b/backend/inspirehep/matcher/api.py
@@ -74,11 +74,34 @@ def match_reference_with_config(reference, config, previous_matched_recid=None):
     return reference
 
 
-def match_reference_config(reference):
+def match_journal_title_config(reference, use_relaxed_titles_matching):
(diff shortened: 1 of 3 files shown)


backend/inspirehep/matcher/api.py
```diff
@@ -159,7 +167,7 @@ def match_reference_control_numbers(reference):
         except KeyError:
             return None
 
-    configs = match_reference_config(reference)
+    configs = match_reference_config(reference, use_relaxed_titles_matching=True)
```
michamos, line 170: I think it would be cleaner to expose this fact to the top-level API functions. Here you're using the fact that `match_reference_control_numbers` is used only in the search API, while the regular `match_reference` is used elsewhere. But that's incidental and not clear from the API. Maybe rename this function to`match_reference_control_numbers_with_relaxed_journal_titles` to be explicit?

michamos (approved):
Small comment regarding API design, looks good otherwise :+1:

PR #2205, opened 2021-09-16
setup: bump invenio-records-rest

oguzdemirbasci:


MJedr (commented):
I am wondering why there are so many changes (normally it happens when you add an optional dependency as a non-optional one, but it's not a case there). Did you have a look if all the references match? And are you using a correct poetry version?

oguzdemirbasci:
@MJedr I am using poetry in the docker container. It should be the correct version.

MJedr (approved):

Issue #2171, opened 2021-08-24
Display `imprints` for books


michamos:
From https://inspirehep.zulipchat.com/#narrow/stream/195298-experts/topic/pubinfo.20for.20books/near/250475668:

Books often contain useful metadata in `imprints`, which is currently not displayed in the search results or detailed record view. We should improve that.

StellaCh:
In the detailed page of the paper, under ISBN add a new line with the following information: 
Published **imprints.date** in **imprints.place** by **imprints.publisher** 

eg: 

> Published May 1, 2021 in Boca Raton by CRC Press

 for https://inspirehep.net/literature/1886833

MJedr:
**TECH NOTES**
1. Create a new component `ImprintInfo` which will take imprint list as the argument and will generate 
Published **imprints.date** in **imprints.place** by **imprints.publisher**
2. Add the new component [here](https://github.com/inspirehep/inspirehep/blob/39bbacf372b720b6cfed7ce577f74abf72256dc5/ui/src/literature/containers/DetailPageContainer/DetailPageContainer.jsx#L193) and pass imprints from metadata

michamos:
Example of record with two imprints: https://inspirehep.net/literature/659635.

StellaCh:
Thank you!
@hanadikaleel The list of imprints should be connected with 'and'

PR #2213, opened 2021-09-22
ui: display imprints from books

hanadikaleel:
#2171

### DIFF
--- a/ui/src/literature/components/ImprintInfo.jsx
+++ b/ui/src/literature/components/ImprintInfo.jsx
@@ -1,0 +1,67 @@
+import React, { Component } from 'react';
+import PropTypes from 'prop-types';
+import { List } from 'immutable';
+import moment from 'moment';
+import { hasDayMonthAndYear, hasMonthAndYear } from '../../common/utils';
+
+import InlineList, { SEPARATOR_AND } from '../../common/components/InlineList';
+
+function formatDate(date) {
+  if (hasDayMonthAndYear(date)) {
+    return ' MMM D, YYYY';
+  }
+
+  if (hasMonthAndYear(date)) {
+    return ' MMM, YYYY';
+  }
+
+  return ' YYYY';
+}
+
+class ImprintInfo extends Component {
+  static renderImprint(imprint) {
+    const date = imprint.get('date');
+    const place = imprint.get('place');
+    const publisher = imprint.get('publisher');
+
+    const datePart = date ? moment(`  ${date} `).format(formatDate(date)) : '';
+    const placePart = place ? ` in ${place} ` : '';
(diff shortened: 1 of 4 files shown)


hanadikaleel:
Screenshot 2021-09-22 at 14 57 55

ui/src/literature/components/ImprintInfo.jsx
```diff
@@ -0,0 +1,52 @@
+import React, { Component } from 'react';
+import PropTypes from 'prop-types';
+import { List } from 'immutable';
+
+import InlineList, { SEPARATOR_AND } from '../../common/components/InlineList';
```
MJedr, line 5: Can you add some tests for this component?

hanadikaleel: updated

ui/src/literature/components/__tests__/ImprintInfo.test.jsx
```diff
@@ -0,0 +1,76 @@
+import React from 'react';
```
MJedr, line 1: Can you add a test when you'll have couple of imprints info in the array?

hanadikaleel: updated

ui/src/literature/components/ImprintInfo.jsx
```diff
@@ -0,0 +1,52 @@
...
+
+class ImprintInfo extends Component {
+  static renderImprints(imprints) {
+    const placePart = imprints.get('place')
```
MJedr, line 9: Maybe you can destructure this object before creating `placePart`, `datePart` etc.? It would be easier to read I think.

ui/src/literature/components/ImprintInfo.jsx
```diff
@@ -0,0 +1,52 @@
+import React, { Component } from 'react';
+import PropTypes from 'prop-types';
+import { List } from 'immutable';
+
+import InlineList, { SEPARATOR_AND } from '../../common/components/InlineList';
+
+class ImprintInfo extends Component {
+  static renderImprints(imprints) {
```
MJedr, line 8: I would change the name of this function to `renderImprint` and an argument to a single `imprint`

michamos:
I believe the date should be formatted in the same way as the date at the top (see screenshot). @StellaCh ?

StellaCh:
yes, consistency is key. It should be displayed as the example I have in the first comment of the issue: https://github.com/inspirehep/inspirehep/issues/2171 , similar to the other date

hanadikaleel:
updated date format
Screenshot 2021-09-23 at 14 33 16

MJedr (approved):

PR #2229, opened 2021-10-01
ui: update help wordpress

drjova:


hanadikaleel (approved):

ui/src/common/constants.js
```diff
@@ -41,7 +41,7 @@ export const INSPIRE_TWITTER_ACCOUNT = 'https://twitter.com/inspirehep';
 export const SURVEY_LINK = 'https://forms.gle/ZQi31GvXXHcsgXgM6';
 export const FEEDBACK_EMAIL = 'feedback@inspirehep.net';
 export const BLOG_URL = 'https://blog.inspirehep.net';
-export const HELP_BLOG_URL = 'https://inspirehep.net/help';
+export const HELP_BLOG_URL = 'https://help.inspirehep.net/';
```
michamos, line 44: don't we want to redirect from `/help/*` instead in order not to break other links?

drjova: We will do that as well

Issue #2185, opened 2021-09-01
URI Validation: URI inputs should be encoded to replace special characters with valid values.


oguzdemirbasci:
Our validator doesn't validate the following URI because of the character `φ`. 
http://caod.oriprobe.com/articles/61619219/Some_characterizations_for_the_exponential_φ_expan.htm

Browsers can automatically encode it, load the page and still show the special characters in the URL bar. So URIs with special characters can be passed to the validator and we need to replace them with the valid ASCII codes. For this example, the encoded URI should be like this:
http://caod.oriprobe.com/articles/61619219/Some_characterizations_for_the_exponential_%CF%86_expan.htm

Following code should work:
```Python
url = url.encode('utf-8')
url = urllib.parse.quote(url)
```

ASCII encoding reference:
https://www.w3schools.com/tags/ref_urlencode.asp

michamos:
@oguzdemirbasci copy/pasting from the browser bar should already encode the URL behind the scenes as you mention. I'm not sure I understand the scenario that would cause the URL to be unencoded and hence invalid.

oguzdemirbasci:
Yes, but It was the input for a workflow. I do not know how it is written there. Maybe not all browsers support it.

michamos:
So looks like we should improve the form or the submission logic, rather than the record logic.

oguzdemirbasci:
**TECH NOTES**
* Add a `transform` function into the `yup` validation schema for [these URLs](https://github.com/inspirehep/inspirehep/blob/master/ui/src/submissions/literature/schemas/sections/links.js) that encodes the valid URL inputs. ([Check out the documentation](https://github.com/jquense/yup#mixedtransformcurrentvalue-any-originalvalue-any--any-schema)).
* Use default URI encoding function of javascript `encodeURI()` in the transform function.

michamos:
I don't think that's sufficient. You somehow need to detect whether the URL is already encoded, to avoid encoding it twice in case it is.

MJedr:
> **TECH NOTES**
> 
> * Add a `transform` function into the `yup` validation schema for [these URLs](https://github.com/inspirehep/inspirehep/blob/master/ui/src/submissions/literature/schemas/sections/links.js) that encodes the valid URL inputs. ([Check out the documentation](https://github.com/jquense/yup#mixedtransformcurrentvalue-any-originalvalue-any--any-schema)).
[...]

PR #2237, opened 2021-10-05
editor: fix uri validation

MJedr:
ref:#2185

### DIFF
--- a/record-editor/src/app/core/services/record-cleanup.service.ts
+++ b/record-editor/src/app/core/services/record-cleanup.service.ts
@@ -34,11 +34,17 @@ export class RecordCleanupService {
         if (this.isEmpty(value[i])) {
           value.splice(i, 1);
         }
+        if (this.isURL(value[i]) && this.isNotEncodedURL(value[i])) {
+          value[i] = encodeURI(value[i]);
+        }
       }
     } else {
       Object.keys(value).forEach((key) => {
         if (typeof value[key] === 'object' && value[key] !== null) {
           this.cleanup(value[key]);
+        }
+        if (this.isURL(value[key]) && this.isNotEncodedURL(value[key])) {
+          value[key] = encodeURI(value[key]);
         }
         if (this.isEmpty(value[key])) {
           delete value[key];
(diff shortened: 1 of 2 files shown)


oguzdemirbasci (approved):
LGTM 👍

record-editor/src/app/core/services/record-cleanup.service.ts
```diff
@@ -61,4 +67,16 @@ export class RecordCleanupService {
     // boolean, number
     return false;
   }
+
+  public isURL(value: any): boolean {
+    if (typeof value !== 'string') {
+      return false;
+    }
+    const pattern = /(http|https):\/\/(\w+:{0,1}\w*)?(\S+)(:[0-9]+)?(\/|\/([\w#!:.?+=&%!\-\/]))?/;
```
drjova, line 75: you could use https://developer.mozilla.org/en-US/docs/Web/API/URL

michamos: Maybe I'm missing sometihng, but isn't the whole point that you want to encode things that are not valid URLs? What if there's a ` ` or a :smiley: in the value, but is otherwise a URL? will your regex detect that? conversely, what if an abstract contains a URL? are you now going to encode the full abstract?

I don't know if it's easy to do, but it seems more robust to do this based on the `format` expected by the schema, instead of the value and try to guess whether it should be a URL.

record-editor/src/app/core/services/record-cleanup.service.ts
```diff
@@ -61,4 +67,20 @@ export class RecordCleanupService {
     // boolean, number
     return false;
   }
+
+  public isURL(value: any): boolean {
+    if (typeof value !== 'string') {
+      return false;
+    }
+    try {
+      const url = new URL(value);
```
michamos, line 76: As I mentioned previously, I don't think a naive approach based only on the value and not taking into account the schema `format` (or other information) can possibly work. 

How can you distinguish a `public_notes.value` containing `https://example.com/foo contains additional information`, from a `documents.url` containing `https://example.com/file name with a space.pdf`?

PR #2262, opened 2021-10-18
assign : assign papers in async task

MJedr:


drjova:
Isn't the bottleneck the HAL push?

Issue #2207, opened 2021-09-20
Concurrency in the editor


StellaCh:
It often happens that 2 curators are in the record editor of the same record. 
What should happen when one saves the record? How to handle changes in the record?

michamos:
after discussing with the curators, it seems the biggest issue is not concurrent edits by another curator, but automated processes running after saving the record (e.g. author disambiguation).

drjova:
TECH NOTES:

EDITOR:
* Use the `editor-lock:control_number@version_id` as a key that expires after configurable time and add a message to `editor`: `{user_email} has opened the record at {time} and hasn't saved yet. Your or their modifications might get lost.` on https://github.com/inspirehep/inspirehep/blob/master/backend/inspirehep/editor/views.py#L68 
* Get all the locks for the specific record and pass them to the payload `locks`
* Create an endpoint to clear the lock of the editor on Save, on browser [close window event and on the warning when you trying to leave the browser with unmodified changes](https://github.com/inspirehep/inspirehep/blob/3c55ba49a38c0af0e073c702957ade5b83808962/record-editor/src/app/core/services/dom-utils.service.ts).
* On the editor read the `locks` payload and display notifications accordingly 

HAL_PUSH:
* Add the `editor-lock:control_number@version_id` as a key that expires after configurable time  and add to `hal_push`: `{task_name} has been scheduled at {time} and hasn't finished yet. Your modifications might get lost.`
* When the task finishes clear the lock

AUTHOR DISAMBIGUATION:
* In the receivers add the `editor-lock:control_number@version_id` as a key and add to `disambiguation`: `{task_name} has been scheduled at {time} and hasn't finished yet. Your modifications might get lost.`
* Pass the `version_id` to the task
* When the task finishes clear the lock

PR #2278, opened 2021-10-27
editor: display warning when record already in edit

MJedr:
ref: #2207


backend/inspirehep/editor/cache.py
```diff
@@ -0,0 +1,89 @@
...
+            self.task_hash_name, key=self.task_name, value=self.task_lock_message
+        )
+        self.redis.expire(self.task_hash_name, time=REDIS_KEY_VALUE_PAIR_EXPIRE_TIME)
+
+    def remove_task_lock(self):
+        self.redis.hdel(self.task_hash_name, self.task_name)
+
+    def remove_user_lock(self):
+        self.redis.hdel(self.user_hash_name, self.user_email)
+
```
michamos, line 64: do you really need to have two copies of every method here? I would think you can detect which case you're in based on the data passed to the `__init__` (`task_name` or `user_email`). Alternatively, you could subclass a base class and have a task version and a user version. Currently the API doesn't make much sense. What happens if you call `add_task_lock` for a task initialized with a `user_email`  but no `task_name`.

backend/inspirehep/editor/cache.py
```diff
@@ -0,0 +1,89 @@
...
+            ]
+        )
+        existing_editor_task_locks = ", ".join(
+            [
+                value.decode()
+                for value in self.redis.hgetall(self.task_hash_name).values()
+            ]
+        )
+        if existing_editor_user_locks:
+            editor_user_lock_msg = f"The record might still be edited by {existing_editor_user_locks}".strip(
```
michamos, line 80: could you use the full message as in the issue?

backend/inspirehep/editor/cache.py
```diff
@@ -0,0 +1,89 @@
...
+                ", "
+            )
+            payload["user_locks"] = editor_user_lock_msg
+        if existing_editor_task_locks:
+            editor_task_lock_msg = f"Those tasks have been already scheduled and haven't finished yet: {existing_editor_task_locks}".strip(
```
michamos, line 85: same

backend/inspirehep/editor/views.py
```diff
@@ -308,3 +314,13 @@ def authorlist_text():
         return jsonify(parsed_authors)
     except Exception as err:
         return jsonify(status=400, message=" / ".join(err.args)), 400
+
+
+@blueprint.route("/<endpoint>/<pid_value>/<version_id>/lock/resolve", methods=["GET"])
```
michamos, line 319:
```suggestion
@blueprint.route("/<endpoint>/<pid_value>/<version_id>/lock/release", methods=["POST"])
```
Not idempotent, so can't be a `GET`, and the standard terminology is `acquire`/`release`.

Or maybe even remove the version from the endpoint, and pass the whole `Etag` as a body, that can be treated as an opaque version identifier by the editor frontend? that would also allow you to write a backend test asserting that if you pass the record `Etag` to the endpoint, the lock gets cleared successfully, and catch the case where the `Etag` changes.

backend/inspirehep/editor/views.py
```diff
@@ -308,3 +314,13 @@ def authorlist_text():
...
+@login_required_with_roles([Roles.cataloger.value])
```
michamos, line 320: this is incorrect, you need to support also the private collection edit roles.

backend/inspirehep/records/receivers.py
```diff
@@ -41,6 +42,14 @@ def index_after_commit(sender, changes):
...
-                    disambiguate_authors.delay(str(model_instance.id))
+                    disambiguate_authors.delay(
+                        str(model_instance.id), model_instance.version_id
+                    )
+                    editor_cache = EditorCache(
+                        recid=model_instance.json["control_number"],
+                        record_version=model_instance.version_id,
+                        task_name=disambiguate_authors.name,
+                    )
+                    editor_cache.add_task_lock()
```
michamos, line 53: move all of this to a new function in `inspirehep.disambiguation.api`?

backend/tests/integration-async/disambiguation/test_disambiguation_tasks.py
```diff
@@ -1253,3 +1255,50 @@ def assert_disambiguation_task():
...
+    db.session.commit()
+    version_id = literature_record.model.version_id
+
+    def assert_lock_in_redis():
+        expected_hash_name = (
+            f"editor-task-lock:{literature_record['control_number']}@{version_id}"
+        )
+        assert redis.hgetall(expected_hash_name)
+
+    retry_until_pass(assert_lock_in_redis, retry_interval=1)
```
michamos, line 1304: won't this be flaky? can you be sure that the disambiguation task doesn't finish before you've checked the condition here (in which case the lock has already been cleared)?

backend/tests/integration/editor/conftest.py
```diff
@@ -0,0 +1,34 @@
...
+
+
+@pytest.fixture(scope="function", autouse=True)
+def clear_redis_editor_locks(app):
+    with app.app_context():
+        redis_url = current_app.config.get("CACHE_REDIS_URL")
+        redis = StrictRedis.from_url(redis_url)
+        matched_keys_values = redis.hscan_iter(name="editor-lock", match="editor-*")
+        for key, _ in matched_keys_values:
+            redis.hdel("editor-lock", key)
```
michamos, line 34: you can just delete the keys with `delete()`, no need to iterate on the values.

backend/inspirehep/editor/cache.py
MJedr: This message doesn't fit many locks for same record.

backend/inspirehep/editor/cache.py
```diff
@@ -0,0 +1,100 @@
...
+        self.task_name = task_name
+
+    @property
+    def redis(self):
+        redis_url = current_app.config.get("CACHE_REDIS_URL")
+        redis = StrictRedis.from_url(redis_url)
+        return redis
+
+    @property
+    def user_lock_key(self):
```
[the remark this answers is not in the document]
MJedr: And what would you do to gracefully get hash from redis for both tasks and editor locks (and separate them) if the hash name would be unified?

backend/inspirehep/editor/views.py
MJedr: Sorry, I don't understand why it is incorrect now, I am taking record version directly from etag in frontend, so what's the difference if I'll pass something like `'W 25'` or 25?

michamos: That's not incorrect (`GET` method is). Currently the editor frontend needs to be aware of how the Etag is constructed, but I believe that's better left as a backend implementation detail.

backend/inspirehep/editor/cache.py
michamos: Good point. Maybe something like
```suggestion
            editor_user_lock_msg = f"Record opened by {existing_editor_user_locks}. Their modifications aren't saved yet, they (or yours) might get lost.".strip(
```
so they get the consequences.

backend/inspirehep/editor/cache.py
michamos: {task_name} has been scheduled at {time} and hasn't finished yet. Your modifications might get lost.
```suggestion
            editor_task_lock_msg = f"Scheduled tasks {existing_editor_task_locks} haven't finished yet. Your modifications might get lost.".strip(
```

backend/inspirehep/editor/views.py
drjova: `etag` was not part of the refinement, and we discussed it all together. Implementing `etag` would need to change the logic and it's like starting from scratch again. I will strongly suggest to implement this feature as we designed and later to improve it if needed.

michamos: @drjova The version was part of the refinement. The only way for the editor to get the version currently is through the Etag. So we can't leave it alone (unless we add some different mechanism to transmit the version to the editor). The question is: what should extract the version from the ETag: the record editor frontend or the backend?

MJedr: Just to add that if we want to post by using etag we'll have to encode and then decode etag in backend, what imo is less clear than doing a request with a version.

michamos: I have different taste here :) You need to decode it anyway, if you do it in the frontend, you couple the implementation of the Etag encoding of the version with the record editor logic.

backend/tests/integration-async/disambiguation/test_disambiguation_tasks.py
MJedr: I don't have any other idea how to test it. For me this test never failed, but I can remove it if you want.

backend/inspirehep/disambiguation/api.py
```diff
@@ -0,0 +1,19 @@
...
+# inspirehep is free software; you can redistribute it and/or modify it under
+# the terms of the MIT License; see LICENSE file for more details.
+
+from inspirehep.disambiguation.tasks import disambiguate_authors
+from inspirehep.editor.cache import EditorSoftLock
+
+
+def author_disambiguation(model_instance):
+    disambiguate_authors.delay(str(model_instance.id), model_instance.version_id)
+    editor_cache = EditorSoftLock(
```
michamos, line 14: maybe also change the var name now :)

backend/inspirehep/disambiguation/tasks.py
```diff
@@ -262,3 +263,9 @@ def disambiguate_authors(self, record_uuid):
         )
         record.update(dict(record))
         db.session.commit()
+        editor_cache = EditorSoftLock(
```
michamos, line 266: same

backend/inspirehep/editor/cache.py
```diff
@@ -0,0 +1,80 @@
...
+
+    def _get_hash_name_for_edited_record(self):
+        return {
+            "user": f"editor-lock:{self.recid}@{self.record_version}",
+            "task": f"editor-task-lock:{self.recid}@{self.record_version}",
+        }
+
+    def add_lock(self):
+        self.redis.hset(self.hash_name, key=self.key, value=self.message)
+        self.redis.expire(self.hash_name, time=EDITOR_LOCK_EXPIRATION_TIME)
```
michamos, line 51: FYI, redis is non transactional by default, so a different task/user might come in between these two actions and do some other things. I don't think it's a problem in this specific case, but something to always keep in mind when you're making several logically related changes to redis.

backend/inspirehep/editor/cache.py
```diff
@@ -0,0 +1,80 @@
+from datetime import datetime
```
michamos, line 1: also the file name should probably be changed

backend/inspirehep/editor/views.py
```diff
@@ -70,17 +71,22 @@ def get_record_and_schema(endpoint, pid_value):
     record = InspireRecord.get_record_by_pid_value(
         pid_value, pid_type, original_record=True
     )
-
     if not check_permissions_for_private_collection_read(
         record.get("_collections", [])
     ):
         return jsonify(message="Unauthorized", code=403), 403
-
+    editor_cache_service = EditorSoftLock(
```
michamos, line 78: same

backend/inspirehep/editor/views.py
```diff
@@ -308,3 +314,24 @@ def authorlist_text():
         return jsonify(parsed_authors)
     except Exception as err:
         return jsonify(status=400, message=" / ".join(err.args)), 400
+
+
+@blueprint.route(
+    "/<endpoint>/<pid_value>/W/<version_id>/lock/release", methods=["POST"]
```
michamos, line 320: minor: maybe put the Etag in the body?

backend/inspirehep/hal/api.py
```diff
@@ -31,6 +33,12 @@ def push_to_hal(record):
                 "record_version_id": record.model.version_id,
             },
         )
+        editor_cache = EditorSoftLock(
```
michamos, line 36: naming

backend/inspirehep/hal/tasks.py
```diff
@@ -60,6 +61,12 @@ def hal_push(self, recid, record_version_id):
             raise StaleDataError
         _hal_push(record)
         LOGGER.info("hal_push task successfully completed.", recid=recid)
+        editor_cache = EditorSoftLock(
```
michamos, line 64: same

backend/tests/integration/editor/test_cache.py
```diff
@@ -0,0 +1,51 @@
...
+# Copyright (C) 2021 CERN.
+#
+# inspirehep is free software; you can redistribute it and/or modify it under
+# the terms of the MIT License; see LICENSE file for more details.
+from helpers.utils import create_record
+
+from inspirehep.editor.cache import EditorSoftLock
+
+
+def test_editor_cache_user_locks(inspire_app):
```
michamos, line 12: naming (also filename)

backend/inspirehep/disambiguation/api.py
```diff
@@ -0,0 +1,19 @@
+# -*- coding: utf-8 -*-
+#
+# Copyright (C) 2021 CERN.
+#
+# inspirehep is free software; you can redistribute it and/or modify it under
+# the terms of the MIT License; see LICENSE file for more details.
+
+from inspirehep.disambiguation.tasks import disambiguate_authors
+from inspirehep.editor.editor_soft_lock import EditorSoftLock
```
michamos, line 9: Did you forget to commit this module?

backend/inspirehep/editor/views.py
```diff
@@ -308,3 +314,23 @@ def authorlist_text():
...
+def remove_editor_lock(endpoint, pid_value):
+    version_id = request.headers["ETag"]
+    record = request.json
+    if not check_permissions_for_private_collection_read_write(
+        record.get("_collections", [])
+    ):
+        return jsonify(message="Unauthorized", code=403), 403
+    version_from_etag = re.findall(r"\d+", version_id)
+    if not version_from_etag:
+        return jsonify(message="Incorrect Etag passed", code=500), 500
```
michamos, line 330: 400 instead of 500?

michamos (approved):
minor comment, but looks very good overall :+1: I didn't check the frontend code though

drjova (approved):

Issue #1230, opened 2020-08-20
Improve autocompletion


StellaCh:
>Affiliation picker only shows 5 choices, can't scroll beyond 5 – It would be nice to have more so as not to have to leave and search Institutions 

>Some Affiliations don’t show in the picker – Example: “Caltech” - Today, I could not get it to give just “Caltech” as a choice. When I typed in “Caltech”, I got all the affiliations that have the word in them, but had to type in “Division of Physics” (without the word “Caltech” at all) to get it to offer “Caltech” as a choice.

It is impossible to add some experiments or affiliations because they are not in the proposed (short) list, even if the exact value is given.
Need better sort and/or no truncation

drjova:
We'll try the search as you type solution, https://www.elastic.co/guide/en/elasticsearch/reference/7.x/search-as-you-type.html

harunurhan:
This should be probably done for all completion fields.

drjova:
**TECH NOTES:**

We're using already ES `7.1.1` on `prod`

* Change the mapping type to `search_as_you_type` for `affiliation_suggest ` https://github.com/inspirehep/inspirehep/blob/b6aa07abab5504eb5744ef74d161a1f5c1e2f208/backend/inspirehep/search/mappings/v7/records/institutions.json#L67
* Remap and reindex institutions

drjova:
You're right it needs `7.2` thanks!

drjova:
ES team will release 7.8 in few weeks. We should revisit this issue when it's done.

MJedr:
**TECH NOTES**
Add `size` parameter [here](https://github.com/inspirehep/inspirehep/blob/636879322e68b4a9b83c40f4d76873351f9f573a/backend/inspirehep/records/config.py#L430) and set it to 20

PR #2288, opened 2021-11-10
records: addition of `search_as_you_type` for exp and ins

drjova:
* Needs remaping and reindexing

* Ref: #1230


MJedr (approved):
Looks good, just a question - are you using a correct poetry version? There are suspiciously many changes in poetry.lock

PR #2294, opened 2021-11-17
records: addition of `search_as_you_type` boost

drjova:


backend/inspirehep/records/config.py
```diff
@@ -313,7 +313,40 @@
...
+                        "field": "accelerator_search_as_you_type",
+                        "boost": 1,
+                    },
+                    {
+                        "field": "collaboration_search_as_you_type",
+                        "boost": 1,
+                    },
+                    {
+                        "field": "experiment_short_name_search_as_you_type",
+                        "boost": 1,
```
michamos, line 327: maybe it would make sense to boost this too? how have you arrived at the current weights?

drjova: I took them from the `completion` weights https://github.com/inspirehep/inspirehep/commit/bd2d14ac715df1ae6ee241a76659fff63983636d#diff-4bd65a4802a80112f9f39963c9487595be8604806969c45388db10686daed263R32-R43. I don't really know which ones to boost since I don't know how it's been used from the curators, any suggestions are welcome

michamos: ok, let's start with this and tune it based on feedback.

backend/inspirehep/search/mappings/v7/records/experiments.json
```diff
@@ -19,6 +19,30 @@
...
+      },
+      "long_name_search_as_you_type": {
+        "type": "search_as_you_type"
+      },
+      "name_variants_search_as_you_type": {
+        "type": "search_as_you_type"
+      },
+      "legacy_name_search_as_you_type": {
+        "type": "search_as_you_type"
+      },
```
michamos, line 45: alternative: define a `search_as_you_type` object, with all the search fields as subfields, so this becomes easier to exclude in the serializers.

michamos (approved):
:+1: minor cosmetic comment, it's up to you

Issue #1928, opened 2021-06-08
Improve hepcrawl logging


michamos:
There are various issues with hepcrawl logging:

* The detailed crawler logs are written to files, which never make it to the central ES monitoring, and are only preserved for 6 harvests.
* The detailed crawler logs are extremely detailed, containing the records themselves multiple times (due to `item` [here](https://github.com/inspirehep/hepcrawl/blob/8a5cd86c821c5480290ec8c78622d453366d090a/hepcrawl/pipelines.py#L80) and [that](https://github.com/inspirehep/hepcrawl/blob/8a5cd86c821c5480290ec8c78622d453366d090a/hepcrawl/pipelines.py#L228-L231).
* As far as I can see, when file download fails in the pipeline, the `results` argument of `item_completed` will contain exception info (elements are tuples, first element of a tuple will be `False`, second one exception class on exception).
Example from logs:
```
results: [(False, <twisted.python.failure.Failure scrapy.pipelines.files.FileException: download-error>)]
```
We should make that exception more visible, even if there is no data loss (the URL is passed along unmodified AFAICT).
* Errors in spiders make it to sentry, but the format is pretty much unusable, see https://sentry.inspirehep.net/inspire-prod/next/issues/29336/?query=is%3Aunresolved and only spider exceptions are concerned, not other parts of scrapy. Instead of using the `scrapy-python` middleware plugin, it seems easy enough to plug into the logging system directly to report all errors and exceptions to sentry, see https://github.com/llonchj/scrapy-sentry/issues/6 and especially https://gist.github.com/Framartin/4e57b57139ed31f36684cfc514037bf6.

MJedr:
**TECH**
In [settings.py](https://github.com/inspirehep/hepcrawl/blob/master/hepcrawl/settings.py) add:
`LOG_FILE = None` (that should cause not logging to file)
`LOGS_ENABLED = True`
`LOG_LEVEL = ‘INFO’` (that should disable debug logs on ES)

Remove [this lines](https://github.com/inspirehep/hepcrawl/blob/8a5cd86c821c5480290ec8c78622d453366d090a/hepcrawl/pipelines.py#L171-L173)

Apply [this](https://gist.github.com/Framartin/4e57b57139ed31f36684cfc514037bf6) extension for sentry logging.

michamos:
We should also log when trying to get attached files which files fail, currently we just get a generic 404 exception from `boto`, which doesn't tell which file is actually missing.

PR #2302, opened 2021-11-23
osti sync

MJedr:
ref: #1928


drjova:
by closing it we'll not lose the code.

PR #2305, opened 2021-11-25
records api: return validation errors from validation

MJedr:


backend/inspirehep/records/api/base.py
```diff
@@ -65,7 +65,12 @@ def strip_empty_values(data):
         return strip_empty_values(data)
 
     def validate(self):
-        schema_validate(self)
+        self._validation_errors = list(get_validation_errors(self))
+        if self._validation_errors:
+            raise self._validation_errors[0]
+
+    def get_validation_errors(self):
+        return self._validation_errors
```
michamos, line 73: this will raise `AttributeError` if this gets called without calling `validate` first. You'de need to check if the attribute exists and recompute it if not.

michamos (approved):

Issue #2307, opened 2021-11-26
Don't match authors with different first names


michamos:
See https://inspirehep.zulipchat.com/#narrow/stream/122202-devel/topic/disambiguation/near/262720253.

Currently, we match on `first_name` in https://github.com/inspirehep/inspirehep/blob/59f01c7c39aa1802cdd044bd8270a0391ffbc932/backend/inspirehep/matcher/config.py#L316-L317. As this does a match query, it will match on any common token, meaning `Evgeny V.` matches `Igor V.` as the `V` tokens match. We should improve the logic to prevent matching incompatible names: a subset or a superset of tokens should match, but as soon as one of the tokens is different it shouldn't match.

MJedr:
**TECH NOTES**
* add a [new validator](https://github.com/inspirehep/inspirehep/blob/master/backend/inspirehep/matcher/validators.py) that will check if all the names from result from ES are in author's name (so if hit `first_name` is `Evgeny G.` and author's that we are trying to disambiguate `first_name` is `Evgeny I.` it doesn't match).
* add a new validator [here](https://github.com/inspirehep/inspirehep/blob/97d3f612f90c7b650ffb6fb6cb7ed60efca983fa/backend/inspirehep/matcher/config.py#L326)

MJedr:
It seems that changing matcher query to use `AND` operator would be enough to prevent incorrect disambiguation.

PR #2320, opened 2021-12-09
disambiguation: don't match authors with different names

MJedr:
ref:#2307


michamos:
@MJedr can we close this? IIRC, you implemented a different approach in #2324 which is merged.

Issue #2314, opened 2021-11-30
OOM when assiging records to authors


michamos:
From https://rt.inspirehep.net/Ticket/Display.html?id=1190446:

Many LIGO/VIRGO records are not disambiguated. After investigation, this seems to be due to memory issues when trying to create new author records. For example, on https://inspirehep.net/literature/1929315, I managed to disambiguate almost all authors, but the pod gets killed when trying to create a record for the BAI `K.Lee.19`. More specifically https://github.com/inspirehep/inspirehep/blob/e06311e90a25d41293efa5eb25128a5848497752/backend/inspirehep/records/api/authors.py#L88 tries to load all records at once which consumes too much memory in this case:
```
In [9]: from inspirehep.records.api import AuthorsRecord                                                                                                                                                           

In [10]: ids = AuthorsRecord.query_author_papers("K.Lee.19")                                                                                                                                                       

In [11]: ids                                                                                                                                                                                                       
Out[11]: <generator object AuthorsRecord.query_author_papers at 0x7fde315adb30>

In [12]:         author_papers_ids = [ 
    ...:             str(record_control_number) 
    ...:             for record_control_number in ids 
    ...:         ]                                                                                                                                                                                                 

In [13]:         author_papers = LiteratureRecord.get_records(author_papers_ids)                                                                                                                                   
command terminated with exit code 137
```
After fixing the issue, disambiguation should be rerun for all non-disambiguated records in https://inspirehep.net/literature?sort=mostrecent&size=25&page=1&q=authors.full_name%3A%2A%20-authors.ids.schema%3A%22INSPIRE%20BAI%22%20_collections%3ALiterature as well as the previously mentioned record.

drjova:
TECH NOTES:
* Create a method in Inspire Records base `get_records_batched` which batches the results as in https://github.com/inspirehep/inspirehep/blob/master/backend/inspirehep/records/api/base.py#L181-L184

PR #2321, opened 2021-12-13
records api: batch `get_records`

MJedr:
ref:#2314

### DIFF
--- a/backend/inspirehep/records/api/base.py
+++ b/backend/inspirehep/records/api/base.py
@@ -26,8 +26,10 @@ from invenio_records.models import RecordMetadata
 from invenio_records.signals import after_record_revert, before_record_revert
 from jsonschema import ValidationError
 from sqlalchemy import tuple_
+from sqlalchemy.dialects.postgresql import JSONB
 from sqlalchemy.orm.attributes import flag_modified
 from sqlalchemy.orm.exc import NoResultFound, StaleDataError
+from sqlalchemy.sql.expression import cast
 from sqlalchemy_continuum import version_class
 
 from inspirehep.indexer.base import InspireRecordIndexer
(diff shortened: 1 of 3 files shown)


backend/inspirehep/records/api/base.py
```diff
@@ -176,6 +176,17 @@ def get_record(cls, record_uuid, with_deleted=False, record_version=None):
...
         return record
 
+    @classmethod
+    def get_records_batched(cls, ids, with_deleted=False, max_batch=100):
+        for batch in chunker(ids, max_chunk_size=max_batch):
+            query = cls.model_cls.query.filter(cls.model_cls.id.in_(batch))
+            if not with_deleted:
+                query = query.filter(
+                    cls.model_cls.json.op("->")("deleted") == None  # noqa
+                )
```
michamos, line 186: this is incorrect, `deleted = False` is allowed in a record. You need to check that `delete: True` doesn't occur.

MJedr: ok, thank you

backend/inspirehep/records/api/base.py
```diff
@@ -176,6 +178,18 @@ def get_record(cls, record_uuid, with_deleted=False, record_version=None):
...
         return record
 
+    @classmethod
+    def get_records_batched(cls, ids, with_deleted=False, max_batch=100):
+        for batch in chunker(ids, max_chunk_size=max_batch):
+            query = cls.model_cls.query.filter(cls.model_cls.id.in_(batch))
+            if not with_deleted:
+                query = query.filter(
+                    (cls.model_cls.json.op("->")("deleted") == None)  # noqa
+                    | (cls.model_cls.json["deleted"] != cast("True", JSONB))
```
michamos, line 188: I believe the first line is not needed here, as Postgres puts `NULL` when the key is absent.

michamos (approved):
Minor comment, looks good :+1:

backend/inspirehep/records/api/base.py
MJedr: It is needed, tests don't pass when it's not there (the only result is the record with `deleted:False`)

Issue #2330, opened 2021-12-17
deduplicate references


MJedr:
Currently, we are deduplicating references only when extracting from pdf in workflows. We should deduplicate them also when updating data from editor.

PR #2346, opened 2022-01-17
deduplicate references

MJedr:
ref: #2330

### DIFF
--- a/backend/inspirehep/editor/views.py
+++ b/backend/inspirehep/editor/views.py
@@ -11,6 +11,7 @@ import structlog
 from flask import Blueprint, current_app, make_response, request
 from flask_login import current_user
 from inspire_schemas.api import load_schema
+from inspire_utils.dedupers import dedupe_list
 from invenio_db import db
 from invenio_records.models import RecordMetadata
 from invenio_records_rest.utils import set_headers_for_record_caching_and_concurrency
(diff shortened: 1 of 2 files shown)


drjova:
@MJedr what about orcid?

MJedr:
I didn't find any literature`update()` call in orcid module

backend/inspirehep/records/api/literature.py
```diff
@@ -191,18 +192,24 @@ def update_refs_to_conferences(data):
...
+        deduplicate_references=True,
         *args,
         **kwargs,
     ):
         with db.session.begin_nested():
             LiteratureRecord.update_authors_signature_blocks_and_uuids(data)
             LiteratureRecord.update_refs_to_conferences(data)
             data = self.add_files(data)
+            if deduplicate_references and "references" in data:
+                self._deduplicate_references(data)
```
michamos, line 212: isn't there a better place to put this (closer to reference extraction), instead of adding it to the general update API, where it's not needed in most cases?

drjova (approved):

backend/inspirehep/records/api/literature.py
MJedr: you mean in record editor? we decided that we don't want to do it there

MJedr:
@michamos here's an update

drjova (approved):

Issue #2334, opened 2022-01-06
`update_references_pointing_to_merged_record` only updates 10 records


michamos:
After merging records, that celery task is invoked to update all record refs mentioning the merged record to point to the surviving record instead. 

However, it doesn't work correctly and only updates 10 records for a given field. This is due to using the matcher, which does an ES query returning 10 results by default https://github.com/inspirehep/inspirehep/blob/4c998b112735c6dab066232cf90e6550e5060b77/backend/inspirehep/records/tasks.py#L84

After fixing the issue, we should rerun the celery task for all records marked as merged in the pidstore.

PR #2353, opened 2022-01-26
records: update references in all the records

MJedr:
ref: #2334

### DIFF
--- a/backend/inspirehep/records/tasks.py
+++ b/backend/inspirehep/records/tasks.py
@@ -8,20 +8,17 @@
 import structlog
 from celery import shared_task
 from elasticsearch import TransportError
+from elasticsearch_dsl import Q
 from flask import current_app
-from inspire_matcher.api import match
 from inspire_schemas.utils import get_refs_to_schemas
 from inspire_utils.record import get_value
 from invenio_db import db
 from sqlalchemy.exc import OperationalError
 from sqlalchemy.orm.exc import StaleDataError
 
-from inspirehep.matcher.utils import (
-    generate_matcher_config_for_nested_reference_field,
-    generate_matcher_config_for_reference_field,
-)
 from inspirehep.pidstore.api import PidStoreBase
 from inspirehep.records.api import InspireRecord, LiteratureRecord
+from inspirehep.search.api import InspireSearch
 from inspirehep.utils import flatten_list
 
 LOGGER = structlog.getLogger()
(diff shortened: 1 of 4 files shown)


backend/inspirehep/records/tasks.py
```diff
@@ -80,11 +78,14 @@ def update_references_pointing_to_merged_record(
...
+        query = get_query_for_given_path(index, path, merged_record_uri)
+        es_index_name = f"records-{index}"
+        matched_records = list(InspireSearch(index=es_index_name).query(query).scan())
         for matched_record in matched_records:
-            matched_inspire_record = InspireRecord.get_record(
-                matched_record["_id"], with_deleted=True
+            matched_inspire_record = InspireRecord.get_record_by_pid_value(
+                matched_record["control_number"],
+                SCHEMA_TO_PID_TYPES[index],
+                with_deleted=True,
```
michamos, line 88: can't you keep using the uuid like previously?

backend/inspirehep/records/tasks.py
```diff
@@ -80,11 +78,14 @@ def update_references_pointing_to_merged_record(
     refs_to_schema, merged_record_uri, new_record_uri
 ):
     for index, path in refs_to_schema:
-        config = get_config_for_given_path(index, path)
-        matched_records = match({"$ref": merged_record_uri}, config)
+        query = get_query_for_given_path(index, path, merged_record_uri)
+        es_index_name = f"records-{index}"
+        matched_records = list(InspireSearch(index=es_index_name).query(query).scan())
```
michamos, line 83: do you need the `list` here?

backend/inspirehep/records/tasks.py
```diff
@@ -80,11 +78,14 @@ def update_references_pointing_to_merged_record(
     refs_to_schema, merged_record_uri, new_record_uri
 ):
     for index, path in refs_to_schema:
-        config = get_config_for_given_path(index, path)
-        matched_records = match({"$ref": merged_record_uri}, config)
+        query = get_query_for_given_path(index, path, merged_record_uri)
+        es_index_name = f"records-{index}"
```
michamos, line 82: does this take into account the index prefix?

backend/inspirehep/records/tasks.py
MJedr: sure, it was only for debugging :)

backend/inspirehep/records/tasks.py
MJedr: yes, otherwise tests would fail cause they're prefixed (I also thought that it wouldn't take prefixes into account, but when I added function `prefix_index` it was prefixed twice)

michamos (approved):
:+1:

Issue #2343, opened 2022-01-14
add `/orcid/{pid_value}` and `/arxiv/{pid_value}` redirects


michamos:
we have pids for ORCID and arXiv, but those are not exposed on the UI, only the API. It would be useful to set up redirects from `/orcid/{pid_value}` and `/arxiv/{pid_value}` so external systems (arXiv, Fermilab) can point to the records directly without needing to go to a search page with one result.

PR #2355, opened 2022-01-28
add redirects for orcid and arxiv

MJedr:
ref:#2343

### DIFF
--- a/backend/inspirehep/legacy/views.py
+++ b/backend/inspirehep/legacy/views.py
@@ -63,7 +63,7 @@ def redirect_merge_profiles():
             302,
         )
 
-    return redirect(f"/authors", 302)
+    return redirect("/authors", 302)
 
 
 @blueprint.route("/author/manage_profile/<bai>", methods=("GET",))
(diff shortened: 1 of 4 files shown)


backend/inspirehep/legacy/views.py
```diff
@@ -115,3 +115,19 @@ def redirect_collection(legacy_collection):
 @blueprint.route("/info/<path:info_path>", methods=("GET",))
 def redirect_info(info_path):
     return redirect(f"{current_app.config['LEGACY_BASE_URL']}/info/{info_path}", 302)
+
+
+@blueprint.route("/legacy/arxiv/<arxiv>", methods=("GET",))
```
michamos, line 120: why `/legacy` in the route?

backend/inspirehep/legacy/views.py
```diff
@@ -115,3 +115,19 @@ def redirect_collection(legacy_collection):
...
+def redirect_to_lit_record_by_arxiv(arxiv):
+    recid = get_pid_for_pid("arxiv", arxiv, "recid")
+    if not recid:
+        abort(404)
+    return redirect(f"/literature/{recid}")
+
+
+@blueprint.route("/legacy/orcid/<orcid>", methods=("GET",))
```
michamos, line 128: same

backend/inspirehep/legacy/views.py
MJedr: cause we will change the route in ingress, otherwise we were not able to keep the route with `/api/orcid/<orcid>` without redirect

michamos (approved):

Issue #2398, opened 2022-02-25
Remove `name_suggest` from mappings


michamos:
Currently, the `name_suggest` field in the `hep` mapping is a `completion` field containing all possible permutations of author fields. This [seems to be unused](https://github.com/inspirehep/inspirehep/search?q=name_suggest) and is the only dependency on `inspire_utils.name.generate_name_variations`.

That approach is really bad (inefficient to generate, bloats the index), so we should try to remove that field. If that field is really unused, it should be removed outright (as well as the Marshmallow schemas that exist only to populate it), otherwise the autocompletion functionality that relies on it should be moved to a simple `search_as_you_type` field based on `authors.full_name`.

MJedr:
**TECH NOTES**
* In hep mapping remove `name_suggest` and add `name_suggest_as_you_type` with type `"type": "search_as_you_type"`.
* Add to es literature serializer a field populated with a method `populate_name_suggest_as_you_type`. Exclude this field from base schema.
* Remove leftovers of `name_suggest` from serializers.
* To add a suggester - in lit config in records/config.py add a config for this file, similarly to this 
* https://github.com/inspirehep/inspirehep/blob/838fa792691561a6338384f8a4d8e34d29b9d4a1/backend/inspirehep/records/config.py#L445-L450

PR #2406, opened 2022-03-01
mappings: remove name_suggest

nooraangelva:
Ref: #2398


drjova:
We should also remove `author_suggest` and replace it with the `suggest_as_you_type` cc @MJedr

MJedr:
but do we use that suggest? I removed this part of the refinement since we don't use that and Micha told to remove it.

drjova:
You are right, my bad 👌

drjova (approved):

Issue #2397, opened 2022-02-25
CERN affiliation should be taken into account only when curated for arXiv export to CDS


michamos:
Currently, when exporting records to CDS, we put all records in the arXiv OAI set that have a CERN affiliation. As that affiliation might be assigned automatically and the record not yet curated, this might cause records not actually associated to CERN but containing an incorrect affiliation to be pushed to CDS.

To avoid that, we should modify the affiliation check https://github.com/inspirehep/inspirehep/blob/838fa792691561a6338384f8a4d8e34d29b9d4a1/backend/inspirehep/oai/utils.py#L36-L44 to also check that `curated=True` in the record. The other clauses in https://github.com/inspirehep/inspirehep/blob/838fa792691561a6338384f8a4d8e34d29b9d4a1/backend/inspirehep/oai/utils.py#L95-L98 are more reliable and don't need to be modified.

PR #2428, opened 2022-03-10
exporting: add curation check

nooraangelva:
#2397

### DIFF
--- a/backend/inspirehep/oai/utils.py
+++ b/backend/inspirehep/oai/utils.py
@@ -37,11 +37,15 @@ def has_cern_affiliation(record):
     """Check if the record should be part of `CERN:arXiv` set."""
     affiliations_values = get_value(record, "authors.affiliations.value", default=[])
     affiliations = chain.from_iterable(affiliations_values)
-    return any(
+
+    if any(
         affiliation
         for affiliation in affiliations
         if affiliation.lower().startswith("cern")
-    )
+    ) and record.get("curated"):
+        return True
+
+    return False
 
 
 def has_cern_collaboration(record):
--- a/backend/tests/unit/oai/test_utils.py
+++ b/backend/tests/unit/oai/test_utils.py
@@ -56,13 +56,22 @@ def test_has_cern_accelerator_experiment_without_field():
 
 @pytest.mark.parametrize(
     "affiliation,expected",
-    [("cern", True), ("CERN", True), ("CERNOIS", True), ("JESSICAJONES", False)],
+    [
+        ("cern", True),
+        ("CERN", True),
+        ("CERNOIS", True),
+        ("JESSICAJONES", False),
+    ],
 )
 def test_has_cern_affiliation(affiliation, expected):
     data = {
+        "curated": True,
         "authors": [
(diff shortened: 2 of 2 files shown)


backend/inspirehep/oai/utils.py
```diff
@@ -37,11 +37,15 @@ def has_cern_affiliation(record):
...
     affiliations_values = get_value(record, "authors.affiliations.value", default=[])
     affiliations = chain.from_iterable(affiliations_values)
-    return any(
+    authors = get_value(record, "authors", default=[])
+    if any(
         affiliation
         for affiliation in affiliations
         if affiliation.lower().startswith("cern")
-    )
+    ) & any(author for author in authors if author["curated_relation"]):
```
MJedr, line 45: this is causing an error in tests cause not all authors have curated_relation. Instead you have to do `author.get('curated_relation`),

backend/inspirehep/oai/utils.py
```diff
@@ -37,11 +37,19 @@ def has_cern_affiliation(record):
...
+    authors = get_value(record, "authors", default=[])
+    if any(
         affiliation
         for affiliation in affiliations
         if affiliation.lower().startswith("cern")
-    )
+    ) & any(
+        author
+        for author in authors
+        if get_value(author, "curated_relation", default=True)
```
MJedr, line 48: don't need to use `get_value` here, this is a wrapper which is useful when you have to get a value of the nested field. Also, default should be `false` which you don't need to specify cause `None` (returned by default) is `false`.

backend/inspirehep/oai/utils.py
```diff
@@ -37,11 +37,19 @@ def has_cern_affiliation(record):
...
+    ):
+        return True
+    else:
+        return False
```
drjova, line 52:
```suggestion
    ) & any(
        author
        for author in authors
        if get_value(author, "curated_relation", default=True)
    ):
        return True
    return False
```

MJedr:
I am not sure if your solution is correct. In the task Micha wrote to check `curated` and not `curated_relation` in authors:
`"to also check that curated=True in the record."`

nooraangelva:
> I am not sure if your solution is correct. In the task Micha wrote to check `curated` and not `curated_relation` in authors:
> `"to also check that curated=True in the record."`

You are right. I had searched for the ```curated``` field from a few records and I didn't find it so I presumed it was the ```curated_relation``` that he meant. I searched a few more records now and found the correct field. I will fix this solution. Thank you for noticing my mishap, when I didn't. :)

backend/tests/unit/oai/test_utils.py
```diff
@@ -56,30 +56,76 @@ def test_has_cern_accelerator_experiment_without_field():
...
+        ],
     }
     record = faker.record("lit", data)
-    assert True == has_cern_affiliation(record)
+    assert has_cern_affiliation(record)
+
+
+def test_has_cern_affiliation_with_currated():
+    data_not_currated = {
+        "curated": False,
```
MJedr, line 100: can you add a test without `curated` key too?

MJedr (approved):

PR #2429, opened 2022-03-11
assign: add option for different profile

MJedr:


ui/src/authors/containers/AssignAllDifferentProfileActionContainer.jsx
```diff
@@ -0,0 +1,37 @@
...
+});
+
+const dispatchToProps = (dispatch) => ({
+  onAssign({
+    from,
+    to,
+    claimingUnclaimedPapersDisabled,
+    claimingClaimedPapersDisabled,
+  }) {
+    if (!claimingUnclaimedPapersDisabled) {
```
drjova, line 25: could we have two different dispatches?

`onAssignWithoutUnclaimed` and `onAssignWithoutClaimed` for example.

ui/src/authors/containers/PublicationSelectContainer.jsx
```diff
@@ -5,14 +5,20 @@ import {
...
 } from '../../actions/authors';
 
 const stateToProps = (state, { recordId }) => ({
   checked: state.authors.get('publicationSelection').has(recordId),
 });
 
-const dispatchToProps = (dispatch, { recordId, claimed }) => ({
+const dispatchToProps = (dispatch, { recordId, claimed, canClaim }) => ({
   onChange(event) {
+    if (!canClaim) {
```
drjova, line 17: I would split it into 3 (or maybe 4) different dispatches

ui/src/authors/containers/AssignAllDifferentProfileActionContainer.jsx
MJedr: but we need to decide somewhere which one we activate and we have only one button for that in the Action, so we need to know which function we need to use `onClick`

ui/src/authors/containers/PublicationSelectContainer.jsx
MJedr: again, we need to have somewhere a logic which actions we want to do `onChange`, I don't know which place would be better

backend/inspirehep/accounts/handlers.py
```diff
@@ -19,15 +20,28 @@ def get_current_user_data():
...
     }
 
+    user_profile_pidstore_record = PersistentIdentifier.query.filter_by(
+        pid_type="orcid", pid_value=orcid
+    ).all()
+
+    if user_profile_pidstore_record:
+        user_profile_recid = PersistentIdentifier.query.filter_by(
+            object_uuid=str(user_profile_pidstore_record[0].object_uuid), pid_type="aut"
+        ).one()
```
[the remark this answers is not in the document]
MJedr: I think that the best thing I can do here is to use `subquery` since I don't have object uuid and I need to get it somehow before querying for recid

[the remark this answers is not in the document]
drjova: Ok let's do that then 👍

ui/src/authors/containers/AssignOneDifferentProfileContainer.jsx
```diff
@@ -0,0 +1,47 @@
...
+    dispatch(setPublicationSelection([recordId], true));
+    if (!canClaimDifferentProfile) {
+      dispatch(setPublicationsCanNotClaimSelection([recordId], true));
+    }
+    if (!disabledAssignAction && canClaimDifferentProfile) {
+      dispatch(setPublicationsUnclaimedSelection([recordId], true));
+      dispatch(assignDifferentProfileUnclaimedPapers({ from, to }));
+    } else {
+      dispatch(setPublicationsClaimedSelection([recordId], true));
+      dispatch(assignDifferentProfileClaimedPapers({ from, to }));
```
drjova, line 39: If this https://github.com/inspirehep/inspirehep/pull/2429/files#diff-d05b30a07ecd69ef8dd8c77c9f48af79e1759380f12ea2f4e7aec7218e5c2dd1R31-R33 happens it will also run this else, is it what we want?

backend/inspirehep/assign/views.py
```diff
@@ -97,6 +99,24 @@ def author_assign_view():
...
+    claimed_literature_recids = body["papers_ids_already_claimed"]
+    not_allowed_to_be_claimed_literature_recids = body["papers_ids_not_matching_name"]
+    create_rt_ticket_for_claiming_action(
+        from_author_recid,
+        to_author_recid,
+        claimed_literature_recids,
+        not_allowed_to_be_claimed_literature_recids,
+    )
+    db.session.commit()
+    return jsonify({"message": "Success"}), 200
```
drjova, line 117: I have two comments about this:
1. Why we don't have it as part of the claiming? A request might fail.
2. I will make the `create_rt_ticket_for_claiming_action` async since it does some querying and takes time especially if the user select for example 100 papers.

ui/src/authors/containers/AssignOneDifferentProfileContainer.jsx
MJedr: yes cause paper can be both claimed and have incompatible author name

backend/inspirehep/assign/views.py
MJedr: I thought that I'll make another endpoint cause we might do 2 actions for this type of claiming (if you have claimed & unclaimed in your selection) and in case of claimedPapers we need to split papersIds into papers that are claimed and the ones with wrong name so the current args in the default assign view don't match the new ones and the logic becomes more complex. Also we don't assign an author in that case, we only create a ticket.

drjova: I see, I would make the `create_rt_ticket_for_claiming_action` async not to wait the request

drjova (approved):
LGTM 👏 let's test it!

PR #2440, opened 2022-03-16
update release action

MJedr:


drjova:
@MJedr you have to update also this file  https://github.com/inspirehep/inspirehep/blob/master/.github/workflows/integration.yml

Issue #2196, opened 2021-09-08
Make the distinction between `Literature` and hidden collections more clear


michamos:
See https://rt.inspirehep.net/Ticket/Display.html?id=1166423 and discussion in https://inspirehep.zulipchat.com/#narrow/stream/195298-experts/topic/CDS.20candidates.20on.20labs

We don't display records from hidden collections (`HAL Hidden`, `Fermilab`, etc.) which are not part of Literature in the default search results, but they are still publicly available, so people can end up there when following external links. This confuses users, so it would be good to make it more clear when records are not part of the main `Literature` collection.

StellaCh:
In the literature detailed page of these records, display an alert message, similar to the one we used for closed jobs, see https://inspirehep.net/jobs/1899313.
The text should be: `This record is not part of the INSPIRE Literature collection. Learn more` 
On Learn more should link to the FAQ page

MJedr:
**TECH NOTES**
1. Create a component `LiteratureCollectionBanner` similar [to this](https://github.com/inspirehep/inspirehep/blob/ba5fe0fdd9aa4daebad985376068487edccf995a/ui/src/jobs/components/JobStatusAlert.jsx)
2. Put it in [`LiteratureDetailedContainer`](https://github.com/inspirehep/inspirehep/blob/ba5fe0fdd9aa4daebad985376068487edccf995a/ui/src/literature/containers/DetailPageContainer/DetailPageContainer.jsx)

PR #2463, opened 2022-03-29
Add a banner to lit for hidden collections

nooraangelva:
Ref: #2196

### DIFF
--- a/ui/src/literature/components/LiteratureCollectionBanner.jsx
+++ b/ui/src/literature/components/LiteratureCollectionBanner.jsx
@@ -1,0 +1,36 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import { Alert } from 'antd';
+import ExternalLink from '../../common/components/ExternalLink.tsx';
+
+function HiddenCollectionAlert({ collections }) {
+  const shouldDisplayAlert = collections !== 'Literature';
+
+  return (
+    shouldDisplayAlert && (
+      <div className="mb2">
+        <Alert
+          type="warning"
+          message={
+            <span>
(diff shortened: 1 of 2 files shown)


ui/src/literature/components/LiteratureCollectionBanner.jsx
```diff
@@ -0,0 +1,36 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import { Alert } from 'antd';
+import ExternalLink from '../../common/components/ExternalLink.tsx';
+
+function HiddenCollectionAlert({ collections }) {
+  const shouldDisplayAlert = collections !== 'Literature';
```
drjova, line 7: ```collections``` is an array 😅 for example
```json
{
"_collections": [
    "Literature",
    "Fermilab"
  ],
}

```

Looking at the code more carefully, we don't make `_collections` [available](https://github.com/inspirehep/inspirehep/blob/140f7323cc0f12c51b334912743bfbb39cb2cb33/backend/inspirehep/records/marshmallow/literature/base.py#L42) to the ui serializer, which means `_collections` is not in the response. What we can do is the following, since we don't want to make this information available to users without permission, we could add a new boolean field in the [ui serializer](https://github.com/inspirehep/inspirehep/blob/master/backend/inspirehep/records/marshmallow/literature/ui.py#L95) called `is_collection_hidden`:


```python

 is_collection_hidden = fields.Method("get_ is_collection_hidden")

def is_collection_hidden(self, data, original_data):
    return True or False
```

Then in the UI we could read this and display or not the banner.

PR #2470, opened 2022-04-01
Add banner for hidden collections in lit

nooraangelva:
Ref: #2196


drjova:
@nooraangelva could you please rebase and upload a screenshot? Thank you!

backend/inspirehep/records/marshmallow/literature/ui.py
```diff
@@ -110,6 +111,12 @@ class Meta:
     thesis_info = fields.Nested(ThesisInfoSchemaV1, dump_only=True)
     dataset_links = fields.Method("get_datasets")
 
+    def is_collection_hidden(self, data):
+        collections = data.get("_collections", [])
```
MJedr, line 115: it can't be empty, record is not valid then

backend/inspirehep/records/marshmallow/literature/ui.py
```diff
@@ -110,6 +111,12 @@ class Meta:
...
+        if "Literature" in collections and len(collections) > 1:
```
MJedr, line 116: I think that this check is not correct. We have following options:
* length of collections is one and the only collection is `Literature` --> not hidden
* length of collections is one but `Literature` is not there --> hidden
* length is more than 1 --> then there is for sure at least one hidden.
For now you're handling correctly only case 3.

MJedr (commented):
Can you add a backend test please?

nooraangelva:
The banner in literature:

Screenshot 2022-04-05 at 10 01 17

ui/src/literature/components/LiteratureCollectionBanner.jsx
```diff
@@ -0,0 +1,34 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import { Alert } from 'antd';
+import ExternalLink from '../../common/components/ExternalLink.tsx';
+
+function HiddenCollectionAlert({ hiddenCollection }) {
```
MJedr, line 6: Can you write a test for this component (containing a snapshot)?

ui/src/literature/components/LiteratureCollectionBanner.jsx
```diff
@@ -0,0 +1,34 @@
...
+              This record is not part of the INSPIRE Literature collection.
+              <ExternalLink
+                as="a"
+                href="https://inspirehep.net/help/knowledge-base/faq/#faq-published"
+              >
+                Learn More
+              </ExternalLink>
+            </span>
+          }
+          showIcon={false}
```
MJedr, line 23: The default is false (you don't set banner mode)
```
false, in banner mode default is true
```

ui/src/literature/containers/DetailPageContainer/DetailPageContainer.jsx
```diff
@@ -104,6 +105,11 @@ function DetailPage({
       />
       <Row className="__DetailPage__" type="flex" justify="center">
         <Col xs={24} md={22} lg={21} xxl={18}>
+          <Row className="mv3" type="flex" justify="center">
+            <Col span={24}>
+              <HiddenCollectionAlert hiddenCollection={hiddenCollection} />
```
MJedr, line 110: put the check if collection hidden here instead

ui/src/literature/components/LiteratureCollectionBanner.jsx
```diff
@@ -0,0 +1,31 @@
...
+            </ExternalLink>
+          </span>
+        }
+      />
+    </div>
+  );
+}
+
+HiddenCollectionAlert.propTypes = {
+  collection: PropTypes.string,
```
MJedr, line 28: not used anymore

PR #2484, opened 2022-04-06
Packages setup

karolina-siemieniuk-morawska:
* added Brewfile with packages that required changes in order to run Inspirehep on Mac with M1 chip


backend/inspirehep/Brewfile
```diff
@@ -0,0 +1,27 @@
...
+brew "pyenv"
+brew "pyenv-virtualenv"
+brew "pyenv-virtualenvwrapper"
+brew "python@3.8"
+brew "python@3.9"
+brew "sshuttle"
+brew "virtualenv"
+brew "yarn"
+brew "zlib"
+brew "int128/kubelogin/kubelogin"
```
MJedr, line 27: not sure if wee need k8s dependencies there @drjova?

drjova: true, could you please @karolina-siemieniuk-morawska remove it?

drjova: Also could you please add another section on the documentation on how to install this file?

drjova:
@karolina-siemieniuk-morawska also i just realize that the `Brewfile` is in the wrong directory, it should be in the root of the repo not inside `backend/inspirehep`

Issue #1945, opened 2021-06-11
Don't reset the `curation collection` facet when doing a search


michamos:
@PaulinaBaranowska reports that it would be more useful to keep that facet active even when doing a search. Currently, it gets reset after changing the search query like all facets.

MJedr:
**TECH NOTES**
Add `curation_collection` [to the `persistedQueryParamsDuringNewSearch` here](https://github.com/inspirehep/inspirehep/blob/13611dfecb47ff2c94d1eb5d138e6bdcd842292d/ui/src/search/config.js#L40)

PR #2486, opened 2022-04-06
ui: do not reset collection facet on new search

karolina-siemieniuk-morawska:
* Keeps the collection facet active on new search

* ref: #1945

### DIFF
--- a/ui/src/search/config.js
+++ b/ui/src/search/config.js
@@ -28,7 +28,7 @@ import {
   onEmbeddedSearchWithoutAggregationsQueryChange,
 } from './queryChange';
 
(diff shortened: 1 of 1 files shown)


ui/src/search/config.js
```diff
@@ -28,7 +28,7 @@ import {
   onEmbeddedSearchWithoutAggregationsQueryChange,
 } from './queryChange';
 
-const defaultPersistedQueryParamsDuringNewSearch = ['size'];
+const defaultPersistedQueryParamsDuringNewSearch = ['size', 'collection'];
```
drjova, line 31: It's better to add it only for `LITERATURE_NS`

Issue #1098, opened 2020-07-08
Desirialization of search results takes up to 25% of the response time


drjova:
The goal is to reduce desirialzation time as much as possible, based on this [profiling](https://inspirehep.zulipchat.com/user_uploads/4640/patlHPW3EKcu9AUzSX3a__yV/wUMJos2K.svg) and improve the overall response time for search queries.

michamos:
spending 20% of the time in `strip_empty_values`. That function is expensive as it's recursive (removes all empty values from leaves, then walks up the tree recursively to remove now empty nodes), and it's called multiple times.

PR #2491, opened 2022-04-08
improve deserialisation response time

nooraangelva:
Ref: #1098


backend/inspirehep/records/marshmallow/fields/nested_without_empty_objects.py
```diff
@@ -5,16 +5,12 @@
 # inspirehep is free software; you can redistribute it and/or modify it under
 # the terms of the MIT License; see LICENSE file for more details.
 
-from inspire_dojson.utils import strip_empty_values
 from marshmallow import fields, missing
 
 
-class NestedWithoutEmptyObjects(fields.Nested):
+class NestedObjects(fields.Nested):
```
MJedr, line 11: I would rename it to `NestedField` or something like that since we don't have Empty Objects inside anymore ;)

MJedr (approved):

PR #2494, opened 2022-04-08
lit: is_collection_hidden only when literature not in collections

MJedr:
ref:cern-sis/issues-inspire#11


backend/inspirehep/records/marshmallow/literature/ui.py
```diff
@@ -113,11 +113,11 @@ class Meta:
...
         collections = data.get("_collections")
-        if collections is None:
+        if not collections:
             return missing
-        if "Literature" in collections and len(collections) == 1:
-            return False
-        return True
+        if "Literature" not in collections:
+            return True
+        return False
```
michamos, line 120:
```suggestion
        return "Literature" not in collections
```
:wink:

michamos (approved):

Issue #2341, opened 2022-01-14
Avoid OOM during disambiguation


michamos:
See also https://cern-rcs-sis.zulipchat.com/#narrow/stream/310485-inspire/topic/redis.20issues/near/267695305:

When running author disambiguation, we might create new author records. When creating a new author record and the corresponding BAI, we check if there are Literature records with that BAI already (assigned by legacy) and try to assign them to the new author record. This potentially affects a large number of records, all of them in the current DB transaction, which might cause the worker to go OOM as all touched models need to be kept in memory as they are needed in the `models_committed` receiver.

We need to find a way to solve this. One possibility would be to retain less information for the receiver, only the primary key instead of the full model (needs a patch to `flask-sqlalchemy` that we have forked already).

Once this is fixed, we need to re-disambiguate the non disambiguated records https://inspirehep.net/literature?sort=mostrecent&size=25&page=1&q=authors.full_name%3A%2A%20-authors.record.%24ref%3A%2A%20and%20da%20%3E%202021%20and%20_collections%3ALiterature.

drjova:
TECH NOTES:

* Add a feature flag for https://github.com/inspirehep/inspirehep/blob/master/backend/inspirehep/records/api/authors.py#L73

PR #2498, opened 2022-04-11
add feature flag for assign author pages

nooraangelva:
Ref: #2341


backend/inspirehep/records/api/authors.py
```diff
@@ -70,7 +71,8 @@ def get_linked_author_records_uuids_if_author_changed_name(self):
     @classmethod
     def create(cls, data, id_=None, *args, **kwargs):
         record = super().create(data, id_, **kwargs)
-        record.assign_author_to_papers()
+        if current_app.config.get("FEATURE_FLAG_ENABLE_ASSIGN_AUTHOR_PAPERS", False):
```
MJedr, line 74: no need for the default `False` here, by default it will return none which is false

MJedr (approved):

PR #2500, opened 2022-04-12
ui: display number of selected papers in authors details

karolina-siemieniuk-morawska:
* fixed missing number of seleted papers when selecting authors publications

* ref: cern-sis/issues-inspire#10


drjova:
@karolina-siemieniuk-morawska thank you for your PR, I would suggest the following:

* I would disconnect it and make it a simple component https://github.com/inspirehep/inspirehep/blob/f8b75af3325c18c260f5c8da4dbb8d0fe46ef577/ui/src/literature/containers/NumberOfResultsWithSelectedItemsNumberContainer.jsx
*  in `LiteratureSearch` I would add an extra prop call selectedKey (default to `literatureSearch`) and will pass 
```js       
<NumberOfResultsWithSelectedItemsNumber
   numberOfSelected={numberOfSelected}
/>
```
and in 

```js
const stateToProps = (state, { namespace, selectedKey }) => ({
  loading: state.search.getIn(['namespaces', namespace, 'loading']),
  loadingAggregations: state.search.getIn([
    'namespaces',
    namespace,
    'loadingAggregations',
  ]),
  numberOfSelected: sate.getIn([namespace, selectedKey]).size
  results: state.search.getIn(['namespaces', namespace, 'results']),
  isCitationSummaryVisible: isCitationSummaryEnabled(state),
  isCatalogerLoggedIn: isCataloger(state.user.getIn(['data', 'roles'])),
});

```

ui/src/literature/components/NumberOfResultsWithSelectedItemsNumber.jsx
```diff
@@ -14,18 +14,18 @@ function getFormattedNumberOfSelectedOrNull(numberOfSelected) {
   );
 }
 
-function NumberOfResults({ namespace, numberOfSelected }) {
+function NumberOfResults({ namespace, numberOfSelected}) {
   return (
     <span>
-      {getFormattedNumberOfSelectedOrNull(numberOfSelected)}
+      {!!numberOfSelected && getFormattedNumberOfSelectedOrNull(numberOfSelected)}
```
drjova, line 20: Why do you need this? anyway if it's undefined or empty it will return `null`.

karolina-siemieniuk-morawska: If the value of _numberOfSelected_ is undefined, _getFormattedNumberOfSelectedOrNull_ does not return null, it only returns null if _numberOfSelected_ is 0. I can either change it to return 0 if the value is undefined, like this: 

```
function getFormattedNumberOfSelectedOrNull(numberOfSelected) {
  if (numberOfSelected === 0 || !numberOfSelected) {
    return null;
  }
```

or leave the above check for truthiness of the value. What do you think :)?

drjova:
@karolina-siemieniuk-morawska could you please add a test, in AuthorsPublicationContainer?

karolina-siemieniuk-morawska:
> 

done 🤓

drjova (approved):

PR #2502, opened 2022-04-12
fulltext: handle file download exception

MJedr:


backend/inspirehep/records/marshmallow/literature/es.py
```diff
@@ -249,7 +250,10 @@ def get_documents_with_fulltext(self, record_data):
...
             ):
-                file_data = download_file_from_url(document["url"])
-                encoded_file = base64.b64encode(file_data).decode("ascii")
-                document["text"] = encoded_file
+                try:
+                    file_data = download_file_from_url(document["url"])
+                    encoded_file = base64.b64encode(file_data).decode("ascii")
+                    document["text"] = encoded_file
+                except DownloadFileError:
+                    continue
```
drjova, line 258: we should log the exception

drjova (approved):

Issue #2442, opened 2022-03-17
Celery requeue tasks that are failing


drjova:
Celery requeue tasks that are failing

MJedr:
**TECH NOTES**
Add [`Reject(requeue=True)`](https://docs.celeryq.dev/en/master/userguide/tasks.html?highlight=Reject#reject) [here](https://github.com/inspirehep/inspirehep/blob/683a7c33c2d62445d2236ae17eae7b05f4787481/backend/inspirehep/celery.py#L28) for other kinds of exception

PR #2504, opened 2022-04-13
Celery requeue tasks

nooraangelva:
Ref: #2442

### DIFF
--- a/backend/inspirehep/celery.py
+++ b/backend/inspirehep/celery.py
@@ -34,6 +34,8 @@ class CeleryTask(AppContextTask):
                     term_log.write(str(exc))
             finally:
                 os.kill(os.getppid(), signal.SIGTERM)
+        else:
+            raise self.retry(exc=exc, max_retries=3)
 
     def __call__(self, *args, **kwargs):
         """Fix for flask-celeryext __call__ override fail.
--- a/backend/tests/integration/submissions/test_submissions_views.py
+++ b/backend/tests/integration/submissions/test_submissions_views.py
@@ -33,7 +33,8 @@ from inspirehep.records.api import (
 )
 
 
-def test_author_submit_requires_authentication(inspire_app):
+@patch("inspirehep.submissions.views.async_create_ticket_with_template")
+def test_author_submit_requires_authentication(ticket_mock, inspire_app):
     with inspire_app.test_client() as client:
         response = client.post(
             "/submissions/authors",
@@ -1279,7 +1280,8 @@ def test_update_job_status_from_open(ticket_mock, inspire_app):
 
 
 @freeze_time("2019-01-31")
-def test_job_update_data_30_days_after_deadline(inspire_app):
+@patch("inspirehep.submissions.views.async_create_ticket_with_template")
+def test_job_update_data_30_days_after_deadline(ticket_mock, inspire_app):
     user = create_user()
     with inspire_app.test_client() as client:
         login_user_via_session(client, email=user.email)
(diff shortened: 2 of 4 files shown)


backend/inspirehep/celery.py
```diff
@@ -34,6 +35,8 @@ def on_failure(self, exc, task_id, args, kwargs, einfo):
                     term_log.write(str(exc))
             finally:
                 os.kill(os.getppid(), signal.SIGTERM)
+        else:
+            Reject(requeue=True)
```
drjova, line 39: Could you please add a test for this?

nooraangelva: Sure. Are there any tests I could use as a reference? I haven't done any tests for celery yet, so I'm not sure where to create one.

michamos: wouldn't you need to `raise` this, as it's an exception? :thinking:

nooraangelva: That is my understanding also from the documentation, but when I raise it I get this error from several tests ```celery.exceptions.Reject: (AttributeError("'NoneType' object has no attribute 'create_ticket'"), True)```

nooraangelva:
> @nooraangelva when everything is passing could you please rename the commit message? It has a typo 🙃

sure thing 👍

backend/tests/integration/submissions/test_submissions_views.py
```diff
@@ -1727,8 +1733,9 @@ def test_confirmation_email_not_sent_when_user_is_cataloger(
 
 
 @patch("inspirehep.submissions.views.send_conference_confirmation_email")
+@patch("inspirehep.submissions.views.async_create_ticket_with_template")
```
drjova, line 1736: You don't need this `patch` it has already :)

nooraangelva: hmm, okay but it then has the same error as the ones that didn't have the ```patch```

drjova: Ok I see, then we have to put this patch as well.

nooraangelva: yep, then just need to figure out a way to 'fix' the ```assertion error``` it creates.

backend/tests/unit/submissions/test_tasks.py
```diff
@@ -31,3 +33,23 @@ def test_async_create_ticket_with_template_logs_on_error(mock_create_ticket):
...
+        {"foo": "bar"},
+        "Test Ticket",
+        12345,
+    ]
+
+    mock_create_ticket.side_effect = InvalidRequestError
+    with pytest.raises(InvalidRequestError):
+        async_create_ticket_with_template(*args)
+
+    mock_create_ticket.side_effect = Reject()
```
drjova, line 53: I would change that to any `Exception` and it should raise the `Reject`

michamos:
I have a stupid question: won't tasks get requeued forever? What about cases where the task actually can't succeed due to various reasons (as happens often during ORCID push or HAL push). Wouldn't that be a problem?

nooraangelva:
That's actually a good point. there is a warning in the documentation about the infinite loop.

I was thinking, that instead of ```Reject()```  we could use ```Retry()``` instead. That can have ```max_retries``` so there is no infinite loop.

nooraangelva:
So what would be the best solution for this? Any thoughts @drjova?

drjova:
We are already retrying the tasks, I think we should keep doing that instead of a global solution.

nooraangelva:
Okay, so this PR and issue can be closed then?

PR #2508, opened 2022-04-14
es fulltext schema: fix get_documents_with_fulltext

MJedr:
* increase timeout for fulltext indexing task


backend/inspirehep/config.py
```diff
@@ -232,6 +232,7 @@ def annotate(self, task):
 INDEXER_DEFAULT_INDEX = "records-hep"
 INDEXER_DEFAULT_DOC_TYPE = "_doc"
 INDEXER_BULK_REQUEST_TIMEOUT = 900
+FULLLTEXT_INDEXER_REQUEST_TIMEOUT = "60s"
```
drjova, line 235: do you need the `s`?

drjova (approved):
Just the comment

backend/inspirehep/config.py
MJedr: I wanted to make sure that it would be in seconds, also
[```Note that newer versions need a unit, e.g. timeout="60s"```](https://stackoverflow.com/a/62800911/13207632) so I thought it would be better to put `s`

PR #2509, opened 2022-04-19
ui: bug in select all checkbox in authors

karolina-siemieniuk-morawska:
* 'select all' checkbox in authors didn't highlight selected papers checkboxes because of missing checked prop. Added missing prop to PublicationsSelect component

* ref: cern-sis/issues-inspire#16

### DIFF
--- a/ui/src/authors/components/__tests__/__snapshots__/PublicationsSelect.test.jsx.snap
+++ b/ui/src/authors/components/__tests__/__snapshots__/PublicationsSelect.test.jsx.snap
@@ -1,5 +1,29 @@
 // Jest Snapshot v1, https://goo.gl/fbAQLP
 
+exports[`PublicationsSelect renders checked when selected 1`] = `
+<Checkbox
+  checked={true}
+  indeterminate={false}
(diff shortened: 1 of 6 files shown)


drjova:
@karolina-siemieniuk-morawska thank you! could you please write some tests cases to make sure that it works?

ui/src/authors/components/__tests__/__snapshots__/PublicationsSelect.test.jsx.snap
```diff
@@ -1,5 +1,29 @@
...
+`;
+
+exports[`PublicationsSelect renders unchecked when not selected 1`] = `
+<Checkbox
+  checked={false}
+  indeterminate={false}
+  onChange={[Function]}
+/>
+`;
+
```
drjova, line 26: could we add also for `Select all`?

karolina-siemieniuk-morawska: Done

drjova (approved):

Issue #1674, opened 2021-01-28
Reorder author name after succesful disambiguation


michamos:
Some authors have several last names, but the grouping of last and first name is wrong on the paper. For example we have `Smith, John Davis`, but the actual name should be `Smith Davis, John`. After performing author disambiguation, we should take the proper grouping of names from the linked author record (just the grouping, not initials/spelling/accents).

MJedr:
**TECH NOTES**
1. Add a function `reorder_literature_author_names` and call it [here](https://github.com/inspirehep/inspirehep/blob/13611dfecb47ff2c94d1eb5d138e6bdcd842292d/backend/inspirehep/disambiguation/tasks.py#L263) for author that was disambiguated to assign a new name to the lit author. In the function:
* get name from `author.name.value` from author record and move corresponding name from lit author to last_name or first_name, depending on the position in author name (on the left comma side - last name). Normalize strings with `unidecode` for comparing. If there's no token in author name, leave it as it was in lit author.

michamos:
Just to make sure everything is clear: the name absolutely has to be the one on the paper, all that's allowed is moving things around.

Also, as this is just a heuristic, we probably should do this only in case there is only one last name in the literature record, if there's more than one, chances are that it's already been split correctly at the source.

drjova:
Thanks for the clarification!

PR #2510, opened 2022-04-19
disambiguation: reorder names

MJedr:
ref:#1674

### DIFF
--- a/backend/inspirehep/disambiguation/tasks.py
+++ b/backend/inspirehep/disambiguation/tasks.py
@@ -13,6 +13,7 @@ from sqlalchemy.orm.exc import NoResultFound, StaleDataError
 from inspirehep.disambiguation.utils import (
     create_new_stub_author,
     link_signatures_to_author,
+    reorder_lit_author_names,
     update_author_names,
 )
 from inspirehep.editor.editor_soft_lock import EditorSoftLock
@@ -21,6 +22,7 @@ from inspirehep.matcher.validators import (
     collaboration_validator,
 )
 from inspirehep.records.api import InspireRecord
+from inspirehep.records.api.authors import AuthorsRecord
 
 LOGGER = structlog.getLogger()
 
@@ -248,19 +250,29 @@ def disambiguate_authors(self, record_uuid, record_version_id):
             assign_bai_to_literature_author(
                 author, matched_author_data.get("author_bai")
             )
+
             updated_authors.append(
                 matched_author_data["author_reference"].split("/")[-1]
             )
         elif "record" not in author:
-            new_author_record = create_new_author(
+            linked_author_record = create_new_author(
                 author["full_name"], record["control_number"]
             )
-            author["record"] = new_author_record["self"]
+            author["record"] = linked_author_record["self"]
             new_author_bai = get_values_for_schema(
-                new_author_record["ids"], "INSPIRE BAI"
+                linked_author_record["ids"], "INSPIRE BAI"
             )[0]
             assign_bai_to_literature_author(author, new_author_bai)
-            updated_authors.append(new_author_record["control_number"])
+            updated_authors.append(linked_author_record["control_number"])
+        if len(author["full_name"].split(",")[0].split(" ")) == 1:
+            if matched_author_data:
+                linked_author_record = AuthorsRecord.get_record_by_pid_value(
+                    matched_author_data["author_reference"].split("/")[-1]
+                )
+            author["full_name"] = reorder_lit_author_names(
+                author["full_name"], linked_author_record["name"]["value"]
+            )
+
     if updated_authors:
         LOGGER.info(
             "Updated references for authors",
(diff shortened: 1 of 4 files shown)


backend/inspirehep/disambiguation/utils.py
```diff
@@ -109,3 +110,30 @@ def update_author_names(author, signatures):
...
+    lit_author_first_names_reordered = []
+    lit_author_last_names_reordered = []
+
+    author_last_names = set(author_name.split(",")[0].split(" "))
+    author_first_names = (
+        set(author_name.split(",")[1].split(" ")) if "," in author_name else []
+    )
+
+    for name_idx, name in enumerate(lit_author_name_tokens):
+        if name in author_last_names:
```
michamos, line 127: you need to compare ASCII- and case-normalized tokens. So `Gómez` should match `gomez` (everywhere).

backend/inspirehep/disambiguation/utils.py
```diff
@@ -109,3 +110,30 @@ def update_author_names(author, signatures):
...
+            lit_author_last_names_reordered.append(name)
+            continue
+        if name in author_first_names:
+            lit_author_first_names_reordered.append(name)
+            continue
+        if name_idx + 1 > len(lit_author_last_names):
+            lit_author_first_names_reordered.append(name)
+        else:
+            author_first_names.append(name)
```
michamos, line 136: why is this (plus normalization) not sufficient?
```suggestion
        else:
            lit_author_first_names_reordered.append(name)
```

backend/inspirehep/disambiguation/tasks.py
```diff
@@ -248,19 +250,33 @@ def disambiguate_authors(self, record_uuid, record_version_id):
...
-            new_author_record = create_new_author(
+            linked_author_record = create_new_author(
                 author["full_name"], record["control_number"]
             )
-            author["record"] = new_author_record["self"]
+            author["record"] = linked_author_record["self"]
+            if len(author["full_name"].split(",")[0]) == 1:
+                author["full_name"] = normalize_lit_author_name(
+                    author["full_name"], linked_author_record["name"]["value"]
+                )
```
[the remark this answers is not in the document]
MJedr: yes, it's a leftover

backend/inspirehep/disambiguation/tasks.py
```diff
@@ -248,19 +250,33 @@ def disambiguate_authors(self, record_uuid, record_version_id):
...
             new_author_bai = get_values_for_schema(
-                new_author_record["ids"], "INSPIRE BAI"
+                linked_author_record["ids"], "INSPIRE BAI"
             )[0]
             assign_bai_to_literature_author(author, new_author_bai)
-            updated_authors.append(new_author_record["control_number"])
+            updated_authors.append(linked_author_record["control_number"])
+        if len(author["full_name"].split(",")[0].split(" ")) == 1:
```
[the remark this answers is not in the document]
MJedr: i wanted to have it here cause I didn't want to query db for author record if it's not needed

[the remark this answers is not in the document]
michamos: :+1: makes sense

backend/inspirehep/disambiguation/tasks.py
```diff
@@ -248,19 +250,33 @@ def disambiguate_authors(self, record_uuid, record_version_id):
...
+            if matched_author_data:
+                linked_author_record = AuthorsRecord.get_record_by_pid_value(
+                    matched_author_data["author_reference"].split("/")[-1]
+                )
```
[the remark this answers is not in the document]
MJedr: yes, we have it but it expects reference to be dict with `ref` key, to me it was clearer to split manually than putting it in dict and extracting from it

backend/inspirehep/disambiguation/utils.py
MJedr: so we assume that lit author name always matches one of the names from author?

michamos: hmmm, good point. I was thinking we apply this only when there's only a single last name, but that doesn't necessarily match either. If there's no match among the last names, we should keep the current last name then (not sure that's what's happening here though).

backend/inspirehep/disambiguation/utils.py
```diff
@@ -109,3 +111,35 @@ def update_author_names(author, signatures):
...
+            lit_author_first_names_reordered.append(name)
+            continue
+        # case when lit author names are not in author names
+        # - we leave name parts as they were
+        if name_idx + 1 > len(lit_author_last_names):
+            lit_author_first_names_reordered.append(name)
+        else:
+            lit_author_last_names_reordered.append(name)
+
+    lit_author_name_reordered = f"{' '.join(lit_author_last_names_reordered)}, {' '.join(lit_author_first_names_reordered)}"
```
drjova, line 144: just for readability I would make the `join`s in separated vars instead of directly in the format str

drjova (approved):
LGTM just the comment

PR #2511, opened 2022-04-19
fulltext pipeline: ignore missing documents field

MJedr:


backend/inspirehep/indexer/cli.py
```diff
@@ -391,6 +391,7 @@ def _put_files_pipeline():
                 {
                     "foreach": {
                         "field": "documents",
+                        "ignore_missing": True,
```
drjova, line 394: is the config the same on tests and here? could we just reuse it instead of modifying it twice?

drjova (approved):

Issue #2441, opened 2022-03-17
Cleanup FEATURE FLAGS


drjova:
We have feature flags everywhere that are not used any more, we should go through them and make the list for the ones we need.

MJedr:
HEP
Same value in local env, qa and prof for:
`FEATURE_FLAG_ENABLE_DISAMBIGUATION`
`FEATURE_FLAG_ORCID_PUSH_WHITELIST_REGEX`
`FEATURE_FLAG_ENABLE_QUERY_PARSER_ENDPOINT`
`FEATURE_FLAG_ENABLE_APPMETRICS`
`FEATURE_FLAG_ENABLE_ASSIGN_AUTHOR_PAPERS`

NEXT
Same value in local env, qa and prof for:
`FEATURE_FLAG_ENABLE_MATCH_REFERENCES_HEP`
`FEATURE_FLAG_ENABLE_ORCID_PUSH`
`FEATURE_FLAG_ENABLE_MERGER`
`FEATURE_FLAG_ENABLE_APPMETRICS`
`FEATURE_FLAG_ENABLE_PUBLISHER_HARVESTS_CARD`
`FEATURE_FLAG_ENABLE_FUZZY_MATCHER`
`FEATURE_FLAG_ENABLE_SENDING_REFERENCES_TO_LEGACY`

UI
Same value in local env, qa and prof for:
`ASSIGN_CONFERENCE_UI_FEATURE_FLAG`
`FIGURES_FEATURE_FLAG`

michamos:
`FEATURE_FLAG_ORCID_PUSH_WHITELIST_REGEX` on QA should be super restrictive to avoid claims getting accidentally pushed to ORCID.

MJedr:
yes, you are right. Small correction:
### flags hep
```
matching --->  FEATURE_FLAG_ENABLE_APPMETRICS local: False, qa: False, prod: False
not matching ---> FEATURE_FLAG_ENABLE_ASSIGN_AUTHOR_PAPERS local: False, qa: False, prod: None
not matching ---> FEATURE_FLAG_ENABLE_AUTHOR_DISAMBIGUATION local: False, qa: True, prod: True
not matching ---> FEATURE_FLAG_ENABLE_BAI_CREATION local: False, qa: True, prod: True
not matching ---> FEATURE_FLAG_ENABLE_BAI_PROVIDER local: False, qa: True, prod: True
[...]

PR #2518, opened 2022-04-22
cleanup feature flags

MJedr:
ref: #2441


backend/inspirehep/orcid/ext.py
```diff
@@ -34,9 +34,6 @@ def init_config(self, app):
         )
 
     def configure_appmetrics(self, app):
-        if not app.config.get("FEATURE_FLAG_ENABLE_APPMETRICS"):
```
michamos, line 37: you're enabling it here, not disabling

backend/inspirehep/orcid/ext.py
```diff
@@ -34,28 +30,4 @@ def init_config(self, app):
         )
 
     def configure_appmetrics(self, app):
```
drjova, line 32: are we using this function anywhere?

MJedr: now it would be useless, we used it in inspire orcid app https://github.com/inspirehep/inspirehep/blob/master/backend/inspirehep/orcid/ext.py

drjova: let's remove it :)

drjova:
I would also suggest to remove the following according to your [analysis](https://github.com/inspirehep/inspirehep/issues/2441) wdyt @MJedr  (some of them might be already in your PR)?


* FEATURE_FLAG_ENABLE_BAI_CREATION
* FEATURE_FLAG_ENABLE_CDS_SYNC
* FEATURE_FLAG_ENABLE_FILES 
* FEATURE_FLAG_ENABLE_LEGACY_VIEW_REDIRECTS
* FEATURE_FLAG_ENABLE_REDIRECTION_OF_PIDS
* FEATURE_FLAG_ENABLE_SELF_CITATIONS
* FEATURE_FLAG_ENABLE_TEXKEY_MINTER
* FEATURE_FLAG_ENABLE_WORKFLOW_ON_AUTHOR_UPDATE
* FEATURE_FLAG_ENABLE_QUERY_PARSER_ENDPOINT

MJedr:
I think that we should leave some flags cause we don't want to do everything locally - otherwise we'll have to mock a lot (for example after removal of files flag). Also for some flags I am not sure why we introduced them, eg. bai and pid creation - wouldn't it impact tests performance (same for self citations)?

drjova (approved):

PR #2520, opened 2022-04-25
experiments_submissions_backend

nooraangelva:
Ref: https://github.com/cern-sis/issues-inspire/issues/18

### DIFF
--- a/backend/inspirehep/submissions/views.py
+++ b/backend/inspirehep/submissions/views.py
@@ -31,6 +31,7 @@ from inspirehep.mailing.api.seminars import send_seminar_confirmation_email
 from inspirehep.records.api import (
     AuthorsRecord,
     ConferencesRecord,
+    ExperimentsRecord,
     JobsRecord,
     SeminarsRecord,
 )
@@ -41,6 +42,7 @@ from inspirehep.utils import get_inspirehep_url
 from .errors import WorkflowStartError
 from .loaders import author_v1 as author_loader_v1
 from .loaders import conference_v1 as conference_loader_v1
+from .loaders import experiment_v1 as experiment_loader_v1
 from .loaders import job_v1 as job_loader_v1
 from .loaders import literature_v1 as literature_loader_v1
 from .loaders import seminar_v1 as seminar_loader_v1
@@ -259,12 +261,24 @@ class ConferenceSubmissionsResource(BaseSubmissionsResource):
         )
 
 
+class ExperimentSubmissionsResource(BaseSubmissionsResource):
+    def load_data_from_request(self):
+        return experiment_loader_v1()
+
+    def post(self):
+        """Adds new experiment record"""
+
+        data = self.load_data_from_request()
+        record = ExperimentsRecord(data=data).create(data)
+        db.session.commit()
+        return (jsonify({"control_number": record["control_number"]}), 201)
+
+
 class SeminarSubmissionsResource(BaseSubmissionsResource):
     def post(self):
         """Adds new conference record"""
 
         data = self.load_data_from_request()
-
         data["acquisition_source"] = self.get_acquisition_source()
         record = SeminarsRecord.create(data)
         db.session.commit()
@@ -555,6 +569,11 @@ blueprint.add_url_rule(
 seminar_submission_view = SeminarSubmissionsResource.as_view("seminar_submission_view")
 blueprint.add_url_rule("/seminars", view_func=seminar_submission_view)
 blueprint.add_url_rule("/seminars/<int:pid_value>", view_func=seminar_submission_view)
+
+experiment_submission_view = ExperimentSubmissionsResource.as_view(
+    "experiment_submission_view"
+)
+blueprint.add_url_rule("/experiments", view_func=experiment_submission_view)
 
 conference_submission_view = ConferenceSubmissionsResource.as_view(
     "conference_submissions_view"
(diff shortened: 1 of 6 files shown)


backend/inspirehep/submissions/views.py
```diff
@@ -259,12 +260,23 @@ def create_ticket(self, record, rt_template):
         )
 
 
+class ExperimentSubmissionsResource(BaseSubmissionsResource):
+    def post(self):
+        """Adds new experiment record"""
+
+        data = self.load_data_from_request()["data"]
```
MJedr, line 267: what if there'll be key error?

I would suggest to create a dict with required fields and then update it with data from request.

backend/inspirehep/submissions/views.py
```diff
@@ -259,12 +260,23 @@ def create_ticket(self, record, rt_template):
...
+        data["_collections"] = ["Experiments"]
+        data["$schema"] = "https://labs.inspirehep.net/schemas/records/experiments.json"
+        record = ExperimentsRecord(data=data).create(data)
+        db.session.commit()
+        return (record, 201)
```
MJedr, line 272: are you sure we want to return a tuple of record and int 201? :p 
I think that you want to return `control_number` and status code using `jsonify`

backend/tests/integration/submissions/test_submissions_views.py
```diff
@@ -2152,3 +2153,48 @@ def test_new_seminar_submission_with_invalid_lit_record(
...
+
+    payload = orjson.loads(response.data)
+    experiment_id = payload["control_number"]
+    experiment_record = ExperimentsRecord.get_record_by_pid_value(experiment_id)
+
+    experiment_record_data = {
+        key: value
+        for (key, value) in experiment_record.items()
+        if key in REQUIRED_EXPERIMENT_RECORD_DATA
+    }
```
MJedr, line 2199: can't you just check if experiment record data equals expected data?

backend/inspirehep/submissions/views.py
```diff
@@ -259,12 +263,31 @@ def create_ticket(self, record, rt_template):
...
+
+    def post(self):
+        """Adds new experiment record"""
+
+        request_data = self.load_data_from_request()
+        data = {
+            "_collections": ["Experiments"],
+            "$schema": "https://inspirehep.net/schemas/records/experiments.json",
+            "legacy_name": request_data["legacy_name"],
+            "project_type": request_data["project_type"],
```
drjova, line 278: they key might be missing 🤔

backend/inspirehep/submissions/views.py
```diff
@@ -259,12 +263,31 @@ def create_ticket(self, record, rt_template):
...
+    def load_data_from_request(self):
+        return experiment_loader_v1()
+
+    def post(self):
+        """Adds new experiment record"""
+
+        request_data = self.load_data_from_request()
+        data = {
+            "_collections": ["Experiments"],
+            "$schema": "https://inspirehep.net/schemas/records/experiments.json",
```
drjova, line 276:
```suggestion
            "$schema":  url_for(
            "invenio_jsonschemas.get_schema",
            schema_path="records/experiments.json",
            _external=True,
        )
```

backend/tests/integration/submissions/test_submissions_views.py
```diff
@@ -2152,3 +2153,48 @@ def test_new_seminar_submission_with_invalid_lit_record(
...
+}
+
+
+@pytest.mark.parametrize(
+    "form_data,expected_record_data",
+    [
+        (deepcopy(EXPERIMENT_FORM_DATA), REQUIRED_EXPERIMENT_RECORD_DATA),
+    ],
+)
+def test_new_experiment_submission(
```
drjova, line 2177: we should have a test case with sending empty data to the API

backend/inspirehep/submissions/views.py
MJedr, line 276: I would put that in marshmallow, also if you'll have a proper serializer `self.load_data_from_request()` should prepare your data in a correct format for creating a record.

backend/inspirehep/submissions/views.py
```diff
@@ -259,12 +263,31 @@ def create_ticket(self, record, rt_template):
...
+            "legacy_name": request_data["legacy_name"],
+            "project_type": request_data["project_type"],
+        }
+        record = ExperimentsRecord(data=data).create(data)
+        db.session.commit()
+        experiment_serializer_v1.dump(record)
```
MJedr, line 282: why you need it?

backend/inspirehep/submissions/views.py
nooraangelva: hmm, but how do I add the ```$schema``` there? $ makes it a bit tricky or do you mean the whole dict?

MJedr: you can have a method for building this data dict (which won't use builder).

nooraangelva: Ah okay. Now I understand. Thank you

backend/inspirehep/submissions/views.py
nooraangelva: Don't I need to serialize the record or is it okay without serialization? On the other hand, I'm returning just the control number and status.

MJedr (approved):

PR #2524, opened 2022-04-27
refextract: create api for next

MJedr:
ref: cern-sis/issues-inspire#1


backend/inspirehep/refextract/utils.py
```diff
@@ -0,0 +1,86 @@
...
+    return title_dict
+
+
+def _write_journal_kb_dict_to_redis(redis):
+    journal_dict = create_journal_dict()
+    LOGGER.info("Writing journal KB dict to redis")
+    redis.hmset("refextract_journal_kb", journal_dict)
+    redis.expire("refextract_journal_kb", JOURNAL_DICT_REDIS_EXPIRATION_PERIOD)
+
+    return journal_dict
```
michamos, line 72: do you really need to store this in redis for performance? I'm a bit concerned about data consistency.

MJedr: On next we have a celery beat which is downloading it to the file every hour. With @drjova we discussed that we won't use that anymore and we'll store data in redis and recalculate with >= 1h

michamos: because the old `refextract` version that `next` is using needs to have the file on disk, so we need to generate it periodically. The new one on `hep` doesn't, and the DB query to generate the data is fast, so I don't believe it's worth it to reintroduce potential inconsistencies and complexity for very little performance gain:
```
In [1]: from inspirehep.matcher.utils import create_journal_dict                                                                                                

In [2]: %timeit create_journal_dict()                                                                                                                           
109 ms ± 3.81 ms per loop (mean ± std. dev. of 7 runs, 10 loops each)
```

drjova: we have already timed it, the point is if we have 1k papers coming from arxiv every morning we don't want to hammer db just for that.

michamos: ok, can we decrease the expiration to something like a few secs - 1 min max then and/or invalidate the cache on `JournalRecord` CRUD?

MJedr: Yep, that's a good idea to expire it there

michamos: mmmh, now I think of it it needs to be done in the DB commit receiver to avoid the race condition where the query happens between calling the CRUD method and the DB commit.

backend/inspirehep/refextract/views.py
```diff
@@ -0,0 +1,66 @@
...
+    try:
+        for publication_info in publication_infos:
+            extracted_publication_info = extract_journal_reference(
+                publication_info["pubinfo_freetext"],
+                override_kbs_files=journal_dict,
+            )
+            if not extracted_publication_info:
+                extracted_publication_info = {}
+            extracted_publication_infos.append(extracted_publication_info)
+    except TimeoutError:
```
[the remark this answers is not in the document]
drjova: let's treat all the errors the same way

backend/inspirehep/refextract/utils.py
drjova: How often do we update/add journals?

michamos: rarely, less than once a day I would say.

backend/inspirehep/refextract/views.py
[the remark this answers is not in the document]
michamos: one issue I see with this approach: refextract for papers with many refs will exceed the HTTP request timeout of `hep` (it can take 5 min sometimes). Also it will put more load on `hep` during the harvest (blocking web workers which are running refextract and are not available to serve user requests).

backend/inspirehep/refextract/utils.py
drjova: So we are talking about a use case that it might or not happen once a day. Then I agree, we can add it to the receiver and avoid the unnecessary queries.

drjova:
@MJedr I'm closing this PR

PR #2530, opened 2022-04-29
refextract: use a service

MJedr:


backend/inspirehep/config.py
```diff
@@ -284,3 +284,7 @@ def annotate(self, task):
 
 # fulltext
 ES_FULLTEXT_PIPELINE_NAME = "file_content"
+
+# refextract
+REFEXTRACT_SERVICE_URL = "https://example:5000"
+ENABLE_REFEXTRACT_SERVICE = False
```
drjova, line 290: It should be prefixed `FEATURE_FLAG_` to be consistent

drjova (approved):

PR #2532, opened 2022-05-03
root table: create api on hep

MJedr:
ref: cern-sis/issues-inspire#3

### DIFF
--- a/backend/inspirehep/records/models.py
+++ b/backend/inspirehep/records/models.py
@@ -7,11 +7,12 @@
 
 """Additional models used by InspireRecords."""
 import enum
+from datetime import datetime
 
 from invenio_db import db
 from invenio_records.models import RecordMetadata
 from sqlalchemy import Boolean, Date, Enum, Text
-from sqlalchemy.dialects.postgresql import ENUM
+from sqlalchemy.dialects.postgresql import ENUM, JSONB
 from sqlalchemy_utils import UUIDType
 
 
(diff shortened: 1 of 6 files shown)


backend/inspirehep/records/models.py
```diff
@@ -262,3 +263,49 @@ class StudentsAdvisors(db.Model):
...
+    created = db.Column(
+        db.DateTime(),
+        default=datetime.utcnow,
+        nullable=True,
+    )
+    updated = db.Column(
+        db.DateTime(),
+        default=datetime.utcnow,
+        nullable=True,
+    )
```
drjova, line 283: No need to define it again `from sqlalchemy_utils.models import Timestamp`

backend/inspirehep/records/views.py
```diff
@@ -82,6 +86,76 @@ def get(self, pid, record):
...
+
+    view_name = "workflows_record_sources"
+    decorators = [
+        login_required_with_roles([Roles.superuser.value, Roles.cataloger.value])
+    ]
+
+    def get(self):
+        record_uuid = request.json.get("record_uuid")
+        if not record_uuid:
+            return jsonify({"message": "record_uuid must be passed"}), 500
```
drjova, line 99:
```suggestion
            return jsonify({"message": "record_uuid must be passed"}), 400
```

backend/inspirehep/records/views.py
```diff
@@ -82,6 +86,76 @@ def get(self, pid, record):
...
+        query = WorkflowsRecordSources.query.filter_by(record_uuid=str(record_uuid))
```
drjova, line 100: isn't it already a string?

backend/inspirehep/records/views.py
```diff
@@ -82,6 +86,76 @@ def get(self, pid, record):
...
+        if source:
+            query = query.filter_by(source=source.lower())
+        results = query.all()
+        if not results:
+            return jsonify({"message": "Workflow source not found"}), 404
+        results_data = [
+            {
+                key: val
+                for key, val in result.__dict__.items()
+                if key in ["created", "json", "record_uuid", "source", "updated"]
```
drjova, line 111: why not just query with enities?

backend/inspirehep/records/models.py
MJedr: yes, I wanted to do that but `SQLAlchemy-Utils timestamp model does not have support for fractional seconds`

backend/inspirehep/records/views.py
```diff
@@ -82,6 +86,76 @@ def get(self, pid, record):
...
+            for result in results
+        ]
+        return jsonify({"workflow_sources": results_data}), 200
+
+    def post(self):
+        record_uuid = request.json.get("record_uuid")
+        source = request.json.get("source")
+        root_json = request.json.get("root_json")
+        if not all([record_uuid, source, root_json]):
+            return jsonify({"message": "all the arguments must be passed"}), 500
```
drjova, line 122: can we just use `webargs`?

backend/inspirehep/records/views.py
MJedr: nope, it's uuid

backend/inspirehep/records/views.py
MJedr: sure, I wanted to be consistent with other views ;)

backend/tests/integration/records/views/test_views.py
```diff
@@ -148,3 +151,117 @@ def test_returns_301_with_proper_location_when_record_redirected_in_chain(inspir
 
     assert response.status_code == 301
     assert response.location.split("/")[-1] == new_cn
+
+
+def test_literature_workflows_record_source(inspire_app):
```
drjova, line 156: just a suggestion, it would be more clear if we have separated tests for each case

drjova, line 156: just a suggestion, it would be more clear if we have separated tests for each case

drjova (approved):
LGTM just the suggestion is up to you.

backend/tests/integration/records/views/test_views.py
MJedr: I thought about it but then I'll have to put data into table all the time, but from the other hand it would be easier to know at which request it failed 🤔

drjova: it will be more clear what failed yes and modifying these tests in the future

PR #2534, opened 2022-05-05
ui: redirect to editor on experiment and institution submission

karolina-siemieniuk-morawska:
* now after submitting one of the two new wndow with recod-editor opens

* ref: cern-sis/issues-inspire#7 cern-sis/issues-inspire#6


drjova (approved):

ui/src/actions/submissions.js
```diff
@@ -59,6 +62,9 @@ export function submit(pidType, data) {
       const response = await http.post(`${SUBMISSIONS}/${pidType}`, { data });
       dispatch(submitSuccess(response.data));
       dispatch(push(`/submissions/${pidType}/new/success`));
+      if (REDIRECT_TO_EDITOR.includes(pidType)) {
+        window.open(`/editor/record/${pidType}/${response.data.control_number}`, "_blank")
```
michamos, line 66: why `_blank`? I don't think a new window is needed here, there is nothing left to do in the current window.

karolina-siemieniuk-morawska: It was guided by the behaviour of editing on other pages, like authors or literature, where editing opens a new page. Thought it will be consistent for the user to open this in new tab also. I can change it of course.

drjova: @karolina-siemieniuk-morawska could you please remove the `_blank`?

PR #2537, opened 2022-05-10
disambiguation: improve name reordering

MJedr:
ref: cern-sis/issues-inspire#34

### DIFF
--- a/backend/inspirehep/disambiguation/utils.py
+++ b/backend/inspirehep/disambiguation/utils.py
@@ -1,10 +1,10 @@
 import datetime
-import re
+from collections import Counter
 
 import structlog
 from flask import url_for
 from inspire_dojson.utils import get_record_ref
-from prometheus_client import Counter
+from prometheus_client import Counter as metrics_counter
 from unidecode import unidecode
 
 from inspirehep.records.api.authors import AuthorsRecord
@@ -12,7 +12,7 @@ from inspirehep.records.api.literature import LiteratureRecord
 
 LOGGER = structlog.getLogger()
 
-disambiguation_changed_signatures = Counter(
+disambiguation_changed_signatures = metrics_counter(
     "disambiguation_changed_signatures", "How many signatures were modified."
 )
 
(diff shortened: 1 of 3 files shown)


backend/tests/unit/disambiguation/test_utils.py
```diff
@@ -57,11 +57,19 @@ def test_update_author_names_doesnt_put_duplicate_name_variants():
...
         ("Qin, Qin", "Qin, Qin", "Qin, Qin"),
         ("Deiana, Allison Mccarn", "Deiana, Allison McCarn", "Deiana, Allison Mccarn"),
         ("Fayyazuddin", "Fayyazuddin", "Fayyazuddin"),
+        ("Garcia, Garcia Miguel", "Garcia Garcia, Miguel", "Garcia, Garcia Miguel"),
+        ("Garcia, Garcia Miguel", "Garcia, Miguel Garcia", "Garcia, Garcia Miguel"),
+        ("Garcia, Gabriel", "Garcia Marquez, Gabriel", "Garcia, Gabriel"),
+        ("Marquez, Gabriel", "Garcia Marquez, Gabriel", "Marquez, Gabriel"),
+        ("Davis, John", "Davis John, John", "Davis, John"),
+        ("Nascimento, J.-D do", "do Nascimento, José Dias", "do Nascimento, J.-D"),
+        ("Axelsen Viktor", "Axelsen, Viktor", "Axelsen, Viktor"),
```
michamos, line 72: could you add all the examples from the issue to make 100% sure they work?

backend/inspirehep/disambiguation/utils.py
```diff
@@ -129,28 +133,28 @@ def reorder_lit_author_names(lit_author_name, author_name):
...
             normalized_name in author_last_names_normalized
-            and name not in lit_author_last_names_reordered
-        ):
-            lit_author_last_names_reordered.append(name)
-            continue
+            and normalized_name not in author_first_names_normalized
+        )
+        no_more_possible_lit_author_first_names_in_author_firsts = (
+            author_first_names_normalized & lit_author_first_names_normalized
+        )
```
michamos, line 148: * I think technically you need a bag (i.e. `collections.Counter`) instead of a set to manage the case of people having more than one repetition like `John John, Smith John`). Not sure it matters in practice though, so this is probably ok, I'll leave it up to you (and you might want to add a comment with a disclaimer if you keep this).
* are you sure about the negation in the naming? if this is true, it looks like there are still possible matching first names.

michamos (approved):
looks great :medal_sports:, some minor comments

backend/tests/unit/disambiguation/test_utils.py
MJedr: I added all from your comment on chat, do you mean those in issue?

michamos: yes, I mean those given by Mathieu and written in the issue.

backend/inspirehep/disambiguation/utils.py
MJedr: I can't come up with any idea when if there might be a match (when using counter)

michamos: it's fine, no need to support super-corner cases, this is good enough :+1:

PR #2539, opened 2022-05-11
fulltext: handle errors in pipeline

MJedr:


backend/tests/unit/records/marshmallow/literature/test_es_schema.py
```diff
@@ -91,3 +91,39 @@ def test_es_schema_removes_fulltext_fields_from_ui_display(
...
+                "attachment": {"content": "this is a text"},
+                "error": {
+                    "message": "Fulltext indexing failed with message expected='>' actual='À' at offset 102005"
+                },
+            }
+        ]
+    }
+    serialized = schema.dump(entry_data).data
+    result = orjson.loads(schema.dumps(serialized).data)["_ui_display"]
+    assert '"error":' not in result
```
drjova, line 129:
```suggestion
    assert "error" not in result
```

backend/inspirehep/records/marshmallow/literature/ui.py
```diff
@@ -256,6 +256,8 @@ def get_documents_without_fulltext(self, data):
                 del document["attachment"]
             if "text" in document:
                 del document["text"]
+            if "error" in document:
+                del document["error"]
```
drjova, line 260: I would put it in the Base literature marshamallow to remove it from all API responses.

MJedr: so maybe let's use private field there, these fields are hidden in public schema then 
https://github.com/inspirehep/inspirehep/blob/243f4378683d4ba6357b3adec76d911f8f1422b3/backend/inspirehep/records/marshmallow/literature/base.py#L51

drjova: I think it does something else https://github.com/inspirehep/inspirehep/blob/243f4378683d4ba6357b3adec76d911f8f1422b3/backend/inspirehep/records/marshmallow/fields/non_hidden.py#L11-L15

drjova (commented):
LGTM, shouldn't we update the mappings as well though?

MJedr:
yes, good point 👍

drjova (approved):

PR #2540, opened 2022-05-12
ui: fix for pagination and banner issue

karolina-siemieniuk-morawska:
* change the usage of storage to support asynchronous values

* ref: cern-sis/issues-inspire#35 & cern-sis/issues-inspire#36

### DIFF
--- a/ui/src/StoreProviderOrNull.jsx
+++ b/ui/src/StoreProviderOrNull.jsx
@@ -1,0 +1,25 @@
+import { Provider } from 'react-redux';
+import React, { useState, useEffect } from 'react';
+import createStore, { REDUCERS_TO_PERSIST } from './store';
+
+import { reHydrateRootStateFromStorage } from './middlewares/statePersister';
+
+export const StoreProviderOrNull = ({ children }) => {
+  const [isBootstrapped, setBootstrapped] = useState(false);
+  const [data, setData] = useState(false);
+
+  const resolveStorage = async () => {
+    setData(await reHydrateRootStateFromStorage(REDUCERS_TO_PERSIST));
+    setBootstrapped(true);
+  };
+
+  useEffect(() => {
+    resolveStorage();
+  }, []);
+
+  if (!isBootstrapped) return null;
+
+  const store = createStore(data);
+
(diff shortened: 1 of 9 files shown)


ui/src/common/storage.js
```diff
@@ -6,59 +6,36 @@ localforage.config({
...
-    return (
-      e instanceof DOMException &&
-      (e.code === 22 ||
-        e.code === 1014 ||
-        e.name === 'QuotaExceededError' ||
-        e.name === 'NS_ERROR_DOM_QUOTA_REACHED') &&
-      (storage && storage.length !== 0)
-    );
-  }
+  localStorage.clear();
```
drjova, line 20: this will delete all of the user's localstorage 🙈

karolina-siemieniuk-morawska: Actually that's what I wanted to do - delete everything that is in the inspire domain's localStorage (I checked and it's separated by domain, so it will not mess with users'  other domains info stored in localStorage), because we don't need it anymore, we only use indexedDB now. And if we keep it, next time the user visits the page localStorage will be once again rewritten to LF and the new search history, page size and everything from the last session will be lost and replaced by legacy info from LS.

ui/src/StoreProviderOrNull.jsx
```diff
@@ -0,0 +1,27 @@
+import { Provider } from 'react-redux';
+import React, { useState, useEffect } from 'react';
+import createStore from './store';
+
+import { reHydrateRootStateFromStorage } from './middlewares/statePersister';
+
+export const StoreProviderOrNull = ({ children }) => {
```
drjova, line 7: the indentation of this file is wrong

ui/src/StoreProviderOrNull.jsx
```diff
@@ -0,0 +1,27 @@
...
+	const [isBootstrapped, setBootstrapped] = useState(false);
+	const [data, setData] = useState(false);
+	
+	const reducersToPersist = ['ui', 'user'];
```
drjova, line 11: we have this in other places as well, let's add it somewhere to use it in both places

PR #2545, opened 2022-05-17
handle gracefully records with no authors in disambiguation

nooraangelva:
Ref: https://github.com/cern-sis/issues-inspire/issues/21

### DIFF
--- a/backend/inspirehep/disambiguation/tasks.py
+++ b/backend/inspirehep/disambiguation/tasks.py
@@ -235,7 +235,7 @@ def disambiguate_authors(self, record_uuid, record_version_id):
             db.session.commit()
         return
     if not literature_in_previous_version:
-        authors = record["authors"]
+        authors = record.get("authors", [])
     else:
         authors = record.get_modified_authors()
     updated_authors = []
--- a/backend/tests/integration-async/disambiguation/test_disambiguation_tasks.py
+++ b/backend/tests/integration-async/disambiguation/test_disambiguation_tasks.py
@@ -16,12 +16,13 @@ from invenio_db import db
 from invenio_pidstore.models import PersistentIdentifier
 from invenio_search import current_search
 from redis import StrictRedis
-
+from flask_sqlalchemy import models_committed
 from inspirehep.disambiguation.tasks import disambiguate_signatures
 from inspirehep.records.api import AuthorsRecord, InspireRecord
 from inspirehep.records.api.literature import LiteratureRecord
 from inspirehep.search.api import AuthorsSearch, InspireSearch
-
+from inspirehep.disambiguation.api import author_disambiguation
+from inspirehep.records.receivers import index_after_commit
 
 def test_signature_linked_by_disambiguation_has_correct_facet_author_name(
     inspire_app, clean_celery_session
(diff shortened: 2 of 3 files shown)


backend/inspirehep/disambiguation/tasks.py
```diff
@@ -235,7 +235,7 @@ def disambiguate_authors(self, record_uuid, record_version_id):
             db.session.commit()
         return
     if not literature_in_previous_version:
-        authors = record["authors"]
+        authors = record.get("authors")
```
drjova, line 238:
```suggestion
        authors = record.get("authors", [])
```

.pre-commit-config.yaml
```diff
@@ -16,3 +16,7 @@ repos:
     hooks:
       - id: flake8
         args: ['--config=backend/setup.cfg']
+  - repo: https://github.com/pre-commit/pre-commit
+    rev: v2.7.1
+    hooks:
+      - id: validate_manifest
```
drjova, line 22: why do we need this change? 😅

.pre-commit-config.yaml
```diff
@@ -1,6 +1,6 @@
 repos:
   - repo: https://github.com/psf/black
-    rev: stable
+    rev: '22.3.0'
```
drjova, line 3: and this one? 😅

backend/inspirehep/disambiguation/tasks.py
drjova: Otherwise the lines after will fail because expects a list

.pre-commit-config.yaml
nooraangelva: I will try to commit without these changes, but I couldn't commit anything before I added these changes. some problems with pre-commit.

drjova: maybe you have to install it again?

.pre-commit-config.yaml
```diff
@@ -1,6 +1,6 @@
 repos:
   - repo: https://github.com/psf/black
-    rev: stable
+    rev: 'stable'
```
drjova, line 3: we don't need the `'` :)

backend/tests/integration-async/disambiguation/test_disambiguation_tasks.py
```diff
@@ -1418,3 +1418,37 @@ def assert_disambiguation_task():
...
+    retry_until_pass(assert_authors_records_exist_in_es)
+
+    literature_data = faker.record("lit", with_control_number=True)
+    literature_record = LiteratureRecord.create(literature_data)
+    db.session.commit()
+
+    def assert_disambiguation_task():
+        assert InspireSearch.get_record_data_from_es(literature_record)
+
+    retry_until_pass(assert_disambiguation_task, retry_interval=2)
```
drjova, line 1454: I would also add a test to explicitly run the `disambiguation` task manually and check that there is no failure without `authors`

backend/tests/integration-async/disambiguation/test_disambiguation_tasks.py
```diff
@@ -1418,3 +1418,16 @@ def assert_disambiguation_task():
...
+
+def test_disambiguation_when_empty_authors(inspire_app, clean_celery_session, enable_disambiguation):
+
+    literature_data = faker.record("lit", with_control_number=True)
+    literature_record = LiteratureRecord.create(literature_data)
+    db.session.commit()
+
+    def assert_disambiguation_task():
+        assert InspireSearch.get_record_data_from_es(literature_record)
+        assert disambiguate_authors(LiteratureRecord.get_uuid_from_pid_value(literature_record['control_number']), '1') is None
```
MJedr, line 1431: It's wrong, you shouldn't trigger this task manually. It is triggered from receiver on db.session.commit() (so it means that after committing it will be executed automatically).

drjova: sorry if I wasn't clear, the idea was to create a separate test to check the task itself and have also the integration test.

nooraangelva: Ah of course. Sorry I misunderstood. 👍  Thanks for clearing it up.

PR #2562, opened 2022-06-01
records: refactor import article view

drjova:


backend/inspirehep/records/api/literature.py
```diff
@@ -659,7 +657,7 @@ def merge_article_with_crossref_data(article):
 
     try:
         crossref_data = import_doi(doi)
-    except (ImportConnectionError, ImportParsingError):
+    except (requests.exceptions.RequestException, ImportParsingError):
```
MJedr, line 660: why we can't import it as `RequestException`?

MJedr (approved):

backend/inspirehep/records/api/literature.py
drjova: indeed just because we are already importing the whole module

drjova:
@michamos is it what we discussed?

michamos (approved):
looks good :+1:

PR #2563, opened 2022-06-01
ui: new account menu

karolina-siemieniuk-morawska:
* add new submenu to header for logged in users

* ref: cern-sis/issues-inspire#43


ui/src/common/layouts/Header/HeaderMenu.jsx
```diff
@@ -19,10 +19,13 @@ import LinkLikeButton from '../../components/LinkLikeButton';
 import './HeaderMenu.scss';
 import { PAPER_SEARCH_URL, HELP_BLOG_URL } from '../../constants';
 import DisplayGuideButtonContainer from '../../containers/DisplayGuideButtonContainer';
+import { CLAIMING_DISABLED_INFO } from '../../../authors/components/AssignNoProfileAction';
 
 class HeaderMenu extends Component {
   render() {
-    const { loggedIn, onLogoutClick, isCatalogerLoggedIn } = this.props;
+    const { loggedIn, onLogoutClick, isCatalogerLoggedIn, orcidId } = this.props;
+    const USER_PROFILE_URL = `https://inspirehep.net/legacy/orcid/${orcidId}`;
```
drjova, line 27: we shouldn't have hardcoded links, the code is been deployed in multiple environments. I just realize that it will work only in a new tab. What can we do instead is the following:

https://github.com/inspirehep/inspirehep/blob/master/backend/inspirehep/accounts/handlers.py#L23-L31

```python
# on the top
from inspirehep.records.utils import get_pid_for_pid

# and change the payload to include `profile_control_number`
    payload = {
        "data": {
            "email": current_user.email,
            "roles": [role.name for role in current_user.roles],
            "orcid": orcid,
            "profile_control_number": get_pid_for_pid('orcid', orcid, 'recid'),
            "allow_orcid_push": orcid_account_extra_data.get("allow_push"),
        }
    }
```

ui/src/common/layouts/Header/HeaderMenu.scss
```diff
@@ -14,4 +14,9 @@
       text-decoration: underline;
     }
   }
+  .ant-btn.ant-btn-background-ghost {
+    padding: 0;
+    color: rgba(255, 255, 255, 0.5);
+    border-color: transparent;
+  }
```
drjova, line 21: why do we need custom css?

drjova (approved):

PR #2566, opened 2022-06-03
fulltext indexer: handle deleted records

MJedr:
* improve indexer logging

### DIFF
--- a/backend/inspirehep/indexer/base.py
+++ b/backend/inspirehep/indexer/base.py
@@ -6,7 +6,7 @@
 # the terms of the MIT License; see LICENSE file for more details.
 
 import structlog
-from elasticsearch import RequestError, TransportError
+from elasticsearch import ConflictError, NotFoundError, RequestError, TransportError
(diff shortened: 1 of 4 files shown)


backend/inspirehep/indexer/base.py
```diff
@@ -142,6 +143,32 @@ def bulk_action(self, record_uuid):
                 "Kombu is not able to process response!", uuid=str(record_uuid)
             )
 
+    def _get_indexing_arguments(slef):
```
drjova, line 146:
```suggestion
    def _get_indexing_arguments(self):
```

backend/inspirehep/indexer/base.py
```diff
@@ -142,6 +143,32 @@ def bulk_action(self, record_uuid):
...
+        """Returns custom arguments for record indexing"""
+        pass
+
+    def index_inspire_record(self, record, force_delete=None, record_version=None):
```
drjova, line 150: can we just add it to the `index`?

drjova (approved):

PR #2567, opened 2022-06-03
Fix: Ed is stripped in LaTeX formats

nooraangelva:
Ref: https://github.com/cern-sis/issues-inspire/issues/53

### DIFF
--- a/backend/tests/integration/records/serializers/test_latex.py
+++ b/backend/tests/integration/records/serializers/test_latex.py
@@ -7,7 +7,7 @@
 
 from freezegun import freeze_time
 from helpers.utils import create_record, create_record_factory
-
+import pytest
 
 @freeze_time("1994-12-19")
 def test_latex_eu(inspire_app):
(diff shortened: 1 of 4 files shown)


drjova:
@nooraangelva thank you for your PR, I believe it would be cleaner if we do the following:

* Parametririze https://github.com/inspirehep/inspire-utils/blob/54cd0488d511fa2f1090efefcd437d9ea3a72ab4/inspire_utils/name.py#L44-L59 to include or not the titles

* On https://github.com/inspirehep/inspire-utils/blob/54cd0488d511fa2f1090efefcd437d9ea3a72ab4/inspire_utils/name.py#L477-L495 add parametririze to initialize ParsedName with `contstant` without titles, we can pass it in the init function https://github.com/inspirehep/inspire-utils/blob/54cd0488d511fa2f1090efefcd437d9ea3a72ab4/inspire_utils/name.py#L80-L81

* On hep just call `format_names(name, without_titles=True)` for that case and that's all 🎉

backend/tests/integration/records/serializers/test_latex.py
```diff
@@ -542,3 +542,37 @@ def test_latex_encodes_non_latex_chars(inspire_app):
...
+
+
+@freeze_time("2020-09-11")
+@pytest.mark.parametrize(
+    "input_author,expected_name",
+    [("Lieber, Ed", "E.~Lieber,"),
+     ('Lieber, Ed Viktor', "E.~V.~Lieber,"),
+     ('Lieber, Ed Jr.', "E.~Lieber, Jr.,"),
+     ('Lieber, Ed Victor Jr.', "E.~V.~Lieber, Jr.,"),
+     ],
```
drjova, line 554: @michamos are the outputs as expected?

michamos: Yes, looks good 👍

drjova (approved):
LGTM 🚀

PR #2568, opened 2022-06-09
ui: allow claiming mix of claimed and unclaimed papers

karolina-siemieniuk-morawska:
* when selecting claimed and unclaimed papers by checking individual checkboxes, the behaviour should be the same as when using select all checkbox (claiming mixed papers is then allowed)

* ref: cern-sis/issues-inspire#57

### DIFF
--- a/ui/src/authors/components/PublicationsSelect.jsx
+++ b/ui/src/authors/components/PublicationsSelect.jsx
@@ -19,10 +19,12 @@ function PublicationsSelect({
     }
     if (claimed) {
       onSelectClaimedPapers(event);
-    } else if (!claimed && canClaim) {
+    }
+    if (!claimed) {
       onSelectUnclaimedPapers(event);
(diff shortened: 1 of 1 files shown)


ui/src/authors/components/PublicationsSelect.jsx
```diff
@@ -19,10 +19,12 @@ function PublicationsSelect({
     }
     if (claimed) {
       onSelectClaimedPapers(event);
-    } else if (!claimed && canClaim) {
+    }
+    if (!claimed) {
```
drjova, line 23: are you sure we should remove the `canClaim`?

karolina-siemieniuk-morawska: after talking to @MJedr I know I can't do it like this. I'll convert the PR into draft and try to do it another way

PR #2569, opened 2022-06-09
ui: fix for always plural version of label and tooltip

karolina-siemieniuk-morawska:
* on single paper claiming change plural version label and tooltip to singular

* ref: cern-sis/issues-inspire#55

### DIFF
--- a/ui/src/authors/components/AssignOwnProfileAction.jsx
+++ b/ui/src/authors/components/AssignOwnProfileAction.jsx
@@ -13,6 +13,7 @@ function AssignOwnProfileAction({
   disabled,
   disabledAssignAction,
   numberOfSelected,
+  claimingTooltip,
 }) {
   const currentAuthorId = Number(useParams().id);
   const onSelfAssign = useCallback(() => {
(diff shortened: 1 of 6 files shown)


ui/src/authors/components/AssignOwnProfileAction.jsx
```diff
@@ -81,6 +82,12 @@ AssignOwnProfileAction.propTypes = {
   disabled: PropTypes.bool,
   disabledAssignAction: PropTypes.bool,
   numberOfSelected: PropTypes.number,
+  claimingTooltip: PropTypes.string,
+};
+
+AssignOwnProfileAction.defaultProps = {
+  claimingTooltip: 'This paper is already claimed',
+  numberOfSelected: 1,
```
drjova, line 90: why do we need `numberOfSelected` 🤔 we are not use it anywhere?

karolina-siemieniuk-morawska: We use it in lines 64 and 71. It's set to default 1 to work correctly with [AssignOneOwnProfileContainer](https://github.com/inspirehep/inspirehep/blob/master/ui/src/authors/containers/AssignOneOwnProfileContainer.jsx) where we only have one selected paper. And in [AssignAllOwnProfileActionContainer](https://github.com/inspirehep/inspirehep/blob/master/ui/src/authors/containers/AssignAllOwnProfileActionContainer.jsx) I pass it as props because there there can be more that one paper selected :)

drjova (approved):

PR #2582, opened 2022-06-28
move disambiguation to record create & update

MJedr:


backend/inspirehep/disambiguation/utils.py
```diff
@@ -27,6 +26,8 @@ def link_signature_to_author(signature_data, author_control_number):
     Returns:
         dict: The signature data from the publication with the linked author.
     """
+    from inspirehep.records.api.literature import LiteratureRecord
```
drjova, line 29: why we have to move this inside?

MJedr: circular import

backend/inspirehep/records/api/literature.py
```diff
@@ -133,6 +135,11 @@ def create(
         else:
             push_to_orcid(record)
             push_to_hal(record)
+        if (
+            not disable_disambiguation
+            and current_app.config["FEATURE_FLAG_ENABLE_AUTHOR_DISAMBIGUATION"]
+        ):
+            disambiguate_authors.delay(str(record.id))
```
drjova, line 142: not sure if it will work, the record is not created yet

MJedr: so here you have nested session: https://github.com/inspirehep/inspirehep/blob/2980d09ea36b03613323eb668eca9baadc83b8d9/backend/inspirehep/records/api/literature.py#L121-L125
and here from `InspireRecord` you add the record to the db session https://github.com/inspirehep/inspirehep/blob/2980d09ea36b03613323eb668eca9baadc83b8d9/backend/inspirehep/records/api/base.py#L272

PR #2583, opened 2022-06-28
assign: add an endpoint for checking names compatibility

MJedr:
ref: cern-sis/issues-inspire#68


backend/inspirehep/assign/utils.py
```diff
@@ -41,27 +43,68 @@ def update_author_bai(to_author_bai, lit_author):
...
             .with_entities(RecordMetadata.json)
             .filter(
-                PersistentIdentifier.pid_value == str(data["control_number"]),
+                PersistentIdentifier.pid_value == str(control_number),
                 PersistentIdentifier.pid_type == "lit",
                 RecordMetadata.id == PersistentIdentifier.object_uuid,
             )
             .one()
         )
+        return lit_record[0]
```
drjova, line 58: why we even have a tuple here? `.one()` should be enough

backend/inspirehep/assign/utils.py
```diff
@@ -71,5 +114,46 @@ def can_claim(data, author_profile_recid):
             for author_name in current_author_profile.get("name.name_variants", [])
         ]
     )
-    lit_author = get_author_by_recid(lit_record[0], int(author_profile_recid))
+    lit_author = get_author_by_recid(lit_record, int(author_profile_recid))
     return author_names & set([lit_author.get("full_name").split(",")[0]])
+
+
+def check_author_compability_with_lit_authors(literature_control_number):
```
drjova, line 121: can we split the logic?

backend/inspirehep/assign/utils.py
MJedr: when you query with entities `one` returns a tuple?

backend/inspirehep/assign/utils.py
MJedr: we can, although imo we have 3 simple checks for 3 cases there

drjova (approved):

PR #2593, opened 2022-07-11
add book publication information to LaTeX formats

nooraangelva:
Ref: https://github.com/cern-sis/issues-inspire/issues/58

### DIFF
--- a/backend/inspirehep/records/marshmallow/literature/latex.py
+++ b/backend/inspirehep/records/marshmallow/literature/latex.py
@@ -8,12 +8,16 @@
 
 import datetime
 
+from idutils import normalize_isbn
 from inspire_utils.name import format_name
 from inspire_utils.record import get_value
+from isbn import ISBNError
 from marshmallow import fields, missing
+
 from ..base import BaseSchema
 from .bibtex import BibTexCommonSchema
 from .utils import latex_encode
+
 
 class LatexSchema(BaseSchema):
     arxiv_eprints = fields.Raw()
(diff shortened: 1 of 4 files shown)


backend/inspirehep/records/marshmallow/literature/latex.py
```diff
@@ -113,3 +122,29 @@ def get_dois(self, data):
         for doi_data in dois:
             doi_data["value"] = latex_encode(doi_data.get("value"))
         return dois or None
+
+    def get_book_publication_info(self, data):
+        if "book" in get_value(data, "document_type", []):
```
MJedr, line 127: better check `if not` - then you can instantly return and remove one indentation level. Also, no need for default get value `[]`

backend/inspirehep/records/marshmallow/literature/latex.py
```diff
@@ -113,3 +122,29 @@ def get_dois(self, data):
...
+            elif publisher:
+                return publisher
+        return missing
+
+    def get_isbn(self, data):
+        def hyphenate_if_possible(no_hyphens):
+            try:
+                return normalize_isbn(no_hyphens)
+            except ISBNError:
+                return no_hyphens
```
MJedr, line 144: why inner function?

backend/inspirehep/records/marshmallow/literature/latex.py
```diff
@@ -113,3 +122,29 @@ def get_dois(self, data):
...
+
+        if "book" in get_value(data, "document_type", []):
```
MJedr, line 146: again, I would return here if it's not in `document_type`. Also, if I am correct only books have `isbns` - if that's the case you can get rid of this line and just check `if isbns`

backend/inspirehep/records/marshmallow/literature/latex.py
nooraangelva: At least when I test the latex_schema the ```[]``` default value is necessary.
```
if "book" not in get_value(data, "document_type"):
E       TypeError: argument of type 'NoneType' is not iterable
```

MJedr: yes, you are right, sorry

MJedr (approved):

PR #2594, opened 2022-07-11
changes for m1

ErnestaP:
Changes for m1 with @MJedr  and @karolina-siemieniuk-morawska  help:
* updated bundle version in UI from ^0.17.0 to 0.18.1
* removed watchmedo, because is not supported on M1
* pulling the ES image separately, because of ES error: `elasticsearch.exceptions.NotFoundError: NotFoundError(404, 'index_not_found_exception', 'no such index [records-hep]', records-hep, index_or_alias)`
* updated ReadMe
* changed the node image

### DIFF
--- a/docker-compose.services.yml
+++ b/docker-compose.services.yml
@@ -50,7 +50,8 @@ services:
   es:
     build:
       context: docker/opensearch
-    restart: 'always'
+      dockerfile: 'opensearch.Dockerfile'
+    restart: always
     environment:
       - cluster.name=opensearch-cluster
       - node.name=opensearch-node1
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -76,8 +76,8 @@ services:
     extends:
       file: docker-compose.services.yml
       service: app
-    entrypoint: watchmedo auto-restart -d . -p '*.py'
-    command: -- celery worker -E -A inspirehep.celery -l INFO --purge --queues celery,migrator,indexer_task,matcher,assign
+    entrypoint: celery
+    command: worker -E -A inspirehep.celery -l INFO --purge --queues celery,migrator,indexer_task,matcher,assign
     depends_on:
       - db
       - es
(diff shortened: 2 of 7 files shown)


docker-compose.services.yml
```diff
@@ -50,7 +50,8 @@ services:
   es:
     build:
       context: docker/opensearch
-    restart: 'always'
+      dockerfile: 'opensearch.Dockerfile'
+    restart: always
```
MJedr, line 54: hm it's weird that you need to add this - `dockerfile` is an optional arg for `build` and I think that since Dockerfile is the default image that docker is searching in the directory specified in `context` it should work (it worked for all the intel users and me on m1). But if it doesn't break anything and makes ES work for you is ok, but then instead of adding a new dockerfile rename the old one :)

docker-compose.yml
```diff
@@ -76,8 +76,8 @@ services:
     extends:
       file: docker-compose.services.yml
       service: app
-    entrypoint: watchmedo auto-restart -d . -p '*.py'
-    command: -- celery worker -E -A inspirehep.celery -l INFO --purge --queues celery,migrator,indexer_task,matcher,assign
+    entrypoint: celery
```
MJedr, line 79: I think that in that case we can just skip entrypoint and leave the command as it was

opensearch/opensearch.Dockerfile
```diff
@@ -0,0 +1,2 @@
+FROM docker.elastic.co/elasticsearch/elasticsearch-oss:7.10.2
+RUN bin/elasticsearch-plugin install -b ingest-attachment analysis-icu
```
MJedr, line 2: as I mentioned, if that's needed better rename old dockerfile

record-editor/Dockerfile
```diff
@@ -1,5 +1,5 @@
-FROM node:10.14.0-jessie as build-stage
-
+# FROM node:10.14.0-jessie as build-stage
+FROM node:10.14.0-stretch as build-stage
```
MJedr, line 2: If the old image doesn't work just replace it instead of commenting ;)

PR #2596, opened 2022-07-12
bug in cv format with fullname

nooraangelva:
Ref: https://github.com/cern-sis/issues-inspire/issues/38
error: https://sentry.sis.cern.ch/inspire/hep/issues/36765/?query=is%3Aunresolved

### DIFF
--- a/backend/tests/integration/records/serializers/test_cv.py
+++ b/backend/tests/integration/records/serializers/test_cv.py
@@ -4,9 +4,13 @@
 #
 # inspirehep is free software; you can redistribute it and/or modify it under
 # the terms of the MIT License; see LICENSE file for more details.
+import mock
+import orjson
 from flask_sqlalchemy import models_committed
 from helpers.utils import create_record
 
+from inspirehep.records.api import LiteratureRecord
+from inspirehep.records.marshmallow.literature.cv import CVSchema
 from inspirehep.records.receivers import index_after_commit
 from inspirehep.search.api import LiteratureSearch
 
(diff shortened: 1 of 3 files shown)


backend/tests/integration/records/serializers/data/2108903.json
```diff
@@ -0,0 +1,40 @@
...
+        "$ref": "https://inspirehep.net/api/authors/2108904"
+      },
+      "full_name": "Brown, Michael",
+      "inspire_roles": [
+        "supervisor"
+      ],
+      "signature_block": "BRANm"
+    }
+  ]
+}
```
drjova, line 40: this record has `full_name` on both authors, hence we are not testing the actual corner case.

.pre-commit-config.yaml
```diff
@@ -1,18 +1,18 @@
 repos:
   - repo: https://github.com/psf/black
-    rev: stable
+    rev: 22.6.0
```
drjova, line 3: why we need to change the versions? 🤔

backend/tests/unit/records/marshmallow/literature/common/test_reference_item.py
```diff
@@ -241,3 +241,39 @@ def test_returns_arxiv_eprints_from_the_resolved_record(
...
+        "urls": [
+            {"value": "http://www.claymath.org/prize-problems"},
+            {"value": "http://www.arthurjaffe.com"},
+        ],
+        "collaborations": [{"value": "CMS"}],
+        "collaborations_with_suffix": [{"value": "ATLAS Team"}],
+    }
+    record = LiteratureRecord(dump)
+    result = schema.dumps(record).data
+    assert expected == orjson.loads(result)
```
drjova, line 279: hmm I'm very confused, this test is not related to the code changes, you changed something in the CVSchema :/ Do I miss something?

.pre-commit-config.yaml
nooraangelva: Sorry, I forgot to change it back. For some reason, if I make changes to python scripts Github doesn't let me commit until I have updated the pre-commit. Usually, afterwards I revert the pre-commit back.

backend/tests/unit/records/marshmallow/literature/common/test_reference_item.py
nooraangelva: Thank you for pointing that out. I don't know how I managed to get so confused with the codes that I wrote the test in the wrong place.

backend/tests/integration/records/serializers/test_cv.py
```diff
@@ -620,3 +624,72 @@ def test_literature_detail_cv_link_alias_format(inspire_app):
...
+                    {
+                        "value": "LIGO - California Institute of Technology - Pasadena - CA 91125 - USA"
+                    }
+                ],
+            }
+        ],
+        "arxiv_eprints": [{"value": "1606.09129", "categories": ["hep"]}],
+    }
+    record = LiteratureRecord(dump)
+    result = schema.dumps(record).data
```
drjova, line 693: I would test it like https://github.com/inspirehep/inspirehep/blob/656f88b0742ac384aa5e080019ec8fca8e91b59d/backend/tests/integration/records/serializers/test_cv.py#L18-L36

PR #2600, opened 2022-07-15
disambiguation: don't create stub author when already existing

MJedr:
ref: cern-sis/issues-inspire#56


backend/tests/integration-async/conftest.py
```diff
@@ -32,7 +32,7 @@ def app():
     app_config = {}
     app_config["DEBUG"] = False
     app_config["CELERY_CACHE_BACKEND"] = "memory"
-    app_config["CELERY_TASK_ALWAYS_EAGER"] = False
+    app_config["CELERY_TASK_ALWAYS_EAGER"] = True
```
drjova, line 35: what's the point to have async tests then :)

PR #2605, opened 2022-07-21
literature: addition of file size limit

drjova:
* ref: cern-sis/issues-inspire#82


backend/inspirehep/records/utils.py
```diff
@@ -89,6 +89,18 @@ def download_file_from_url(url):
...
+        file_size_limit = current_app.config["FILES_SIZE_LIMIT"]
+        content_length = request.headers.get("content-length")
+
+        if content_length and isinstance(content_length, str):
+            content_length = int(content_length)
+
+        if check_file_size and content_length and content_length > file_size_limit:
+
+            raise FileSizeExceededError(
+                f"File size larger {content_length} is larger than the limit {file_size_limit}."
```
karolina-siemieniuk-morawska, line 102: Not very relevant comment, but shouldn't it be `f"File size {content_length} is larger than the limit {file_size_limit}."` ? 🤓

PR #2606, opened 2022-07-25
Retry disambiguation task for authors PIDAlreadyExistsError

nooraangelva:
ref: https://github.com/cern-sis/issues-inspire/issues/78

### DIFF
--- a/backend/inspirehep/disambiguation/tasks.py
+++ b/backend/inspirehep/disambiguation/tasks.py
@@ -21,6 +21,7 @@ from inspirehep.matcher.validators import (
     affiliations_validator,
     collaboration_validator,
 )
+from inspirehep.pidstore.errors import PIDAlreadyExistsError
 from inspirehep.records.api import InspireRecord
 from inspirehep.records.api.authors import AuthorsRecord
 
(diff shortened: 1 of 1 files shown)


backend/tests/integration-async/disambiguation/test_disambiguation_tasks.py
```diff
@@ -1608,3 +1613,60 @@ def assert_authors_records_exist_in_es():
     db.session.commit()
 
     assert mock_disambiguate_authors.call_count == 0
+
+
+@mock.patch(
+    "inspirehep.disambiguation.tasks.disambiguate_authors.retry", side_effect=Retry
+)
+def test_disambiguate_retry_when_PIDAlreadyExistsError(
```
drjova, line 1621: @nooraangelva  the easiest way to check the retry would be to mock the [_disambiguate_authors](https://github.com/inspirehep/inspirehep/blob/5f0302dd06319538c2a3ad99802e8f1f8cc6b755/backend/inspirehep/disambiguation/tasks.py#L221) to return  `PIDAlreadyExistsError`. This functions is raising this exception because we are trying to create new `BAIS` there.

drjova: the mock function is wrong :) it should be with `_`

nooraangelva: I tried with that locally and it didn't give me the error or get the task to retry. I presume it's because the error originates really from [here](https://github.com/inspirehep/inspirehep/blob/master/backend/inspirehep/pidstore/minters/base.py#L69-L70). It also looks like my test is not "going through" ```disambiguate_authors()``` at all.

MJedr (approved):

PR #2607, opened 2022-07-26
disambiguation: improve logging & retry

MJedr:
* don't run disambiguation for deleted records instead of handling it with `NoResultFound` exception


nooraangelva:
@MJedr Can there be tests for this or is it simpler to test in QA?

MJedr:
@nooraangelva added some tests

nooraangelva (approved):
Thanks for the tests. Looks good to me 👍

PR #2611, opened 2022-08-02
disambiguation: assign to stub author when ambiguous match

MJedr:
ref: cern-sis/issues-inspire#97

### DIFF
--- a/backend/inspirehep/disambiguation/tasks.py
+++ b/backend/inspirehep/disambiguation/tasks.py
@@ -102,15 +102,14 @@ def get_reference_and_bai_if_unambiguous_match(matched_refs_ids):
         }
 
 
-def get_reference_and_bai_if_unambiguous_literature_author_match(matched_records):
+def get_reference_and_ids_for_literature_author_match(matched_records):
     matched_refs_ids = {
         get_value(
             matched_author, "inner_hits.authors.hits.hits[0]._source.record.$ref"
         ): get_value(matched_author, "inner_hits.authors.hits.hits[0]._source.ids", [])
         for matched_author in matched_records
     }
-    matched_author_data = get_reference_and_bai_if_unambiguous_match(matched_refs_ids)
-    return matched_author_data
+    return matched_refs_ids
 
 
 def _filter_out_initials(name):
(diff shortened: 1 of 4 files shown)


backend/inspirehep/disambiguation/tasks.py
```diff
@@ -93,23 +93,51 @@ def match_literature_author_with_config(author_data, matcher_config):
...
+    author_bais = get_values_for_schema(author_ids, "INSPIRE BAI")
+    return {
+        "author_reference": author_reference,
+        "author_bai": author_bais[0] if author_bais else None,
+    }
+
+
+def get_reference_and_bai_for_stub_author_if_ambiguous_match(matched_refs_ids):
+    if len(matched_refs_ids) < 1:
+        return
```
michamos, line 108: do you need this? won't the `for` loop afterwards iterate on an empty dict and return `None` implicitly in this case anyway?

backend/inspirehep/disambiguation/tasks.py
```diff
@@ -218,15 +252,25 @@ def _get_not_disambiguated_authors(authors):
...
     updated_authors = []
     for author in authors_to_disambiguate:
         if author.get("curated_relation"):
             continue
         assigned_author_recid = None
-        matched_author_data = match_author(author)
+        matched_author_data = match_author(author, record)
         if not matched_author_data:
-            matched_author_data = match_literature_author(author, record)
+            matched_author_data = match_author(author, record, True)
```
michamos, line 273: wouldn't it be better to avoid doing all the queries on ES twice, and instead remember all the encountered author record IDs in the first pass, to later check for `stub`ness in case of ambiguity? I think it would also make the logic more clear. Now it's pretty hard to understand what's happening.

backend/inspirehep/disambiguation/tasks.py
```diff
@@ -218,15 +229,43 @@ def _get_not_disambiguated_authors(authors):
             yield author
 
 
+def find_stub_author_in_matched_authors(matched_authors_references):
+    for author_reference in matched_authors_references:
+        author = AuthorsRecord.get_record_by_pid_value(
```
michamos, line 234: maybe it woould be better to use `get_records_by_pids` to avoid doing N queries here, or even push the stub filtering to the DB query?

backend/inspirehep/disambiguation/tasks.py
```diff
@@ -218,15 +229,43 @@ def _get_not_disambiguated_authors(authors):
...
+                matched_authors_ids_literature_author_match,
+            ) = match_literature_author(author, record)
         if not matched_author_data:
-            matched_author_data = match_literature_author(author, record)
+            matched_authors_references = set(
+                [
+                    *matched_authors_ids_author_match,
+                    *matched_authors_ids_literature_author_match,
+                ]
+            )
```
michamos, line 265:
```suggestion
            matched_authors_references = {
                    *matched_authors_ids_author_match,
                    *matched_authors_ids_literature_author_match,
            }
```

michamos (approved):
minor comments, but looks good :+1:

backend/inspirehep/disambiguation/tasks.py
MJedr: Yes, with stub filtering it might be worth doing one db query

nooraangelva (approved):
LGTM 👍

PR #2612, opened 2022-08-02
assign: fix name compability check when no match found

MJedr:
ref: cern-sis/issues-inspire#100


nooraangelva (approved):
Looks good to me 👍

PR #2616, opened 2022-08-04
ui: move claim button to right on literature

karolina-siemieniuk-morawska:
* ref: cern-sis/issues-inspire#98


ui/src/literature/containers/DetailPageContainer/DetailPageContainer.jsx
```diff
@@ -149,9 +149,9 @@ function DetailPage({
                         text="links"
                         trackerEventId="LiteratureFileLink"
                       />
-                    )}
-                    {displayClaimingButton()}
+                    )}  
                     <CiteModalActionContainer recordId={controlNumber} />
+                    {displayClaimingButton()}
```
MJedr, line 154: it doesn't require updating the snapshot?

karolina-siemieniuk-morawska: there is no test for this component 🙈

MJedr: in cypress I mean

karolina-siemieniuk-morawska: In cypress the snapshot is ok, but I added unit test :)

MJedr: The tests probably won't fail cause the difference is pretty small, but maybe it would be worth having an updated view for this component?

karolina-siemieniuk-morawska: You are right - snapshots updated 🫡

MJedr (approved):

PR #2619, opened 2022-08-04
ui: fix for broken ui when paper doesn't have authors

karolina-siemieniuk-morawska:
* ref: cern-sis/issues-inspire#107

### DIFF
--- a/ui/src/literature/components/NoAuthorsClaimingButton.tsx
+++ b/ui/src/literature/components/NoAuthorsClaimingButton.tsx
@@ -1,0 +1,22 @@
+import React from 'react';
+import { FileDoneOutlined } from '@ant-design/icons';
+import { Button, Tooltip } from 'antd';
+
+import DropdownMenu from '../../common/components/DropdownMenu';
+import IconText from '../../common/components/IconText';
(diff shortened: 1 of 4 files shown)


ui/src/authors/components/AssignNoProfileAction.jsx
```diff
@@ -13,14 +13,14 @@ export const CLAIMING_DISABLED_INFO = (
   </p>
 );
 
-function AssignNoProfileAction() {
+function AssignNoProfileAction(tooltip) {
```
MJedr, line 16: Hmm does this problem show only in the no profile case? What about assigning when you have a profile and a paper doesn't have authors?

Also IMO this is not the best naming of the prop - you want to pass the tooltip title there and the name `tooltip` is a bit misleading. If the tooltip title is fixed maybe you can add a variable `TOOLTIP_TITLE` and use it when a boolean prop will be passed to this component?

karolina-siemieniuk-morawska: I reasued the AssignNoProfileAction component to not create another one almost identical to it, with just different tooltip, but you are right that it looks misleading. I will create another component for this no authors case.

PR #2626, opened 2022-08-18
Redirection of merged records

nooraangelva:
ref: https://github.com/cern-sis/issues-inspire/issues/103

### DIFF
--- a/backend/inspirehep/records/tasks.py
+++ b/backend/inspirehep/records/tasks.py
@@ -21,6 +21,9 @@ from inspirehep.pidstore.api import PidStoreBase
 from inspirehep.records.api import InspireRecord, LiteratureRecord
 from inspirehep.search.api import InspireSearch
 from inspirehep.utils import flatten_list
+from inspire_utils.dedupers import dedupe_list_of_dicts, dedupe_list
+from dict_deep import deep_set
+
 
 LOGGER = structlog.getLogger()
 
@@ -96,10 +99,13 @@ def update_references_pointing_to_merged_record(
             referenced_records_in_path = flatten_list(
                 get_value(matched_inspire_record, path[: -len(".$ref")], [])
             )
+
             for referenced_record in referenced_records_in_path:
-                update_reference_if_reference_uri_matches(
-                    referenced_record, merged_record_uri, new_record_uri
-                )
+                update_reference_if_reference_uri_matches(referenced_record, merged_record_uri, new_record_uri)
+            deduped_matched_inspire_record = remove_duplicate_refs_from_record(matched_inspire_record, path)
+            
+            if deduped_matched_inspire_record:
+                matched_inspire_record = deduped_matched_inspire_record
             matched_inspire_record.update(dict(matched_inspire_record))
             LOGGER.info(
                 "Updated reference for record", uuid=str(matched_inspire_record.id)
@@ -135,3 +141,15 @@ def regenerate_author_records_table_entries(uuids_to_regenerate):
         record.update_authors_records_table()
         record.update_self_citations()
         db.session.commit()
+
+
+def remove_duplicate_refs_from_record(matched_inspire_record, path):
+    references_path = ".".join(path.split(".")[:-2])
+    references = flatten_list(get_value(matched_inspire_record, references_path, []))
+
+    deduped_references = dedupe_list_of_dicts(references)
+    if len(references) == len(deduped_references):
+        return None
+
+    deep_set(matched_inspire_record, references_path, deduped_references)
+    return matched_inspire_record
--- a/backend/poetry.lock
+++ b/backend/poetry.lock
@@ -571,6 +571,14 @@ ordered-set = ">=3.1.1"
 murmur = ["mmh3"]
 
 [[package]]
+name = "dict-deep"
+version = "4.1.2"
+description = "Very simple deep_set and deep_get functions to access nested dicts (or any object) using 'dotted strings' as key."
+category = "main"
+optional = false
+python-versions = ">=3.5,<4.0"
+
+[[package]]
 name = "dictdiffer"
 version = "0.8.1"
 description = "Dictdiffer is a library that helps you to diff and patch dictionaries."
@@ -1553,7 +1561,7 @@ werkzeug = ">=0.15"
 admin = ["Flask-Admin (>=1.3.0)"]
 all = ["Flask-Admin (>=1.3.0)", "Sphinx (>=1.4.2,<1.6)", "check-manifest (>=0.25)", "coverage (>=4.0)", "isort (>=4.3.0)", "mock (>=2.0.0)", "pydocstyle (>=1.0.0)", "pytest (>=4.0.0,<5.0.0)", "pytest-cov (>=1.8.0)", "pytest-flask (>=0.10.0)", "pytest-pep8 (>=1.0.6)", "selenium (>=3.0.1)"]
 docs = ["Sphinx (>=1.4.2,<1.6)"]
-mysql = ["invenio-db[versioning,mysql] (>=1.0.0)"]
+mysql = ["invenio-db[mysql,versioning] (>=1.0.0)"]
 postgresql = ["invenio-db[versioning,postgresql] (>=1.0.0)"]
 sqlite = ["invenio-db[versioning] (>=1.0.0)"]
 tests = ["check-manifest (>=0.25)", "coverage (>=4.0)", "isort (>=4.3.0)", "mock (>=2.0.0)", "pydocstyle (>=1.0.0)", "pytest (>=4.0.0,<5.0.0)", "pytest-cov (>=1.8.0)", "pytest-flask (>=0.10.0)", "pytest-pep8 (>=1.0.6)", "selenium (>=3.0.1)"]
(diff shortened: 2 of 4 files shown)


backend/inspirehep/utils.py
```diff
@@ -174,3 +174,48 @@ def _exit_handler(record_uuids, task_name, signum, frame):
...
+
+
+def __getter(getter: callable, accessor: callable):
+    def __default_getter(o, k):
+        if isinstance(o, list):
+            return [accessor(i, k) for i in o]
+        else:
+            return accessor(o, k)
+
+    return __default_getter if getter is None else getter
```
drjova, line 206: @nooraangelva is this copy-paste from somewhere else?

backend/inspirehep/records/tasks.py
```diff
@@ -96,10 +98,15 @@ def update_references_pointing_to_merged_record(
...
             for referenced_record in referenced_records_in_path:
-                update_reference_if_reference_uri_matches(
-                    referenced_record, merged_record_uri, new_record_uri
-                )
+                update_reference_if_reference_uri_matches(referenced_record, merged_record_uri, new_record_uri)
+            
+            referenced_records_in_path = dedupe_list_of_dicts(referenced_records_in_path)
+            updated_matched_inspire_record = remove_duplicate_refs_from_record(matched_inspire_record, referenced_records_in_path, path)
+            if updated_matched_inspire_record:
+                matched_inspire_record = updated_matched_inspire_record
```
MJedr, line 108: I don't get why you can't simply assign the deduped list to the correct path and update a record with it?

backend/inspirehep/utils.py
nooraangelva: Yes, from this repository -> https://github.com/mbello/dict-deep/tree/7c05652c5c845886ea3b36d2fb8b95d5618e1652

backend/inspirehep/records/tasks.py
nooraangelva: How would you assign/update it? I had a lot of trouble using the path since it's a string separated with . and if I simply removed the duplicates from the list it doesn't remove them from the ```matched_inspire_record```.
That's why I found the ```set_value_by_path``` function so I could update the affiliations once I have removed the duplicates from "the old version" of the record.

backend/inspirehep/utils.py
drjova: I would suggest installing it as a package https://pypi.org/project/dict-deep/ :)

nooraangelva: Much better idea. I should have asked if I could install it. 😄 My last internship place didn't really like open source packages.

backend/inspirehep/records/tasks.py
```diff
@@ -135,3 +143,18 @@ def regenerate_author_records_table_entries(uuids_to_regenerate):
         record.update_authors_records_table()
         record.update_self_citations()
         db.session.commit()
+
+
+def remove_duplicate_refs_from_record(matched_inspire_record, referenced_records_in_path, path):
```
drjova, line 148: looks much cleaner now :) Could you please also remove the extra empty line (149)?

backend/inspirehep/records/tasks.py
```diff
@@ -135,3 +142,17 @@ def regenerate_author_records_table_entries(uuids_to_regenerate):
...
+    affiliations_path = ".".join(path.split(".")[:-2])
+    affiliations = flatten_list(get_value(matched_inspire_record, affiliations_path, []))
```
MJedr, line 149: Can you change the naming, please? These are references, not affiliations :) 
Can we stick to the one version of extracting path to be consistent? So if you want get rid of `record.$ref` you do (imo it will be more readable)
```
referenced_records_in_parent_path = flatten_list(
                get_value(matched_inspire_record, path[: -len(".record.$ref")], [])
            )
```

backend/inspirehep/records/tasks.py
```diff
@@ -135,3 +142,17 @@ def regenerate_author_records_table_entries(uuids_to_regenerate):
...
+
+    if len(affiliations) == len(referenced_records_in_path):
+        return None
+
+    for affiliation in affiliations:
+        if flatten_list(affiliations).count(affiliation) > 1:
+            affiliations.remove(affiliation)
+    deep_set(matched_inspire_record, affiliations_path, affiliations)
```
drjova, line 157: I'm checking the api of `dict-deep` and I'm not sure if it accepts arrays, shouldn't we iterate through affiliations and paths and call `deep_set`? Maybe it is working in the tests because actually is only one in the list

backend/inspirehep/records/tasks.py
```diff
@@ -135,3 +142,17 @@ def regenerate_author_records_table_entries(uuids_to_regenerate):
...
+def remove_duplicate_refs_from_record(matched_inspire_record, referenced_records_in_path, path):
+    affiliations_path = ".".join(path.split(".")[:-2])
+    affiliations = flatten_list(get_value(matched_inspire_record, affiliations_path, []))
+
+    if len(affiliations) == len(referenced_records_in_path):
+        return None
+
+    for affiliation in affiliations:
+        if flatten_list(affiliations).count(affiliation) > 1:
+            affiliations.remove(affiliation)
```
MJedr, line 156: now you can use `dedupe_list` here instead of the loop, can't you?

backend/inspirehep/records/tasks.py
```diff
@@ -96,10 +99,14 @@ def update_references_pointing_to_merged_record(
             referenced_records_in_path = flatten_list(
                 get_value(matched_inspire_record, path[: -len(".$ref")], [])
             )
+
             for referenced_record in referenced_records_in_path:
-                update_reference_if_reference_uri_matches(
-                    referenced_record, merged_record_uri, new_record_uri
-                )
+                update_reference_if_reference_uri_matches(referenced_record, merged_record_uri, new_record_uri)
+            referenced_records_in_path = dedupe_list_of_dicts(referenced_records_in_path)
```
MJedr, line 105: I think it's not needed anymore with the logic inside `remove_duplicate_refs_from_record`, you iterate through the list of refs twice then

backend/inspirehep/records/tasks.py
```diff
@@ -135,3 +142,17 @@ def regenerate_author_records_table_entries(uuids_to_regenerate):
...
         record.update_self_citations()
         db.session.commit()
+
+
+def remove_duplicate_refs_from_record(matched_inspire_record, referenced_records_in_path, path):
+    affiliations_path = ".".join(path.split(".")[:-2])
+    affiliations = flatten_list(get_value(matched_inspire_record, affiliations_path, []))
+
+    if len(affiliations) == len(referenced_records_in_path):
+        return None
```
MJedr, line 152: I think it's awkward and as I wrote above I would remove dict deduping from `update_references_pointing_to_merged_record` since in this function you iterate once again over the list

backend/inspirehep/records/tasks.py
```diff
@@ -135,3 +142,17 @@ def regenerate_author_records_table_entries(uuids_to_regenerate):
...
+
+    for affiliation in affiliations:
+        if flatten_list(affiliations).count(affiliation) > 1:
```
MJedr, line 155: isn't `affiliations` already flatten?

backend/inspirehep/records/tasks.py
```diff
@@ -135,3 +142,17 @@ def regenerate_author_records_table_entries(uuids_to_regenerate):
...
+
+
+def remove_duplicate_refs_from_record(matched_inspire_record, referenced_records_in_path, path):
+    affiliations_path = ".".join(path.split(".")[:-2])
+    affiliations = flatten_list(get_value(matched_inspire_record, affiliations_path, []))
+
+    if len(affiliations) == len(referenced_records_in_path):
+        return None
+
+    for affiliation in affiliations:
```
MJedr, line 154: So now you are iterating over the list of dicts in, let's say, publication_info and you are removing items from it if they're already on the list. Why it can't be done with dedupe list of dicts instead of this whole function?

backend/inspirehep/records/tasks.py
nooraangelva: Hmm, I added an extra institution to the tests to see what happens it seems to work fine. I also added an assert to check that there are two institutions in the records references.

backend/inspirehep/records/tasks.py
nooraangelva: I changed the naming, but I'm using the ```".".join(path.split(".")[:-2])``` for a reason.
Some of the reference paths don't have ```".record.$ref"``` record in them.
for example:
```
'publication_info.parent_record.$ref'
'publication_info.journal_record.$ref'
'publication_info.conference_record.$ref'
```
So I wanted it to be dynamic.

MJedr: ok, I see. But that means that you take path parent, so maybe take the first element only?

nooraangelva: I would, but then the second problem is that some of the paths have 4 "steps" in them.😅 
For example:
```
'authors.affiliations.record.$ref'
'thesis_info.institutions.record.$ref'
```

MJedr (commented):
💯 rebase master :)

MJedr (approved):
good job!

PR #2627, opened 2022-08-22
ui: add pagination for assign literature item drawer

karolina-siemieniuk-morawska:
* ref: cern-sis/issues-inspire#99

### DIFF
--- a/ui/src/literature/components/AssignLiteratureItemDrawer.tsx
+++ b/ui/src/literature/components/AssignLiteratureItemDrawer.tsx
@@ -1,8 +1,7 @@
-import React, { useCallback, useState } from 'react';
+import React, { useCallback, useState, useEffect } from 'react';
 import { SelectOutlined } from '@ant-design/icons';
 
-import { Drawer, Radio, Row, Col, Button } from 'antd';
-import { RadioChangeEvent } from 'antd/lib/radio';
+import { Drawer, Radio, Row, Col, Button, List, Input } from 'antd';
 import ResultItem from '../../common/components/ResultItem';
 import Author from '../../common/components/Author';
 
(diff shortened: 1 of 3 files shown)


ui/src/literature/components/AssignLiteratureItemDrawer.tsx
```diff
@@ -70,6 +45,13 @@ function AssignLiteratureItemDrawer({
   paperId,
 }: AssignLiteratureItemDrawerProps) {
   const [selectedAuthorId, setSelectedAuthorId] = useState<string>();
+  const [availableAuthors, setAvailableAuthors] = useState<IAuthorResult[]>([]);
+  
+  useEffect(() => {
+    // @ts-ignore
+    setAvailableAuthors(authors.toArray());
```
drjova, line 52: why we need an array?

drjova: can we just read authors from the state directly?

ui/src/literature/components/AssignLiteratureItemDrawer.tsx
```diff
@@ -93,7 +86,43 @@ function AssignLiteratureItemDrawer({
       <p>
         <strong>Select the author to claim:</strong>
       </p>
-      {renderAuthorItem(authors, onSelectedAuthorChange)}
+      {/* @ts-ignore */}
+      {authors.size > 20 && (
```
drjova, line 90: by default ant-pagination can handle this case

ui/src/literature/components/AssignLiteratureItemDrawer.tsx
karolina-siemieniuk-morawska: Antd component `List` that I use for displaying list with pagination requires array of elements as input

ui/src/literature/components/AssignLiteratureItemDrawer.tsx
```diff
@@ -83,6 +64,17 @@ function AssignLiteratureItemDrawer({
...
+  const onAuthorSearch = (value: string): void => {
+    const filteredAuthors = authors.filter(
+        (author: IAuthorResult) => author
+          .get('full_name')
+          .toString()
+          .toLowerCase()
+          .includes(value.toLowerCase())
+      );
+    setAvailableAuthors(filteredAuthors);
+  };
```
drjova, line 76: what if it's an empty search?

karolina-siemieniuk-morawska: It will return all of the authors (the same behaviour as if we make an empty search in collections). 
+ I noticed a bug that the pagination wasn't working after search, fixed it :)

ui/src/literature/components/__tests__/__snapshots__/AssignLiteratureItemDrawer.test.jsx.snap
```diff
@@ -12,168 +12,33 @@ exports[`AssignLiteratureItemDrawer renders authors list 1`] = `
       Select the author to claim:
     </strong>
   </p>
+  <Search
+    className="pb3"
+    enterButton={true}
+    onSearch={[Function]}
+    placeholder="Search for author"
```
drjova, line 19:
```suggestion
    placeholder="Search for an author"
```

drjova (approved):
LGTM just the small comment

PR #2629, opened 2022-08-23
e2e: fix for firefox test

karolina-siemieniuk-morawska:
* ref: cern-sis/issues-inspire#85


drjova:
@karolina-siemieniuk-morawska we have to merge it on master to check if it works

PR #2648, opened 2022-09-05
indexer cli: handle fulltext in reindex when passing many pids

MJedr:
ref: cern-sis/issues-inspire#129

### DIFF
--- a/backend/inspirehep/indexer/cli.py
+++ b/backend/inspirehep/indexer/cli.py
@@ -6,7 +6,6 @@
 # the terms of the MIT License; see LICENSE file for more details.
 
 import re
-from os import makedirs, path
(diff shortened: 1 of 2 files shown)


backend/inspirehep/indexer/cli.py
```diff
@@ -193,44 +205,39 @@ def reindex_records(
         click.secho(f"Successfully reindexed record {pid}", fg="green")
         ctx.exit(0)
 
-    if not set(pidtype) <= set(allowed_pids):
+    pidtype = set(pidtype)
```
drjova, line 208:
```suggestion
    pidtypes = set(pidtype)
```

drjova (approved):

PR #2650, opened 2022-09-06
fulltext: add highlightning

MJedr:
ref:cern-sis/issues-inspire#132


backend/inspirehep/serializers.py
```diff
@@ -211,6 +212,14 @@ def populate_curated_relation(hits):
                 hit["_source"]["curated_relation"] = True
         return hits
 
+    @staticmethod
+    def populate_highlight(hits):
+        for hit in hits:
```
drjova, line 217: I see that we iterate twice for the populating `hightlights` and `curated_relation` can we move the logic in the `serialize_search` and iterate only once?

drjova (approved):

Issue #1659, opened 2021-01-21
Improve PDG identifiers


michamos:
On legacy, PDG identifiers are expanded into their human-friendly name and link back to the PDG site. On labs we display the PDG codes among the keywords and don't link them. Compare 
![Screenshot_2021-01-21 Observation of a New Boson at a Mass of 125 GeV with the CMS Experiment at the LHC - INSPIRE](https://user-images.githubusercontent.com/2895793/105375960-83c32f00-5c09-11eb-8b8b-c6599791b291.png)
 on https://inspirehep.net/literature/1124338 and 
![Screenshot_2021-01-21 Observation of a New Boson at a Mass of 125 GeV with the CMS Experiment at the LHC - INSPIRE-HEP](https://user-images.githubusercontent.com/2895793/105376007-93427800-5c09-11eb-89b6-e42166cf32d7.png)
on https://old.inspirehep.net/record/1124338 

We should improve that.

drjova:
A solution is to filter out PDG keywords and display them like in the screenshot:
![Image](https://user-images.githubusercontent.com/4287772/188115461-8637419b-ddb3-4dd3-89b2-7c0e1fd21f15.png)

michamos:
Translation from PDG identifiers to long descriptions is in https://pdg.lbl.gov/2022/pdgid/PDGIdentifiers-2022v0.json.

PR #2658, opened 2022-09-14
ui: improve PDG identifiers

karolina-siemieniuk-morawska:
* filter PDG identifiers from keywords and move them from abstract section to top section
* add backend function to receive identifier description
* display identifier in its human understandable version in the modal

* ref: inspirehep/inspirehep#1659


backend/inspirehep/records/marshmallow/literature/ui.py
```diff
@@ -260,6 +263,29 @@ def get_documents_without_fulltext(self, data):
                 del document["_error"]
         return documents
 
+    def get_pdg_keywords(self, data):
+        keywords = data.get("keywords", [])
+        pdg_keywords = []
+
+        for keyword in keywords:
+            for pdg_identifier in PDG_IDENTIFIERS:
```
MJedr, line 271: it's wrong cause you iterate through all 11k of identifiers :p

backend/inspirehep/records/marshmallow/literature/ui.py
```diff
@@ -260,6 +263,29 @@ def get_documents_without_fulltext(self, data):
...
+        return pdg_keywords
+    
+    def get_keywords(self, data):
+        keywords = data.get("keywords", [])
+
+        for keyword in keywords:
+            for pdg_identifier in PDG_IDENTIFIERS:
+                if keyword['value'] == pdg_identifier:
+                    keywords.remove(keyword)
+
```
MJedr, line 286:
```suggestion
keywords_without_pdg = [keyword_entry for keyword_entry in keywords if keyword_entry['schema'] != 'PDG']
```

MJedr (commented):
you need backend tests also ;)

ui/src/literature/components/PDGKeywords.tsx
```diff
@@ -0,0 +1,67 @@
...
+      </ExternalLink>
+    );
+  };
+
+  return (
+    <div>
+      PDG:{' '}
+      <>
+        {/* @ts-ignore */}
+        {renderKeyword(keywords.toArray()[0])}
```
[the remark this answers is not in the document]
karolina-siemieniuk-morawska: I do it because Immutable.List[index] returns undefined, so I first cast it to array to use some array methods such as [index] in this case :)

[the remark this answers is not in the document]
MJedr: Not really, you can use [`get`](https://immutable-js.com/docs/v3.8.2/List/#first()).
Converting to array might be expensive and has no reason here if you can use immutable methods [see the comment](https://stackoverflow.com/a/56034395)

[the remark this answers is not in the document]
karolina-siemieniuk-morawska: Thank you for catching it, fixed! :)

backend/inspirehep/records/marshmallow/literature/ui.py
```diff
@@ -260,6 +263,29 @@ def get_documents_without_fulltext(self, data):
                 del document["_error"]
         return documents
 
+    def get_pdg_keywords(self, data):
+        keywords = data.get("keywords", [])
+        pdg_keywords = []
+
+        for keyword_entry in keywords:
+            # case when it's another schema - we continiue iteration
+            if 'schema' not in keyword_entry or keyword_entry['schema'] != 'PDG':
```
MJedr, line 272: it will raise key error if `schema` is not in keyword because with OR both conditions will be checked (in that case AND clause should be used).
However, you can avoid it by checking `if keyword_entry.get('schema') != 'PDG'` instead

backend/inspirehep/records/marshmallow/literature/ui.py
```diff
@@ -260,6 +263,29 @@ def get_documents_without_fulltext(self, data):
                 del document["_error"]
         return documents
 
+    def get_pdg_keywords(self, data):
+        keywords = data.get("keywords", [])
+        pdg_keywords = []
+
+        for keyword_entry in keywords:
+            # case when it's another schema - we continiue iteration
```
MJedr, line 271: you can remove the comment

backend/tests/unit/records/marshmallow/literature/test_ui.py
```diff
@@ -279,3 +279,31 @@ def test_documents_doesnt_contain_error_from_fulltext(
...
+
+def test_filter_pdg_keywords():
+    data = {
+        "keywords": [
+            { "schema": "PDG", "source": "PDG", "value": "M013WX" },
+            { "source": "publisher", "value": "Strong Interactions" },
+        ]
+    }
+    expected_data = [
+        { "pid": "M013WX", "value": "f(2)'(1525) WIDTH" }
```
MJedr, line 305: hm is that correct? it is not test case `data`

karolina-siemieniuk-morawska: It is the correct result of  this function to filter out and map pdg keyword id to its' description `pdg_keywords.append({'pid': pdg_identifier, 'value': PDG_IDS_TO_DESCRIPTION_MAPPING[pdg_identifier]})`, and the tests' `serialized["pdg_keywords"]` equals exactly this, so I think it's correct, but I might be missing something.

MJedr: ok, I overlooked it but I think it's wrong to call the keyword value `pid` cause IMO that should be reserved for persistent identifiers of our records. I would add instead an entry with keys (`value`, `description`) instead

backend/inspirehep/records/marshmallow/literature/ui.py
```diff
@@ -111,6 +112,8 @@ class Meta:
     )
     thesis_info = fields.Nested(ThesisInfoSchemaV1, dump_only=True)
     dataset_links = fields.Method("get_datasets")
+    pdg_keywords = fields.Method("get_pdg_keywords")
+    keywords = fields.Method("get_keywords")
```
MJedr, line 116: can you add `dump_only` to those fields?

MJedr (commented):
looks good, just a small comment

PR #2669, opened 2022-09-27
ui: add claim button to literature search page items

karolina-siemieniuk-morawska:
* ref: cern-sis/issues-inspire#114


ui/src/literature/components/AssignLiteratureItemDrawer.tsx
```diff
@@ -24,7 +24,7 @@ interface IAuthorResult {
 }
 
 interface AssignLiteratureItemDrawerProps {
-  visible: boolean;
+  visible: string;
```
drjova, line 27: shouldn't be `bool`?

ui/src/literature/components/LiteratureItem.jsx
```diff
@@ -79,9 +89,13 @@ function LiteratureItem({ metadata, searchRank, isCatalogerLoggedIn }) {
     !assignOwnProfileView &&
     !assignAuthorView &&
     !assignDifferentProfileView;
-  
+
   const assignNotLoggedInView = useContext(AssignViewNotLoggedInContext);
-  const assignNotLoggedInViewCondition = assignNotLoggedInView && !assignNoProfileViewCondition;
+  const assignNotLoggedInViewCondition =
+    assignNotLoggedInView && !assignNoProfileViewCondition;
+  const assignLiteratureItemCondition =
```
drjova, line 96:
```suggestion
  const assignLiteratureViewCondition =
```

ui/src/literature/containers/AssignLiteratureItemDrawerContainer.tsx
```diff
@@ -11,20 +11,20 @@ interface RootState {
...
+  visible: state.literature.get('assignDetailViewDrawerVisible'),
   currentUserRecordId: Number(state.user.getIn(['data', 'recid'])),
 });
 
 export const dispatchToProps = (
   dispatch: MapDispatchToPropsParam<any, any>
 ) => ({
   onDrawerClose() {
-    dispatch(setAssignDetailViewDrawerVisibility(false));
+    dispatch(setAssignDetailViewDrawerVisibility(null));
```
drjova, line 27: hmm I don't understand why not `bool`

ui/src/literature/components/AssignLiteratureItemDrawer.tsx
karolina-siemieniuk-morawska: I had to change this logic a little bit now that the drawer can also be displayed on literature search page and not just on details page - if this value would stay just a true/false then when ```setAssignDetailViewDrawerVisibility``` action is triggered it would render a drawer for every literature item on search page, not just the one that was clicked. So to render just one drawer component for the desired literature item I set this visible parameter to record id, and then on drawer component I check if visible parameter in state equals the selected literature item id

ui/src/literature/containers/AssignLiteratureItemDrawerContainer.tsx
karolina-siemieniuk-morawska: Explanation above ⤴️ 😀

ui/src/literature/components/AssignLiteratureItemDrawer.tsx
drjova: if it's an id, can we use integer?

karolina-siemieniuk-morawska: You're right, changed to type ```number``` ✅

ui/src/literature/components/AssignLiteratureItemDrawer.tsx
```diff
@@ -81,7 +81,7 @@ function AssignLiteratureItemDrawer({
       className="search-drawer"
       placement="right"
       onClose={onDrawerClose}
-      visible={visible}
+      visible={literatureId === paperId}
```
drjova, line 84: this is confusing 🤔 `itemLiteratureId === literatureId`? WDYT?

ui/src/literature/components/LiteratureClaimButton.tsx
```diff
@@ -0,0 +1,26 @@
...
+}: {
+  loggedIn: boolean;
+  hasAuthorProfile: boolean;
+  authors: [];
+  controlNumber: string;
+}) => (
+    <>
+      {!loggedIn && <ClaimingDisabledButton />}
+      {!hasAuthorProfile && <AssignNoProfileAction />}
+      {!authors && <NoAuthorsClaimingButton />}
```
drjova, line 21: !authors is not correct, what exactly you would like to check? Perhaps if `authors.length === 0`?

drjova (approved):
LGTM

PR #2674, opened 2022-10-04
search: handle citedby query

MJedr:
ref: cern-sis/issues-inspire#137


backend/inspirehep/search/api.py
```diff
@@ -192,15 +225,21 @@ def query_by_user_role(self, query_string):
         fulltext_highlight_options = dict(
             fragment_size=160, type="fvh", number_of_fragments=1, order="score"
         )
+        query = IQ(query_string, self)
+
+        if "citedby" in query_string:
+            query = query.to_dict()
+            self.replace_recid_in_citedby_query(query)
```
drjova, line 232: Why we have this logic inside the [execute](https://github.com/inspirehep/inspirehep/pull/2674/files#diff-19b96149fef1fdaddf2df9e639f18cc6b2fd728e0d2f7a6e5feb1f0514fece0cL179)?

MJedr: cause we parse the query here and I need a parsed query to replace recid

backend/inspirehep/search/factories/query.py
```diff
@@ -4,37 +4,58 @@
...
+        if isinstance(obj, dict):
+            for key, val in obj.items():
+                if isinstance(val, list):
+                    _replace_recid_in_citedby_query(val)
+                elif isinstance(val, dict):
+                    # check if we are in citedby query and replace
+                    # recid with uuid
+                    if (
+                        citedby_query_fields == set(val.keys())
+                        and self_ref_field in obj.keys()
```
drjova, line 31: I would move the obj.keys() outside, since it's always the same and we don't have to calculate it in every iteration

drjova (approved):

PR #2682, opened 2022-10-05
ui: change the url for assigning from different profile

karolina-siemieniuk-morawska:


ui/src/actions/literature.js
```diff
@@ -166,7 +166,7 @@ export function assignLiteratureItem({ from, to, literatureId }) {
     try {
       assigning(ASSIGNING_NOTIFICATION_LITERATURE_ITEM_KEY);
       const { data } = await http
-        .post('/assign/literature/assign', {
+        .post('/assign/literature/assign-different-profile', {
```
drjova, line 169: I'm not sure that this is the case, this is the same as `assignLiteratureItemNoNameMatch`. Maybe we should change the logic triggering the action. @MJedr WDYT?

MJedr: IMO it should work because:
1. when there's a name it doesn't mean that it's own profile claiming, it can be a different profile claiming
2. even if it is own profile claiming the behavior should be correct cause names match (?)

However, if we want to be consistent maybe we can add a check if `from` and `to` authors have the same recid and if yes send it to `/assign/literature/assign` - it will be safe cause we have a check for editing in backend

MJedr (approved):

PR #2685, opened 2022-10-10
jobs: users should see closed jobs

MJedr:
cern-sis/issues-inspire#146

### DIFF
--- a/backend/inspirehep/search/api.py
+++ b/backend/inspirehep/search/api.py
@@ -387,13 +387,13 @@ class JobsSearch(InspireSearch):
         """
         if not is_superuser_or_cataloger_logged_in():
             if not query_string:
-                user_query = Q("term", status="open")
+                user_query = Q("terms", status=["open", "closed"])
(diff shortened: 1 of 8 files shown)


backend/inspirehep/search/api.py
```diff
@@ -387,13 +387,13 @@ def query_from_iq(self, query_string):
         """
         if not is_superuser_or_cataloger_logged_in():
             if not query_string:
-                user_query = Q("term", status="open")
+                user_query = Q("terms", status=["open", "closed"])
```
drjova, line 390: we need by default to display only open, do we have to change it?

MJedr: yes we do https://github.com/inspirehep/inspirehep/pull/2685/files#diff-2bbc000ac49ab94dc52c7931324b1490f0ab382bc8f67e9da203b444d0c6d8e1R71-R74

drjova (approved):

PR #2686, opened 2022-10-11
latex & bibtex: strip mathml

MJedr:
ref: cern-sis/issues-inspire#147

### DIFF
--- a/backend/inspirehep/records/marshmallow/literature/bibtex.py
+++ b/backend/inspirehep/records/marshmallow/literature/bibtex.py
@@ -11,6 +11,7 @@ from idutils import is_arxiv_post_2007, normalize_isbn
 from inspire_schemas.readers.conference import ConferenceReader
 from inspire_schemas.readers.literature import LiteratureReader
 from inspire_utils.date import PartialDate
+from inspire_utils.helpers import remove_tags
 from inspire_utils.record import get_value
 from isbn import ISBNError
 from marshmallow import fields, pre_dump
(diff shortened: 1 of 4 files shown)


backend/inspirehep/records/marshmallow/literature/bibtex.py
```diff
@@ -234,7 +235,8 @@ def get_title(self, data):
         if "subtitle" in title_dict:
             title_parts.append(title_dict["subtitle"])
         return ": ".join(
-            f"{{{latex_encode(part, contains_math=True)}}}" for part in title_parts
+            f"{{{latex_encode(remove_tags(remove_tags(part, allowed_tags='DUMMYROOTTAG')), contains_math=True)}}}"
```
drjova, line 238: can we do it in a loop? it's getting a hard to read having everything in the formatted string

backend/inspirehep/records/marshmallow/literature/bibtex.py
```diff
@@ -233,9 +234,17 @@ def get_title(self, data):
...
-        )
+
+        formatted_parts = []
+        for part in title_parts:
+            part_witouth_mathml = remove_tags(
+                remove_tags(part, allowed_tags="DUMMYROOTTAG")
+            )
+            latex_encoded_part = (
+                f"{{{latex_encode(part_witouth_mathml, contains_math=True)}}}"
+            )
```
drjova, line 245: hmm I don't understand ` f"{{{ }}}"` need

MJedr: me neither, but it was like that before and I don't know latex and bibtex at all so I was afraid to remove it 😅 I can remove it

drjova: Let’s try to remove it and see :)

MJedr: doesn't work, it's about adding `{}` to the encoded text, I'll leave it as it was

drjova (approved):

PR #2690, opened 2022-10-11
ui: add reference search button to literature item

karolina-siemieniuk-morawska:
* ref: cern-sis/issues-inspire#125


ui/src/common/components/ReferenceSearchLinkAction.tsx
```diff
@@ -0,0 +1,20 @@
+import React from 'react';
+import { FileSearchOutlined } from '@ant-design/icons';
+import { Link } from 'react-router-dom';
+
+import IconText from './IconText';
+import { LITERATURE } from '../routes';
+import ListItemAction from './ListItemAction';
+import EventTracker from './EventTracker';
+
+const ReferenceSearchLinkAction = ({ linkQuery }: { linkQuery: string }) => (
```
MJedr, line 10: since the query is always the same for this component, maybe we can have it in the component and we can pass only recid?

drjova (approved):

PR #2694, opened 2022-10-13
journals: add associative table & number_of_papers property

MJedr:
ref: cern-sis/issues-inspire#159


backend/tests/integration/records/api/test_api_mixins.py
```diff
@@ -810,3 +811,71 @@ def test_hard_delete_advisor_clears_entries_in_students_advisors_table(
...
+        "lit", data={"publication_info": [{"journal_record": journal_rec["self"]}]}
+    )
+    assert JournalLiterature.query.filter_by(literature_uuid=rec.id).count() == 1
+
+
+def test_journal_literature_table_is_populated_when_multiple_lit_records_added(
+    inspire_app,
+):
+    journal_rec = create_record("jou")
+    journal_rec_2 = create_record("jou")
```
drjova, line 828: I know we are using it on other places as well, but actually `create_record` is a helper function for tests and changing it might affect the tests. Hence, it would be better to use JournalRecord.create()` instead to make sure we are testing the `create` code and avoid interference with test artefacts.

MJedr: I can change it, but actually `create_record` calls the `create` method from `InspireRecord`, which is using a class method for a specified class

drjova: The point is that indeed `create_record` just calls `InspireRecord` but it's a test logic because it's a helper function. We can leave it like that for now, since it's everywhere.

drjova (approved):
Just the `pass` comment

backend/tests/integration/records/api/test_api_mixins.py
MJedr: it's ok, I've already changed it :)

backend/inspirehep/records/cli.py
```diff
@@ -192,3 +193,32 @@ def close_expired_jobs(notify):
...
+                    }
+                },
+            }
+        }
+    }
+    records = LiteratureSearch().from_dict(query).params().scan()
+
+    for record in records:
+        record = LiteratureRecord.get_record(record.meta.id)
+        record.update_journal_relations()
```
drjova, line 224: Let's chunk them and add a try except in case something fails not to interrupt the process

drjova (approved):

PR #2698, opened 2022-10-17
bibtex & latex: handle `&` in title

MJedr:


drjova:
@MJedr thanks for the PR, can we double check that both cases are returning a valid result?

MJedr:
What do you mean? Serialization is correct cause we get the response for the request.

drjova:
I mean if the output it's a valid latex and bibtex for this corner cases

drjova (approved):

PR #2701, opened 2022-10-20
prometheus exporter: add label depending on request headers

MJedr:
cern-sis/kubernetes#592

### DIFF
--- a/backend/inspirehep/logger/ext.py
+++ b/backend/inspirehep/logger/ext.py
@@ -78,7 +78,13 @@ class InspireLogger:
 
         prefix = app.name
         metrics_flask = GunicornInternalPrometheusMetrics.for_app_factory(
-            defaults_prefix=prefix, group_by=url_rule
+            defaults_prefix=prefix,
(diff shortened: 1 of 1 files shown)


backend/inspirehep/logger/ext.py
```diff
@@ -78,7 +78,13 @@ def init_prometheus_flask_exporter(self, app):
 
         prefix = app.name
         metrics_flask = GunicornInternalPrometheusMetrics.for_app_factory(
-            defaults_prefix=prefix, group_by=url_rule
+            defaults_prefix=prefix,
+            group_by=url_rule,
+            static_labels={
+                "source": lambda request: "external"
+                if "X-Unique-ID" in request.headers
+                else "internal"
```
drjova, line 86: @benjamin-bergia is it going to work like this?

benjamin-bergia: I guess but I don't really understand where is request is coming from

drjova (approved):

PR #2702, opened 2022-10-21
disambiguation cli: cast to string record uuid

MJedr:


backend/inspirehep/disambiguation/cli.py
```diff
@@ -108,7 +108,7 @@ def _get_all_not_disambiguated_records_search():
 
 def _send_celery_group_disambiguation_task(uuids, batch_size):
     records = LiteratureRecord.get_records(uuids)
-    input_data = ((record.id, record.model.version_id, True) for record in records)
+    input_data = ((str(record.id), record.model.version_id, True) for record in records)
```
drjova, line 111: could we add a test for that?

MJedr: what exactly do you want to test in casting from uuid type to string?

drjova: yes it's a cast, but we should have caught this during tests if it's causing errors. WDYT?

MJedr: yes, I agree, the problem is that we have a test covering this case and it's working correctly, so I don't know what else can I add

PR #2705, opened 2022-10-24
refactor indexer 

MJedr:


backend/inspirehep/indexer/base.py
```diff
@@ -72,6 +75,10 @@ def _process_bulk_record_for_index(
             "_version_type": version_type,
             "_source": self._prepare_record(record, index, doc_type),
         }
+        if current_app.config["FEATURE_FLAG_ENABLE_FULLTEXT"]:
+            ingestion_pipeline_name = current_app.config["ES_FULLTEXT_PIPELINE_NAME"]
+            payload["pipeline"] = ingestion_pipeline_name
```
drjova, line 80: shouldn't this be only for literature?

backend/inspirehep/indexer/api.py
```diff
@@ -21,6 +21,7 @@ def get_references_to_update(record):
             uuids_to_reindex |= (
                 record.get_all_connected_records_uuids_of_modified_collaborations()
             )
+        uuids_to_reindex |= record.get_modified_journal_uuids()
```
drjova, line 24: shouldn't we have it for experiments, institutions etc?

drjova: actually we just need to update this one instead https://github.com/inspirehep/inspirehep/blob/master/backend/inspirehep/records/api/literature.py#L555-L575

MJedr: probably we should

drjova: As we discussed let's move this into `record.get_linked_papers_if_reference_changed`

drjova (approved):

PR #2712, opened 2022-11-01
assign: add logging & refactor get_author_by_recid

MJedr:


backend/inspirehep/assign/utils.py
```diff
@@ -26,11 +30,21 @@ def is_assign_view_enabled():
...
-        author
-        for author in literature_record.get("authors")
-        if get_recid_from_ref(author.get("record")) == author_recid
-    )
+    lit_authors_recids = {
+        get_recid_from_ref(author["record"]): author
+        for author in literature_record.get("authors", [])
+        if "record" in author
+    }
+    maybe_int(author_recid) in lit_authors_recids
```
drjova, line 38: is this a left over?

backend/inspirehep/assign/utils.py
```diff
@@ -26,11 +30,20 @@ def is_assign_view_enabled():
...
+    }
+    lit_author_found = lit_authors_recids.get(maybe_int(author_recid))
+
+    if not lit_author_found:
+        LOGGER.warning(
+            "Author not found in literature authors",
+            lit_authors_recids=lit_authors_recids.keys(),
+            author_recid=author_recid,
+        )
+    return lit_author_found
```
drjova, line 46: if not `lit_author_found` it will return `None`, do we handle that case?

MJedr: in general yes, no exception should be raised, the only question might be about this line: maybe in case of a missing author, we should add a name from the author record? https://github.com/inspirehep/inspirehep/blob/d21dc0f11ad352313647c351ce2d567979be26df/backend/inspirehep/assign/views.py#L59

drjova: What about not include it at all if it's `None`?

MJedr: but then we create author without name which is stupid

drjova: indeed, let's add the name from the record

drjova (approved):

PR #2713, opened 2022-11-01
add celery queue limits to disambiguation cli

MJedr:


backend/inspirehep/disambiguation/cli.py
```diff
@@ -19,6 +20,9 @@
 from inspirehep.records.models import RecordsAuthors
 from inspirehep.search.api import AuthorsSearch, LiteratureSearch
 
+MAX_INDEXER_QUEUE_LEN = 100000
```
drjova, line 23: shouldn't this be configurable?

drjova (approved):

PR #2721, opened 2022-11-02
fulltext: disable scanned documents

MJedr:
ref: cern-sis/issues-inspire#56


backend/inspirehep/records/api/literature.py
```diff
@@ -507,6 +508,7 @@ def add_file(
             new_key = hash_data(file_data)
             mimetype = magic.from_buffer(file_data, mime=True)
             file_data = BytesIO(file_data)
+            is_scanned = is_document_scanned(file_data)
```
drjova, line 511: can we somehow narrow down the usage of this? Can we get some performance benchmark for files with many pages?

MJedr: 1) Do you have any suggestions?
2) maybe we can add a prometheus metric

MJedr: @drjova

backend/inspirehep/records/api/literature.py
```diff
@@ -507,6 +508,9 @@ def add_file(
             new_key = hash_data(file_data)
             mimetype = magic.from_buffer(file_data, mime=True)
             file_data = BytesIO(file_data)
+            is_scanned = None
+            if "fulltext" not in file_data and mimetype == "application/pdf":
```
drjova, line 512: `file_data` is `BytesIO` why checking if `fulltext` is not there?

MJedr: `fulltext` is a mark and we wanted to avoid double-checking if they're scanned for the documents that we already checked.

backend/inspirehep/records/utils.py
```diff
@@ -152,3 +154,16 @@ def remove_author_bai_from_id_list(author):
...
+def is_document_scanned(file_data):
+    try:
+        with pdfplumber.open(file_data) as pdf:
+            for page in pdf.pages:
+                page_text = page.extract_text()
+                if page_text:
+                    return False
+    # case when the pdf is invalid
+    except PDFException:
+        pass
```
drjova, line 168: let's add a log to have more info for the missing pdfs

MJedr: do we want to raise an error then (eg. `UnsupportedFileError`)? And what value of the`is_scanned` do we put in that case?

backend/inspirehep/records/api/literature.py
```diff
@@ -510,6 +513,17 @@ def add_file(
             filename = filename or key
             if not filename:
                 filename = new_key
+            is_scanned = None
+            if not fulltext and mimetype == "application/pdf":
+                try:
+                    is_scanned = is_document_scanned(file_data)
+                # case when the pdf is invalid
```
drjova, line 520: we don't need the comment :)

PR #2730, opened 2022-11-10
ui: change conference info label depending on document type

karolina-siemieniuk-morawska:
* ref: cern-sis/issues-inspire#175

### DIFF
--- a/ui/src/literature/components/ConferenceInfoList.jsx
+++ b/ui/src/literature/components/ConferenceInfoList.jsx
@@ -7,11 +7,11 @@ import ConferenceInfo from './ConferenceInfo';
 
 class ConferenceInfoList extends Component {
   render() {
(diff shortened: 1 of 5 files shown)


ui/src/literature/components/ConferenceInfoList.jsx
```diff
@@ -7,11 +7,11 @@ import ConferenceInfo from './ConferenceInfo';
 
 class ConferenceInfoList extends Component {
   render() {
-    const { conferenceInfo, wrapperClassName } = this.props;
+    const { conferenceInfo, wrapperClassName, proceedings } = this.props;
```
MJedr, line 10: Maybe call it `isProceeding` since it is boolean

MJedr (approved):

PR #2731, opened 2022-11-14
innovation: rewrite bibliographyGenerator and user to ts

karolina-siemieniuk-morawska:


ui/src/bibliographyGenerator/latexExample.ts
```diff
@@ -0,0 +1,153 @@
+export const latexExample1 = `
```
drjova, line 1: since it's a exported const `LATEX_EXAMPLE_1` :) etc

ui/src/user/components/SignUpPage.tsx
```diff
@@ -1,18 +1,31 @@
...
+  ) => void | Promise<any>) &
+    Function;
+  loading: boolean;
+  error: Error;
+}) {
   return (
     <>
       <DocumentHead title="Sign up" />
-      <Row className="h-100" type="flex" justify="center" align="middle">
+      <Row className="h-100" justify="center" align="middle">
```
drjova, line 24: why are we removing `flex`?

karolina-siemieniuk-morawska: [Antd removed it from <Row /> component](https://ant.design/components/grid/#Row) and flex is now it's default property, TS recognizes this and highlights type="flex" as error, that's why I have to remove it.

ui/src/bibliographyGenerator/latexExample.ts
drjova: @karolina-siemieniuk-morawska ☝️

drjova (approved):

PR #2742, opened 2022-11-22
ui: remove force render parameter to trigger fetching searchAggregations

karolina-siemieniuk-morawska:
* restoring  lazy render after clicking on tabs in order to trigger fetching search aggregations 

* ref: cern-sis/issues-inspire#187

### DIFF
--- a/ui/src/authors/containers/DetailPageContainer/DetailPageContainer.tsx
+++ b/ui/src/authors/containers/DetailPageContainer/DetailPageContainer.tsx
@@ -211,7 +211,6 @@ function DetailPage({
                     </Tooltip>
                   }
                   key="2"
-                  forceRender
                 >
                   <ContentBox className="remove-top-border-of-card">
                     <AuthorCitationsContainer />


drjova:
Hey @karolina-siemieniuk-morawska, thank you for the PR. Could you please explain a bit about this change?

karolina-siemieniuk-morawska:
> Hey @karolina-siemieniuk-morawska, thank you for the PR. Could you please explain a bit about this change?

Sure. This bug with date facet is not the same as the one we had previously, when aggregations fetch was not triggered after back button click or when navigating through collections. Here, aggregations are after all fetched properly, but because of force rendering both tabs at once (not only focused default tab), aggregations from the last/slowest facets request (`citation-by-year`) that is made for 'research works' tab "covers" aggregations for `hep-author-citations` 
Screenshot 2022-11-22 at 14 33 48

When using lazy loading to load tab's content, requests are then made in a short time separation and doesn't conflict with each other which is enough to fix the issue.
Screenshot 2022-11-22 at 14 38 43

My solution has the same number of requests as there was before, just the order of their execution is different, plus now all three tabs (research work, cited by, seminars) are rendered with the same mechanism.

drjova (approved):

Issue #2186, opened 2021-09-01
Remove BAIs from papers


michamos:
On literature records we have two ways of associating signatures to an author profile: through a link to the record (`authors.record`) and through the BAI (`authors.ids.value` with `schema="INSPIRE BAI"`). In principle, both ways should point to the same record in case there is one, meaning that the record referenced `authors.record`should have that BAI among its `ids`.

In practice, this is not always the case, as BAIs on author records might have been changed by curators.

Discussing with @drjova, we concluded it would be better to remove BAIs from the literature records, and populate them from the linked author records when generating the `AuthorsRecords` objects (which are used for self-citation determination), then use that table to put the BAI back among the `ids` in the ES marshmallow serializer (similarly to what we do for `referenced_authors_bais`). That would allow to simplify a lot of code that currently special cases BAIs, especially in the `disambiguation` module (which can become BAI agnostic).

We will need to remove all existing BAIs from records, which will trigger disambiguation for all signatures that are not currently associated with an author record (legacy disambiguation linking it to an author profile with no matching HEPNames record). This requires some more thought to do properly. We might for example first remove all BAIs (which will re-disambiguate all of INSPIRE), then change the code, then remove all BAIs again.

drjova:
1. Change the `RecordsAuthors` instead of BAI to use record id
2. Remove `BAI` from literature records and populate them on reindexing
3. Trigger reindexing of literature records when BAI changes

MJedr:
will require also changes in disambiguation

MJedr:
**TECH NOTES**
1. Modify all the functions/methods that use BAI for fetching author/literature records:
* https://github.com/inspirehep/inspirehep/blob/b3095875eb9a0d6d6a261e5692368eda97cb5994/backend/inspirehep/records/api/authors.py#L101
* https://github.com/inspirehep/inspirehep/blob/48826133a5b9cae7b18f05b461da81870d01ea53/backend/inspirehep/disambiguation/cli.py#L72
* https://github.com/inspirehep/inspirehep/blob/56524ac979d5d44e4883d5dc37526d08ca99d269/backend/inspirehep/records/marshmallow/literature/es.py#L69

PR #2746, opened 2022-11-23
records: remove bai from lit record

MJedr:
ref: #2186


backend/inspirehep/disambiguation/tasks.py
```diff
@@ -142,36 +134,34 @@ def match_literature_author(author, record):
...
             matched_records
         )
-        matched_author_data = get_reference_and_bai_if_unambiguous_match(
-            matched_refs_ids
-        )
+        matched_author_reference = get_reference_if_unambiguous_match(matched_refs)
 
-        if matched_author_data:
+        if matched_author_reference:
             break
```
drjova, line 143: wouldn't be sufficient to check the size? instead of using `get_reference_if_unambiguous_match`

backend/inspirehep/disambiguation/tasks.py
```diff
@@ -253,43 +224,37 @@ def _disambiguate_authors(authors_to_disambiguate, record):
...
-            author["record"] = {"$ref": matched_author_data["author_reference"]}
-            assign_bai_to_literature_author(
-                author, matched_author_data.get("author_bai")
-            )
-            assigned_author_recid = matched_author_data["author_reference"].split("/")[
-                -1
-            ]
+        if matched_author_reference:
+            author["record"] = {"$ref": matched_author_reference}
+            assigned_author_recid = matched_author_reference.split("/")[-1]
```
drjova, line 246: why not using the function that we have for that?

backend/inspirehep/records/api/mixins.py
```diff
@@ -284,10 +253,11 @@ def update_refs_in_citation_table(self, save_every=100):
...
 
-    def get_authors_bais(self):
-        return get_values_for_schema(
-            flatten_list(self.get_value("authors.ids", [])), "INSPIRE BAI"
-        )
+    def get_authors_recids(self):
+        return [
+            str(get_recid_from_ref(recid))
+            for recid in get_value(self, "authors.record", [])
+        ]
```
drjova, line 260:
```suggestion
        return [
            str(get_recid_from_ref(ref))
            for ref in get_value(self, "authors.record", [])
        ]
```

drjova (approved):
Let's test it

PR #2768, opened 2023-01-12
ui: fix for missing authors when claiming from search page

karolina-siemieniuk-morawska:
* ref: cern-sis/issues-inspire#182


MJedr (commented):
it looks good, but can we have more tests? especially e2e might be useful

karolina-siemieniuk-morawska:
> it looks good, but can we have more tests? especially e2e might be useful

added unit tests, for e2e there already is a separate task [Add e2e tests for claiming](https://github.com/cern-sis/issues-inspire/issues/156) that I could take next

ui/src/literature/components/__tests__/AuthorResult.test.jsx
```diff
@@ -0,0 +1,76 @@
...
+    });
+
+    const wrapper = shallow(<AuthorResult item={authors} page="Page" />);
+
+    expect(wrapper.find(Radio)).toHaveProp({
+      value: undefined,
+    });
+  });
+
+  it('renders enabled radio button if author has id', () => {
```
MJedr, line 48: Can we assert for radio button is enabled in previous tests since it's the same case

MJedr (commented):
just a small comment but looks very good

MJedr (approved):

PR #2769, opened 2023-01-17
accounts: add endpoint for email change

MJedr:
ref cern-sis/issues-inspire#222

### DIFF
--- a/backend/inspirehep/accounts/views.py
+++ b/backend/inspirehep/accounts/views.py
@@ -6,11 +6,14 @@
 # the terms of the MIT License; see LICENSE file for more details.
 
 from flask import Blueprint, redirect, request, session
+from flask_login import current_user
 from flask_security.utils import login_user, logout_user, verify_password
 from invenio_accounts.models import User
 from invenio_db import db
 from invenio_oauthclient import current_oauthclient
 from sqlalchemy.exc import IntegrityError
+from webargs import fields
+from webargs.flaskparser import FlaskParser
 
 from inspirehep.orcid.tasks import push_account_literature_to_orcid
 from inspirehep.serializers import jsonify
@@ -18,6 +21,8 @@ from inspirehep.serializers import jsonify
 from .api import get_current_user_remote_orcid_account
 from .decorators import login_required
 from .handlers import get_current_user_data
+
+parser = FlaskParser()
 
 blueprint = Blueprint(
     "inspirehep_accounts", __name__, template_folder="templates", url_prefix="/accounts"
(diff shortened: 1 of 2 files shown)


backend/inspirehep/accounts/views.py
```diff
@@ -104,3 +109,21 @@ def set_orcid_push_setting():
...
+@parser.use_args(
+    {
+        "new_email": fields.String(required=True),
+    },
+    locations=("json",),
+)
+def set_email(args):
+    new_email = args["new_email"]
+    try:
+        current_user.email = new_email
```
drjova, line 125: I'm not sure this is enough, it's a proxy object

backend/inspirehep/accounts/views.py
```diff
@@ -104,3 +109,21 @@ def set_orcid_push_setting():
...
+
+@blueprint.route("/settings/update-email", methods=["POST"])
+@login_required
+@parser.use_args(
+    {
+        "new_email": fields.String(required=True),
+    },
+    locations=("json",),
+)
+def set_email(args):
```
drjova, line 122:
```suggestion
def update_email(args):
```

backend/inspirehep/accounts/views.py
MJedr: I was not sure either, found this source where they set `current_user` params like that https://stackoverflow.com/questions/62361924/flask-sqlalchemy-update-the-current-user (there's also a test in the pr). But to be on the safe side I'll do it with db table query.

backend/tests/integration/accounts/test_views.py
```diff
@@ -225,3 +225,18 @@ def test_orcid_push_setting_without_user(inspire_app):
...
+    user = create_user(role="user", orcid=orcid)
+    with inspire_app.test_client() as client:
+        login_user_via_session(client, email=user.email)
+        response = client.post(
+            "/accounts/settings/update-email",
+            data=orjson.dumps({"new_email": new_email}),
+            content_type="application/json",
+        )
+    assert response.status_code == 200
+    assert user.email == new_email
```
drjova, line 242: can we add it as async and query db directly?

MJedr: Do we really need it async? It doesn't seem like a very complex operation and it won't be used super often probably.
`query db directly?` -> isn't it what I did?

drjova: Since we don't commit any changes on integration, it's better to test it on async to make sure the change is in the db too.

drjova (approved):

PR #2778, opened 2023-01-23
ui: keep error causing url in history

karolina-siemieniuk-morawska:
* add error causing url to history and display it on error error pages

* ref: cern-sis/issues-inspire#196

Screenshot 2023-01-23 at 14 31 42

### DIFF
--- a/ui/src/errors/components/Error401.jsx
+++ b/ui/src/errors/components/Error401.jsx
@@ -4,10 +4,12 @@ import ErrorPage from './ErrorPage';
 
 class Error401 extends Component {
   render() {
+    const previousURL = document.referrer;
     return (
       <ErrorPage
         message="Sorry, you are not authorised to view this page."
         imageSrc={error401Image}
+        url={previousURL}
       />
     );
   }
--- a/ui/src/errors/components/ErrorPage.jsx
+++ b/ui/src/errors/components/ErrorPage.jsx
@@ -1,10 +1,11 @@
 import React, { Component } from 'react';
 import PropTypes from 'prop-types';
 import { Row, Col } from 'antd';
+import { Link } from 'react-router-dom';
 
 class ErrorPage extends Component {
   render() {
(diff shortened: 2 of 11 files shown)


ui/src/errors/components/Error401.jsx
```diff
@@ -4,10 +4,12 @@ import ErrorPage from './ErrorPage';
 
 class Error401 extends Component {
   render() {
+    const previousURL = document.referrer;
```
drjova, line 7: is it always there?

ui/src/errors/components/ErrorPage.jsx
```diff
@@ -19,6 +19,13 @@ class ErrorPage extends Component {
             </Col>
           </Row>
         )}
+        {url && (
+          <Row>
+            <Col span={24}>
+              <p className="f5 normal">Redirected from: {url}</p>
```
drjova, line 25: shouldn't be a link back? instead of string?

karolina-siemieniuk-morawska: I talked with Micha about how it should work and purposely put the url string here instead of our standard go back link, as it makes it easy to copy the broken page url.

ui/src/errors/components/Error401.jsx
karolina-siemieniuk-morawska: If the page was entered directly (like by pasting http://inspirehep.net/errors/404) and there will be no referrer, it will return empty string and the page will look like it used to look previously without redirect url (lack of referrer won't cause the page to break :) ).
Screenshot 2023-01-25 at 17 03 05

ui/src/errors/components/ErrorPage.jsx
drjova: why not both? a link which has as a text the url?

drjova (approved):

PR #2781, opened 2023-01-25
parallelize e2e tests

MJedr:
cern-sis/issues-inspire#225


drjova (approved):
there is only way to try :shipit:

drjova (approved):

PR #2782, opened 2023-01-26
fix alembic migration for records_authors

MJedr:


backend/inspirehep/alembic/3637cb5551a8_update_values_in_authorschematype.py
```diff
@@ -42,10 +42,44 @@
...
+        "ix_authors_records_record_id",
+        "tmp_records_authors",
+        ["record_id"],
+        unique=False,
+    )
+    op.create_index(
+        "ix_records_authors_id_type_authors_id",
+        "tmp_records_authors",
+        ["id_type", "author_id"],
+        unique=False,
```
drjova, line 71: shouldn't this be unique?

backend/inspirehep/alembic/3637cb5551a8_update_values_in_authorschematype.py
```diff
@@ -42,10 +42,44 @@
...
+    )
+    op.create_index(
+        "ix_records_authors_id_type_record_id",
+        "tmp_records_authors",
+        ["id_type", "record_id"],
```
drjova, line 76: and this one?

backend/inspirehep/alembic/3637cb5551a8_update_values_in_authorschematype.py
MJedr: this is copy-paste from the previous migrations, actually from my understanding it shouldn't be unique cause you can have same author with same kind of id (recid) in different records

backend/inspirehep/alembic/3637cb5551a8_update_values_in_authorschematype.py
MJedr: again, it is copy-paste, and again I think it's correct cause you can have many records with same id_type but different value (so many authors, all have recids -> id_type)

PR #2784, opened 2023-01-31
add snow API integration

MJedr:


backend/inspirehep/editor/views.py
```diff
@@ -188,52 +210,63 @@ def create_rt_ticket(endpoint, pid_value):
...
-    except NoUsersFound:
-        LOGGER.warning(
-            "Cannot find user in RT",
-            ticket_id=ticket_id,
-            pid_value=pid_value,
-            email=current_user.email,
-        )
+    # TODO: remove it after implementing snow
+    if current_app.config.get("FEATURE_FLAG_ENABLE_SNOW"):
+        snow_instance.resolve_ticket(ticket_id, current_user.email)
```
drjova, line 215: `try/except`?

backend/inspirehep/snow/api.py
```diff
@@ -0,0 +1,571 @@
...
+    def get_user(self, user_id):
+        """Get user by id
+        Args:
+            user_id (str): id of the user
+        Returns:
+            User name (str) if the user exists
+            None if the user doesn't exist
+        """
+        try:
+            user = self.get_users_from_cache().get(user_id, {}).get("user.name")
```
drjova, line 124: the key is `user.name`?

backend/inspirehep/snow/api.py
```diff
@@ -0,0 +1,571 @@
...
+        response = requests.put(
+            f"{self.url}/u_request_fulfillment/{ticket_id}",
+            orjson.dumps(payload),
+            headers=self.headers,
+        )
+        response.raise_for_status()
+        return get_response_result(response)
+
+    @relogin_if_needed
+    def search_snow(self, endpoint, query=None, fields=None):
```
drjova, line 151:
```suggestion
    def search(self, endpoint, query=None, fields=None):
```

backend/inspirehep/snow/api.py
MJedr: yes :| it's `{"user.name": "Test Name"}`

backend/inspirehep/snow/api.py
```diff
@@ -0,0 +1,571 @@
...
+            exclude_resolved (boolean): should resolved tickets be excluded from the response
+        Returns:
+            tickets_for_recid (list): list of found tickets
+        """
+        third_party_search_query_string = f"u_third_party_ticket_id={recid}"
+        tickets_search_parameters = dict(u_third_party="INSPIRE")
+        if exclude_resolved:
+            tickets_search_parameters[
+                "u_current_task_state"
+            ] = f"{self.ticket_status_mapping['waiting']}^ORu_current_task_state={self.ticket_status_mapping['assigned']}"
```
drjova, line 332:
```suggestion
            ] = f"{self.ticket_status_mapping['waiting']}^ORu_current_task_state={self.ticket_status_mapping['assigned']}"
```

👆 is `^OR` correct?

backend/inspirehep/snow/api.py
```diff
@@ -0,0 +1,571 @@
...
+                (
+                    user["id"]
+                    for user in self.get_formatted_user_list()
+                    if user["email"] == user_email
+                ),
+                None,
+            )
+            if user_email
+            else None
+        )
```
drjova, line 485: let's splitted to be easier to read

backend/tests/integration/editor/cassettes/test_create_snow_ticket_returns_500_on_error.yaml
```diff
@@ -0,0 +1,104 @@
...
+      Strict-Transport-Security:
+      - max-age=63072000; includeSubDomains
+      Transfer-Encoding:
+      - chunked
+      WWW-Authenticate:
+      - BEARER realm="Service-now"
+      content-length:
+      - '113'
+    status:
+      code: 401
```
drjova, line 102: this doesn't return `500` is it on purpose?

backend/tests/integration/editor/cassettes/test_get_snow_tickets_for_record.yaml
```diff
@@ -0,0 +1,579 @@
...
+      - Tue, 31 Jan 2023 13:37:33 GMT
+      Expires:
+      - '0'
+      Pragma:
+      - no-store,no-cache
+      Server:
+      - snow_adc
+      Server-Timing:
+      - sem_wait;dur=0, sesh_wait;dur=0
+      Set-Cookie:
```
drjova, line 82: let's drop the cookies as well from the cassettes

drjova:
we can follow this guide for docs  strings https://sphinxcontrib-napoleon.readthedocs.io/en/latest/example_google.html please also check the spacing between sections :)

backend/inspirehep/snow/api.py
```diff
@@ -0,0 +1,571 @@
...
+            "client_id": self.client_id,
+            "client_secret": self.secret,
+        }
+        headers = {
+            "Content-Type": "application/x-www-form-urlencoded",
+        }
+        try:
+            response = requests.post(self.auth_url, data=login_payload, headers=headers)
+            response.raise_for_status()
+            self.token = response.json()["access_token"]
```
drjova, line 53: do we know for how long we keep the token? do we have to send the request to get it? Couldn't we just set it on our configmaps?

drjova: If that's possible, it's much easier, we can just drop all the flask ext/proxy

backend/inspirehep/snow/api.py
```diff
@@ -0,0 +1,571 @@
...
+        return get_response_result(response)["sys_id"]
+
+    @relogin_if_needed
+    def get_ticket(self, ticket_id, params=None):
+        """Get ticket by id
+        Args:
+            params (str|dict): params to be passed to the request query string
+        """
+        if not params:
+            params = {}
```
drjova, line 80: we don't need that, we can leave it with the default `None`, it will work

backend/inspirehep/snow/api.py
```diff
@@ -0,0 +1,571 @@
...
+            category_id: SNOW sys_id of the category
+         Returns:
+             human-readable category name (str)
+             None if the category is not found
+        """
+        try:
+            category = (
+                self.get_functional_categories_from_cache()
+                .get(category_id, {})
+                .get("sys_name")
```
drjova, line 102: We have this type of code in many places, shall we add the logic to `self.get_functional_categories_from_cache` and pass the key to return? we can use something like `get_value` and we don't need the nested `get` with defaults.

backend/inspirehep/snow/api.py
```diff
@@ -0,0 +1,571 @@
...
+            LOGGER.error("Can't find tickets for recid", recid=recid)
+            tickets_for_recid = []
+        return tickets_for_recid
+
+    def create_inspire_ticket(
+        self,
+        user_email=None,
+        functional_category=None,
+        subject=None,
+        description="",
```
drjova, line 365:
```suggestion
        description=None,
```

backend/inspirehep/snow/api.py
MJedr: yes, unfortunately to add or condition we need to link it with `^OR`, from docs: [`^OR: Logically OR multiple query statements.`](https://docs.servicenow.com/bundle/tokyo-application-development/page/integrate/inbound-rest/concept/c_RESTAPI.html)

backend/inspirehep/snow/api.py
MJedr: token is valid for 20 min, we have to send a request to get it

backend/inspirehep/snow/api.py
MJedr: this is on purpose actually, we use later `_strip_lines` on it which requires a string, if you prefer I can modify this function

backend/tests/integration/editor/cassettes/test_create_snow_ticket_returns_500_on_error.yaml
MJedr: no, the cassette is wrong, thanks

backend/inspirehep/snow/api.py
MJedr: We do it twice, I didn't want to create a function that will store the same data as `get_functional_categories_from_cache` but in a different data structure. I can use get `get_value` instead to get rid 
of chaining `get`.

backend/inspirehep/snow/api.py
MJedr: > If that's possible, it's much easier, we can just drop all the flask ext/proxy

I think we can, I thought it was our coding convention 😉

backend/inspirehep/snow/api.py
drjova: yes, let's move do a check in the `strip_lines`, it's better to have immutable defaults in the signature

MJedr: strings are immutable ;)

drjova: you are absolute right, I had in mind array for some reason while replying 😅. Still let's move the logic inside the `strip_lines`

backend/inspirehep/snow/api.py
```diff
@@ -0,0 +1,635 @@
...
+                    del self.__dict__["token"]
+                return f(self, *args, **kwargs)
+
+        return decorated_function
+
+    @cached_property
+    def token(self):
+        token = current_cache.get("snow-token-value")
+        if not token:
+            token = self.get_token()
```
drjova, line 65: why we have to run twice `get_token`? It's inside the `write_token_to_cache` as well

backend/inspirehep/snow/api.py
```diff
@@ -0,0 +1,645 @@
...
+        url=None,
+        client_id=None,
+        secret=None,
+        auth_url=None,
+        verify_cert=False,
+        inspire_fe_id=None,
+        inspire_user_id=None,
+    ):
+        super().__init__(url, client_id, secret, auth_url, verify_cert)
+        self.inspire_fe_id = inspire_fe_id or current_app.config.get("SNOW_INSPIRE_FE")
```
drjova, line 312: small thing, because we are going to forget what `fe` means, can we use the full name `functional_element`? Also we can drop `inspire_` prefix from vars, it's already clear from the classname that we are using it for INSPIRE project

drjova (approved):
LGTM let's test it

PR #2794, opened 2023-02-02
workflow: add save build step

karolina-siemieniuk-morawska:


.github/workflows/integration.yml
```diff
@@ -130,6 +130,14 @@ jobs:
         with:
           image: inspirehep/ui
           context: ui/
+  
+  save_build_folder:
+    - name: Save build folder
```
MJedr, line 135: 1) is there a build folder already? If yes, you need to specify it with the path cause you are in the root of our project, not in the ui directory
2) if there's no build folder you need to use the checkout action before as described here https://docs.cypress.io/guides/continuous-integration/github-actions#Install-Job

MJedr (commented):
You need to reuse the build artifact, otherwise this step doesn't make sense, here's the instruction https://docs.cypress.io/guides/continuous-integration/github-actions#Worker-Jobs

PR #2800, opened 2023-02-03
add jupyter docker image and update integration workflow

MJedr:


docker/jupyter/jupyter_requirements.txt
```diff
@@ -0,0 +1,5 @@
+# jupyterhub==2.2.1
```
drjova, line 1: remove completely?

MJedr: yes, thanks,  a leftover ;)

drjova (approved):
LGTM, just can we check if there is a way to parallelise that and not to slowdown the rest of the process?

MJedr:
Actually, everything happens after images push, so basically it shouldn't affect any other part of the workflow

PR #2804, opened 2023-02-07
bai removal: add cli to remove it from lit authors

MJedr:


backend/inspirehep/records/tasks.py
```diff
@@ -172,3 +172,29 @@ def populate_journal_literature(uuids):
...
+            if not author_ids:
+                continue
+            new_ids = [
+                id_dict for id_dict in author_ids if id_dict["schema"] != "INSPIRE BAI"
+            ]
+            if new_ids:
+                author["ids"] = new_ids
+            else:
+                del author["ids"]
+        record.update_authors_records_table()
```
drjova, line 198: no need it's already running https://github.com/inspirehep/inspirehep/blob/master/backend/inspirehep/records/api/literature.py#L229

drjova (approved):

PR #2810, opened 2023-02-09
ui: added e2e tests for claiming - part 1

karolina-siemieniuk-morawska:
* ref: cern-sis/issues-inspire#156


e2e/cypress/integration/claiming.test.js
```diff
@@ -0,0 +1,354 @@
+describe('Author collection', () => {
```
MJedr, line 1: Shouldn't we add a case for `no profile`? also, don't we want to add tests for selecting multiple papers (for example with select all) and claiming them by the button on the top of the box?

Also, don't we want a test when the `authors` request fails on the literature search?

PR #2811, opened 2023-02-09
stub authors: verify if author is really stub before deleting

MJedr:


backend/tests/integration/disambiguation/test_cli.py
```diff
@@ -80,3 +81,24 @@ def test_clean_stub_authors_query_all_stub_authors(override_config, inspire_app,
...
+        control_numbers = []
+        for _ in range(2):
+            rec = create_record("aut", data={"stub": True}, with_control_number=True)
+            control_numbers.append(("aut", str(rec["control_number"])))
+
+        cli.invoke(["disambiguation", "clean_stub_authors"])
+
+        results = InspireRecord.get_records_by_pids(control_numbers)
+        for result in results:
+            assert not result.get("deleted")
```
drjova, line 104: let's have a case with no linked papers

MJedr: but we already do those cases in other tests

MJedr: yes, it returns False in other tests

PR #2814, opened 2023-02-10
ui: redirect to 500 page on serializer error

karolina-siemieniuk-morawska:


ui/src/middlewares/redirectToErrorPage.js
```diff
@@ -9,6 +9,9 @@ export default function() {
       // INFO: 'push' and 'replace' methods from 'connected-react-router' 
       // remove error causing url from history. To be able to keep in 
       // and retrieve it from history in error pages we use window API
+      if (error.status === 400 && error.message === "The search result can't be serialized") {
```
drjova, line 12: can we redirect all errors there but the malformed queries?

ui/src/middlewares/redirectToErrorPage.js
```diff
@@ -9,6 +9,9 @@ export default function() {
       // INFO: 'push' and 'replace' methods from 'connected-react-router' 
       // remove error causing url from history. To be able to keep in 
       // and retrieve it from history in error pages we use window API
+      if (error.status === 400) {
+        window.location.assign(`${ERRORS}/500`);
```
drjova, line 13: wouldn't this always redirect to `500`, event if it's an `The syntax of the search query is invalid.`?

karolina-siemieniuk-morawska: No, because this middleware only redirects errors which have redirectableError: true property, and in case of invalid query errors I intercept them in search action and set this flag to false for them so they are not redirectable.

drjova (approved):

PR #2824, opened 2023-02-20
remove bai cli: catch validation error

MJedr:


drjova (approved):

backend/inspirehep/records/tasks.py
```diff
@@ -195,5 +196,12 @@ def remove_bai_from_literature_authors(uuids):
                 author["ids"] = new_ids
             else:
                 del author["ids"]
-        record.update(dict(record), disable_disambiguation=True)
-        db.session.commit()
+        try:
+            record.update(dict(record), disable_disambiguation=True)
+            db.session.commit()
+        except ValidationError:
```
drjova, line 202: shall we catch all errors gracefully? otherwise it will blow-up all the time

MJedr: I don't know, I can't think of other errors that might appear and do not repeat the task

PR #2829, opened 2023-02-21
fix snow ui url

MJedr:


backend/inspirehep/snow/api.py
```diff
@@ -643,4 +643,4 @@ def get_ticket_simplified_response(self, ticket):
 
     def get_ticket_link(self, ticket_id):
         """Return ticket API url."""
-        return f"{self.base_url}/nav_to.do?uri=/u_request_fulfillment.do?sys_id=/{ticket_id}"
+        return f"{self.base_url}/nav_to.do?uri=/u_request_fulfillment.do?sys_id={ticket_id}"
```
drjova, line 646: I thought SNOW needed it

drjova (approved):

backend/inspirehep/snow/api.py
MJedr: yes, it's confusing cause in `uri` param it needs it :/

PR #2834, opened 2023-02-23
ui: add link to twitter in homepage

karolina-siemieniuk-morawska:
* ref: cern-sis/issues-inspire#177

Screenshot 2023-02-23 at 13 00 03


drjova:
@michamos is this ok for you?

michamos:
yes :+1:

drjova:
@karolina-siemieniuk-morawska why we have such difference in the screenshots?

karolina-siemieniuk-morawska:
@drjova on matchSnapshots custom command we set viewport width to established (1920px) width so that all the screenshots are equal, but for some reason it is not stable and some snapshots are generated with different screen resolution. I noticed it before and didn't find the cause after brief search, yet another cypress instability. For example, this are the results of ONE test run - it created two different screenshots with different resolutions - wider, with smaller resolution on first attempt and correct, 1920 high resolution on second attempt
Screenshot 2023-02-27 at 10 11 07

I believe this should be investigated further in separate task.

drjova:
@karolina-siemieniuk-morawska let's create an issue and investigate what's going on, it's weird to have this inconsistency

drjova (approved):

PR #2843, opened 2023-02-28
snow: use different user ids for caller and assigned_to

MJedr:


backend/inspirehep/snow/api.py
```diff
@@ -457,19 +458,10 @@ def create_inspire_ticket(
...
-                    if user["email"] == user_email
-                ),
-                None,
-            )
-            caller_id = assignee = snow_user_id
-        else:
-            caller_id = self.user_id
-            assignee = ""
+        caller_id = self._get_user_by_email(user_email) if user_email else self.user_id
+        assigned_to_id = (
```
drjova, line 462:
```suggestion
        assignee_id = (
```

backend/inspirehep/snow/api.py
```diff
@@ -627,6 +612,17 @@ def get_formatted_user_list(self):
...
+    def _get_user_by_email(self, user_email):
+        found_user = next(
+            (
+                user["id"]
+                for user in self.get_formatted_user_list()
+                if user["email"] == user_email
+            ),
+            None,
+        )
+        return found_user
```
drjova, line 624: should we add a `try/except`?

drjova: would that make sense?

MJedr: Do you mean `StopIteration` instead of `None`? I prefer this way tbh :p

backend/tests/integration/snow/test_snow.py
```diff
@@ -93,7 +94,7 @@ def test_get_ticket_by_recid(mocked_inspire_snow, inspire_app, teardown_cache):
     found_tickets = InspireSnow().get_tickets_by_recid(control_number)
     assert len(found_tickets) == 1
     assert found_tickets[0]
-    assert len(found_tickets[0].keys()) == 3
+    assert len(found_tickets[0].keys()) == 7
```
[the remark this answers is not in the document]
drjova: could we test the exact fields? Otherwise everytime we will have to add a field update the length

drjova (approved):
:shipit:

PR #2849, opened 2023-03-03
jupyter: update pre-commit hook to generate requirements.txt

MJedr:


docker/jupyter/CERN_Root_Certification_Authority_2.pem
```diff
@@ -0,0 +1,38 @@
+-----BEGIN CERTIFICATE-----
```
drjova, line 1: can we reuse the same one as hep?

MJedr: it is difficult cause it's not in docker context and either I have to copy this or I have to copy files from `docker/jupyter` to `inspirehep/backend`

drjova (approved):

PR #2850, opened 2023-03-06
headers: dont add link header to search & record response

MJedr:
ref: cern-sis/issues-inspire#221

### DIFF
--- a/backend/inspirehep/serializers.py
+++ b/backend/inspirehep/serializers.py
@@ -15,6 +15,7 @@ from inspire_utils.record import get_value
 from invenio_records_rest.serializers.json import (
     JSONSerializer as InvenioJSONSerializer,
 )
+from invenio_records_rest.utils import set_headers_for_record_caching_and_concurrency
 from invenio_search.utils import build_alias_name
 
 from inspirehep.accounts.api import is_user_logged_in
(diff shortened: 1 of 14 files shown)


backend/inspirehep/serializers.py
```diff
@@ -321,3 +322,59 @@ def dumps(obj, app=None, **kwargs):
...
+
+        if headers is not None:
+            response.headers.extend(headers)
+
+        return response
+
+    return view
+
+
+def search_responsify(serializer, mimetype):
```
drjova, line 350: can we add a param to be able to include links in the headers if we need to?

MJedr: we can

backend/inspirehep/serializers.py
```diff
@@ -321,3 +322,76 @@ def dumps(obj, app=None, **kwargs):
...
+                links=links,
+                item_links_factory=item_links_factory,
+            ),
+            mimetype=mimetype,
+        )
+        response.status_code = code
+        if headers is not None:
+            response.headers.extend(headers)
+
+        if links:
```
drjova, line 381: why we have different for search and record? shouldn't be something like `links_to_header True or False`

drjova (approved):

PR #2854, opened 2023-03-06
ui: fix for wrong margin in header

karolina-siemieniuk-morawska:


ui/src/common/layouts/Header/Header.jsx
```diff
@@ -45,7 +48,7 @@ function Header({ isHomePage, isSubmissionsPage, isBetaPage }) {
           </Row>
         </Layout.Header>
       </div>
-      <div className="non-sticky">
+      <div className={`non-sticky ${banners.length > 0 ? 'big-margin' : 'small-margin'}`}>
```
drjova, line 51: we could use `classNames` :)

drjova: we could use also the existing functions from tachyons, as well

drjova (approved):

PR #2867, opened 2023-03-14
add top level workflows

benjamin-bergia:


.github/workflows/build.yml
```diff
@@ -43,5 +43,9 @@ jobs:
           image: ${{ inputs.image }}
           context: ${{ inputs.context }}
           stage: ${{ inputs.stage }}
+          tags: |
```
benjamin-bergia, line 46: Here is the doc for this
https://github.com/marketplace/actions/docker-metadata-action#typeref

PR #2883, opened 2023-03-30
ui: clean up self curation testing features

karolina-siemieniuk-morawska:


drjova:
@karolina-siemieniuk-morawska lgtm, we have to update also the config.js with the flags

karolina-siemieniuk-morawska:
> @karolina-siemieniuk-morawska lgtm, we have to update also the config.js with the flags

it's updated on both envs :)

drjova (approved):

PR #2885, opened 2023-03-30
e2e: update host in records

MJedr:


backend/inspirehep/records/cli.py
```diff
@@ -36,11 +37,30 @@
 LOGGER = structlog.getLogger()
 
 
-def _create_record(data, save_to_file=False):
+def _replace_host_for_current_app_host(value):
+    value = value.replace(
+        "https://inspirebeta.net",
```
drjova, line 42: i would add regex to match prod and qa urls

drjova (approved):

PR #2888, opened 2023-03-30
tests: fix faker's `control_number`

MJedr:


backend/tests/helpers/providers/record_provider.py
```diff
@@ -241,7 +241,7 @@ def record(
 
         if data:
             record.update(data)
-        if with_control_number and "control_number" not in record:
+        if with_control_number or "control_number" in record:
```
drjova, line 244:
```suggestion
        if with_control_number or "control_number" not in record:
```

MJedr: actually let's wait for the CI fix and we can see if it works with smaller `randint` range

drjova (approved):
Let’s try it

PR #2889, opened 2023-03-31
docker: simplify ui images for local development

drjova:
* Adds a `Make` file with simple commands to run the project.


Makefile
```diff
@@ -0,0 +1,32 @@
...
+
+stop:
+	docker-compose down
+	kill -9 $(lsof -ti:8080)
+	kill -9 $(lsof -ti:8000)
+	kill -9 $(lsof -ti:3000)
+	kill -9 $(lsof -ti:4572)
+	kill -9 $(lsof -ti:4566)
+	kill -9 $(lsof -ti:6379)
+	kill -9 $(lsof -ti:5432)
```
MJedr, line 14: do we need to kill those processes?

Makefile
```diff
@@ -0,0 +1,32 @@
...
+sleep:
+	sleep 10
+
+ui:
+	docker-compose up -d ui
+	sleep 5
+
+setup:
+	docker-compose -f docker-compose.yml exec hep-web ./scripts/setup
+	docker-compose -f docker-compose.yml exec next-web inspirehep db create
```
MJedr, line 26: I would put next setup in a separate command since we're not using next that often I think

Makefile
```diff
@@ -0,0 +1,36 @@
...
+	echo -e "\033[0;32m Starting setup Inspire. \033[0m"
+	docker-compose -f docker-compose.yml exec hep-web ./scripts/setup
+	echo -e "\033[0;32m Finished setup Inspire. \033[0m"
+
+load-records:
+	echo -e "\033[0;32m Loading records. \033[0m"
+	docker-compose exec hep-web inspirehep importer demo-records
+
+next-setup:
+	docker-compose -f docker-compose.yml exec next-web inspirehep db create
```
MJedr, line 31: maybe we can exec all those setup steps actually (these are from `setup_inspire_next` script)? Cause for example we don't create HP indices with hep commands etc.
```
poetry run inspirehep db drop --yes-i-know
poetry run inspirehep alembic upgrade
poetry run inspirehep index destroy --force --yes-i-know
poetry run inspirehep index init
poetry run inspirehep fixtures init
```

MJedr (approved):
🚀

PR #2897, opened 2023-04-04
fix test

MJedr:
fixes `failed to parse field [control_number] of type [integer] in document with id '34ef628b-04cd-4737-b574-e7de7c956bc1'. Preview of field's value: '9999912588'` error in workflows


drjova (approved):
🚀

PR #2900, opened 2023-04-05
snow: add views and reply ticket option

MJedr:


backend/inspirehep/snow/views.py
```diff
@@ -0,0 +1,121 @@
...
+@parser.use_args(
+    {
+        "functional_category": fields.String(required=True),
+        "recid": fields.String(required=False),
+        "subject": fields.String(required=True),
+        "description": fields.String(required=False),
+        "caller_email": fields.String(required=False),
+    }
+)
+@login_required_with_roles([Roles.superuser.value])
```
drjova, line 27: shouldn't be curators?

backend/inspirehep/rt/templates/rt/curation_core.html
```diff
@@ -0,0 +1,38 @@
+{#
+## This file is part of INSPIRE.
+## Copyright (C) 2015 CERN.
```
drjova, line 3: let's update the licenses as well :)

drjova (commented):
great thank you! I would just rename the module to `tickets` or `curation_tickets` just not to do the same mistake (`rt` and now `snow` and in the future something else).

MJedr:
@drjova When creating snow module I thought we could clean up everything related to tickets (removing rt module, moving templates to snow module etc). Can we rename it afterward instead?

backend/inspirehep/snow/views.py
```diff
@@ -0,0 +1,121 @@
...
+        ticket = snow_instance.create_inspire_ticket_with_template(
+            template_context=args.get("template_context"),
+            template_path=template_path,
+            user_email=args.get("caller_email"),
+            functional_category=args.get("functional_category"),
+            subject=args.get("subject"),
+        )
+        return jsonify({"ticket_id": ticket}), 200
+    except (CreateTicketException, EditTicketException):
+        return jsonify({"message": "Can't create SNOW ticket!"}), 500
```
drjova, line 67: can we log what's going on in these cases?

drjova:
@MJedr we can do it after switching to snow 👍

backend/inspirehep/snow/views.py
MJedr: we can log if it's create ticket exception or edit ticket exception, I don't know if it would be very useful

drjova: It would be helpful to understand if something is wrong on our side or snow

drjova (approved):

PR #2911, opened 2023-04-19
fix logging in integration-async tests

MJedr:
ref: cern-sis/issues-inspire#283


backend/tests/helpers/utils.py
```diff
@@ -218,3 +219,13 @@ def before_record_response(response):
         return response
 
     return before_record_response
+
+
+def disable_low_level_logging(f):
```
drjova, line 224: Instead of this, wouldn't this work https://tenacity.readthedocs.io/en/latest/#before-and-after-retry-and-logging?

drjova: I see, my main concern is that we have to remember every time to add two decorators for this. We can add an auto-fixture for these modules that does this without us adding extra code.

MJedr: the issue is that I want to change the logging level for the assertion function only, we can try wrapping retry with this function, but I am not sure if that's a better solution

MJedr: The last option is adding a custom function that will change the level (I haven't tested it) https://tenacity.readthedocs.io/en/latest/#my_before_sleep, but then we'll need to pass so many parameters to `retry` so I am not convinced to this idea

drjova: We can combine both functions into 1 decorator and use only this one

backend/tests/helpers/utils.py
```diff
@@ -218,3 +220,26 @@ def before_record_response(response):
...
+    def wrapper(*args):
+        logging.disable(logging.ERROR)
+        result = f(*args)
+        logging.disable(logging.NOTSET)
+        return result
+
+    return wrapper
+
+
+def retry_with_low_level_logging_disabled(stop=None, wait=None):
```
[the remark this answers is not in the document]
MJedr: we can

drjova (approved):

PR #2916, opened 2023-04-25
script: add legacy_records xml upload script

PascalEgn:


backend/scripts/marcxml_export.py
MJedr, line 1: For writing cli we use click, you can have a look at how other cli's are done. Can you refactor your code to use it? 
Some other comments:
* Can you make arguments configurable?
* You can add the script to the existing file, like `records/cli.py`.
* Add tests for the functionality too.
* Could you use logger instead of printing?

backend/inspirehep/records/cli.py
```diff
@@ -338,3 +341,84 @@ def remove_bai_from_literature_records(
...
+    show_default=True,
+)
+@with_appcontext
+def export_xml(dir_path):
+    if not os.path.exists(dir_path):
+        try:
+            os.mkdir(dir_path)
+        except Exception as e:
+            LOGGER.info(f"There was a problem creating the dir: {e}")
+    if os.path.exists(dir_path):
```
MJedr, line 365: can it happen that it doesn't exist if you create it in other cases? then, you can remove if block

backend/inspirehep/records/cli.py
```diff
@@ -338,3 +341,84 @@ def remove_bai_from_literature_records(
...
+    default="/tmp/legacy_xmls",
+    show_default=True,
+)
+@with_appcontext
+def export_xml(dir_path):
+    if not os.path.exists(dir_path):
+        try:
+            os.mkdir(dir_path)
+        except Exception as e:
+            LOGGER.info(f"There was a problem creating the dir: {e}")
```
MJedr, line 364: Return here. It would be nice to catch a particular exception. Also, not all the exceptions stringify well.

backend/inspirehep/records/cli.py
```diff
@@ -338,3 +341,84 @@ def remove_bai_from_literature_records(
...
+    if os.path.exists(dir_path):
+        try:
+            query = LegacyRecordsMirror.query.all()
+            for row in query:
+                file_path = os.path.join(dir_path, f"{row.recid}.xml")
+                xml = open(file_path, "wb")
```
MJedr, line 370: why not to use a context manager `with open`?

backend/inspirehep/records/cli.py
```diff
@@ -338,3 +341,84 @@ def remove_bai_from_literature_records(
...
+                xml.write(row.marcxml)
+                xml.close()
+        except Exception as e:
```
MJedr, line 373: again, better to catch a specific exception

backend/inspirehep/records/cli.py
```diff
@@ -338,3 +341,84 @@ def remove_bai_from_literature_records(
...
+            LOGGER.info(f"There was a problem exporting the column: {e}")
+
+    LOGGER.info(
+        f"{len([f for f in os.listdir(dir_path) if os.path.isfile(os.path.join(dir_path, f))])} marcxml rows exported to xml."
```
MJedr, line 377: We use structlog, so you wirte a message like LOGGER.info("marcxml rows exported to xml.", rows_number=rows_number"). Can you also extract this list comprehension to a variable and make it more readable?

backend/inspirehep/records/cli.py
```diff
@@ -338,3 +341,84 @@ def remove_bai_from_literature_records(
...
+    default="inspire-qa-legacy",
+    show_default=True,
+)
+@with_appcontext
+def upload_xmls(dir_path, bucket):
+    mime_type = "application/xml"
+    acl = current_app.config["S3_FILE_ACL"]
+    try:
+        for filename in os.listdir(dir_path):
+            file = open(os.path.join(dir_path, filename), "rb")
```
MJedr, line 400: can you use context manager?

backend/inspirehep/records/cli.py
```diff
@@ -338,3 +341,84 @@ def remove_bai_from_literature_records(
...
+            key = "legacy_xml_" + filename
+            LOGGER.info(f"Uploading {key} to {bucket} bucket")
+            current_s3_instance.upload_file(file, key, filename, mime_type, acl, bucket)
+            file.close()
+        LOGGER.info("All files uploaded successfully.")
+    except Exception as e:
```
MJedr, line 406: same as before

backend/tests/integration/records/test_cli.py
```diff
@@ -232,3 +235,56 @@ def test_close_expired_jobs_ignores_deleted_records(inspire_app, cli):
...
+@mock.patch("click.confirm")
+def test_legacy_records_mirror_xml_cleanup(mock_click, cli, tmp_path):
+    xml_path = os.path.join(tmp_path, "legacy_records_xml")
+    os.mkdir(xml_path)
+    mock_click.return_value = "y"
+    open(os.path.join(xml_path, "1.xml"), "w").close
+    open(os.path.join(xml_path, "2.xml"), "w").close
+    open(os.path.join(xml_path, "3.xml"), "w").close
+    cli.invoke(["legacy_records", "cleanup_dir", "--dir-path", xml_path])
+    assert os.path.exists(xml_path) is False
```
MJedr, line 290:
```suggestion
    assert not os.path.exists(xml_path) 
```

backend/tests/integration/records/test_cli.py
```diff
@@ -232,3 +235,56 @@ def test_close_expired_jobs_ignores_deleted_records(inspire_app, cli):
...
+            "--dir-path",
+            xml_path,
+            "--bucket",
+            "inspire-tmp",
+        ]
+    )
+    result = current_s3_instance.file_exists(
+        key="legacy_xml_1.xml", bucket="inspire-tmp"
+    )
+    assert result is True
```
MJedr, line 278:
```suggestion
    assert result 
```

backend/inspirehep/records/cli.py
PascalEgn: Okay with that comment in mind, I can remove the if in the next line

backend/inspirehep/records/cli.py
```diff
@@ -338,3 +341,85 @@ def remove_bai_from_literature_records(
...
+@with_appcontext
+def export_xml(dir_path):
+    if not os.path.exists(dir_path):
+        try:
+            os.mkdir(dir_path)
+        except OSError as e:
+            LOGGER.info(f"There was a problem creating the dir: {e}")
+            return
+    try:
+        query = LegacyRecordsMirror.query.all()
```
MJedr, line 367: actually, it might be a lot of records, can you use `yield_per` instead of `all`?

MJedr (approved):

PR #2918, opened 2023-04-25
ui: self curation diff view

karolina-siemieniuk-morawska:
* ref: cern-sis/issues-inspire#254

View with linked current reference :
Screenshot 2023-04-27 at 16 38 53

View with unlinked current reference (don't display the 'to' box):
Screenshot 2023-04-27 at 16 36 19


drjova:
@michamos could you please check if the UI is ok for you?

karolina-siemieniuk-morawska:
@michamos I added better labels and fixed the mix-up with prev and current referenced data. Updated the screenshots, can you have a look?

michamos:
looks good now :+1:

ui/src/literature/index.tsx
```diff
@@ -0,0 +1,56 @@
...
+         * RouteOrRedirect with additional validation for loggedIn does the same job with success.
+         */}
+        {loggedIn && (
+          <RouteOrRedirect
+            redirectTo={ERROR_401}
+            path={`${LITERATURE}/:id/diff/:old..:new`}
+            condition={isUserAuthorized}
+            component={ReferenceDiffInterfaceContainer}
+          />
+        )}
```
drjova, line 45: We could use `<PrivateRoute>` which does the same thing :)

ui/src/actions/literature.ts
```diff
@@ -158,6 +185,29 @@ export function fetchLiteratureReferences(
   };
 }
 
+export function fetchReferencesDiff(  
+  recordId: number,
+  prevId: number,
```
MJedr, line 190: maybe you can name those vars like `prevVersionNumber` or `prevRevisionId` so that it won't be confusing with any kind of record ID

MJedr (approved):

drjova (approved):

PR #2932, opened 2023-05-03
add functions to set refereed

MJedr:
closes: cern-sis/issues-inspire#275


backend/inspirehep/curation/utils.py
```diff
@@ -0,0 +1,100 @@
...
+        not_a_conference_paper and published_in_a_refereed_journal_with_proceedings
+    ):
+        record["refereed"] = True
+    elif published_exclusively_in_non_refereed_journals:
+        record["refereed"] = False
+
+    if published_only_in_proceedings and published_only_in_non_refereed_journals:
+        try:
+            record["document_type"].remove("article")
+            record["document_type"].append("conference paper")
```
drjova, line 92: are we sure we don't produce duplicates?

MJedr: it's why there's try/catch (this part it's a copy-paste from next, I believe when there's an article we don't have a conference paper)

drjova (approved):

PR #2933, opened 2023-05-04
legacy_records: add xml export celery task

PascalEgn:

### DIFF
--- a/backend/inspirehep/records/cli.py
+++ b/backend/inspirehep/records/cli.py
@@ -6,16 +6,15 @@
 # the terms of the MIT License; see LICENSE file for more details.
 
 import datetime
+import math
 import os
 import re
-from io import BytesIO
 from itertools import islice
 
 import click
 import orjson
 import requests
 import structlog
-from botocore.exceptions import ClientError
 from flask import current_app
 from flask.cli import with_appcontext
 from flask_celeryext.app import current_celery_app
@@ -25,19 +24,19 @@ from invenio_records.api import RecordMetadata
 from sqlalchemy import DateTime, cast, not_, or_, type_coerce
 from sqlalchemy.dialects.postgresql import JSONB
 
-from inspirehep.files.api import current_s3_instance
 from inspirehep.mailing.api.jobs import send_job_deadline_reminder
 from inspirehep.migrator.models import LegacyRecordsMirror
 from inspirehep.pidstore.api import PidStoreBase
 from inspirehep.records.api import InspireRecord, JobsRecord
 from inspirehep.records.models import RecordsAuthors
 from inspirehep.records.tasks import (
+    export_legacy_recids,
     populate_journal_literature,
     regenerate_author_records_table_entries,
     remove_bai_from_literature_authors,
 )
 from inspirehep.search.api import LiteratureSearch
-from inspirehep.utils import chunker
+from inspirehep.utils import chunker, flatten_list
 
 LOGGER = structlog.getLogger()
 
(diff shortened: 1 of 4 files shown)


backend/inspirehep/records/cli.py
```diff
@@ -350,37 +350,65 @@ def legacy_records():
...
+    for count, recid_chunk in enumerate(recid_chunks):
+        bucket_name = f"{bucket}-{(count % bucket_count)}"
+        export_task = export_legacy_recids.s(
+            bucket_name, flatten_list(recid_chunk), db_batch_size
+        ).set(queue=task_queue)
+        export_tasks.append(export_task)
+
+    export_task_group = group(export_tasks)
+    export_group_result = export_task_group()
+    export_group_result.join()
```
MJedr, line 413: Why you're grouping tasks? Is there any particular reason for that? In my opinion, you can create tasks with `export_legacy_recids.delay()`, grouping in this case is not needed (without it you'll create ~12000 tasks, which is totally fine) and makes the code more complex. You can just send a task with a list of recids and then no need for chunking in the task.

backend/inspirehep/records/tasks.py
```diff
@@ -236,3 +242,40 @@ def reference_self_curation(
...
+            upload_counter += 1
+        except ClientError as e:
+            LOGGER.warning(
+                "There was an issue uploading to s3",
+                filename=filename,
+                bucket_name=bucket_name,
+                stacktrace=e,
+            )
+        if upload_counter % math.ceil(recid_len / 10) == 0:
+            LOGGER.info("files uploaded", upload_count=upload_counter)
```
MJedr, line 281: can't you just log in in the end of your CLI?

backend/inspirehep/records/tasks.py
```diff
@@ -236,3 +242,40 @@ def reference_self_curation(
...
+            current_s3_instance.upload_file(
+                fileob, key, filename, mime_type, acl, bucket_name
+            )
+            upload_counter += 1
+        except ClientError as e:
+            LOGGER.warning(
+                "There was an issue uploading to s3",
+                filename=filename,
+                bucket_name=bucket_name,
+                stacktrace=e,
```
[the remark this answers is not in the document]
PascalEgn: They use the `__str__` function in their own [documentation](https://botocore.amazonaws.com/v1/documentation/api/latest/client_upgrades.html#error-handling) . But also mention `e.msg`, i will change it :+1:

backend/inspirehep/records/tasks.py
PascalEgn: Doesn't only the tasks know if the upload was successful? thats why i log it there.

backend/inspirehep/records/cli.py
PascalEgn: I used grouping to know when every task is finished, so the script only finishes if every task comes to an end. I agree this might cause problems so yeah i will remove it and use `.delay()`. 
But i dont really understand why the chunking is not needed, i thought its quiet useful to get a list of for example 1000 recids each iteration.

backend/inspirehep/records/tasks.py
MJedr: yes, but I meant to remove the check, why do you need that? If you process all the documents they're uploaded or am I wrong?

PascalEgn: No you are right, its just to see the upload progress of each task.

MJedr: I still don't understand why you can't just log `files are updated` at the end of the loop, without the `if` condition. Then, you don't need a counter (also, you can use enumerate, it's more "elegant" I would say), especially, since you will upload 100 files only in one batch.

backend/inspirehep/records/cli.py
```diff
@@ -350,37 +349,61 @@ def legacy_records():
 
 
 @legacy_records.command(help="Upload marcxml column to S3.")
+@click.option(
+    "--task-queue",
+    help="Celery task queue",
+    default="xml_export",
```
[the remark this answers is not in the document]
PascalEgn: Harris wanted it to be flexible so we can have a special queue for it. I used `migrator` in the tests, so it will run without creating a new queue. I will change the default to `migrator` as well.

backend/inspirehep/records/tasks.py
PascalEgn: I will remove it. And I agree, enumerate would be more elegant but I think I would not be able to use it here because I only want to count the successful uploads.

backend/inspirehep/records/cli.py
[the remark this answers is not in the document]
MJedr: It's ok, you can have a special queue, but then please test it with a new setting. Otherwise, it's not tested properly.

MJedr (approved):

PR #2940, opened 2023-05-10
migrator: remove migrator module & tests

PascalEgn:


drjova:
@PascalEgn it looks good, could we please create also an alembic recipe that drops the table?

backend/inspirehep/alembic/3fd6471bb960_remove_legacy_records_mirror_table.py
```diff
@@ -0,0 +1,51 @@
...
+        "legacy_records_mirror",
+        ["valid", "collection"],
+        unique=False,
+    )
+    op.create_index(
+        "ix_legacy_records_mirror_last_updated",
+        "legacy_records_mirror",
+        ["last_updated"],
+        unique=False,
+    )
```
drjova, line 51: we don't have this on the definition, we don't need it.

backend/inspirehep/alembic/3fd6471bb960_remove_legacy_records_mirror_table.py
```diff
@@ -0,0 +1,51 @@
...
+from sqlalchemy.dialects import postgresql
+
+# revision identifiers, used by Alembic.
+revision = "3fd6471bb960"
+down_revision = "72d010d89702"
+branch_labels = ()
+depends_on = None
+
+
+def upgrade():
```
drjova, line 21: We should also update https://github.com/inspirehep/inspirehep/blob/f28e872812dcde800f6b8090691a687c79896b9a/backend/tests/integration/alembic/test_alembic.py#L15 :)

backend/inspirehep/alembic/3fd6471bb960_remove_legacy_records_mirror_table.py
PascalEgn: We need it because this downgrade will fail if its missing: https://github.com/inspirehep/inspirehep/blob/master/backend/inspirehep/alembic/5ce9ef759ace_add_lecacy_records_mirror_table.py#LL48C1-L59C35

drjova (approved):
LGTM, let's wait first to finish with files uploading before merging it.

PR #2944, opened 2023-05-15
move matching from next to hep

MJedr:


backend/inspirehep/matcher/views.py
```diff
@@ -1,13 +1,48 @@
...
+@blueprint.route("/fuzzy-match", methods=["GET"])
+@login_required_with_roles([Roles.cataloger.value])
+@parser.use_args(
+    {
+        "data": fields.Dict(required=True),
+    },
+    locations=("json",),
+)
+def fuzzy_match(args):
+    match_result = fuzzy_worfklow_match(args["data"])
```
drjova, line 47: can we add metrics as well? To check the execution time etc?

backend/inspirehep/matcher/api.py
```diff
@@ -239,3 +242,62 @@ def get_affiliations_from_pdf(url, **kwargs):
...
+
+    Arguments:
+        data: data used for matching.
+
+    Returns:
+        List: List of matched ids
+
+    """
+    exact_match_workflow_matcher_config = current_app.config[
+        "EXACT_WORKFLOW_MATCH_CONFIG"
```
drjova, line 264: Let's rename everything with workflow to generic names, we might use it in other cases too.

backend/inspirehep/matcher/api.py
```diff
@@ -239,3 +242,62 @@ def get_affiliations_from_pdf(url, **kwargs):
...
+
+
+def exact_match_workflow_data(data):
+    """Return matched record ids if the record is already present in the system.
+
+    Uses the default configuration of the ``inspire-matcher`` to find
+    duplicates of the current workflow object in the system.
+
+    Also sets the ``matches.exact`` property in ``extra_data`` to the list of
+    control numbers that matched.
```
drjova, line 254: is that still true?

backend/inspirehep/matcher/api.py
```diff
@@ -239,3 +242,62 @@ def get_affiliations_from_pdf(url, **kwargs):
...
+
+
+def fuzzy_match_workflow_data(data):
+    """Return list of similar records are found in the system.
+
+    Uses a custom configuration for ``inspire-matcher`` to find records
+    similar to the current workflow object's payload in the system.
+
+    Also sets the ``matches.fuzzy`` property in ``extra_data`` to the list of
+    the brief of first 5 record that matched.
```
drjova, line 278: ditto

backend/inspirehep/matcher/utils.py
```diff
@@ -123,3 +123,39 @@ def normalize_title(raw_title):
     normalized_title = normalized_title.upper()
 
     return normalized_title
+
+
+def get_hep_record_summary(hep_record):
```
drjova, line 128: would it be hard to use marshmallow?

backend/inspirehep/matcher/api.py
MJedr: Well, I am not sure, from the one hand I know what you mean (hep is a separate service), but on the other hand, if we don't add a prefix it might be confusing - exact match or fuzzy match what (which index)?

drjova: hmm I see, what if we  do something like `EXACT_LITERATURE_MATCH_CONFIG`?

PR #2950, opened 2023-05-23
search: update institutions mapping

MJedr:
ref: cern-sis/issues-inspire#311


docker-compose.services.yml
```diff
@@ -57,7 +57,7 @@ services:
       - discovery.seed_hosts=opensearch-node1
       - bootstrap.memory_lock=true
       - discovery.type=single-node
-      - 'OPENSEARCH_JAVA_OPTS=-Xms1024m -Xmx1024m'
+      - 'ES_JAVA_OPTS=-Xms512m -Xmx512m'
```
drjova, line 60: is it really the case? :/

PR #2951, opened 2023-05-23
ui: fix for self curation

karolina-siemieniuk-morawska:
* ref: cern-sis/issues-inspire#237

### DIFF
--- a/ui/src/actions/literature.ts
+++ b/ui/src/actions/literature.ts
@@ -42,7 +42,7 @@ import {
   assignLiteratureItemError,
   ASSIGNING_NOTIFICATION_KEY,
   ASSIGNING_NOTIFICATION_LITERATURE_ITEM_KEY,
-  CURATING_NOTIFICATION_KEY
+  CURATING_NOTIFICATION_KEY,
 } from '../literature/assignNotification';
 import { LITERATURE_REFERENCES_NS } from '../search/constants';
 import { searchQueryUpdate } from './search';
@@ -303,16 +303,18 @@ export function setCurateDrawerVisibility(referenceId: number | null) {
 
 export function curateReference({
   recordId,
+  recordUuid,
   revisionId,
   referenceId,
   newReferenceId,
 }: {
   recordId: number;
+  recordUuid: string;
   revisionId: number;
   referenceId: number;
   newReferenceId: number;
 }): (
-  dispatch: ActionCreator<Action>,
+  dispatch: any,
   getState: () => RootStateOrAny,
   http: HttpClientWrapper
 ) => Promise<void> {
@@ -320,12 +322,15 @@ export function curateReference({
     try {
       curating(CURATING_NOTIFICATION_KEY);
       await http.post('/literature/reference-self-curation', {
-        record_id: recordId.toString(),
+        record_id: recordUuid,
         revision_id: revisionId,
         reference_index: referenceId,
         new_reference_recid: newReferenceId,
       });
-      curationSuccess();
+      // fetch record data to refresh revision_id
+      dispatch(fetchLiterature(recordId)).then(() => {
+        curationSuccess();
+      });
     } catch (error) {
       curationError(CURATING_NOTIFICATION_KEY);
     }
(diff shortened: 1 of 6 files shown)


ui/src/actions/literature.ts
```diff
@@ -302,30 +302,35 @@ export function setCurateDrawerVisibility(referenceId: number | null) {
...
 export function curateReference({
-  recordId,
+  id,
+  recordUuid,
   revisionId,
   referenceId,
   newReferenceId,
 }: {
-  recordId: number;
+  id: number;
```
drjova, line 311: isn't this the `record_id`? We are reading it from the router right?

ui/src/literature/containers/CurateReferenceDrawerContainer.tsx
```diff
@@ -17,21 +18,29 @@ const stateToProps = (state: RootStateOrAny, { namespace }: { namespace: string
...
     newReferenceId: number;
   }) {
-    dispatch(curateReference({recordId, revisionId, referenceId, newReferenceId}));
+    dispatch(curateReference({id, recordUuid, revisionId, referenceId, newReferenceId}));
+  },
+
+  fetchSearchResults(namespace: string) {
+    dispatch(searchBaseQueriesUpdate(namespace, {
+      baseQuery: { q: '' },
+    }));
```
drjova, line 43: I think here we can go back to the initial state of the drawer

drjova: actually, with `destroyOnClose` do we even need it?

karolina-siemieniuk-morawska: It's to prepopulate the drawer with literature search results on initial enter, without the additional step of having to click search on empty search input. I thought it would be better this way, so the user sees all literature search results once the drawer opens, and the searchbar will only help to narrow the results.

ui/src/actions/literature.ts
karolina-siemieniuk-morawska: We need two parameters here -  _recordUuuid_ is uuid, and _id_ is control_number. Should I rename id to recordId?

drjova: it will be more clear :)

ui/src/literature/containers/CurateReferenceDrawerContainer.tsx
drjova: In all other drawers we have it like that, AFAIR. Could we please double check, if that's the case let's keep it consistent

karolina-siemieniuk-morawska: sure 👍

ui/src/literature/components/CurateReferenceDrawer/CurateReferenceDrawer.tsx
```diff
@@ -66,11 +66,13 @@ export function renderReferenceItem(result: Map<string, any>) {
 
 interface CurateReferenceDrawerProps {
   recordId: number;
+  recordUuid: number;
   revisionId: number;
   referenceId: number;
   onDrawerClose: () => void;
   onCurate: (args: {
     recordId: number;
+    recordUuid: number;
```
drjova, line 75: the uuid it's string :)

ui/src/actions/literature.ts
```diff
@@ -303,29 +303,34 @@ export function setCurateDrawerVisibility(referenceId: number | null) {
...
+  dispatch: any,
   getState: () => RootStateOrAny,
   http: HttpClientWrapper
 ) => Promise<void> {
   return async (dispatch, getState, http) => {
     try {
       curating(CURATING_NOTIFICATION_KEY);
       await http.post('/literature/reference-self-curation', {
-        record_id: recordId.toString(),
+        record_id: recordUuid.toString(),
```
drjova, line 325: we don't need the `toString` 🙃

drjova (approved):

PR #2970, opened 2023-06-13
Refresh references after edit

karolina-siemieniuk-morawska:


ui/src/actions/literature.ts
```diff
@@ -324,13 +324,14 @@ export function curateReference({
       await http.post('/literature/reference-self-curation', {
         record_id: recordUuid,
         revision_id: revisionId,
-        reference_index: referenceId,
+        reference_index: referenceId - 1,
```
drjova, line 327: why we need the `-1`?

karolina-siemieniuk-morawska: On the backend the reference id is indexed from 0 while on ui from 1, so when I'm sending the reference id from ui it changes the next one in line on the backend

drjova: isn't a list on UI as well? can we just pass the index of the reference? Are we using something else?

karolina-siemieniuk-morawska: The list is on UI, but if an element has an index of 1 on UI on backend it has index 0. That's why when I was passing index straight as it is on UI it was editing the wrong reference.

karolina-siemieniuk-morawska: Screenshot 2023-06-13 at 14 23 40
We get the reference id like this from the backend, it doesn't send us the "real" index which is counted from 0, that's why I need to subtract 1

karolina-siemieniuk-morawska: You're right, I instead using label I will use index

ui/src/literature/components/ReferenceList.tsx
```diff
@@ -36,11 +36,13 @@ function ReferenceList({
       <ReferenceItem
         key={reference.getIn(['titles', 0, 'title']) || String(index)}
         reference={reference}
+        referenceId={index}
+        total={total}
```
drjova, line 40: can we already calculate this here? instead of passing down the `total`

karolina-siemieniuk-morawska: I pass total so that the query for refreshing references after edit has correct page. Example: if we go to page 123 of references in one record, and then go to another record that has only one page of references and edit, value of query.page in state would still be 123 and result in incorrect request, so I do this:
Screenshot 2023-06-14 at 14 27 50

karolina-siemieniuk-morawska:
@drjova I deleted your comment instead of mine by accident 🥲. I deleted unnecessary total and now PR is ready for review.

drjova (approved):

PR #2975, opened 2023-06-22
e2e: add tests for self curation

karolina-siemieniuk-morawska:
* ref: cern-sis/issues-inspire#312

### DIFF
--- a/e2e/cypress/integration/references.test.js
+++ b/e2e/cypress/integration/references.test.js
@@ -1,0 +1,317 @@
+import { onlyOn } from '@cypress/skip-test';
+
+describe('References', () => {
+  describe('Reference Search', () => {
+    onlyOn('headless', () => {
+      it('matches image snapshot for reference search', () => {
+        cy.registerRoute();
+        cy.visit(
+          '/literature?sort=mostrecent&size=25&page=1&q=citedby%3Arecid%3A1322719'
+        );
+        cy.waitForRoute();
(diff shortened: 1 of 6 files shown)


e2e/cypress/integration/references.test.js
```diff
@@ -0,0 +1,233 @@
...
+        cy.get('[data-test-id="curate-button"').click();
+
+        cy.waitForLoading();
+        cy.wait(5000);
+
+        cy.get('.ant-notification-notice-message').should(
+          'have.text',
+          'Reference Successfully Modified!'
+        );
+      });
```
MJedr, line 122: is it possible to check if the reference text changed?

e2e/cypress/integration/references.test.js
MJedr, line 1: can we have a test case with successful edit of the reference not from the first page, in which we'll verify that the index sent to backend is correct?

karolina-siemieniuk-morawska: it will be possible after https://github.com/cern-sis/issues-inspire/issues/345 is done. Also in line 124 we have this test case: 'curates reference successfully when page is changed' which tests that

e2e/cypress/integration/references.test.js
karolina-siemieniuk-morawska: added

drjova (approved):

PR #2998, opened 2023-07-26
Opensearch

MJedr:


backend/inspirehep/pidstore/models.py
```diff
@@ -144,5 +144,6 @@ def delete(self):
             self.original_pid.delete()
             db.session.delete(self)
             db.session.expire(self)
+            db.session.flush([self])
```
drjova, line 147: that's weird, we need to do that for opensearch? I just saw

PR #3002, opened 2023-07-27
add chrome version

karolina-siemieniuk-morawska:


.github/workflows/e2e-tests.yml
```diff
@@ -184,6 +185,12 @@ jobs:
       - name: import dataset
         run: >
           docker exec hep-web inspirehep importer demo-records
+      
+      - name: Setup chrome
+        id: setup-chrome
+        uses: browser-actions/setup-chrome@v1
+        with:
+          chrome-version: ${{ matrix.chrome }}
```
MJedr, line 193: I think you don't need to put it in the matrix, you can just pass the version directly since we're not planning to test multiple versions.

MJedr (approved):

PR #3003, opened 2023-07-27
Fix cypress

karolina-siemieniuk-morawska:


MJedr (approved):
you have to rebase ;

.github/workflows/e2e-tests.yml
```diff
@@ -189,7 +189,7 @@ jobs:
         id: setup-chrome
         uses: browser-actions/setup-chrome@v1
         with:
-          chrome-version: '114'
+          chrome-version: '999988'
```
drjova, line 192: what's that? :)

MJedr: chromium by commit position https://github.com/browser-actions/setup-chrome#parameters

karolina-siemieniuk-morawska: With this action [you can only specify chrome version as stable, beta, dev, and canary](https://github.com/browser-actions/setup-chrome#setup-chrome), directly specifying chrome version is not implemented yet. But you can specify chromium commit number which represents version, so I chose this one.

PR #3016, opened 2023-08-15
e2e: updated to cypress 12

karolina-siemieniuk-morawska:
1. Updated Cypress from 8 to 12, Node to 16
2. Added Firefox tests back
3. Snapshot tests don't work due to 1-5 pixel differences between those generated in CI and in local docker/locally
4. Submission tests don't work - might be related to https://github.com/cypress-io/cypress/issues/27415 and https://github.com/cypress-io/cypress/issues/26900, needs more investigating

* ref: cern-sis/issues-inspire#358


drjova:
@karolina-siemieniuk-morawska the commit message is wrong, it should be cypress 12 :)

PR #3023, opened 2023-08-18
add chatbot role

MJedr:


backend/inspirehep/accounts/roles.py
```diff
@@ -13,3 +13,4 @@ class Roles(Enum):
     hermescoll = "hermescoll"
     hermescurator = "hermescurator"
     jlabcurator = "jlabcurator"
+    chatbot = "chatbot"
```
drjova, line 16: not sure if we need it here :)

drjova (approved):

PR #3069, opened 2023-09-28
search: fix jobs field of interest filter

MJedr:
ref: cern-sis/issues-inspire#396

### DIFF
--- a/backend/tests/integration/search/test_aggregations.py
+++ b/backend/tests/integration/search/test_aggregations.py
@@ -414,6 +414,8 @@ def test_jobs_field_of_interest_aggregation_and_filter(inspire_app, override_con
         expected_record = create_record("job", data)
         data = {"arxiv_categories": ["hep-ex"], "status": "open"}
         create_record("job", data)
+        data = {"status": "open"}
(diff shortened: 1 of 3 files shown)


backend/tests/integration/search/test_aggregations.py
```diff
@@ -434,6 +437,13 @@ def test_jobs_field_of_interest_aggregation_and_filter(inspire_app, override_con
             response["hits"]["hits"][0]["metadata"]["control_number"]
             == expected_record["control_number"]
         )
+        with inspire_app.test_client() as client:
+            response = client.get("/jobs?field_of_interest=Other").json
```
drjova, line 441: do we have a test case with the Unknown?

MJedr: It's an old functionality which is working, and there's already a test for it https://github.com/inspirehep/inspirehep/blob/0be447100b82bfc965f46e0836a14dc8fa903959/backend/tests/integration/search/test_aggregations.py#L354-L355

drjova (approved):

PR #3081, opened 2023-10-19
snow: add user to watchlist

MJedr:


backend/inspirehep/snow/views.py
```diff
@@ -104,6 +108,8 @@ def reply_ticket(args):
     snow_instance = InspireSnow()
     try:
         snow_instance.comment_ticket(args["ticket_id"], args["reply_message"])
+        if args["user_email"]:
```
drjova, line 111: the `user_email` is not required and might return `KeyError`

backend/inspirehep/snow/views.py
```diff
@@ -86,6 +87,8 @@ def reply_ticket_with_template(args):
         snow_instance.comment_ticket_with_template(
             args["ticket_id"], template_path, args.get("template_context", {})
         )
+        if args["user_email"]:
```
drjova, line 90: same

drjova (approved):

PR #3084, opened 2023-10-23
records: add filesize exceeded error code

PascalEgn:
Add http error code to `FileSizeExceededError`. This should lead to [except block](https://github.com/inspirehep/inspire-next/blob/fc6287f551d1cd770207549cb126ab8c1dce0170/inspirehep/modules/workflows/tasks/upload.py#L136) in inspire-next getting triggered for better error handling instead of running into the keyError.

ref: https://github.com/cern-sis/issues-inspire/issues/399


MJedr (commented):
could you please add a test (actually assertion to already existing test)?

PascalEgn:
> could you please add a test (actually assertion to already existing test)?

Added a assertion for the HTTP error code and removed some unused variables from the tests (because black will complain otherwise)

MJedr (approved):

PR #3093, opened 2023-11-08
add missing snow template

MJedr:


drjova:
hmm It doesn't seem that we're using this template anywhere

drjova (approved):

PR #3114, opened 2024-01-25
e2e: improve local setup

karolina-siemieniuk-morawska:
* ref: cern-sis/issues-inspire#393


e2e/docker-compose.cypress.dev.yml
```diff
@@ -2,7 +2,7 @@ version: '2.1'
 
 services:
   cypress:
-    image: cypress/included:12.17.0
+    image: cypress/included:13
```
drjova, line 5: do we need both compose files?

drjova:
Also @karolina-siemieniuk-morawska could we cleanup the cypress-*.sh files?

e2e/docker-compose.cypress.dev.yml
karolina-siemieniuk-morawska: It says [here in readme](https://github.com/inspirehep/inspirehep/blob/e869894d75c61878112e71277156409cf70b40de/README.md#:~:text=To%20enable%20mounting%20backend%20code%20and%20live%20update%2C%20just%20use%20e2e/docker%2Dcompose.cypress.dev.yml%20instead.) that it was created with a purpose, but tbh I never used it for anything so I can delete

drjova: ok, let's also cleanup the README file please

karolina-siemieniuk-morawska:
> Also @karolina-siemieniuk-morawska could we cleanup the cypress-*.sh files?
What kind of clean up are you thinking about? Like merging them all into one file?

drjova:
see what we need actually, do we need them all? does it make sense to have the separately etc Happy to discuss IRL

drjova (approved):

PR #3123, opened 2024-02-01
hal: fix exception raise on _hal_create

PascalEgn:
ref: https://github.com/cern-sis/issues-inspire/issues/403


backend/inspirehep/hal/tasks.py
```diff
@@ -108,6 +109,8 @@ def _hal_create(tei, record):
         if "duplicate-entry" in message:
             hal_id = re.findall("hal-[0-9]{8}", message)[0]
             receipt = _hal_update(tei, hal_id, record)
+        else:
+            raise HALCreateException(message)
```
drjova, line 113: can we add please a test case for that?

drjova (approved):

PR #3155, opened 2024-03-19
holdingpen: initial cleanup

karolina-siemieniuk-morawska:
- Rewrite files in ```/holdingpen``` directory to Typescript
- Rewrite tests to use RTL and TS
- Add new routes for search and detail pages
- Add new components and unit tests 

* ref: cern-sis/issues-inspire#458


ui/src/common/routes.ts
```diff
@@ -23,10 +23,9 @@ export const USER_SIGNUP = `${USER}/signup`;
 export const USER_LOCAL_LOGIN = `${USER_LOGIN}/local`;
 export const USER_SETTINGS = `${USER}/settings`;
 
-// TODO: rename `holdingpen` to `admin`
 export const HOLDINGPEN = '/holdingpen';
 export const HOLDINGPEN_DASHBOARD = `${HOLDINGPEN}/dashboard`;
-export const HOLDINGPEN_INSPECT = `${HOLDINGPEN}/inspect`;
```
drjova, line 29: actually, we have to keep the inspect, is a tool for curators. Any particular reason why we removed it?

karolina-siemieniuk-morawska: I thought it's not needed, I'll revert :)

PR #3166, opened 2024-04-16
matcher: request explicitly XML format in grobid

drjova:
* ref: cern-sis/issues-inspire#462


PascalEgn:
Don't we also need it here?
https://github.com/inspirehep/inspire-next/blob/fab1c19d33d7a5c6ba543aedcc9f9ddef814787a/inspirehep/modules/workflows/tasks/actions.py#L1094

PR #3179, opened 2024-06-04
holdingpen: author submission UI prototype

karolina-siemieniuk-morawska:
* ref: cern-sis/issues-inspire#468

In this PR I revert changes made in the old ```/holdingpen``` directory to the stage before any changes, and move everything related to new holdingpen to new directory, ```/holdingpen-new```. This is to ensure it will not collide with the currently used tool once we merge to QA. New directory is only accessible for ```superuser``` and ```cataloger``` user roles. 

localhost_3000_holdingpen_author_123456


drjova:
@karolina-siemieniuk-morawska have we deployed it?

PR #3180, opened 2024-06-04
update PDG identifiers to use latex description

karolina-siemieniuk-morawska:
* ref: cern-sis/issues-inspire#471

Screenshot 2024-06-04 at 15 16 53


backend/inspirehep/records/marshmallow/literature/ui.py
```diff
@@ -43,7 +43,7 @@
     PublicationInfoItemSchemaV1,
     ThesisInfoSchemaV1,
 )
-from .pdg_identifiers import PDG_IDS_TO_DESCRIPTION_MAPPING
```
drjova, line 46: do we use anywhere else the `pdg_identifiers` file? if not, we can remove it.

drjova (approved):

PR #3184, opened 2024-06-13
assign: fix author `can_claim` check

drjova:
* ref: cern-sis/issues-inspire#467

### DIFF
--- a/backend/inspirehep/assign/utils.py
+++ b/backend/inspirehep/assign/utils.py
@@ -4,6 +4,7 @@
 #
 # inspirehep is free software; you can redistribute it and/or modify it under
 # the terms of the MIT License; see LICENSE file for more details.
+
 import structlog
 from flask import request
 from inspire_dojson.utils import get_recid_from_ref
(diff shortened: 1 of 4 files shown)


backend/inspirehep/assign/utils.py
```diff
@@ -108,30 +110,35 @@ def can_claim(data, author_profile_recid):
...
+        if "," not in name:
+            names = name.split()
+        else:
+            names = parsed_name.last_list
+
+        last_names = set()
+        for last_name in names:
+            last_name = unidecode(last_name)
+            last_names.add(last_name)
+        return last_names
```
michamos, line 129:
```suggestion
        last_names = set()
        for last_name in names:
            last_name = unidecode(last_name)
            last_names.add(last_name)
        return last_names
```
```suggestion
        return {unidecode(name) for name in last_names}
```

backend/inspirehep/assign/views.py
```diff
@@ -166,7 +166,7 @@ def assign_different_profile(args):
 
     for literature_id in literature_ids:
         record = LiteratureRecord.get_record_by_pid_value(literature_id)
-        if record.get("curated") and not is_from_author_stub:
+        if record.get("curated_relation") and not is_from_author_stub:
```
michamos, line 169: this is incorrect, you need to check the `curated_relation` field from the matched author, there is no `curated_relation` in the record. It might be good to add a test too for this.

michamos (commented):
the name logic seems correct now :+1: but the check of whether the signature has been claimed already isn't.

backend/inspirehep/assign/views.py
drjova: Ok, I see thanks. I misunderstood your requirement

michamos (approved):
looks good now :+1:

PR #3186, opened 2024-06-14
Add FEATURE_FLAG_ENABLE_SEND_TO_BACKOFFICE flag

DonHaul:
A few things:
send_post_request_to_inspire_next has now 3 extra parameters:
- url: to define if its whea the next or backoffice
- token: to pass either the next or backoffice
- bearer_keyword: because currently its Bearer for next and Token for backoffice (default keyword in django) - should it be changed also to Bearer?

Wrapped around try except so that does'st break normal functionality in case it fails


backend/inspirehep/config.py
```diff
@@ -32,6 +32,7 @@
 FEATURE_FLAG_ENABLE_ORCID_PUSH = False
 FEATURE_FLAG_ENABLE_REDIRECTION_OF_PIDS = False
 FEATURE_FLAG_ENABLE_ASSIGN_AUTHOR_PAPERS = False
+FEATURE_FLAG_ENABLE_SEND_TO_BACKOFFICE = True
```
drjova, line 35: This should be always `False` by default

backend/inspirehep/submissions/views.py
```diff
@@ -191,7 +200,34 @@ def load_data_from_request(self):
     def start_workflow_for_submission(self, record):
         record["acquisition_source"] = self.get_acquisition_source()
         payload = {"data": record}
-        return self.send_post_request_to_inspire_next("/workflows/authors", payload)
+
+        if current_app.config.get("FEATURE_FLAG_ENABLE_SEND_TO_BACKOFFICE"):
+            try:
+                payload_backoffice = {
+                    "data": record,
+                    "workflow_type": "AUTHOR_CREATE",
```
drjova, line 208: this function is used for both create an update

backend/inspirehep/submissions/views.py
```diff
@@ -191,7 +200,34 @@ def load_data_from_request(self):
...
+                    "/api/workflows/",
+                    payload_backoffice,
+                    current_app.config["BACKOFFICE_BEARER_TOKEN"],
+                    bearer_keyword="Token",
+                )
+            except WorkflowStartError:
+                print(
+                    f"Error creating record in {current_app.config['INSPIRE_BACKOFFICE_URL']} ",
+                    str(payload_backoffice),
+                )
```
drjova, line 224: we don't need to catch this error. In general we don't use print, instead we log the errors because it will send them to monitoring OpenSearch

backend/inspirehep/submissions/views.py
DonHaul: Okay, besides passing a parameter to `start_workflow_for_submission`  for create/update depending if method POST or PUT is used, is there any way to discern the two?

DonHaul:
Added request_mock for the failing testing. To confirm why other similar tests are not failing namely 
- test_new_author_submit_with_workflows_error
- test_new_author_submit

backend/inspirehep/submissions/views.py
```diff
@@ -191,7 +200,28 @@ def load_data_from_request(self):
     def start_workflow_for_submission(self, record):
         record["acquisition_source"] = self.get_acquisition_source()
         payload = {"data": record}
-        return self.send_post_request_to_inspire_next("/workflows/authors", payload)
+
+        if current_app.config.get("FEATURE_FLAG_ENABLE_SEND_TO_BACKOFFICE"):
+            payload_backoffice = {
+                "data": record,
+                "workflow_type": "AUTHOR_CREATE",
```
drjova, line 207: what about updates? 🤔

backend/inspirehep/submissions/views.py
```diff
@@ -191,7 +200,28 @@ def load_data_from_request(self):
...
+                "status": "running",
+                "core": False,
+                "is_update": True,
```
drjova, line 210: These seem unnecessary, `is_update` is conflicting with `AUTHOR_CREATE`

backend/inspirehep/submissions/views.py
```diff
@@ -130,7 +139,13 @@ def get(self, pid_value):
 
     def post(self):
         data = self.load_data_from_request()
-        return self.start_workflow_for_submission(data)
+        payload = {
+            "data": data,
+            "workflow_type": "AUTHOR_CREATE",
+            "core": False,
+            "is_update": False,
```
drjova, line 146: aren't these the default values?

backend/inspirehep/submissions/views.py
```diff
@@ -148,7 +163,13 @@ def put(self, pid_value):
         db.session.commit()
 
         if current_app.config.get("FEATURE_FLAG_ENABLE_WORKFLOW_ON_AUTHOR_UPDATE"):
-            self.start_workflow_for_submission(record)
+            payload = {
+                "data": record,
+                "workflow_type": "AUTHOR_UPDATE",
+                "core": False,
```
drjova, line 169: we can omit this, it's authors

backend/inspirehep/submissions/views.py
```diff
@@ -386,7 +426,13 @@ def start_workflow_for_submission(self, control_number=None):
         }
         submission_data["acquisition_source"] = self.get_acquisition_source()
         payload = {"data": submission_data, "form_data": form_data}
-        return self.send_post_request_to_inspire_next("/workflows/literature", payload)
+
+        return self.send_post_request_to_inspire_next(
+            current_app.config["INSPIRE_NEXT_URL"],
+            "/workflows/literature",
+            payload,
```
drjova, line 433: here we need only record, not the rest

drjova (approved):
It looks good!

PR #3188, opened 2024-07-03
tests: remove fixed ids from integration and async tests

PascalEgn:
Remove most of the fixed control numbers from integration and integration-async tests. 

ref: https://github.com/cern-sis/issues-inspire/issues/476


karolina-siemieniuk-morawska (approved):
LGTM, and all the tests passed at first attempt 🎉🎉

PR #3197, opened 2024-07-19
ui: set up and use redux in new holdingpen

karolina-siemieniuk-morawska:
* ref: cern-sis/issues-inspire#481


ui/src/actions/holdingpen.ts
```diff
@@ -0,0 +1,230 @@
...
+    if (axios.isAxiosError(error)) {
+      if (
+        error.response?.status === 403 &&
+        (!(error.config as any)._retry as boolean)
+      ) {
+        (error.config as any)._retry = true;
+        try {
+          const accessToken = await refreshToken();
+          axios.defaults.headers.common.Authorization = `Bearer ${accessToken}`;
+          return httpClient(error.config!);
```
drjova, line 61:
```suggestion
          return httpClient(error.config!);
```
It has an extra `!`

karolina-siemieniuk-morawska: it's actually a typescript operator (non-null assertion) to remove null and undefined from type (Axios's error.config has a type of ```InternalAxiosRequestConfig<any> | undefined```, and httpClient doesn't accept undefined, thus this operator assures that it can accept it

drjova (approved):

ui/src/actions/holdingpen.ts
karolina-siemieniuk-morawska: it could be changed to ```return httpClient(error.config || {});```, but then the try catch wouldn't catch ```tokenError``` (that's what I thought when writing this code, but now I'm not sure, please advise 😅)

drjova: Let's keep it like that with the operator

PR #3202, opened 2024-07-26
compose: correct mq img tag

PascalEgn:
* Replace rabbitmq docker image tag with one that offers arm as well as linux as a plattform


drjova:
do we need this?

PascalEgn:
> do we need this?

Either this or we put `platform: linux/amd64` into the compose, otherwise I wont work on the M1/M2/M3... Macs

drjova (approved):

PR #3204, opened 2024-08-01
ui: remove `rel="noopener noreferrer"` from external links

drjova:
* ref: cern-sis/issues-inspire#506


karolina-siemieniuk-morawska (approved):
Snapshots need to be updated for unit tests to pass

PR #3228, opened 2024-08-26
ui: connect search in new holdingpen

karolina-siemieniuk-morawska:
* ref: cern-sis/issues-inspire#480

### DIFF
--- a/ui/src/holdingpen-new/utils/utils.tsx
+++ b/ui/src/holdingpen-new/utils/utils.tsx
@@ -5,15 +5,35 @@ import {
   HourglassOutlined,
   LoadingOutlined,
 } from '@ant-design/icons';
+import { push } from 'connected-react-router';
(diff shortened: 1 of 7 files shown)


ui/src/holdingpen-new/utils/utils.tsx
```diff
@@ -70,3 +78,18 @@ export const resolveDecision = (decision: string | number) => {
...
+
+export const handleSearch = (
+  dispatch: ActionCreator<Action>,
+  type: string,
+  searchValue: string
+) => {
+  const query = {
+    page: 1,
+    search: searchValue,
+    ...(type !== 'ALL' ? { workflow_type: type } : {}),
```
drjova, line 90: I'm wondering, if we can simplify it. I see many hardcoded check for types etc

drjova (approved):

PR #3229, opened 2024-08-27
editor: close curated record tickets on save

PascalEgn:
ref: https://github.com/cern-sis/issues-inspire/issues/486


record-editor/src/app/core/services/record-api.service.ts
```diff
@@ -71,6 +74,41 @@ export class RecordApiService extends CommonApiService {
...
+        }
+      } catch (error) {
+        if (error.status === 403) {
+          console.error('Logged in user cannot access tickets');
+        } else {
+          console.error('Could not load tickets');
+        }
+      }
+    }
+  }
```
drjova, line 110: It's better to use use guard clauses, to make the code more readable :)
```suggestion
    if (!record['curated']) {
       return ...
    }
    [....]
```

drjova (changes requested):
Just a small comment, otherwise it looks good!

record-editor/src/app/core/services/record-api.service.ts
```diff
@@ -71,6 +74,42 @@ export class RecordApiService extends CommonApiService {
...
+          console.error('Could not close ticket', ticket['sys_id']);
+        }
+      }
+    } catch (error) {
+      if (error.status === 403) {
+        console.error('Logged in user cannot access tickets');
+      } else {
+        console.error('Could not load tickets');
+      }
+    }
```
drjova, line 110: On more thing that I missed before, if we don't throw it will not display an error message [here](https://github.com/inspirehep/inspirehep/pull/3229/files#diff-5f33fad6fbd3534ab05c5eb40f989ae204853d7846d9b056fe579f5aa1d60321R158)

PascalEgn: Was actually on purpose (because I don't know how likely it is to have errors here and I didn't wanted to interfere with the "normal" saving workflow)

drjova: indeed, you have a point, we shouldn't fail saving a workflow because snow is down to close the ticket for example.

drjova (approved):

PR #3241, opened 2024-09-06
ui: fix display of tables on small screen

karolina-siemieniuk-morawska:
* ref: cern-sis/issues-inspire#562

Screenshot 2024-09-06 at 10 55 42

### DIFF
--- a/backend/data/records/conferences/1794610.json
+++ b/backend/data/records/conferences/1794610.json
@@ -35,8 +35,8 @@
     "_collections": [
         "Conferences"
     ],
-    "closing_date": "2024-09-10",
-    "opening_date": "2024-09-05",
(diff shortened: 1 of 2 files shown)


e2e/package.json
```diff
@@ -5,7 +5,7 @@
   "main": "index.js",
   "scripts": {
     "test:dev": "cypress open",
-    "test": "cypress run",
```
DonHaul, line 8: Why was this no required before?

DonHaul (approved):

e2e/package.json
karolina-siemieniuk-morawska: that's for local development, forgot to delete it, thanks for catching!

PR #3246, opened 2024-09-20
record-editor: open backoffice workflows in editor

karolina-siemieniuk-morawska:
* ref: cern-sis/issues-inspire#538

### DIFF
--- a/backend/data/records/jobs/1813119.json
+++ b/backend/data/records/jobs/1813119.json
@@ -25,7 +25,7 @@
             }
         }
     ],
-    "deadline_date": "2024-09-20",
+    "deadline_date": "2025-09-20",
     "control_number": 1813119,
     "contact_details": [
         {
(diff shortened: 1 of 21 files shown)


backend/data/records/jobs/1813119.json
PascalEgn, line 1: can we do like 2135, otherwise I see ourselves confused why the tests are failing next year

PascalEgn (approved):

backend/data/records/jobs/1813119.json
karolina-siemieniuk-morawska: Would've been nice, unfortunately the requirement for job offer in our form is that the deadline has to be within a year 😂 so yeah, same issue will appear next year.

PascalEgn: well alright...

PR #3261, opened 2024-09-26
github actions: avoid hep tests when modified bo

DonHaul:


PascalEgn (changes requested):
Lets use https://docs.github.com/en/actions/writing-workflows/choosing-when-your-workflow-runs/triggering-a-workflow#example-excluding-paths

And ignore backoffice-repo

PascalEgn (approved):

PR #3265, opened 2024-09-27
ruff: merged ruff files

DonHaul:


.pre-commit-config.yaml
PascalEgn, line 1: Shouldn't that be `backoffice-repo`?

DonHaul: yes it should

PascalEgn (approved):

PR #3269, opened 2024-09-30
stardandizing | workflows -> authors

DonHaul:


drjova (changes requested):
shouldn't we change the app name in settings too?

DonHaul:
> shouldn't we change the app name in settings too?

It should be already done
https://github.com/inspirehep/inspirehep/pull/3269#pullrequestreview-2387817073

drjova (approved):

PR #3272, opened 2024-09-30
backoffice: add validation errors in response

PascalEgn:
ref: https://github.com/cern-sis/issues-inspire/issues/585
ref: https://github.com/cern-sis/issues-inspire/issues/578

### DIFF
--- a/backoffice/.envs/local/.django
+++ b/backoffice/.envs/local/.django
@@ -20,7 +20,7 @@ OPENSEARCH_HOST=es:9200
 OPENSEARCH_INDEX_PREFIX=backoffice-backend-local
 
 # Airflow
-AIRFLOW_BASE_URL=http://airflow-webserver:8070
+AIRFLOW_BASE_URL=http://airflow-webserver:8080
 AIRFLOW_TOKEN=YWlyZmxvdzphaXJmbG93
 
 SERVICENOW_URL=https://cerntraining.service-now.com
--- a/backoffice/backoffice/workflows/api/utils.py
+++ b/backoffice/backoffice/workflows/api/utils.py
@@ -1,11 +1,22 @@
-from backoffice.workflows.api.serializers import DecisionSerializer
+from backoffice.workflows.api import serializers
 
 
 def add_decision(workflow_id, user, action):
-    serializer_class = DecisionSerializer
+    serializer_class = serializers.DecisionSerializer
     data = {"workflow": workflow_id, "user": user, "action": action}
 
     serializer = serializer_class(data=data)
     if serializer.is_valid(raise_exception=True):
         serializer.save()
         return serializer.data
+
+
+def render_validation_error_response(validation_errors):
+    validation_errors_messages = [
+        {
+            "message": error.message,
+            "path": list(error.path),
+        }
+        for error in validation_errors
+    ]
+    return validation_errors_messages
--- a/backoffice/backoffice/workflows/tests/cassettes/TestAuthorWorkflowViewSet.test_validate_not_valid_record.yaml
+++ b/backoffice/backoffice/workflows/tests/cassettes/TestAuthorWorkflowViewSet.test_validate_not_valid_record.yaml
@@ -1,29 +1,5 @@
 interactions:
 - request:
-    body: '{"index":{"_id":"00000000-0000-0000-0000-000000000000","_index":"backoffice-backend-test-workflows"}}
-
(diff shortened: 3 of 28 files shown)


DonHaul:
Looks good but I think it would be better if we add this logic to the workflow serializer.

backoffice/backoffice/workflows/api/utils.py
```diff
@@ -1,11 +1,22 @@
-from backoffice.workflows.api.serializers import DecisionSerializer
+from backoffice.workflows.api import serializers
 
 
 def add_decision(workflow_id, user, action):
-    serializer_class = DecisionSerializer
```
DonHaul, line 5: why this change

backoffice/backoffice/workflows/tests/cassettes/TestAuthorWorkflowViewSet.test_validate_not_valid_record.yaml
```diff
@@ -1,4 +1,28 @@
 interactions:
+- request:
```
DonHaul, line 2: es responses should not be recorded. in the `conftest.py` could you rename in the `ignore_hosts` `opensearch` to  `es` or alternatively find all instances of `es` and rename the to `opensearch`, and then rerun the tests? (my bad on this one, forgot to change this naming during the monorepoing)

record-editor/src/app/backoffice-editor/backoffice-editor.component.ts
```diff
@@ -86,8 +86,16 @@ export class BackofficeEditorComponent
...
   }
 
+  private injectValidationErrorsIntoExtraData() {
+    this.workflowObject._extra_data.validation_errors = this.workflowObject.validation_errors;
+  }
+
   get workflowExtraData() {
-    return this.workflowObject._extra_data || {};
+    this.workflowObject._extra_data = this.workflowObject._extra_data || {};
+    if (this.workflowObject.validation_errors) {
```
DonHaul, line 95: why do we now require this?

record-editor/src/app/shared/interfaces/workflow-object.ts
```diff
@@ -3,6 +3,7 @@ import { JsonPatch } from './json-patch';
 
 export interface WorkflowObject<T = object> {
   id: number;
+  validation_errors?: Array<ValidationError>;
   _extra_data?: {
     validation_errors?: Array<ValidationError>;
```
DonHaul, line 8: should this field be removed?

backoffice/backoffice/workflows/api/utils.py
PascalEgn: Its a circular import otherwise

record-editor/src/app/backoffice-editor/backoffice-editor.component.ts
PascalEgn: As described in https://github.com/cern-sis/issues-inspire/issues/585, the plan is to add the `validation_errors` field into the backend response (because we dont want to have this `extra_data` field anymore). But as we dont want to change much in the record editor (as it need to work with next and backoffice records), we inject the `validation_errors` field into the `extra_data` field

record-editor/src/app/shared/interfaces/workflow-object.ts
PascalEgn: field was added because we send the `validation_errors` directly instead of a field in `extra_data` in case of backoffice responses. But we cant remove it from `extra_data` either because we send in like this in case of next responses

backoffice/.envs/local/.django
```diff
@@ -20,7 +20,7 @@ OPENSEARCH_HOST=es:9200
 OPENSEARCH_INDEX_PREFIX=backoffice-backend-local
 
 # Airflow
-AIRFLOW_BASE_URL=http://airflow-webserver:8070
+AIRFLOW_BASE_URL=http://airflow-webserver:8080
```
drjova, line 23: wouldn't this be a conflict with ui? Hence we would not be able to run both inspirehep UI and airflow at the same time

PascalEgn: It doesnt because it's the container port (so internally) and not the exposed one. So from the outside you can reach ui under port 8080 and airflow under 8070

backoffice/backoffice/workflows/api/utils.py
```diff
@@ -1,11 +1,22 @@
-from backoffice.workflows.api.serializers import DecisionSerializer
+from backoffice.workflows.api import serializers
```
drjova, line 1: since it's one, I wouldn't import the whole module

drjova (approved):
Only a small change, otherwise LGTM

backoffice/backoffice/workflows/api/utils.py
PascalEgn: If I only import one, Its a circular import otherwise, therefore the change

PR #3243, opened 2024-09-09
ui: integrate login with orcid on ui and add local login page

karolina-siemieniuk-morawska:
⚠️⚠️ **NEEDS IMPLEMENTING SESSION ID AS AUTHORISATION TOKEN** ⚠️⚠️ 

* ref: cern-sis/issues-inspire#509

PR #3294, opened 2024-10-28
ui: orcid login integration with backoffice and local login 

miguelgrc:
* ref: cern-sis/issues-inspire/issues/509

Supersedes #3243 

~Some ports have been swapped in order for backoffice to use 8080, since 8001 is not in the orcid sandbox redirect whitelist.~


ui/src/common/PrivateRoute.tsx
```diff
@@ -10,9 +10,9 @@ interface PrivateRouteProps extends ComponentPropsWithoutRef<any> {
   loggedIn: boolean;
   userRoles: List<string>;
   authorizedRoles: List<string>;
+  loggedInToHoldingpen: boolean;
```
drjova, line 13: we don't use Holdingpen any more, shouldn't this be  `loggedInToBackoffice`

drjova (changes requested):

backoffice/config/urls.py
```diff
@@ -1,4 +1,4 @@
-from allauth.socialaccount.providers.orcid.views import oauth2_login
+from allauth.socialaccount.providers.orcid.views import oauth2_login, oauth2_callback
```
DonHaul, line 1: I believe that these changes need to be reverted.

This is because.
for the sandbox orcid the route is indeed `api/oauth/authorized/orcid/`
however for the orcid we have in QA and will be in prod, the route is the default one (`/accounts/orcid/login/callback/`) so no need for this. 

As an alternative we can make it so that `api/oauth/authorized/orcid/` is also a valid route in our orcid in qa @drjova ?

docker-compose.yml
```diff
@@ -120,8 +120,8 @@ services:
       - ./backoffice/.envs/local/.django
       - ./backoffice/.envs/local/.postgres
     ports:
-      - '8001:8001'
```
DonHaul, line 123: if you want to change the ports. they will also have to change in more places:

backoffice/.envs/local/.django
`FRONTEND_LOGIN_SUCCESS=http://localhost:8080/backoffice` to use new port of ui

`workflows/scripts/connections/connections.json` to use new port of backoffice
`backoffice_conn` to use the new port

backoffice/config/urls.py
drjova: yes I will request adding / or verify that on orcid production as well. I will send an email

miguelgrc: So we use prod ORCID in QA and PROD and sandbox only in local?
It will take long to get a reply, right? Otherwise maybe I should make this route dynamic depending on the environment.

DonHaul: @drjova could you also check that the prod environment is one of the allowed domains in the orcid prod app?
`https://backoffice.inspirehep.net`  `/api/oauth/authorized/orcid/` or `/accounts/orcid/login/callback/`

DonHaul:
seems the port in the airflow tests `cassette` must be changed to the newer one

miguelgrc:
> seems the port in the airflow tests `cassette` must be changed to the newer one

No need now after reverting back to the previous ports :)

PR #3299, opened 2024-10-31
backoffice: allow cookies in cors requests

DonHaul:
*ref: cern-sis/issues-inspire/issues/509


drjova:
@DonHaul do we really need that?

DonHaul:
> @DonHaul do we really need that?

Yes, else the communication between ui and backoffice through sessions wont work.
This has been added temporarly to `kubernetes-inspire` for it to work at the moment:
https://github.com/cern-sis/kubernetes-inspire/blob/main/backoffice/environments/qa/production.py#L12

PR #3300, opened 2024-10-31
worfklows: update author routes

DonHaul:
*ref: cern-sis/issues-inspire/issues/582


DonHaul:
We should find a way to configure this routes on a global level so that we only need to change them in one single place across backoffice, workflows and inspirehep

drjova (approved):

PR #3309, opened 2024-11-11
backoffice: add author-details api tests

PascalEgn:
ref: https://github.com/cern-sis/issues-inspire/issues/602

### DIFF
--- a/backoffice/backoffice/authors/tests/test_views.py
+++ b/backoffice/backoffice/authors/tests/test_views.py
@@ -370,6 +370,73 @@ class TestAuthorWorkflowViewSet(BaseTransactionTestCase):
         self.assertIn("id", response.json())
 
     @pytest.mark.vcr
+    def test_get_author(self):
+        self.api_client.force_authenticate(user=self.curator)
+        data = {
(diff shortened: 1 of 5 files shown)


backoffice/backoffice/authors/tests/test_views.py
```diff
@@ -369,6 +369,73 @@ def test_create_author(self):
...
+    def test_get_non_existent_author(self):
+        self.api_client.force_authenticate(user=self.curator)
+        detail_url = reverse(
+            "api:authors-detail", kwargs={"pk": "THISISFORSURENOTANID"}
+        )
+        detail_response = self.api_client.get(detail_url)
+        self.assertEqual(detail_response.status_code, 404)
+        self.assertEqual(detail_response.json()["detail"], "Not found.")
+
+    @pytest.mark.vcr
```
drjova, line 406: Since the application is calling itself, I don’t think we need to record a cassette unless it’s interacting with external services.

PascalEgn: Actually they all need the cassettes as they all in a way interact with airflow in the background (so in case its not running there will be a error message)

drjova (approved):

PR #3314, opened 2024-11-21
backoffice: addition of processing status

drjova:


DonHaul:
should we also add a INITIAL state, for when the workflow is in the queue?

DonHaul (approved):

drjova:
@DonHaul added Processing as the default

DonHaul (approved):
cool

PR #3322, opened 2024-11-29
basic harvest implementation

DonHaul:


workflows/dags/data/data_harvest.py
```diff
@@ -0,0 +1,221 @@
...
+            )
+            hepdata_response.raise_for_status()
+            payload = hepdata_response.json()
+
+            record = {"base": payload}
+            for version in range(1, payload["record"]["version"]):
+                response = generic_http_hook.call_api(
+                    endpoint=f"/record/ins{id}?format=json&version={version}"
+                )
+                record[version] = response.json()
```
drjova, line 76: what if it fails? we should add `raise_for_status()` with a useful message.

workflows/dags/data/data_harvest.py
```diff
@@ -0,0 +1,221 @@
...
+    def collect_ids():
+        """Collects the ids of the records that have been updated in the last two days.
+
+        Returns: list of ids
+        """
+
+        from_date = (datetime.datetime.now().date() - timedelta(days=1)).strftime(
+            "%Y-%m-%d"
+        )
+        # http sensor
```
drjova, line 41: redundant

workflows/dags/data/data_harvest.py
```diff
@@ -0,0 +1,221 @@
...
+        payload = {"inspire_ids": True, "last_updated": from_date, "sort_by": "latest"}
```
drjova, line 42: pre-commit?

workflows/dags/data/data_harvest.py
```diff
@@ -0,0 +1,221 @@
...
+                    if keyword == "cmenergies":
+                        if len(item) >= 1 and "lte" in item[0] and "gte" in item[0]:
+                            builder.add_keyword(
+                                f"{keyword}: {item[0]['lte']}-{item[0]['gte']}"
+                            )
+                    elif keyword == "observables":
+                        builder.add_keyword(f"{keyword}: {','.join(item)}")
+                    else:
+                        for value in item:
+                            builder.add_keyword(value)
```
drjova, line 123: can we move somewhere separate? for example utils?

workflows/dags/data/data_harvest.py
```diff
@@ -0,0 +1,221 @@
...
+        def load_record(new_record):
+            """
+            Load the record to inspirehep.
+            """
+
+            try:
+                response = inspire_http_record_management_hook.get_record(
+                    pid_type="doi", control_number=new_record["dois"][0]["value"]
+                )
+            except RetryError:
```
drjova, line 194: hmm `RetryError`?

workflows/dags/data/data_harvest.py
```diff
@@ -0,0 +1,221 @@
...
+            revision_id = response.get("revision_id", 0)
+            old_record.update(new_record)
+            logger.info(f"Updating Record: {old_record['control_number']}")
+            response = inspire_http_record_management_hook.update_record(
+                data=old_record,
+                pid_type="data",
+                control_number=old_record["control_number"],
+                revision_id=revision_id + 1,
+            )
+            return response.json()
```
drjova, line 212: handling error?

workflows/tests/test_data_tasks.py
```diff
@@ -0,0 +1,102 @@
...
+import pytest
+from airflow.models import DagBag
+from airflow.utils.context import Context
+
+dagbag = DagBag()
+
+
+class TestDataHarvest:
+    print("holup partner")
+    print(dagbag.dags.keys())
```
drjova, line 12: we don't need the prints :)

drjova (changes requested):
Thanks @DonHaul, few comments

workflows/dags/data/data_harvest.py
DonHaul: Sorry I dont understand - how else can we get the date?

workflows/dags/data/data_harvest.py
DonHaul: pre-commit fomatted it like this

workflows/dags/data/data_harvest.py
DonHaul: I will try, because its a virtual env I am not completely sure if its possible

workflows/dags/data/data_harvest.py
drjova: the comment :)

workflows/dags/data/data_harvest.py
drjova: ok, I see. We can leave as is for now

workflows/dags/data/data_harvest.py
DonHaul: its to detect if the request fails, this is the generic error it throws when the doi does not exist. Will check if `raise_for_status` works in its place

workflows/dags/data/data_harvest.py
DonHaul: indeed I checked, and in the `sys.path` of this task only the following exist:
`['/tmp/venv-calliyoq65lm', '/usr/local/lib/python311.zip', '/usr/local/lib/python3.11', '/usr/local/lib/python3.11/lib-dynload', '/tmp/venvicyq6h_l/lib/python3.11/site-packages']`

its possible to do `sys.path.append('/opt/airflow/plugins'])` and it would work. although it would look jankier

workflows/dags/data/data_harvest.py
DonHaul: it will fail before

workflows/dags/data/data_harvest.py
drjova: it's ok we can leave it like that, no need to append stuff to the path

drjova (approved):

PR #3340, opened 2024-12-18
workflows: fix author submissions not erroring

DonHaul:
also some small config fixes for easy testing with docker

* ref: https://github.com/cern-sis/issues-inspire/issues/643

### DIFF
--- a/workflows/dags/author/author_create/author_create_init.py
+++ b/workflows/dags/author/author_create/author_create_init.py
@@ -21,13 +21,14 @@ logger = logging.getLogger(__name__)
     params={
         "workflow_id": Param(type="string", default=""),
         "data": Param(type="object", default={}),
+        "collection": Param(type="string", default=AUTHORS),
     },
     start_date=datetime.datetime(2024, 5, 5),
     schedule=None,
     catchup=False,
     # TODO: what if callback fails? Data in backoffice not up to date!
     on_failure_callback=set_workflow_status_to_error,
-    tags=["authors"],
+    tags=[AUTHORS],
 )
 def author_create_initialization_dag():
     """
--- a/Makefile
+++ b/Makefile
@@ -9,6 +9,7 @@ run: services start-inspirehep sleep setup-inspirehep start-backoffice sleep set
 run-inspirehep: services start-inspirehep sleep setup-inspirehep
 run-backoffice: services start-backoffice sleep setup-backoffice
 
+start: services start-inspirehep start-backoffice
 
 start-inspirehep:
 	echo -e "\033[0;32m Starting HEP. \033[0m"
--- a/README.md
+++ b/README.md
@@ -62,6 +62,11 @@ Then depending on the service you are testing:
 - inspire-hep: ?
 - backend: ?
 
+### Adding global variables:
+There are two ways of setting environment variables on hep:
+- `backend/inspirehep/config.py` 
+- `docker-compose.services.yml` - `INVENIO_` prefix must be added. Variables here overwrite `config.py`
+
 ## Running Locally
 For running the enviroment locally you have the following prerequirements:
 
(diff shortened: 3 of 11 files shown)


backend/inspirehep/config.py
```diff
@@ -51,8 +51,8 @@
 INSPIRE_NEXT_URL = "http://localhost:5000"
 LEGACY_BASE_URL = "https://old.inspirehep.net"
 LEGACY_RECORD_URL_PATTERN = "http://inspirehep.net/record/{recid}"
-INSPIRE_BACKOFFICE_URL = "https://backoffice.dev.inspirebeta.net"
-AUTHENTICATION_TOKEN_BACKOFFICE = "CHANGE_ME"
+INSPIRE_BACKOFFICE_URL = "http://host.docker.internal:8001"
```
DonHaul, line 54: so we dont modify it when we are doing tests every single time

workflows/dags/author/author_create/author_create_init.py
```diff
@@ -17,17 +17,22 @@
...
 
+def task_failure_alert(context):
+    print(f"Task has failed, task_instance_key_str: {context['task_instance_key_str']}")
+
+
 @dag(
     params={
         "workflow_id": Param(type="string", default=""),
         "data": Param(type="object", default={}),
+        "collection": Param(type="string", default=AUTHORS),
```
DonHaul, line 28: because there's no generic `/api/workflows` api anymore, we have to pass this to know which collection we want to change the workflow of.

Should we create this generic endpoint for this holistic  workflow actions such as setting its status to `ERROR` ?

backend/inspirehep/config.py
```diff
@@ -31,7 +31,7 @@
 FEATURE_FLAG_ENABLE_ORCID_PUSH = False
 FEATURE_FLAG_ENABLE_REDIRECTION_OF_PIDS = False
 FEATURE_FLAG_ENABLE_ASSIGN_AUTHOR_PAPERS = False
-FEATURE_FLAG_ENABLE_SEND_TO_BACKOFFICE = False
+FEATURE_FLAG_ENABLE_SEND_TO_BACKOFFICE = True
```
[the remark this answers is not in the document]
drjova: I would say it's quite dangerous, that's why we have them all `False`. Tou can simple set it only for the tests here https://github.com/inspirehep/inspirehep/blob/39428998c0177e70335db03bac9ed0dc740743bb/backend/tests/integration/conftest.py#L43

backend/inspirehep/config.py
drjova: 👍 We have to make sure it's set on k8s

workflows/dags/author/author_create/author_create_init.py
drjova: Let's create a task for it

backend/inspirehep/config.py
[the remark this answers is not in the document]
DonHaul: okay, I will take it away for now, as we are not using it for testing atm

backend/inspirehep/config.py
DonHaul: Added to the `hep-globals`
https://github.com/cern-sis/kubernetes-inspire/commit/b19919d806529302b4dd807de564f63512b05708

workflows/dags/author/author_create/author_create_init.py
DonHaul: https://github.com/cern-sis/issues-inspire/issues/645 . So for the time being I undo this change?

drjova: No we need it to be able to test it, we can refactor it with the other task

drjova (approved):

PR #3344, opened 2024-12-20
workflows: fix data harvesting

DonHaul:
- correctly format observables
- add date parameter to dag
- fetch collections ids relative to first run

* ref: cern-sis/issues-inspire/issues/633


workflows/dags/data/data_harvest.py
```diff
@@ -115,7 +118,8 @@ def add_keywords(record, builder):
                                 f"{keyword}: {item[0]['lte']}-{item[0]['gte']}"
                             )
                     elif keyword == "observables":
-                        builder.add_keyword(f"{keyword}: {','.join(item)}")
+                        for value in item:
+                            builder.add_keyword(f"observables: {value}")
```
drjova, line 122: @michamos do we need to add this prefix?

michamos: yes, is it a problem?

drjova: No, I’m just thinking it’s a bit dirty for metadata. Can we add a `type` for keywords? Or did I misunderstand the entire thing?

michamos: that's how we encode hierarchical keywords on INSPIRE, see https://inspirehep.net/literature/1998859 for example

drjova (approved):

PR #3349, opened 2025-01-06
backoffice: replace run_with_advanced_retry calls

PascalEgn:
ref: https://github.com/cern-sis/issues-inspire/issues/646


drjova:
@PascalEgn no test need update? Interesting

PascalEgn:
> @PascalEgn no test need update? Interesting

It basically still calls the same function:
https://github.com/inspirehep/inspirehep/blob/master/workflows/plugins/hooks/generic_http_hook.py#L62

drjova (approved):

PR #3350, opened 2025-01-06
backoffice: set processing state when restarting

PascalEgn:
ref: https://github.com/cern-sis/issues-inspire/issues/647

### DIFF
--- a/backoffice/backoffice/authors/airflow_utils.py
+++ b/backoffice/backoffice/authors/airflow_utils.py
@@ -2,8 +2,6 @@ import logging
 from os import environ
 
 import requests
-from django.http import JsonResponse
-from requests.exceptions import RequestException
 from rest_framework import status
 import json
 from django.core.serializers.json import DjangoJSONEncoder
@@ -22,7 +20,7 @@ def trigger_airflow_dag(dag_id, workflow_id, extra_data=None, workflow=None):
 
     :param dag_id: name of the dag to run
     :param workflow_id: id of the workflow being triggered
-    :returns: request response
+    :returns: request response content
     """
 
     data = {"dag_run_id": str(workflow_id), "conf": {"workflow_id": str(workflow_id)}}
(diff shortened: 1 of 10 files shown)


drjova:
Thanks @PascalEgn, do we need to change anything on UI?

PascalEgn:
> Thanks @PascalEgn, do we need to change anything on UI?

Good Point, forgot that we dont really have tests yet 🫠  I will check

backoffice/backoffice/authors/api/views.py
```diff
@@ -219,17 +246,46 @@ def resolve(self, request, pk=None):
...
+            else:
+                AuthorDecision.objects.filter(workflow=workflow).delete()
+                response = airflow_utils.restart_workflow_dags(
+                    workflow.id, workflow.workflow_type, request.data.get("params")
+                )
+                if response is None:
+                    return Response(
+                        {"error": "No run configuration found. Skipping restart."},
+                        status=status.HTTP_400_BAD_REQUEST,
+                    )
```
drjova, line 272: `if`s 🍝 :)

backoffice/backoffice/authors/airflow_utils.py
```diff
@@ -134,24 +123,19 @@ def delete_workflow_dag(dag_id, workflow_id):
...
-            {"error": "Failed to delete DAG"}, status=status.HTTP_424_FAILED_DEPENDENCY
-        )
+    logger.info(
+        "Deleting dag Failed Tasks of DAG %s with no data and %s",
+        dag_id,
+        url,
+    )
+    response = requests.delete(url, headers=AIRFLOW_HEADERS)
+    response.raise_for_status()
+    return response
```
drjova, line 138: I’m quite skeptical about returning the entire response object. In our case, we only need the `data` and `code`. Additionally, there’s a risk that the requests API might change in the future, requiring us to modify all the methods instead of just one place here.

drjova (changes requested):
thanks @PascalEgn, few comments

drjova (approved):

PR #3359, opened 2025-01-09
workflows test PR (ubuntu 24.04 and actions/upload-artifact v4)

Lorenzovagliano:


drjova:
@Lorenzovagliano I think we can close that right?

PR #3367, opened 2025-01-13
backoffice-ui: refactor wf status logic

PascalEgn:
ref: https://github.com/cern-sis/issues-inspire/issues/666

### DIFF
--- a/ui/src/backoffice/containers/DetailPageContainer/AuthorDetailPageContainer.tsx
+++ b/ui/src/backoffice/containers/DetailPageContainer/AuthorDetailPageContainer.tsx
@@ -28,7 +28,7 @@ import {
   columnsAdvisors,
 } from './columnData';
 import { getConfigFor } from '../../../common/config';
-import { resolveDecision } from '../../utils/utils';
+import { getWorkflowStatusInfo, resolveDecision } from '../../utils/utils';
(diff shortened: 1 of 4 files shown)


ui/src/backoffice/containers/DetailPageContainer/AuthorDetailPageContainer.tsx
```diff
@@ -114,7 +115,11 @@ const AuthorDetailPageContainer: React.FC<AuthorDetailPageContainerProps> = ({
                         status === 'error' ? 'white' : ''
                       } w-100`}
                     >
-                      <p className="b f3 tc pv2">{status?.toUpperCase()}</p>
+                      <p className="b f3 tc pv2">
+                        {status && getWorkflowStatusInfo(status.toLowerCase())
```
ioannistsanaktsidis, line 119: To make it more readable, you can extract this part before the return with the other variables. You can have something like 

```
const statusInfo = status ? getWorkflowStatusInfo(status.toLowerCase()) : null;

```
and then here you just do 
```
{statusInfo ? statusInfo.text : 'Unknown Status'}
```

ioannistsanaktsidis (changes requested):
Some small improvements, otherwise nice! 👍

ioannistsanaktsidis (approved):
All good!

PR #3380, opened 2025-01-14
feat[workflows]: Bump inspire-schemas and Access DOI value directly when adding literature

Lorenzovagliano:
- Dependant on: https://github.com/inspirehep/inspire-schemas/pull/476

ref: https://github.com/cern-sis/issues-inspire/issues/631


PascalEgn:
@Lorenzovagliano You can bump `inspire-schemas` to `61.6.9` now and we see if the tests pass :)

Lorenzovagliano:
@PascalEgn This function actually gets inspire-schemas in a virtual env using `@task.virtualenv(requirements=["inspire-schemas"], system_site_packages=False,)`, so just rerunning gets the latest version.

We do still need to bump it though.

PascalEgn (approved):

PR #3384, opened 2025-01-15
data: set source to HEPData

DonHaul:
* ref: cern-sis/issues-inspire/issues/675

### DIFF
--- a/workflows/dags/data/data_harvest.py
+++ b/workflows/dags/data/data_harvest.py
@@ -128,7 +128,7 @@ def data_harvest_dag():
                         for value in item:
                             builder.add_keyword(value)
 
-            builder = DataBuilder(source="inspirehep")
+            builder = DataBuilder(source="HEPData")
 
             builder.add_creation_date(datetime.datetime.now(datetime.UTC).isoformat())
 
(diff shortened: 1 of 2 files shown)


PascalEgn (approved):

workflows/dags/data/data_harvest.py
```diff
@@ -128,7 +128,7 @@ def add_keywords(record, builder):
                         for value in item:
                             builder.add_keyword(value)
 
-            builder = DataBuilder(source="inspirehep")
+            builder = DataBuilder(source="HEPData")
```
drjova, line 131: `acquisition_source.source which should be inspirehep.` Do I miss something?

PascalEgn: Had the same thought but Joao talked with Micha and the source should be "HEPData" and the method "inspirehep", its wrong in the issue

DonHaul: I've talked after in person with @michamos and there was some confusion.
all sources should be `HEPData`, only the method should be `inspirehep`

drjova: ok thanks, I just saw the issue and I was confused.

michamos: yes, sorry about the confusion, I've updated the issue now

PR #3385, opened 2025-01-15
fix[ui][backoffice]: Make order of the columns under Project membership history the same as Institution history

Lorenzovagliano:
ref: https://github.com/cern-sis/issues-inspire/issues/671

### DIFF
--- a/ui/src/backoffice/containers/DetailPageContainer/columnData.tsx
+++ b/ui/src/backoffice/containers/DetailPageContainer/columnData.tsx
@@ -44,6 +44,16 @@ export const columnsProjects = [
     dataIndex: 'name',
   },
   {
+    title: 'Start date',
+    dataIndex: 'start_date',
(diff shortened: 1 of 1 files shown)


ioannistsanaktsidis (approved):

DonHaul:
this is introducing a regression by undoing this PR
https://github.com/inspirehep/inspirehep/pull/3372

ioannistsanaktsidis:
@DonHaul I don't see the part of regression. The PR #3372 is setting the default value of "End date" to "-". This PR here just re-orders the columns. Is there something else we are missing ?

DonHaul:
> @DonHaul I don't see the part of regression. The PR #3372 is setting the default value of "End date" to "-". This PR here just re-orders the columns. Is there something else we are missing ?

You are right, is just because I saw the `Present` there I thought something was wrong. I guess it just wasnt rebased before hand but thats fine

PR #3388, opened 2025-01-17
fix[ui][backoffice]: remove ORCID from the author links box

Lorenzovagliano:
ref: https://github.com/cern-sis/issues-inspire/issues/678

### DIFF
--- a/ui/src/backoffice/authors/containers/__tests__/AuthorDetailPageContainer.test.tsx
+++ b/ui/src/backoffice/authors/containers/__tests__/AuthorDetailPageContainer.test.tsx
@@ -1,6 +1,6 @@
 import React from 'react';
 import { fromJS } from 'immutable';
-import { fireEvent, render, screen, waitFor } from '@testing-library/react';
+import { fireEvent, render, screen, waitFor} from '@testing-library/react';
 import { Provider } from 'react-redux';
 import { MemoryRouter } from 'react-router-dom';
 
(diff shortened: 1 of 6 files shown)


PascalEgn (changes requested):
Lets add a ui test for it, otherwise looks good :)

ioannistsanaktsidis (changes requested):
@Lorenzovagliano we also have this method here https://github.com/inspirehep/inspirehep/blob/master/ui/src/backoffice/components/Author/AuthorMainInfo.tsx#L19 . Very similar to what you are doing here. What we can do so as to normalize everything is to move these methods to the utils file here https://github.com/inspirehep/inspirehep/blob/master/ui/src/backoffice/utils/utils.tsx and reuse them everywhere. That way we can have consistency and more readability.

ui/src/backoffice/authors/containers/__tests__/AuthorDetailPageContainer.test.tsx
```diff
@@ -50,6 +53,14 @@ describe('AuthorDetailPageContainer', () => {
     expect(asFragment()).toMatchSnapshot();
   });
 
+  it('should display ORCID link only once', () => {
+    renderComponent();
+
+    const link = screen.getByRole('link', {name: '0000-0002-6357-9297'});
+    expect(link).toBeInTheDocument();
+    expect(link).toHaveAttribute('href', expect.stringContaining('/orcid.org/0000-0002-6357-9297'));
```
ioannistsanaktsidis, line 61: if you have the whole url already you  do not need to do stringContaining. You can check against the full href. 

E.g 
```
expect(link).toHaveAttribute('href', 'https://www.springer.com/journal/526');
```

ioannistsanaktsidis (changes requested):

ioannistsanaktsidis (approved):
Looks good!

PascalEgn (approved):

PR #3389, opened 2025-01-17
backoffice: separation of concerns

ioannistsanaktsidis:
ref https://github.com/cern-sis/issues-inspire/issues/677


ui/src/backoffice/search/containers/SearchPageContainer.tsx
```diff
@@ -17,7 +17,16 @@ import ResponsiveView from '../../../common/components/ResponsiveView';
 import DrawerHandle from '../../../common/components/DrawerHandle';
 import AggregationBox from '../../../common/components/AggregationBox';
 import { isSuperUser } from '../../../common/authorization';
-import WorkflowResultItem from '../../components/ResultItem/ResultItem';
+import WorkflowResultItem from '../../authors/components/ResultItem/ResultItem';
```
drjova, line 20: isn't this just for authors?

ioannistsanaktsidis: Indeed for now. As discussed, we leave it like this and we refactor the  Component when we have multiple collections

PR #3394, opened 2025-01-20
backoffice-ui: fix breadcrumb icon and link

PascalEgn:
ref: https://github.com/cern-sis/issues-inspire/issues/683

### DIFF
--- a/ui/src/backoffice/common/components/Breadcrumbs/__tests__/Breadcrumbs.test.tsx
+++ b/ui/src/backoffice/common/components/Breadcrumbs/__tests__/Breadcrumbs.test.tsx
@@ -1,5 +1,5 @@
 import React from 'react';
-import { render } from '@testing-library/react';
+import { render, screen } from '@testing-library/react';
 import { Provider } from 'react-redux';
 import { MemoryRouter } from 'react-router-dom';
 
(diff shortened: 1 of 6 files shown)


ui/src/backoffice/common/components/Breadcrumbs/__tests__/Breadcrumbs.test.tsx
```diff
@@ -35,4 +35,22 @@ describe('Breadcrumbs', () => {
...
+            href1="backoffice"
+            title2="Author Detail"
+            href2="1234"
+          />
+        </MemoryRouter>
+      </Provider>
+    );
+
+    const link = screen.getByText('Author Detail').closest('a');
+    expect(link).toHaveAttribute('href', expect.stringContaining('/backoffice/1234'));
```
ioannistsanaktsidis, line 54: what about something like this ? 

```
    const link = screen.getByRole('link', { name: 'Author detail' });
    expect(link).toBeInTheDocument();
    expect(link).toHaveAttribute( ...
```
as getByRole is usually used to find a specific link in a doc for a test

ioannistsanaktsidis (changes requested):
Left a comment related to the test. Not very important, but if we can achieve it like this is a bit cleaner.

ioannistsanaktsidis (approved):
All good! Nice!

PR #3395, opened 2025-01-20
ui: remove enzyme from journals

ioannistsanaktsidis:
ref https://github.com/cern-sis/issues-inspire/issues/687


ui/src/journals/components/__tests__/JournalPapers.test.jsx
DonHaul, line 1: why has this test been removed?

ui/src/journals/components/__tests__/JournalTitlesListModal.test.jsx
```diff
@@ -1,18 +1,41 @@
 import React from 'react';
-import { shallow } from 'enzyme';
+import { render, screen } from '@testing-library/react';
+import '@testing-library/jest-dom/extend-expect';
```
DonHaul, line 3: to what is this import being used?

ui/src/journals/components/__tests__/JournalPapers.test.jsx
ioannistsanaktsidis: This is removed because creating a snapshot to it is actually useless(we have a lot of tests like this). In order to make it meaningful, we first need to write a test for the https://github.com/inspirehep/inspirehep/blob/master/ui/src/literature/containers/LiteratureSearchContainer.jsx and then have something meaningful for this one. We can do that though in a later step as it requires quite some work.

ui/src/journals/components/__tests__/JournalTitlesListModal.test.jsx
ioannistsanaktsidis: good question !  This is used in order to utilize the extended matches from jest(e.g .toBeInTheDocument()). But from what I see now, we already import it in our [setupTests.js ](https://github.com/inspirehep/inspirehep/blob/master/ui/src/setupTests.js#L7) which  is already enough. I am going to remove it from this test.

miguelgrc (approved):

ui/src/journals/containers/DetailPageContainer.tsx
```diff
@@ -65,7 +65,7 @@ export const DetailPage = ({
   return (
     <>
       <DocumentHead title={shortTitle} />
-      <Row justify="center">
+      <Row justify="center" data-testid="journal-detail-page-container">
```
drjova, line 68:
```suggestion
      <Row justify="center" data-testid="journals-detail-page-container">
```

ui/src/journals/containers/SearchPageContainer.tsx
```diff
@@ -50,7 +52,12 @@ export const JournalSearchPage = ({
   return (
     <>
       <DocumentHead title={TITLE} description={META_DESCRIPTION} />
-      <Row className="mt3" gutter={SEARCH_PAGE_GUTTER} justify="center">
+      <Row
+        className="mt3"
+        gutter={SEARCH_PAGE_GUTTER}
+        justify="center"
+        data-testid="journal-search-page-container"
```
drjova, line 59:
```suggestion
        data-testid="journals-search-page-container"
```

drjova (approved):

PR #3396, opened 2025-01-20
backoffice-ui: remove decision-pill border color

PascalEgn:
remove decision-pill border color and fix typo

ref: https://github.com/cern-sis/issues-inspire/issues/682

### DIFF
--- a/ui/src/backoffice/authors/components/ResultItem/ResultItem.less
+++ b/ui/src/backoffice/authors/components/ResultItem/ResultItem.less
@@ -37,13 +37,20 @@
     line-break: anywhere;
   }
 
-  .decission-pill {
-    border-color: red;
+  .decision-pill {
     font-weight: 400;
     margin-left: 0.5rem;
   }
 
-  .decission-pill.bg-halted {
+  .decision-pill.bg-halted {
     border-color: #f39c12;
+  }
+
+  .decision-pill.bg-completed {
+    border-color: #16a085;
+  }
+
+  .decision-pill.bg-error {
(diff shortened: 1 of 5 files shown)


ui/src/backoffice/authors/components/ResultItem/ResultItem.less
```diff
@@ -37,13 +37,12 @@
     line-break: anywhere;
   }
 
-  .decission-pill {
-    border-color: red;
+  .decision-pill {
```
ioannistsanaktsidis, line 40: It is better to not remove the border-color in the default class, as you are also removing it from other states  too (reject). What would work better is actually to add a class for **completed** like it is  for halted https://github.com/inspirehep/inspirehep/pull/3396/files#diff-fb2f0f3ba33ee9488dbf4b0b5b7180c270429ebc33285565e26dd7799179a1feR45 . So by default is red, and then it gets overwritten here https://github.com/inspirehep/inspirehep/blob/e1eea05f07c03cea0850f86bbb85afdbecece540/ui/src/backoffice/utils/utils.tsx#L87. Same way will happen for completed.

ioannistsanaktsidis (changes requested):

ui/src/backoffice/authors/components/ResultItem/ResultItem.less
PascalEgn: Removed it from the base class and added the correct color for each status:

image

ioannistsanaktsidis (approved):
Even better!

PR #3399, opened 2025-01-21
ui: fix citation count in data

drjova:


ui/src/data/components/DataItem.jsx
```diff
@@ -51,15 +51,15 @@ function DataItem({ metadata, page }) {
         }
         rightActions={
           <>
-          {citationCount ? (
+          {citationCount != null && (
```
ioannistsanaktsidis, line 54: lets make a strict check for consistency -> !==

Also, should we check against undefined too so as to be more precise ?

ui/src/data/components/DataItem.jsx
```diff
@@ -51,15 +51,15 @@ function DataItem({ metadata, page }) {
...
               <IncomingLiteratureReferencesLinkAction
                 itemCount={citationCount}
                 referenceType="citation"
                 linkQuery={getReferencingPapersQueryString(recordId)}
                 trackerEventId="Citations link"
                 eventCategory="Data search"
               />
-            ) : <></>
```
ioannistsanaktsidis, line 62: No need to remove this. We can remove the upper one, line 53 and 64. So if there is a citationCount it render otherwise it passes the fragment

ui/src/data/containers/DetailPageContainer/DetailPageContainer.tsx
```diff
@@ -72,15 +72,15 @@ const DetailPage = ({
             }
             rightActions={
               <>
-                {citationCount ? (
+                {citationCount != null && (
```
ioannistsanaktsidis, line 75: Same here. Also lets remove the fragment on line 74

ui/src/data/containers/DetailPageContainer/DetailPageContainer.tsx
```diff
@@ -72,15 +72,15 @@ const DetailPage = ({
...
                   <IncomingLiteratureReferencesLinkAction
                     itemCount={citationCount}
                     referenceType="citation"
                     linkQuery={getReferencingPapersQueryString(recordId)}
                     trackerEventId="Citations link"
                     eventCategory="Data search"
                   />
-                ) : <></>
```
ioannistsanaktsidis, line 83: Same here

ioannistsanaktsidis (changes requested):
small improvements

ui/src/data/components/DataItem.jsx
drjova: Correct me if I'm wrong, wouldn't the immutable Map  always  give back `null` and not undefined?

ioannistsanaktsidis: I think it is the other way around, no ? https://immutable-js.com/docs/v3.8.2/Map/#get()

ui/src/data/components/DataItem.jsx
```diff
@@ -75,7 +72,7 @@ function DataItem({ metadata, page }) {
                 <LiteratureTitle title={title} />
               </Link>
             </div>
-          </div>
+          </div>w
```
ioannistsanaktsidis, line 75: small leftover

ioannistsanaktsidis (approved):
All good!

PR #3405, opened 2025-01-22
use harbor registry for base images

PascalEgn:
ref: https://github.com/cern-sis/issues-inspire/issues/689

This update will reduce the amount of pulls from dockerhub so we hopefully dont run into quota limits anymore


drjova (approved):

DonHaul:
Shoulndt we also do it in the github workflows?
https://github.com/inspirehep/inspirehep/blob/master/.github/workflows/backend-tests.yml#L50

PR #3406, opened 2025-01-22
backoffice-ui: generify result item

ioannistsanaktsidis:


ui/src/backoffice/common/components/ResultItem/ResultItem.tsx
```diff
@@ -68,17 +57,9 @@ const WorkflowResultItem = ({ item }: { item: any }) => {
...
-            </p>
-            <p className="waiting">
-              {data?.getIn(['acquisition_source', 'source'])}
-            </p>
-            <p className="waiting mb0">
-              {data?.getIn(['acquisition_source', 'email'])}
-            </p>
+            <p className="waiting">{acquisitionSourceDatetime}</p>
+            <p className="waiting">{acquisitionSourceSource}</p>
+            <p className="waiting mb0">{acquisitionSourceEmail}</p>
```
drjova, line 62: this might not be always there

ui/src/backoffice/common/components/ResultItem/ResultItem.tsx
```diff
@@ -23,43 +25,30 @@ const renderWorkflowStatus = (status: string) => {
...
-                    >
-                      {resolveDecision(decision?.get('action'))?.text}
-                    </UnclickableTag>
-                  )}
-                </div>
-                <span className="dib ml2">
-                  {data?.getIn(['name', 'value'])}
-                </span>
-              </div>
+              {isAuthor && <AuthorResultItem item={item} />}
```
drjova, line 51: Can we have a factory for that with cases? Because if we add more, we will add more and more lines like this one.

ioannistsanaktsidis:
@drjova comments addressed as requested.

drjova (approved):

PR #3407, opened 2025-01-23
Cosmetic fixes - Author Detail Page

DonHaul:
How it looks like now

image


ui/src/backoffice/authors/components/AuthorMainInfo.tsx
```diff
@@ -29,23 +29,23 @@ const AuthorMainInfo = ({ data }: AuthorMainInfoProps) => {
...
       )}
       {nameVariants && (
-        <p>
+        <p className='mb0'>
           <b>Name variants:</b> {nameVariants.join('; ')}
         </p>
       )}
       {status && (
-        <p className={classNames({ mb0: !orcidIdExists })}>
+        <p className={classNames({ mb0: orcidIdExists })}>
```
ioannistsanaktsidis, line 48: why removing the negetion here ?

DonHaul: It was wrong. if an `orcid` id exists below, we want to to have no margin between the `status` and the `orcid`.
we only want a `margin-bottom` on the last element on the last element being rendered from what I understood.

ioannistsanaktsidis: Yes you are correct!

ioannistsanaktsidis (approved):
Looks good!

PR #3408, opened 2025-01-23
ui: fix linting and prettier addition

ioannistsanaktsidis:
ref https://github.com/cern-sis/issues-inspire/issues/673


ui/.eslintrc
```diff
@@ -1,13 +1,14 @@
 {
   "root": true,
   "parser": "babel-eslint",
-  "plugins": ["react-hooks"],
-  "extends": ["airbnb", "prettier", "prettier/react"],
+  "plugins": ["react-hooks", "prettier"],
+  "extends": ["airbnb",  "prettier"],
```
miguelgrc, line 5: I believe you can just append [eslint-plugin-prettier/recommended](https://github.com/prettier/eslint-plugin-prettier?tab=readme-ov-file#configuration-legacy-eslintrc) to extends and remove `prettier` from everywhere else (plugins, extends and rules). That one should take care of all that

miguelgrc (commented):
Looks good and the formatting seems right. I just left a small comment regarding the config in `.eslintrc`.

ui/.eslintrc
ioannistsanaktsidis: I think you are correct! Let me check and come back !

ioannistsanaktsidis:
@miguelgrc changes updated based on your comment!

ioannistsanaktsidis:
Hmm actually it looks that it is not picking everything correctly with the recommended only version. The explicit prettier config seems to be fixing the problem.

PR #3409, opened 2025-01-24
Enzyme is dead seminar

DonHaul:
Wasn't able to convert test
`dispatches SEARCH_QUERY_UPDATE onChange with start_date and timezone if specific date` to rtl

### DIFF
--- a/ui/src/seminars/containers/__tests__/SeminarStartDateFilterContainter.test.jsx
+++ b/ui/src/seminars/containers/__tests__/SeminarStartDateFilterContainter.test.jsx
@@ -1,14 +1,11 @@
 import React from 'react';
-import { mount } from 'enzyme';
+import { render, fireEvent, within } from '@testing-library/react';
 import { fromJS } from 'immutable';
+import { mount } from 'enzyme';
 import { Provider } from 'react-redux';
 
-import {
-  getStoreWithState,
-  getStore,
-  mockActionCreator,
-} from '../../../fixtures/store';
-import { SEMINARS_NS, AUTHOR_SEMINARS_NS } from '../../../search/constants';
+import { getStore, mockActionCreator } from '../../../fixtures/store';
+import { SEMINARS_NS } from '../../../search/constants';
 import * as constants from '../../../common/constants';
 import SeminarStartDateFilterContainer from '../SeminarStartDateFilterContainer';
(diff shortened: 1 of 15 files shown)


ui/src/seminars/containers/__tests__/SeminarStartDateFilterContainter.test.jsx
```diff
@@ -42,56 +39,48 @@ describe('SeminarStartDateFilterContainer', () => {
...
           namespace={namespace}
           switchTitle="title"
         />
       </Provider>
     );
-    const onChange = wrapper.find(EventStartDateFilter).prop('onChange');
-    onChange(constants.START_DATE_ALL);
 
-    const query = {
+    const quertAll = {
```
ioannistsanaktsidis, line 60: small typo here

ui/src/seminars/containers/__tests__/SeminarStartDateFilterContainter.test.jsx
```diff
@@ -42,56 +39,48 @@ describe('SeminarStartDateFilterContainer', () => {
...
-    const expectedActions = [searchQueryUpdate(namespace, query)];
+    const expectedActions = [
+      searchQueryUpdate(namespace, quertAll),
+      searchQueryUpdate(namespace, queryUpcoming),
+    ];
+
+    const childButton = within(
+      getByTestId('event-start-date-filter')
+    ).getByRole('switch');
+    fireEvent.click(childButton);
```
ioannistsanaktsidis, line 80: I would propose that we keep the same approach as before by splitting the tests in 2 as they were. One test will render and check that the default state is dispatched. One other will do the click and check that the other state is dispatched. I would avoid merging them in one.

ioannistsanaktsidis (changes requested):
Some small comments!

ui/src/seminars/containers/__tests__/SeminarStartDateFilterContainter.test.jsx
ioannistsanaktsidis: Agreed we can have one for the default state and one like you have it. In this one that you combine both , lets also leave a small comment for the the click actions so as to know from a first reading in which state you should be before performing the action.

ioannistsanaktsidis (approved):
Looks good!

PR #3410, opened 2025-01-27
ui[experiments]: replace enzyme by rtl

DonHaul:
* ref: cern-sis/issues-inspire/issues/698


ui/src/experiments/__tests__/index.test.jsx
```diff
@@ -1,45 +1,138 @@
...
+
+    expect(
+      getByTestId('experiments-detail-page-container')
+    ).toBeInTheDocument();
+    const experimentCollaboration = getAllByRole('link')[0];
+    const experimentAssociatedArticlesLink = getAllByRole('link')[1];
+    const experimentCollaborationArticlesLink = getAllByRole('link')[2];
+    const experimentCollaborationMembersLink = getAllByRole('link')[3];
+
+    expect(experimentCollaboration.textContent).toBe('CERN-LHC-ATLAS');
```
ioannistsanaktsidis, line 76: Very nice! Could be better if we can explicitly check every link by name and attribute as this example
```
const link = screen.getByRole('link', { name: 'Author Detail' });
expect(link).toBeInTheDocument();
expect(link).toHaveAttribute(
  'href',
   "<LINK_HREF")
);
```

ioannistsanaktsidis (changes requested):
Small comment, otherwise it looks good! Nice job!

ui/src/experiments/__tests__/index.test.jsx
DonHaul: cool didn know we coudl do it like that

ioannistsanaktsidis (approved):
Looks good!

PR #3420, opened 2025-01-30
backend: add data facet for author agg

PascalEgn:
* Add endpoint to filter for Authors in Data Collections
* Add facet to fetch Data Records of Author
* Reindex data records if their linked Literature updates

ref: https://github.com/cern-sis/issues-inspire/issues/720

### DIFF
--- a/backend/inspirehep/records/api/literature.py
+++ b/backend/inspirehep/records/api/literature.py
@@ -35,7 +35,6 @@ from inspirehep.records.api.base import InspireRecord
 from inspirehep.records.api.mixins import (
     CitationMixin,
     ConferencePaperAndProceedingsMixin,
-    DataPapersMixin,
     ExperimentPapersMixin,
     InstitutionPapersMixin,
     JournalPapersMixin,
@@ -51,6 +50,7 @@ from inspirehep.records.marshmallow.literature import (
     LiteratureElasticSearchSchema,
     LiteratureFulltextElasticSearchSchema,
 )
+from inspirehep.records.models import DataLiterature
 from inspirehep.records.utils import (
     download_file_from_url,
     get_literature_earliest_date,
@@ -82,7 +82,6 @@ class LiteratureRecord(
     ExperimentPapersMixin,
     InstitutionPapersMixin,
     JournalPapersMixin,
-    DataPapersMixin,
     InspireRecord,
 ):
     """Literature Record."""
@@ -107,7 +106,6 @@ class LiteratureRecord(
         self.update_institution_relations()
         self.update_experiment_relations()
         self.update_journal_relations()
-        self.update_data_relations()
 
     @classmethod
     def create(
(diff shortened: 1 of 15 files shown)


backend/tests/integration/records/views/test_views_data.py
```diff
@@ -118,3 +121,81 @@ def test_data_returns_301_when_pid_is_redirected(inspire_app):
     assert response.status_code == 301
     assert response.location.split("/")[-1] == str(record.control_number)
     assert response.location.split("/")[-2] == "data"
+
+
+def test_data_facets(inspire_app):
+    create_record("dat")
+    with inspire_app.test_client() as client:
+        response = client.get("/data/facets")
+    response_data = orjson.loads(response.data)
```
ioannistsanaktsidis, line 130: I see in some places we do not use loads, and we access directly with response.json....  Is there a reason you used that one over the other ?

PascalEgn: No specific reason, just used it because we used it in the other tests too 👀

backend/inspirehep/records/tasks.py
```diff
@@ -193,3 +194,19 @@ def remove_bai_from_literature_authors(uuids):
...
+
+@shared_task(
+    ignore_results=False,
+    acks_late=True,
+    retry_backoff=True,
+    retry_kwargs={"max_retries": 8},
+)
+def reindex_linked_data(control_number):
+    """Reindex linked data records."""
+    linked_data_records = (
```
drjova, line 207: I would query DataLiterature table as THE source of truth for the relations.

backend/inspirehep/records/api/literature.py
```diff
@@ -550,6 +550,23 @@ def get_linked_papers_if_reference_changed(self):
...
+        authors = self.get_modified_authors()
+        if not authors:
+            return set()
+        linked_data_records = (
+            DataSearch()
+            .query_from_iq(f"literature.record.$ref:{self.control_number}")
+            .execute()
+        )
+        uuids = set([record._id for record in linked_data_records["hits"]["hits"]])
+        return uuids
```
drjova, line 568:
```suggestion
        return set([record._id for record in linked_data_records["hits"]["hits"]])
```

backend/inspirehep/records/marshmallow/utils.py
```diff
@@ -71,3 +71,38 @@ def get_acquisition_source_without_email(data):
...
+            author["control_number"]
+            for author in authors_with_record
+            if author.get("control_number")
+        ]
+    )
+    authors_without_record = [
+        author
+        for author in record.get("authors", [])
+        if "record" not in author
+        or int(PidStoreBase.get_pid_from_record_uri(author["record"].get("$ref"))[1])
```
drjova, line 95: why [1] ?

drjova (changes requested):
Few minor comments

backend/inspirehep/records/marshmallow/utils.py
PascalEgn: The code is 6 years old, i just moved it to have it cleaner, but:
`get_pid_type_from_recid` returns `[pid_type, recid]` and we want the `recid` in that case

backend/inspirehep/records/api/literature.py
```diff
@@ -550,6 +550,22 @@ def get_linked_papers_if_reference_changed(self):
         LOGGER.info("No references changed", uuid=str(self.id))
         return set()
 
+    def get_linked_datas_if_authors_changed(self):
```
drjova, line 553: can we do something similar as https://github.com/inspirehep/inspirehep/blob/3fd6eaa721e6ccab303500343aef018564a67cb5/backend/inspirehep/records/api/mixins.py#L706, so we don't rely on ES? check also `get_linked_papers_if_reference_changed`

PascalEgn: added the DataLiterature mixin and fetch the linked data records from its table now 👍

drjova (approved):
LGTM

ioannistsanaktsidis (approved):

ioannistsanaktsidis:
Do we need to trigger a full re-index of dat when this is merged? cc @PascalEgn @drjova  ?

drjova:
We need to populate the table, and reindex

PR #3427, opened 2025-02-05
fix(ui): Remove enzyme from Bibliography Generator

Lorenzovagliano:
ref: https://github.com/cern-sis/issues-inspire/issues/701

### DIFF
--- a/ui/src/bibliographyGenerator/__tests__/BibliograpyGeneratorPageContainer.test.jsx
+++ b/ui/src/bibliographyGenerator/__tests__/BibliograpyGeneratorPageContainer.test.jsx
@@ -1,12 +1,16 @@
 import React from 'react';
-import { mount } from 'enzyme';
+import { render } from '@testing-library/react';
 import { Provider } from 'react-redux';
 import { fromJS } from 'immutable';
-
 import BibliographyGeneratorPageContainer from '../BibliographyGeneratorPageContainer';
-import { getStoreWithState } from '../../fixtures/store';
 import BibliographyGenerator from '../BibliographyGenerator';
+import { getStore } from '../../fixtures/store';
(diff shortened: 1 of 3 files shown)


ui/src/bibliographyGenerator/__tests__/BibliograpyGeneratorPageContainer.test.jsx
```diff
@@ -17,7 +21,18 @@ describe('BibliographyGeneratorPageContainer', () => {
...
+
+  it('passes props to BibliographyGenerator component', () => {
+    const expectedState = {
+      data: fromJS({ download_url: 'https://google.com' }),
+      citationErrors: fromJS([{ message: 'Citation error 1' }]),
+      error: fromJS({ message: 'Error' }),
+      loading: false,
+    };
+
     const store = getStoreWithState({
```
ioannistsanaktsidis, line 36: Since you are touching the files, replace this with `getStore`as `getStoreWithState` is deprecated and will be removed.

ui/src/bibliographyGenerator/__tests__/BibliograpyGeneratorPageContainer.test.jsx
```diff
@@ -56,19 +76,23 @@ describe('BibliographyGeneratorPageContainer', () => {
...
       }),
     });
+
+    render(
+      <Provider store={store}>
+        <BibliographyGeneratorPageContainer />
+      </Provider>
+    );
+
+    const { onSubmit } = BibliographyGenerator.mock.calls[0][0];
```
ioannistsanaktsidis, line 86: Here it can be better if we can actually simulate the submission with fireEvent from rtl so as to simulate better the user behavior

ioannistsanaktsidis (changes requested):

ioannistsanaktsidis (approved):

PR #3434, opened 2025-02-11
fix(ui)(backoffice): Use ISO format for dates on author workflows

Lorenzovagliano:
ref: https://github.com/cern-sis/issues-inspire/issues/750

### DIFF
--- a/ui/src/backoffice/authors/containers/AuthorDetailPageContainer.tsx
+++ b/ui/src/backoffice/authors/containers/AuthorDetailPageContainer.tsx
@@ -31,6 +31,7 @@ import {
   getWorkflowStatusInfo,
   resolveDecision,
   filterByProperty,
+  formatDateTime,
 } from '../../utils/utils';
 import DeleteWorkflow from '../../common/components/DeleteWorkflow/DeleteWorkflow';
 import EmptyOrChildren from '../../../common/components/EmptyOrChildren';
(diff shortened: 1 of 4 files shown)


ui/src/backoffice/authors/containers/AuthorDetailPageContainer.tsx
```diff
@@ -75,9 +75,12 @@ const AuthorDetailPageContainer = ({
   const urls = data?.get('urls');
   const filteredIds = filterByProperty(data, 'ids', 'schema', 'ORCID', false);
   const acquisitionSourceEmail = data?.getIn(['acquisition_source', 'email']);
-  const acquisitionSourceDateTime = new Date(
-    data?.getIn(['acquisition_source', 'datetime'])
-  )?.toLocaleDateString();
+  const rawDateTime = data?.getIn(['acquisition_source', 'datetime']);
+  const acquisitionSourceDateTime = rawDateTime
```
ioannistsanaktsidis, line 79: we can make this a bit more elegant. We can create a small utility function and construct the date similar to   this example

```
export const formatDateTime = (isoString: string): string => {
  const [datePart, timePart] = isoString.split('T');
  const formattedTimePart = timePart.slice(0, 5);
  return `${datePart} ${formattedTimePart}`;
};
```

then in line 79 you can check if rawDateTime exists . If yes you call the function and you format the date, if no, you return undefined. 

Since we are already doing that, it would be nice to conditionally render the `ContentBox` that we show this information -> https://github.com/inspirehep/inspirehep/blob/master/ui/src/backoffice/authors/containers/AuthorDetailPageContainer.tsx#L284

as for the user does not make sense to see something like, `Submitted by ioannis on Invalid Date`. In that case if the date is invalid(which is very unlikely) we can show `Submitted by ioannis`

ioannistsanaktsidis (approved):

ui/src/backoffice/authors/containers/AuthorDetailPageContainer.tsx
```diff
@@ -75,9 +76,15 @@ const AuthorDetailPageContainer = ({
   const urls = data?.get('urls');
   const filteredIds = filterByProperty(data, 'ids', 'schema', 'ORCID', false);
   const acquisitionSourceEmail = data?.getIn(['acquisition_source', 'email']);
-  const acquisitionSourceDateTime = new Date(
-    data?.getIn(['acquisition_source', 'datetime'])
-  )?.toLocaleDateString();
+  const rawDateTime = data?.getIn(['acquisition_source', 'datetime']);
+
+  let acquisitionSourceDateTime: string | undefined;
+  if (rawDateTime) {
```
ioannistsanaktsidis, line 82: const acquisitionSourceDateTime = rawDateTime ? format.... : undefined

PR #3437, opened 2025-02-11
fix(ui)(backoffice): return separate date and time in formatDateTime()

Lorenzovagliano:
ref: https://github.com/cern-sis/issues-inspire/issues/750

Related: https://github.com/inspirehep/inspirehep/pull/3434


ioannistsanaktsidis (approved):
Nice!

PR #3439, opened 2025-02-12
fix(ui): fix fake-collections-menu on header

Lorenzovagliano:
ref: https://github.com/cern-sis/issues-inspire/issues/752

## Before:
https://github.com/user-attachments/assets/44362ec0-36c3-4964-9d97-9c441db72133

### Desktop

https://github.com/user-attachments/assets/7d1fe293-8284-4b72-b7ac-013748d1583e

### Mobile

https://github.com/user-attachments/assets/b793eaab-bafe-4a91-8c45-d1550e57ff21

### DIFF
--- a/ui/src/common/layouts/Header/Header.less
+++ b/ui/src/common/layouts/Header/Header.less
@@ -40,8 +40,18 @@
     }
   }
 
-  &.holdingpen .fake-collections-menu {
+  &.backoffice .fake-collections-menu {
     width: 100%;
     height: 10px;
+    background-color: @layout-header-background;
+    position: fixed;
+    top: @layout-header-height;
+    left: 0;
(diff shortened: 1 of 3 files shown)


ui/src/common/layouts/Header/Header.less
```diff
@@ -43,5 +43,10 @@
   &.holdingpen .fake-collections-menu {
```
drjova, line 43: since we are here, could you please also rename `.holdingpen` class to `.backoffice`? It might need to touch also some other files. 🥺

drjova (approved):

ui/src/common/layouts/Header/Header.less
Lorenzovagliano: Fixed this!

PR #3441, opened 2025-02-12
workflows[data]: add a yearly data harvest

DonHaul:
* ref: cern-sis/issues-inspire/issues/760

N.B. I havent added tests and this dag is only intended to be use temporarily. lmk if it makes sense to add them

### DIFF
--- a/workflows/plugins/hooks/tenacity_config.py
+++ b/workflows/plugins/hooks/tenacity_config.py
@@ -5,5 +5,5 @@ def tenacity_retry_kwargs():
     return {
         "wait": tenacity.wait_exponential(),
         "stop": tenacity.stop_after_attempt(5),
-        "retry": tenacity.retry_if_exception_type(Exception),
+        "retry": tenacity.retry_if_exception(lambda e: str(e) != "404:NOT FOUND"),
     }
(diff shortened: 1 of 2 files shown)


workflows/plugins/hooks/tenacity_config.py
```diff
@@ -1,9 +1,13 @@
...
+def should_retry(exception):
+    return str(exception) != "404:NOT FOUND"
+
+
 def tenacity_retry_kwargs():
     return {
         "wait": tenacity.wait_exponential(),
         "stop": tenacity.stop_after_attempt(5),
-        "retry": tenacity.retry_if_exception_type(Exception),
+        "retry": tenacity.retry_if_exception(should_retry),
```
drjova, line 12: you can use the `~` and `&` operators from py 3.10. This will ensure that we don't retry when the error is coming from the requests 404. Isn't it the case? We do the requests using the module.

```
        "retry": tenacity.retry_if_exception_type(Exception) & 
                 ~tenacity.retry_if_exception(lambda e: isinstance(e, requests.exceptions.HTTPError) and 
                                                       e.response is not None and 
                                                       e.response.status_code == 404),
```

DonHaul: I will check this implementation but in general it seems the Exception returned is an `AirflowException` so all of this `status.code` stuff dont seem to be directly available

drjova (approved):

PR #3450, opened 2025-02-14
workflows[data]: catch cases where update record is not found

DonHaul:


ioannistsanaktsidis (approved):

DonHaul:
may actually not be required, it was detecting a literature record, where a hepdata doi had been added as a material

DonHaul:
reference was removed this is no longer needed

PR #3452, opened 2025-02-17
workflows[data]: fix creation date

DonHaul:
* ref: cern-sis/issues-inspire/issues/767

### DIFF
--- a/workflows/dags/data/bulk_data_harvest.py
+++ b/workflows/dags/data/bulk_data_harvest.py
@@ -146,9 +146,9 @@ def bulk_data_harvest_dag():
         for payload in payloads:
             builder = DataBuilder(source="HEPData")
 
-            builder.add_creation_date(datetime.datetime.now(datetime.UTC).isoformat())
-
             base_record = payload["base"]
+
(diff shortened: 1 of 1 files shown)


workflows/dags/data/bulk_data_harvest.py
```diff
@@ -146,10 +146,10 @@ def add_keywords(record, builder):
         for payload in payloads:
             builder = DataBuilder(source="HEPData")
 
-            builder.add_creation_date(datetime.datetime.now(datetime.UTC).isoformat())
-
             base_record = payload["base"]
 
+            builder.add_creation_date(base_record["record"]["creation_date"])
+
```
drjova, line 152: don't we need to do that also in the normal harvest?

drjova (changes requested):
Thanks, a small clarification comment

workflows/dags/data/bulk_data_harvest.py
DonHaul: yes it's already there. I had forgotten to fix it here.

Should be okay now. if we end up requiring running these dag again in the future, we should merge the tasks so the code is shared

PR #3454, opened 2025-02-17
ui[data]: add creation date

DonHaul:
* ref: cern-sis/issues-inspire/issues/770


DonHaul:
Result
image

backend/inspirehep/records/marshmallow/data/ui.py
```diff
@@ -30,6 +32,13 @@ def get_number_of_authors(self, data):
             authors = get_authors(data)
         return self.get_len_or_missing(authors)
 
+    def get_date(self, data):
```
ioannistsanaktsidis, line 35: Lets add a small test in `json/test_data.py` for this.

ui/src/data/containers/DetailPageContainer/DetailPageContainer.tsx
```diff
@@ -100,6 +102,7 @@ const DetailPage = ({
                     page="Data detail"
                   />
                 </div>
+                <LiteratureDate date={date} />
```
ioannistsanaktsidis, line 105: this will render an empty span if no date here. Although it is very unlikely, we can do {date && ....}

ioannistsanaktsidis (dismissed):
Some minor improvements and we are good to 🚢

ioannistsanaktsidis (changes requested):
Based on the previous minor comments

backend/tests/integration/records/serializers/json/test_data.py
```diff
@@ -400,3 +401,8 @@ def test_data_elasticsearch_schema_serialize_literature(inspire_app):
         response_data = orjson.loads(response.data)
         response_data_hit = response_data["hits"]["hits"][0]["metadata"]
         assert response_data_hit["literature"] == expected_literature
+
+
+def test_data_format_date():
+    data_detail_schema = DataDetailSchema()
+    assert data_detail_schema.get_date({"creation_date": "2022-01-01"}) == "Jan 1, 2022"
```
ioannistsanaktsidis, line 408: Here it would be better if we can achieve the whole flow-> similar to how we do here https://github.com/inspirehep/inspirehep/blob/master/backend/tests/integration/records/serializers/json/test_data.py#L255, so as to simulate the whole flow of creating a record that gets serialized correctly when requesting /data/<recid>

ioannistsanaktsidis (changes requested):

ioannistsanaktsidis (approved):
Looks good! 👍

PR #3458, opened 2025-02-18
data-harvest-fix

DonHaul:


workflows/dags/data/bulk_data_harvest.py
```diff
@@ -210,11 +212,13 @@ def normalize_collaborations(records):
...
         """
         normalized_records = []
+
+        logger.info(f"Normalizing {len(records)} records")
         for record in records:
             collaborations = record.get("collaborations", [])
 
             if not collaborations:
-                continue
+                normalized_records.append(record)
```
ioannistsanaktsidis, line 221: do we actually need this at all ? we are doing it either way in line 240. Why was it skipped in the first place ?

ioannistsanaktsidis (changes requested):

workflows/dags/data/bulk_data_harvest.py
DonHaul: you have a point I would say we can do 
```
            if not collaborations:
                normalized_records.append(record)
                continue
```

to avoid also making unnecessary requests

workflows/dags/data/bulk_data_harvest.py
```diff
@@ -210,12 +212,11 @@ def normalize_collaborations(records):
...
         normalized_records = []
+
+        logger.info(f"Normalizing {len(records)} records")
         for record in records:
             collaborations = record.get("collaborations", [])
 
-            if not collaborations:
-                continue
-
             response = inspire_http_hook.call_api(
```
ioannistsanaktsidis, line 220: If you would like to avoid making extra calls is missing collaborations you can wrpa this part to an
`if collaborations:`

If we have collaborations, we do the normalization, if we do not have we do nothing.

Either way we append the record.

ioannistsanaktsidis (approved):
Looks good!

workflows/dags/data/bulk_data_harvest.py
```diff
@@ -293,7 +299,10 @@ def load_records(new_records):
 
     hepdata_record_versions = download_record_versions(collect_ids())
     records = build_records(
-        data_schema=data_schema, inspire_url=url, payloads=hepdata_record_versions
+        data_schema=data_schema,
+        inspire_url=url,
+        payloads=hepdata_record_versions,
+        logger=logger,
```
ioannistsanaktsidis, line 305: Just a question for my understanding, why we need to pass this here since it is declared on top ?

DonHaul: its a virtual environment task. in these tasks we have to reimport everything you need, or pass it in in order to use it

PR #3461, opened 2025-02-19
bump: inspire-schemas

DonHaul:
* ref: cern-sis/issues-inspire/issues/778


drjova:
@DonHaul we will have to update the default factory for `data` records https://github.com/inspirehep/inspirehep/blob/35d9bb8734abfa8e0b067d73e5dcd042d30be71c/backend/tests/helpers/providers/record_provider.py#L142 and also update any test records we load e.g. `1.json`

ioannistsanaktsidis (approved):
Looks good to me, although as discussed I am a bit confused with this test here `test_create_record_from_db_depending_on_its_pid_type ` . Maybe @drjova has an idea what we are trying to test and why we create a record using the base class + the specific one

backend/tests/helpers/providers/record_provider.py
```diff
@@ -143,6 +143,12 @@ def data_record():
         return {
             "$schema": "http://localhost:5000/schemas/records/data.json",
             "_collections": ["Data"],
+            "titles": [{"title": fake.sentence()}],
+            "dois": [
+                {
+                    "value": RecordProvider.doi(),
```
drjova, line 149: is dois required?

drjova: if not, I wouldn't add it here. It should be added in the tests where it's needed.

ioannistsanaktsidis:
@DonHaul LGTM, you can also trigger a PROD release after merging!

PR #3463, opened 2025-02-20
enable data in main bar + dropdown

ioannistsanaktsidis:
ref https://github.com/cern-sis/issues-inspire/issues/764
ref https://github.com/cern-sis/issues-inspire/issues/765

### DIFF
--- a/ui/src/common/layouts/CollectionsMenu/CollectionsMenuContainer.tsx
+++ b/ui/src/common/layouts/CollectionsMenu/CollectionsMenuContainer.tsx
@@ -1,16 +1,9 @@
 import { connect, RootStateOrAny } from 'react-redux';
 
 import CollectionsMenu from './CollectionsMenu';
-import { isCataloger, isSuperUser } from '../../authorization';
 
-const stateToProps = (state: RootStateOrAny) => {
-  const isCatalogerLoggedIn = isCataloger(state.user.getIn(['data', 'roles']));
-  const isSuperUserLoggedIn = isSuperUser(state.user.getIn(['data', 'roles']));
-  return {
-    canAccessDataCollection: isCatalogerLoggedIn || isSuperUserLoggedIn,
-    currentPathname: state.router.location.pathname,
-    currentHash: state.router.location.hash,
-  };
-};
+const stateToProps = (state: RootStateOrAny) => ({
+  currentPathname: state.router.location.pathname,
+});
 
 export default connect(stateToProps)(CollectionsMenu);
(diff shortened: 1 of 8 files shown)


ioannistsanaktsidis:
Can be reviewed already, but will be merged when we have the 🟢 from @michamos to go public

michamos:
Just to make sure: we don't want to show it to all users yet, we want to wait until all features have been implemented and tested.

ui/src/common/layouts/CollectionsMenu/CollectionsMenuContainer.tsx
```diff
@@ -1,16 +1,9 @@
...
-  const isCatalogerLoggedIn = isCataloger(state.user.getIn(['data', 'roles']));
-  const isSuperUserLoggedIn = isSuperUser(state.user.getIn(['data', 'roles']));
-  return {
-    canAccessDataCollection: isCatalogerLoggedIn || isSuperUserLoggedIn,
-    currentPathname: state.router.location.pathname,
-    currentHash: state.router.location.hash,
-  };
-};
+const stateToProps = (state: RootStateOrAny) => ({
+  currentPathname: state.router.location.pathname,
```
DonHaul, line 6: we dont need `currentHash`  anymore?

ui/src/common/layouts/CollectionsMenu/__tests__/CollectionsMenuContainer.test.jsx
```diff
@@ -28,73 +27,4 @@ describe('CollectionsMenuContainer', () => {
...
-    });
-
-    const { getByText } = render(
-      <Provider store={store}>
-        <MemoryRouter>
-          <CollectionsMenuContainer onHeightChange={jest.fn()} />
-        </MemoryRouter>
-      </Provider>
-    );
-    expect(getByText('Data')).toBeInTheDocument();
```
DonHaul, line 54: should a test be added to check if the CollectiomMenu has now a Data option?

ui/src/common/layouts/CollectionsMenu/CollectionsMenuContainer.tsx
ioannistsanaktsidis: nope, it is not used inside the component nor referenced anywhere. Must be a leftover from back in the days

ui/src/common/layouts/CollectionsMenu/__tests__/CollectionsMenuContainer.test.jsx
ioannistsanaktsidis: should be already covered by https://github.com/inspirehep/inspirehep/pull/3463/files#diff-b419d59ee8a9df5d5fbdfcdfd9f7ad31e45e37f216657f949f0b4aad1d15fac5R102 where we check each individual render + the snapshot that renders it https://github.com/inspirehep/inspirehep/pull/3463/files#diff-ac48b651ebab23f1b16b65ed61964597de8f7b2ea8e5ad4083d5b0f5f6f7b725R202. Since we don't have more tests than snapshots for this(we didn't have from before), I would say it is ok to go this way for now.

DonHaul (approved):

PR #3464, opened 2025-02-20
ui: datasets tab in author

ioannistsanaktsidis:
* ref https://github.com/cern-sis/issues-inspire/issues/721


ui/src/authors/containers/DetailPageContainer/DetailPageContainer.tsx
```diff
@@ -121,6 +123,8 @@ function DetailPage({
     [metadata]
   );
 
+  const canAccessDataTab = isCatalogerLoggedIn || isSuperUserLoggedIn;
+
   let tabItems = [
     {
       label: (
```
drjova, line 130: This should be renamed to `Literature`

drjova (approved):
LGTM 🥇 Just the order of the tables (e.g. Cited by at the end) and renaming to Literature.

PR #3473, opened 2025-02-24
feat(ui): don't open dataset links in new tab

heureka-code:
(Still using external url, in the future this needs to change or some components won't work correctly (Link))

ref: https://github.com/cern-sis/issues-inspire/issues/766

### DIFF
--- a/ui/src/literature/components/LiteratureItem.jsx
+++ b/ui/src/literature/components/LiteratureItem.jsx
@@ -136,6 +136,7 @@ function LiteratureItem({
                 text="datasets"
                 trackerEventId="Dataset links"
                 page={page}
+                isTargetBlank={false}
               />
             )}
             {canEdit && (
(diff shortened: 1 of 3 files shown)


heureka-code:
> Looks good! What do you mean though by this [#3473 (comment)](https://github.com/inspirehep/inspirehep/pull/3473#issue-2875287654) ?

In our tests we changed the isTargetBlank option to false for components. Inside of the url implementation there is a switch based on isTargetBlank that expands to a Link variant with _blank or a link without. The version without _blank uses a different kind of url setting (to instead of href) so it assumes the link to be internal. Currently this links is external and we ran in errors like a link like "/literature/https://www.hepdata.net/record/ins2872501" that was inserted into the href of an a tag in the final webpage view. We were told that the link will point to an internal page in the future so this won't be an issue.

PR #3479, opened 2025-02-27
ui: dataset count on load 

ioannistsanaktsidis:
ref https://github.com/cern-sis/issues-inspire/issues/797


ioannistsanaktsidis:
Note I already fire the request and hide the tab if the user is not admin/curator as we  will soon expose it to the public

drjova (approved):
LGTM

PR #3485, opened 2025-03-04
ui: facets for data + embedded

ioannistsanaktsidis:
* ref https://github.com/cern-sis/issues-inspire/issues/796

### DIFF
--- a/ui/src/middlewares/syncLocationWithSearch.js
+++ b/ui/src/middlewares/syncLocationWithSearch.js
@@ -14,11 +14,14 @@ import {
   LITERATURE_NS,
   BACKOFFICE_SEARCH_NS,
   PATHNAME_TO_NAMESPACE,
+  DATA_NS,
 } from '../search/constants';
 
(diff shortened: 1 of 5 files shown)


ui/src/middlewares/syncLocationWithSearch.js
```diff
@@ -78,7 +81,7 @@ export default function ({ dispatch, getState }) {
         isSearchPage(prevLocation)
       ) {
         dispatch(newSearch(prevNamespace));
-        if (nextNamespace === LITERATURE_NS) {
+        if (ALLOWED_NAMESPACES.includes(nextNamespace)) {
```
drjova, line 84: let's add a more clear name, because it's too generic and we will forget.

drjova (changes requested):

ui/src/middlewares/syncLocationWithSearch.js
ioannistsanaktsidis: Done!

PR #3488, opened 2025-03-04
ui: fix tooltip on range aggs

ioannistsanaktsidis:


ui/src/common/components/RangeAggregation/RangeAggregation.jsx
```diff
@@ -133,12 +144,40 @@ function getSliderMarks([lower, upper], [min, max]) {
...
+    case LITERATURE_NS:
+      return {
+        selectedTitle: 'Selected Papers',
+        totalTitle: 'Total Papers',
+      };
+    default:
+      return {
+        selectedTitle: 'Selected Papers',
+        totalTitle: 'Total Papers',
+      };
```
drjova, line 170:
```suggestion
    case LITERATURE_NS:
    default:
      return {
        selectedTitle: 'Selected Papers',
        totalTitle: 'Total Papers',
      };
```

drjova: Also we could add https://github.com/inspirehep/inspirehep/blob/78fbb7fd6f865d46b50e0405833ba80bc24c9826/ui/src/common/utils.ts#L181

drjova (changes requested):
Small changes

PascalEgn (approved):

ui/src/common/components/RangeAggregation/RangeAggregation.jsx
ioannistsanaktsidis: @drjova updated! For the pluralize, I am not single if it makes sense in this particular case. for the `totalTitle` for sure no. Could be for the selected one ?

drjova (approved):

PR #3489, opened 2025-03-05
backoffice[authors]: show all tickets

DonHaul:
* ref: https://github.com/cern-sis/issues-inspire/issues/620
image


drjova (changes requested):
Could we add them in a unordered list?

DonHaul:
> Could we add them in a unordered list?

look as an ordered list:
image

ioannistsanaktsidis (approved):

PR #3490, opened 2025-03-05
ui: hotfix for dataset links

ioannistsanaktsidis:


drjova (approved):
LGTM, does it work with `target=_blank`?

PR #3491, opened 2025-03-05
backend: add zulip dependency

PascalEgn:
ref: https://github.com/cern-sis/issues-inspire/issues/776

depends on: https://github.com/cern-sis/kubernetes-inspire/pull/246

### DIFF
--- a/backend/inspirehep/utils.py
+++ b/backend/inspirehep/utils.py
@@ -5,6 +5,7 @@
 # the terms of the MIT License; see LICENSE file for more details.
 
 import hashlib
+import os
 import resource
 import signal
 from contextlib import contextmanager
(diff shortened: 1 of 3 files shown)


backend/inspirehep/utils.py
```diff
@@ -5,19 +5,35 @@
...
 
 import structlog
+import zulip
 from flask import current_app
 from flask_celeryext.app import current_celery_app
 from redis import StrictRedis
 from redis_lock import Lock
 
 LOGGER = structlog.getLogger()
+zulip_client = zulip.Client()
```
drjova, line 23: I wouldn't initialize it in module level, we can either add it to the the function itself or as a proxy (not sure it worth it though). WDYT?

drjova (changes requested):
small comment

backend/inspirehep/utils.py
PascalEgn: i moved it into the function for now 👍

drjova (approved):
LGTM

PR #3493, opened 2025-03-06
backend: add zulip env vars

PascalEgn:
ref: https://github.com/cern-sis/issues-inspire/issues/776

depends on: https://github.com/cern-sis/kubernetes-inspire/pull/246

### DIFF
--- a/backend/inspirehep/utils.py
+++ b/backend/inspirehep/utils.py
@@ -24,7 +24,11 @@ LOGGER = structlog.getLogger()
 
 def send_zulip_notification(message: str):
     try:
-        zulip_client = zulip.Client()
+        zulip_client = zulip.Client(
+            api_key=os.environ.get("ZULIP_NOTIFICATION_API_KEY"),
+            email=os.environ.get("ZULIP_NOTIFICATION_EMAIL"),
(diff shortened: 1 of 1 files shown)


drjova (approved):

backend/inspirehep/utils.py
```diff
@@ -24,7 +24,11 @@
 
 def send_zulip_notification(message: str):
     try:
-        zulip_client = zulip.Client()
+        zulip_client = zulip.Client(
+            api_key=os.environ.get("ZULIP_NOTIFICATION_API_KEY"),
+            email=os.environ.get("ZULIP_NOTIFICATION_EMAIL"),
+            site=os.environ.get("ZULIP_NOTIFICATION_SITE"),
+        )
```
ioannistsanaktsidis, line 31: As we want this to run only in PROD, should we exclude adding the config on qa -> https://github.com/cern-sis/kubernetes-inspire/pull/246?

ioannistsanaktsidis: It's not a bad idea, in that case I guess we can send the messages to inspire-qa-tasks. WDYT @drjova ? Would QA make sense in your opinion ?

ioannistsanaktsidis (approved):

PR #3498, opened 2025-03-06
docker: separate postgres in different dbs 

DonHaul:


drjova (approved):
if it works would be great! LGTM, the tests need to be some attention.

DonHaul:
the tests do work. all the push development tests are passing

PR #3505, opened 2025-03-13
workflows: update creation date logic

ioannistsanaktsidis:
* ref https://github.com/cern-sis/issues-inspire/issues/810


workflows/dags/data/bulk_data_harvest.py
```diff
@@ -143,6 +146,20 @@ def add_keywords(record, builder):
                     for value in item:
                         builder.add_keyword(value)
 
+        def add_date(record, builder):
+            """Add date to the record."""
+            creation_date = record["creation_date"]
+            last_updated = record["last_updated"]
```
drjova, line 152: Is the key always there? Otherwise, it will raise `KeyError`. Also, it's contradicting with `if last_updated`. I guess we don't expect it to be `None`.

To avoid it you can use `.get`

workflows/dags/data/bulk_data_harvest.py
```diff
@@ -143,6 +146,20 @@ def add_keywords(record, builder):
                     for value in item:
                         builder.add_keyword(value)
 
+        def add_date(record, builder):
+            """Add date to the record."""
+            creation_date = record["creation_date"]
```
drjova, line 151: ditto

drjova (changes requested):

workflows/dags/data/bulk_data_harvest.py
ioannistsanaktsidis: I think it is always there since they populated it with the dummy 1970, from what I understood from this comment -> https://github.com/cern-sis/issues-inspire/issues/810#issuecomment-2704061840. Will update it though to make it fail safe 👍

ioannistsanaktsidis:
@drjova comments addressed!

workflows/dags/data/bulk_data_harvest.py
```diff
@@ -143,6 +146,20 @@ def add_keywords(record, builder):
...
+            last_updated = record.get("last_updated", None)
```
drjova:
```suggestion
            last_updated = record.get("last_updated")
```

by default it's None, no need to specify

drjova (approved):

PR #3506, opened 2025-03-14
monitoring: error messages for record-indexer

ioannistsanaktsidis:
* ref https://github.com/cern-sis/issues-inspire/issues/711

### DIFF
--- a/backend/inspirehep/logger/config.py
+++ b/backend/inspirehep/logger/config.py
@@ -8,8 +8,11 @@ import logging
 import sys
 
 import structlog
+from celery import shared_task
 from celery.signals import setup_logging, task_failure, task_postrun, task_prerun
 from structlog_sentry import SentryJsonProcessor
+
+from inspirehep.utils import send_zulip_notification
 
 # Sentry
 # ======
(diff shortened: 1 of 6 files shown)


backend/inspirehep/logger/config.py
```diff
@@ -89,8 +91,27 @@ def setup_basic_logging(*args, **kwargs):
     logging.basicConfig(format="%(message)s", stream=sys.stdout, level=logging.INFO)
 
 
+def construct_failure_message(task_name, exception, affected_records):
+    """Construct a failure message based on the task name, exception, and affected records."""
+    return f"**Task name**: `{task_name}`\n **Affected record(s)**: {affected_records}\n **Error message**: {exception}"
```
drjova, line 96: could we do something like:

Task name:

Error Message:

Affected records:
* 
* 
*

backend/inspirehep/logger/config.py
```diff
@@ -108,3 +129,7 @@ def log_error(
         task_args=args,
         task_kwargs=kwargs,
     )
+    task_name = sender.name if sender else "Unknown Task"
+    message = get_failure_message_by_task(task_name, exception, kwargs)
+    if message:
+        send_zulip_notification(message)
```
drjova, line 135: my only concern is that it's still blocking. A solution will be to send the zulip notification as a task. wdyt?

drjova (changes requested):

drjova:
Thanks @ioannistsanaktsidis few comments

ioannistsanaktsidis:
@drjova comments addressed as per our discussion!

drjova (approved):

PR #3510, opened 2025-03-18
monitoring: error messagees for hal push

ioannistsanaktsidis:
* ref https://github.com/cern-sis/issues-inspire/issues/822

### DIFF
--- a/backend/inspirehep/utils.py
+++ b/backend/inspirehep/utils.py
@@ -69,42 +69,40 @@ def format_failure_message_for_no_records(task_name, exception):
     return f"**Task name**: `{task_name}`\n\n" f"**Error message**: {exception}"
 
 
-def get_failure_message_for_batch_index(task_name, exception, *args, **kwargs):
+def get_failure_message_for_batch_index(task_name, exception, **kwargs):
     return format_failure_message_for_multiple(
         task_name, exception, kwargs.get("records_uuids", [])
     )
 
 
-def get_failure_message_for_index_record(task_name, exception, *args, **kwargs):
+def get_failure_message_for_index_record(task_name, exception, **kwargs):
     return format_failure_message_for_single(task_name, exception, kwargs.get("uuid"))
 
 
-def extract_uuid_list(*args):
-    """Safely extracts the first inner list of UUIDs from the given args."""
-    try:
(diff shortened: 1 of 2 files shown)


backend/inspirehep/utils.py
```diff
@@ -101,9 +101,30 @@ def get_failure_message_for_match_references_by_uuids(
...
+        return recid, record_version
+    return None, None
+
+
+def get_failure_message_for_hal_push(task_name, exception, *args, **kwargs):
+    recid, record_version = extract_recid_and_version(*args)
+    if recid:
+        if record_version:
+            recid = f"{recid} (version {record_version})"
+        return format_failure_message_for_single(task_name, exception, recid)
```
drjova, line 127: we can simplify this, since we know what we want to get we could just read them directly and not access args, kwargs etc


```python
def get_failure_message_for_redirect_references_to_merged_record(
    task_name, exception, uuid=None, **kwargs
):
    if uuid:
        return format_failure_message_for_single(task_name, exception, uuid)
    return format_failure_message_for_no_records(task_name, exception)


def get_failure_message_for_hal_push(
    task_name, exception, recid=None, record_version=None, **kwargs
):
    if recid:
        if record_version:
            recid = f"{recid} (version {record_version})"
        return format_failure_message_for_single(task_name, exception, recid)
    return format_failure_message_for_no_records(task_name, exception)
    
```
    
   also it's better to return early and avoid the nested ifs for example we can do something like:
    
```python
def get_failure_message_for_hal_push(task_name, exception, recid=None, record_version=None, **kwargs):
    if not recid:
        return format_failure_message_for_no_records(task_name, exception)
    
    if record_version:
        recid = f"{recid} (version {record_version})"
    
    return format_failure_message_for_single(task_name, exception, recid)
    ```

drjova (changes requested):

backend/inspirehep/utils.py
ioannistsanaktsidis: Thanks for the comments! Will address the improvements suggestions!

ioannistsanaktsidis:
@drjova comments addressed as requested!

drjova (approved):

PR #3513, opened 2025-03-19
actions: add lint dependency to inspirehep tests

PascalEgn:
Add dependency to only run tests if linting was successful


ioannistsanaktsidis (approved):
👍

PR #3517, opened 2025-03-20
monitoring: error messages for disambiguation

DonHaul:
* ref: cern-sis/issues-inspire/issues/710

### DIFF
--- a/backend/inspirehep/utils.py
+++ b/backend/inspirehep/utils.py
@@ -85,6 +85,36 @@ def get_failure_message_for_index_record(task_name, exception, **kwargs):
     return format_failure_message_for_single(task_name, exception, kwargs.get("uuid"))
 
 
+def get_failure_message_for_disambiguate_records(
+    task_name, exception, uuid=None, record_version_id=None, **kwargs
+):
+    if isinstance(uuid, list):
+        if not record_version_id:
+            return format_failure_message_for_multiple(task_name, exception, uuid)
+        uuid = [
+            f"{u} (version {rec_id})"
+            for u, rec_id in zip(uuid, record_version_id, strict=False)
+        ]
+        return format_failure_message_for_multiple(task_name, exception, uuid)
+    if record_version_id:
+        uuid = f"{uuid} (version {record_version_id})"
+
+    return format_failure_message_for_single(task_name, exception, uuid)
+
+
+def get_failure_message_for_celery_starmap(task_name, exception, *args, **kwargs):
+    if not kwargs.get("task"):
+        return
+    task_name = kwargs.get("task", {})["task"]
+    items = kwargs.get("it", [])
+    uuids = [arg[0] for arg in items]
+    version_ids = [arg[1] for arg in items]
+
+    return get_failure_message_by_task(
+        task_name, exception, uuids, version_ids, *args, **kwargs
+    )
+
+
 def get_failure_message_for_orcid_push(task_name, exception, **kwargs):
     return format_failure_message_for_single(
         task_name, exception, kwargs.get("rec_id"), kwargs.get("orcid")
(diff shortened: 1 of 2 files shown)


backend/inspirehep/utils.py
```diff
@@ -85,6 +85,21 @@ def get_failure_message_for_index_record(task_name, exception, **kwargs):
     return format_failure_message_for_single(task_name, exception, kwargs.get("uuid"))
 
 
+def get_failure_message_for_disambiguate_records(task_name, exception, *args, **kwargs):
```
ioannistsanaktsidis, line 88: As we already know the arguments passed, you can unpack it directly. 

`uuid=None, record_version_id=None`

Also you can skip `**kwargs`  as you are not using it

backend/inspirehep/utils.py
```diff
@@ -85,6 +85,21 @@ def get_failure_message_for_index_record(task_name, exception, **kwargs):
...
+    if isinstance(args[0], list):
+        return format_failure_message_for_multiple(task_name, exception, args[0])
+    return format_failure_message_for_single(task_name, exception, args[0])
+
+
+def get_failure_message_for_celery_starmap(task_name, exception, *args, **kwargs):
+    task_name = kwargs.get("task")["task"]
```
ioannistsanaktsidis, line 95: if task is not present, `get('task')` will return `None`, and by doing `["task"]` will throw a `TypeError`. You can make it that part fail safe and only proceed if task name exists

backend/inspirehep/utils.py
```diff
@@ -85,6 +85,21 @@ def get_failure_message_for_index_record(task_name, exception, **kwargs):
...
+    uuids = []
+    for arg in kwargs.get("it"):
+        uuids.append(arg[0])
```
ioannistsanaktsidis, line 98: You can make it as a comprehension `uuids = [arg[0] for arg in kwargs.get("it", [])]`

ioannistsanaktsidis (changes requested):
Some comments

backend/inspirehep/utils.py
DonHaul: after checking `**kwargs` is required to catch arguments that are being received and not used such as `disambiguate_all_not_disambiguated` or `tasks`

backend/inspirehep/utils.py
```diff
@@ -85,6 +85,33 @@ def get_failure_message_for_index_record(task_name, exception, **kwargs):
...
+
+    return format_failure_message_for_single(task_name, exception, uuid)
+
+
+def get_failure_message_for_celery_starmap(task_name, exception, *args, **kwargs):
+    if not kwargs.get("task"):
+        return
+    task_name = kwargs.get("task", {})["task"]
+    uuids = [arg[0] for arg in kwargs.get("it", [])]
+    version_ids = [arg[1] for arg in kwargs.get("it", [])]
```
ioannistsanaktsidis, line 108: you can directly unpack before,

```
items = kwargs.get("it", [])
uuids = [arg[0] for arg in items]
....
```

for better readablility

backend/inspirehep/utils.py
```diff
@@ -85,6 +85,33 @@ def get_failure_message_for_index_record(task_name, exception, **kwargs):
     return format_failure_message_for_single(task_name, exception, kwargs.get("uuid"))
 
 
+def get_failure_message_for_disambiguate_records(
+    task_name, exception, uuid=None, record_version_id=None, **kwargs
+):
+    if isinstance(uuid, list):
+        uuid = [
+            f"{u} (version {rec_id})"
+            for u, rec_id in zip(uuid, record_version_id, strict=False)
```
ioannistsanaktsidis, line 94: Here if record_version_id is None, the zip will return empty list. you can check that before for safety. Also if record_version_id is None, it will throw an error. Right now it is not the case but better to make it safer.

ioannistsanaktsidis (changes requested):
Just a minor change and we are good!

ioannistsanaktsidis (approved):
All good!

PR #3518, opened 2025-03-20
monitoring: add error messages for airflow

ioannistsanaktsidis:
* ref https://github.com/cern-sis/issues-inspire/issues/790


ioannistsanaktsidis:
2 questions for the reviewers: 

* In the URL section of the message, do you know if the URL will be resolved correctly ?
* Does it make sense to move this https://github.com/inspirehep/inspirehep/pull/3518/files#diff-eed7e25e9047fda838f807638aa1db54fa86150d02c32410f04b266a7c833947R9 to inspire-utils library? Or you think it does not worth the effort ?

ioannistsanaktsidis:
@drjova yes this is a possibility. We can do it that way if you prefer the email instead of zulip. I was more in favor of having everything concentrated in zulip so as we don't get lost. But thats my 2 cents. If you are in favor of emails, I will change the PR.

drjova:
What I mean is we can do the same thing without the Zulip SDK. Because Zulip can receive emails

ioannistsanaktsidis:
@DonHaul we already have one that we can use, so it should be out of the box without having to install the sdk! I think it is more straightforward since the email is supported + we already have the smtp.

DonHaul:
> @DonHaul we already have one that we can use, so it should be out of the box without having to install the sdk! I think it is more straightforward since the email is supported + we already have the smtp.

okay I see! As a sidenote, shouldn't we also this smtp for the monitoring we used for the `hep` monitoring?

ioannistsanaktsidis:
@drjova @DonHaul comments addressed!

PR #3522, opened 2025-03-24
workflows: fix log url

ioannistsanaktsidis:
NOTE: Hostname from k8s points to the pod it self. We need the base url for the log to be pointed correctly.

### DIFF
--- a/workflows/plugins/include/utils/alerts.py
+++ b/workflows/plugins/include/utils/alerts.py
@@ -1,5 +1,6 @@
 import logging
 import os
+from urllib.parse import urlparse, urlunparse
 
 from airflow.utils.email import send_email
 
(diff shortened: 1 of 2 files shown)


workflows/plugins/include/utils/alerts.py
```diff
@@ -14,10 +15,20 @@ def task_failure_alert(context):
         logger.error("Cannot send email.")
         return
 
+    base_url = os.environ.get("AIRFLOW_BASE_URL")
+    task_instance = context["task_instance"]
+    parsed_log_url = urlparse(task_instance.log_url)
```
drjova, line 20: so by default airflow returns the log_url as with `localhost` as a host?

ioannistsanaktsidis: yes by default it returns it with localhost:8080. On the other hand I just saw that, https://airflow.apache.org/docs/apache-airflow/stable/configurations-ref.html#base-url . Let me try to replace it with the appropriate varialble, maybe it does the job out of the box for us

ioannistsanaktsidis: that will do the trick . Will close this pr and re-open another one

ioannistsanaktsidis:
Closing as superseded by https://github.com/inspirehep/inspirehep/pull/3523

PR #3523, opened 2025-03-24
workflows: fix webserver url

ioannistsanaktsidis:


docker-compose.yml
```diff
@@ -70,6 +70,7 @@ x-airflow-common: &airflow-common
     context: workflows
     dockerfile: Dockerfile.local
   environment: &airflow-common-env
+    AIRFLOW__WEBSERVER__BASE_URL: ${AIRFLOW__WEBSERVER__BASE_URL}
```
drjova, line 73: this is just for local though, right? We can set it on k8s directly

ioannistsanaktsidis: yes we will set it as extraEnv in k8s

PR #3524, opened 2025-03-25
ui: increase max dropdown height

PascalEgn:
ref: https://github.com/cern-sis/issues-inspire/issues/817

### DIFF
--- a/ui/src/common/components/SearchBoxNamespaceSelect.tsx
+++ b/ui/src/common/components/SearchBoxNamespaceSelect.tsx
@@ -13,6 +13,7 @@ const SearchBoxNamespaceSelect = ({
   onSearchScopeChange: Function;
 }) => (
   <SelectBox
+    listHeight={400}
     popupClassName="header-dropdown"
     onChange={onSearchScopeChange}
     value={searchScopeName}


ui/src/common/components/SelectBox.jsx
```diff
@@ -3,7 +3,12 @@ import { Select } from 'antd';
 
 function SelectBox({ options, virtualScroll = false, ...selectProps }) {
   return (
-    <Select dropdownMatchSelectWidth={virtualScroll} showArrow {...selectProps}>
+    <Select
+      listHeight={400}
```
ioannistsanaktsidis, line 7: By doing this, it will increase the height for every component that references the `SelectBox`. You can pass it as a prop only to the one we want to increase the listHeight. It will work out of the box as we have the `...selectProps`

ioannistsanaktsidis (changes requested):
Small change

ui/src/common/components/SelectBox.jsx
PascalEgn: should be fine now :)

ioannistsanaktsidis (approved):
LGTM

PR #3526, opened 2025-03-25
workflows[data]: fix data harvest

DonHaul:
* ref: cern-sis/issues-inspire/issues/824


ioannistsanaktsidis (approved):
Looks good, since we are here, I see some duplicate keys in `version3` json. Could you please remove them ?  Other than that it is ok!

PR #3528, opened 2025-03-27
backend: add halpush duplicate header

PascalEgn:
ref: https://github.com/cern-sis/issues-inspire/issues/831

### DIFF
--- a/backend/tests/integration/hal/test_core_sword.py
+++ b/backend/tests/integration/hal/test_core_sword.py
@@ -19,6 +19,8 @@
 # granted to it by virtue of its status as an Intergovernmental Organization
 # or submit itself to any jurisdiction.
 
+import contextlib
+
 import mock
 import orjson
 import pytest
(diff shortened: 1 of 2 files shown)


backend/tests/integration/hal/test_core_sword.py
```diff
@@ -231,3 +235,32 @@ def side_effect(*args, **kwargs):
...
+@mock.patch("inspirehep.hal.core.sword.HttpLib2LayerIgnoreCert.request", autospec=True)
+def test_update_sends_force_doublon_by_title_header(mock_request, inspire_app):
+    mock_request.side_effect = lambda self, *args, **kwargs: _original_request(
+        self, *args, **kwargs
+    )
+    with contextlib.suppress(Exception):
+        update("<xml>test</xml>", "123", None)
+    mock_request.assert_called_once()
+
+    headers = mock_request.call_args[1]["headers"]
```
ioannistsanaktsidis, line 265: I see here you are grabbing it from differently. Why is that ?

ioannistsanaktsidis (approved):
Just a question for my understanding!

backend/tests/integration/hal/test_core_sword.py
PascalEgn: just a relict from a past version i forgot, gonna unify it real quick, good that you spotted it 👌

PR #3529, opened 2025-03-28
chore: silent return when no api key in zulip

ioannistsanaktsidis:


PascalEgn:
Is this to easily disable the notifications? Or why we only doing it for the API key and not email and channel?

ioannistsanaktsidis:
It is to easily disable the notifications. I put it only for the API_KEY for simplicity, we can put it though for the whole config so as to be consistent.

PascalEgn (approved):

ioannistsanaktsidis:
I have updated, I think it is nicer checking them all as they are all required for the config

PascalEgn:
> I have updated it, I think it is nicer checking them all as they are all required for the config, just for consistency

In that case we dont need to catch a config exception anymore

PR #3530, opened 2025-04-01
ui: center the no results message on search page

karolina-siemieniuk-morawska:
* ref: cern-sis/issues-inspire#820

https://github.com/user-attachments/assets/72ae11f1-ad38-443d-8b3e-cab12e265b9e

### DIFF
--- a/ui/src/conferences/containers/SearchPageContainer.jsx
+++ b/ui/src/conferences/containers/SearchPageContainer.jsx
@@ -15,9 +15,13 @@ import DocumentHead from '../../common/components/DocumentHead';
 import { CONFERENCES_NS } from '../../search/constants';
 import ConferenceItem from '../components/ConferenceItem';
 import ConferenceStartDateFilterContainer from './ConferenceStartDateFilterContainer';
-import { SEARCH_PAGE_GUTTER } from '../../common/constants';
+import {
+  SEARCH_PAGE_GUTTER,
+  SEARCH_PAGE_COL_SIZE_WITH_FACETS,
+} from '../../common/constants';
 import { APIButton } from '../../common/components/APIButton';
 import { isSuperUser } from '../../common/authorization';
+import EmptyOrChildren from '../../common/components/EmptyOrChildren';
 
 const META_DESCRIPTION = 'Find conferences in High Energy Physics';
 const TITLE = 'Conferences Search';
@@ -30,6 +34,7 @@ function ConferenceSearchPage({
   loading,
   loadingAggregations,
   isSuperUserLoggedIn,
+  results,
 }) {
   const renderAggregations = useCallback(
     () => (
@@ -50,7 +55,7 @@ function ConferenceSearchPage({
     <>
       <DocumentHead title={TITLE} description={META_DESCRIPTION} />
       <Row>
-        <Col xs={24} lg={22} xl={20} xxl={18}>
+        <Col {...SEARCH_PAGE_COL_SIZE_WITH_FACETS}>
           <Row
             className="mt3"
             gutter={SEARCH_PAGE_GUTTER}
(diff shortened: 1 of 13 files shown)


ioannistsanaktsidis (changes requested):
This is fine, but it is only for the literature collection, currently data collection also uses the same message style for 0 results, so we should center it there too. Since we are on it, we should take the chance and apply the same `0 results` message style in the other collections too (jobs/authors etc) so as everything looks unified

karolina-siemieniuk-morawska:
All collections are now changed accordingly to their layout (with or without facets)

Screenshot 2025-04-01 at 14 13 20
Screenshot 2025-04-01 at 14 13 11
Screenshot 2025-04-01 at 14 12 33
Screenshot 2025-04-01 at 14 12 45
Screenshot 2025-04-01 at 14 12 23
Screenshot 2025-04-01 at 14 13 31
Screenshot 2025-04-01 at 14 12 54
Screenshot 2025-04-01 at 14 13 02

ui/src/conferences/containers/SearchPageContainer.jsx
```diff
@@ -61,42 +66,44 @@ function ConferenceSearchPage({
...
-                      max="md"
-                      render={() => (
-                        <DrawerHandle
-                          className="mt2"
-                          handleText="Filter"
-                          drawerTitle="Filter"
-                        >
-                          {renderAggregations()}
-                        </DrawerHandle>
+              <EmptyOrChildren data={results} title="0 conferences">
```
ioannistsanaktsidis, line 69: Lets capitalize them all for consistency -> `0 Conferences` etc

ui/src/conferences/containers/SearchPageContainer.jsx
```diff
@@ -50,7 +55,7 @@ function ConferenceSearchPage({
     <>
       <DocumentHead title={TITLE} description={META_DESCRIPTION} />
       <Row>
-        <Col xs={24} lg={22} xl={20} xxl={18}>
+        <Col {...SEARCH_PAGE_COL_SIZE_WITH_FACETS}>
```
ioannistsanaktsidis, line 58: What is the reason of not utilizing your utility function here and you directly assign that one ? We do not need the SEARCH_PAGE_COL_SIZE_NO_RESULTS here in case there are no results ?

ui/src/seminars/components/SearchPage.jsx
```diff
@@ -14,7 +15,7 @@ function SearchPage() {
     <>
       <DocumentHead title={TITLE} description={META_DESCRIPTION} />
       <Row>
-        <Col xs={24} lg={22} xl={20} xxl={18}>
+        <Col {...SEARCH_PAGE_COL_SIZE_WITH_FACETS}>
```
ioannistsanaktsidis, line 18: Same here. Btw that can also be a container in case we want to utilize your funciton.

ioannistsanaktsidis (commented):
One comment and one question for clarification

ui/src/conferences/containers/SearchPageContainer.jsx
karolina-siemieniuk-morawska: it is due to difference in layout in those two pages - facets are always visible, no matter if there are results or not, unlike to other pages. my function takes numberOfResults as first argument to decide which size to display, I could use it here as ```...columnSize(1, true)``` (use fake number of results) to make it work, but I think in those two cases it's better to use ```SEARCH_PAGE_COL_SIZE_WITH_FACETS``` directly

ui/src/seminars/components/SearchPage.jsx
karolina-siemieniuk-morawska: replied above

ui/src/conferences/containers/SearchPageContainer.jsx
ioannistsanaktsidis: ok, thanks for clarifying, we can have it like this

ioannistsanaktsidis (approved):
All good!

PR #3531, opened 2025-04-02
celery: task parameters cleanup

DonHaul:
* ref: https://github.com/cern-sis/issues-inspire/issues/821


backend/inspirehep/disambiguation/tasks.py
```diff
@@ -225,7 +225,7 @@ def disambiguate_authors(
         LOGGER.warning(
             "Received stale data",
             recent_version_id=version_id,
-            record_version_id=record.model.version_id,
+            record_version_id=record.mgitodel.version_id,
```
ioannistsanaktsidis, line 228: leftover?

backend/inspirehep/assign/tasks.py
```diff
@@ -130,7 +130,7 @@ def assign_conference(record, conference_ref, cnum):
 
 
 @shared_task(
-    ignore_result=False,
+    ignore_result=True,
```
ioannistsanaktsidis, line 133: This is the export_papers_to_cds one. Even though declared as a task, I can only find it here https://github.com/inspirehep/inspirehep/blob/master/backend/inspirehep/assign/views.py#L226 called synchronously. Is that the intended behavior ?  @drjova any insights on this ?

ioannistsanaktsidis (commented):
A minor comment and a question  mostly for @drjova . In general, the changes look valid, we should just ensure no regression since almost every task is changed

backend/inspirehep/assign/tasks.py
drjova: It's a mistake, it should be triggered as a task in the views

ioannistsanaktsidis: Thank you! We can have a small extra PR to fix that @DonHaul

ioannistsanaktsidis (approved):
Looks good! Any doubts @drjova ? we never utilize the result of the task, so these changes should improve resources also

backend/inspirehep/disambiguation/tasks.py
```diff
@@ -209,10 +209,10 @@ def _disambiguate_authors(authors_to_disambiguate, record):
 
 
 @shared_task(
-    ignore_result=False,
```
drjova, line 212: why removing it from here?

backend/inspirehep/orcid/tasks.py
```diff
@@ -245,7 +245,7 @@ def _find_user_matching(orcid, email):
     return User.query.filter_by(email=email).one_or_none()
 
 
-@shared_task
+@shared_task(ignore_result=True)
```
drjova, line 248: acks_late=True

backend/inspirehep/records/tasks.py
```diff
@@ -27,13 +27,12 @@
 
 
 @shared_task(
-    bind=True,
```
drjova, line 30: ignore_result=True

backend/inspirehep/submissions/tasks.py
```diff
@@ -10,7 +10,7 @@
 from inspirehep.snow.api import InspireSnow
 
 
-@shared_task(ignore_result=False, max_retries=5)
+@shared_task(ignore_result=True, max_retries=5)
```
drjova, line 13: acks_late=True,

backend/inspirehep/disambiguation/tasks.py
drjova: Yep!

DonHaul: if we set it to `True` the following line fails:
https://github.com/inspirehep/inspirehep/blob/dbbe2338631ab178af3cab3bfdd23e60e835e768/backend/tests/integration-async/disambiguation/test_disambiguation_tasks.py#L1310

I removed it because by default its already `False`

ioannistsanaktsidis: It hangs because with True the result is never reported back as we ignore it. Wondering if we can just skip this assertion

drjova: yes, it doesn't make sense to keep the result just for that

backend/inspirehep/indexer/tasks.py
```diff
@@ -15,8 +15,8 @@
 LOGGER = structlog.getLogger()
 
 
-@shared_task(ignore_result=False, bind=True)
-def batch_index(self, records_uuids, request_timeout=None):
+@shared_task
```
ioannistsanaktsidis, line 18: Do we need the result in that one ?

backend/inspirehep/matcher/tasks.py
```diff
@@ -23,7 +23,6 @@
 
 
 @shared_task(
-    ignore_result=False,
```
drjova, line 26: ignore_result=True,

drjova (changes requested):
few more comments :)

backend/inspirehep/matcher/tasks.py
DonHaul: it would fail as its being used here
https://github.com/inspirehep/inspirehep/blob/master/backend/inspirehep/matcher/cli.py#L50

backend/inspirehep/assign/tasks.py
```diff
@@ -197,14 +197,13 @@ def _get_claimed_author_name_for_paper(from_author_recid, paper_authors):
 
 
 @shared_task(
+    ignore_result=True,
     queue="assign",
-    bind=True,
     retry_backoff=2,
```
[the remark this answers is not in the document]
DonHaul: added

backend/inspirehep/indexer/tasks.py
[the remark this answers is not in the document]
ioannistsanaktsidis: I think it should be fine like this no? Basically the only change is that bind is now False, making it not accessible by `self` as @DonHaul said. The part commented is related to `ignore_result` in my opinion, which correctly remains `False`.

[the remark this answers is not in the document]
drjova: 👍

drjova (approved):

PR #3532, opened 2025-04-02
ui: add search feedback form

miguelgrc:
- Removes old feedback floating button and form
- Adds search feedback card at the bottom of search results and when no results
- The feedback is sent to an endpoint in the AI backend
- Added feature flag to enable/disable
- Added matomo events
- The collected feedback is to be used by Jaén to develop the new AI tools

Screenshot 2025-04-02 at 17 02 54
Screenshot 2025-04-02 at 17 03 07


ui/src/App.less
```diff
@@ -239,3 +238,10 @@ body {
     }
   }
 }
+
+.ant-message {
```
ioannistsanaktsidis, line 242: I assume this is the default class coming directly from ant. Can we have a custom class for the overrides? In case we add another component that uses this class, these changes will take effect without noticing and they are very specific to this usecase

ui/public/config.js
```diff
@@ -10,6 +10,7 @@
     BACKOFFICE_URL: 'http://localhost:8001',
     EDITOR_SCHEMA_URL: 'https://inspirebeta.net/schemas/records/authors.json',
     EDITOR_BACKOFFICE_API_URL: 'http://localhost:8001/api',
+    SEARCH_FEEDBACK_CARD_FEATURE_FLAG: true,
```
ioannistsanaktsidis, line 13: Usually we have it by default false, and we override in k8s

ui/src/common/components/SearchFeedback/SearchFeedback.tsx
```diff
@@ -0,0 +1,122 @@
...
+  const [feedbackModalVisible, setFeedbackModalVisible] = useState(false);
+  const [form] = useForm();
+
+  const question = useWatch('question', form);
+
+  const handleSubmit = async () => {
+    try {
+      const values = form.getFieldsValue();
+
+      const response = await fetch('/ai/v1/search-feedback', {
```
ioannistsanaktsidis, line 18: I see you are using fetch. We mainly use axios and HttpClientWrapper, is it possible to utilize it for consistency e.g https://github.com/inspirehep/inspirehep/blob/master/ui/src/common/http.ts#L145?

ioannistsanaktsidis (changes requested):
Some minor comments!

ui/src/App.less
miguelgrc: Yes, the problem is that without some of these settings (especially the z-index) the messages don't show (as they are covered by the header and the footer), and anyway since we have the header visible in all pages we can't really put this at the top since it would look a bit messy over the header, so I think if we use messages for this or any other other use case (and to have some consistency) it would be better to have them at the bottom

PR #3540, opened 2025-04-08
editor: ensure default user_email

ioannistsanaktsidis:
* ref https://github.com/cern-sis/issues-inspire/issues/834

### DIFF
--- a/backend/inspirehep/editor/views.py
+++ b/backend/inspirehep/editor/views.py
@@ -117,9 +117,14 @@ def get_revisions(endpoint, pid_value):
         revisions = []
         for revision in reversed(record.revisions):
             transaction_id = revision.model.transaction_id
-
-            user = Transaction.query.filter(Transaction.id == transaction_id).one().user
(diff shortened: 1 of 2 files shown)


backend/inspirehep/editor/views.py
```diff
@@ -117,9 +117,14 @@ def get_revisions(endpoint, pid_value):
         revisions = []
         for revision in reversed(record.revisions):
             transaction_id = revision.model.transaction_id
-
-            user = Transaction.query.filter(Transaction.id == transaction_id).one().user
-            user_email = user.email if user else "system"
+            transaction = Transaction.query.filter(
```
drjova, line 120: no tests for that 😿

drjova (approved):

backend/inspirehep/editor/views.py
ioannistsanaktsidis: Updated! The test added should cover the erroneous scenario !

drjova (approved):

PR #3543, opened 2025-04-17
orcid: fix titles with xml

drjova:
* ref: https://github.com/cern-sis/issues-inspire/issues/829

### DIFF
--- a/backend/inspirehep/orcid/builder.py
+++ b/backend/inspirehep/orcid/builder.py
@@ -6,6 +6,7 @@
 
 from lxml import etree
 from lxml.builder import ElementMaker
+from lxml.html import fragment_fromstring
 from six import text_type
 
 _NAMESPACES = {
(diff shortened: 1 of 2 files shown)


DonHaul (approved):
Seems good 👍

backend/inspirehep/orcid/builder.py
```diff
@@ -18,6 +19,8 @@
 
 _ELEMENT_MAKERS = {"work": _WORK, "common": _COMMON}
 
+XML_TAG_RE = re.compile(r"<[^>]+>")
```
michamos, line 22: I don't think this is correct, what about a title like `Hubble tension between z < 5 and z > 9`?

drjova: You are right, I didn't thought about this case. Thanks for the suggestion.

PR #3548, opened 2025-04-29
tasks: catch KeyError on redirect references

ioannistsanaktsidis:
* ref https://github.com/cern-sis/issues-inspire/issues/882

### DIFF
--- a/backend/inspirehep/records/tasks.py
+++ b/backend/inspirehep/records/tasks.py
@@ -74,8 +74,16 @@ def update_records_relations(self, uuids):
 )
 def redirect_references_to_merged_record(self, uuid):
     record = InspireRecord.get_record(uuid, with_deleted=True)
-    new_record_ref = record["new_record"]["$ref"]
-    deleted_record_ref = record["self"]["$ref"]
+    new_record = record.get("new_record")
+    deleted_record = record.get("self")
+    if not (new_record and deleted_record):
+        LOGGER.error(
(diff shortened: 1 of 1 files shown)


backend/inspirehep/records/tasks.py
```diff
@@ -74,8 +74,16 @@ def update_records_relations(self, uuids):
...
-    new_record_ref = record["new_record"]["$ref"]
-    deleted_record_ref = record["self"]["$ref"]
+    new_record = record.get("new_record")
+    deleted_record = record.get("self")
+    if not (new_record and deleted_record):
+        LOGGER.error(
+            "Missing 'new_record' or 'self' for uuid.",
+            uuid=str(uuid),
+        )
+        return
```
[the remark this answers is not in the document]
ioannistsanaktsidis: Sure @drjova! Updated!

backend/inspirehep/records/tasks.py
```diff
@@ -74,8 +74,22 @@ def update_records_relations(self, uuids):
...
+            "redirect_references_to_merged_record: Missing 'new_record' field for record with uuid.",
+            uuid=str(uuid),
+        )
+        return
+    if not deleted_record:
+        LOGGER.error(
+            "redirect_references_to_merged_record: Missing 'self' field for record with uuid.",
+            uuid=str(uuid),
+        )
+        return
```
drjova, line 90: if both are missing it will never go to `deleted_record`, right?

ioannistsanaktsidis: Updated to the first solution with more granular error message

PR #3549, opened 2025-05-07
isbn: catch correctly the exception

ioannistsanaktsidis:
* ref https://github.com/cern-sis/issues-inspire/issues/902


drjova (approved):
LGTM

PR #3551, opened 2025-05-12
bump: inspire-matcher, inspire-dojson and inspire-schemas

DonHaul:


ui/package.json
```diff
@@ -55,7 +55,7 @@
     "history": "^4.7.2",
     "idlejs": "^3.0.0",
     "immutable": "^4.1.0",
-    "inspire-schemas": "61.5.26",
+    "inspire-schemas": "^61.6.16",
```
ioannistsanaktsidis, line 58: As discussed regarding this, for some reason this commit -> https://github.com/inspirehep/inspirehep/commit/64f385edd3aeda0b4a8821405a59a3072db7cf1a pins it explicitly to a specific version (not sure why), so lets validate that this was not done on purpose since the version is quite old. Also, from what I can see, we only use the schemas js in the ui to get the `countryName` https://github.com/inspirehep/inspirehep/blob/dce9da97ded78f569a6dac8d0a4147922db4b710/ui/src/submissions/common/schemas/constants.js#L1, hence I don't expect any problem. Lets validate though so as to be 100%

ioannistsanaktsidis (commented):
Looks good, lets just validate the point in the comment and we are good to go

ioannistsanaktsidis (approved):
Discussed In Person with @drjova , lets go ahead!

PR #3555, opened 2025-05-16
exception: robust to None request

DonHaul:
* ref: cern-sis/issues-inspire/issues/881

### DIFF
--- a/backend/inspirehep/orcid/tasks.py
+++ b/backend/inspirehep/orcid/tasks.py
@@ -209,7 +209,8 @@ def orcid_push(self, orcid, rec_id, oauth_token, kwargs_to_pusher=None):
             message = (exc.args[0:1] or ("",))[0]
             with contextlib.suppress(AttributeError):
                 message += f"\nResponse={exc.response.content}"
-            message += f"\nRequest={exc.request.method} {exc.request.url}"
+            if exc.request:
+                message += f"\nRequest={exc.request.method} {exc.request.url}"
             exc.args = (message,) + exc.args[1:]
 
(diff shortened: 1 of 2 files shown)


backend/inspirehep/orcid/tasks.py
```diff
@@ -209,7 +209,8 @@ def orcid_push(self, orcid, rec_id, oauth_token, kwargs_to_pusher=None):
             message = (exc.args[0:1] or ("",))[0]
             with contextlib.suppress(AttributeError):
                 message += f"\nResponse={exc.response.content}"
-            message += f"\nRequest={exc.request.method} {exc.request.url}"
+                if exc.request is not None:
```
ioannistsanaktsidis, line 212: shouldn't this be outside the context manager ? like before ?

DonHaul: yes you're right

backend/inspirehep/orcid/tasks.py
```diff
@@ -209,7 +209,8 @@ def orcid_push(self, orcid, rec_id, oauth_token, kwargs_to_pusher=None):
             message = (exc.args[0:1] or ("",))[0]
             with contextlib.suppress(AttributeError):
                 message += f"\nResponse={exc.response.content}"
-            message += f"\nRequest={exc.request.method} {exc.request.url}"
+            if exc.request is not None:
```
ioannistsanaktsidis, line 212: we can just do `if exc.request:` no ?

DonHaul: True, I often forget the simple syntax

ioannistsanaktsidis (approved):
LGTM

PR #3565, opened 2025-05-26
cds: migrate to airflow

ioannistsanaktsidis:
* ref https://github.com/cern-sis/issues-inspire/issues/926


workflows/dags/cds/cds_harvest.py
```diff
@@ -0,0 +1,140 @@
...
+    inspire_http_record_management_hook = InspireHTTPRecordManagementHook()
+
+    @task(task_id="get_cds_data")
+    def get_cds_data(**context):
+        """Starts harvesting cds.
+
+        Returns: data from cds
+        """
+        ds = context["ds"]
+        since = context["params"]["since"] or ds_add(ds, -1)
```
ioannistsanaktsidis, line 41: @drjova is it enough if `since` is not provided to fallback to previous day ? We can also have 
`
since = context["params"]["since"] or context.get("prev_execution_date_success") or ds_add(ds, -1)
`
which fallsback first to the last successful run

workflows/plugins/include/utils/cds.py
```diff
@@ -0,0 +1,122 @@
...
+    """
+    Build an Elasticsearch‐style query URL that matches any of the provided
+    control_numbers, arxivs or dois.
+
+    e.g. q=control_number:1234 OR control_number:5678 OR arxiv:1901.1234 OR ...
+    """
+    clauses = []
+    for cn in control_numbers:
+        clauses.append(f"control_number:{cn}")
+    for ax in arxivs:
```
drjova, line 98:
```suggestion
    for arxiv_id in arxiv_ids:
```

workflows/plugins/include/utils/cds.py
```diff
@@ -0,0 +1,122 @@
...
+    dois,
+):
+    """
+    Build an Elasticsearch‐style query URL that matches any of the provided
+    control_numbers, arxivs or dois.
+
+    e.g. q=control_number:1234 OR control_number:5678 OR arxiv:1901.1234 OR ...
+    """
+    clauses = []
+    for cn in control_numbers:
```
drjova, line 96:
```suggestion
    for control_number in control_numbers:
```

workflows/dags/cds/cds_harvest.py
drjova: we can just `-1`

workflows/plugins/include/utils/cds.py
```diff
@@ -0,0 +1,122 @@
...
+    hits = resp.get("hits", {}).get("hits", [])
+    logger.info(
+        f"Found {len(hits)} hits for query",
+    )
+    if len(hits) == 1:
+        record = hits[0]["metadata"]
+        control_number = record.get("control_number")
+        logger.info(f"{log_message} Control number: {control_number}")
+        return control_number
+    return None
```
drjova, line 58: Just to make the code a bit more readable (less nested), I would re-write it with early return

```suggestion
    if not len(hits) == 1:
        return None
    record = hits[0]["metadata"]
    control_number = record.get("control_number")
    logger.info(f"{log_message} Control number: {control_number}")
    return control_number

```

workflows/plugins/include/utils/cds.py
```diff
@@ -0,0 +1,122 @@
...
+    if lit_url:
+        recid = search_and_return_single(
+            hook=inspire_http_record_management_hook,
+            query_url=lit_url,
+            log_message="Matched record by `control_number`, `arxiv`, or `doi`.",
+        )
+        if recid:
+            return recid
+
+    rn_url = build_report_search_url(report_numbers)
```
drjova, line 74: It's better to use full names just for readability

workflows/plugins/include/utils/cds.py
```diff
@@ -0,0 +1,122 @@
...
+        logger.info(f"{log_message} Control number: {control_number}")
+        return control_number
+    return None
+
+
+def get_record_for_provided_ids(
+    inspire_http_record_management_hook, control_numbers, arxivs, dois, report_numbers
+):
+    lit_url = build_literature_search_url(control_numbers, arxivs, dois)
+    if lit_url:
```
drjova, line 65: why if lit_url? Can we return no url?

workflows/dags/cds/cds_harvest.py
```diff
@@ -0,0 +1,140 @@
...
+                f"Updating record with {payload['updated_record']} "
+                f"and with revision ID {revision_id}"
+            )
+            response = inspire_http_record_management_hook.update_record(
+                data=updated_record,
+                pid_type="literature",
+                control_number=control_number,
+                revision_id=revision_id + 1,
+            )
+            response.raise_for_status()
```
drjova, line 128: what happens if it blows up?

drjova (changes requested):

ioannistsanaktsidis:
@drjova comments addressed as discussed. Could you give it another go ?

workflows/plugins/include/utils/cds.py
```diff
@@ -0,0 +1,98 @@
+import logging
+
+from idutils import is_arxiv
+from inspire_utils.record import get_value
+
+logger = logging.getLogger(__name__)
+
+LITERATURE_PID_TYPE = "literature"
```
drjova, line 8: we could put this in constants?

workflows/plugins/include/utils/cds.py
```diff
@@ -0,0 +1,98 @@
...
+        "fields": "control_number",
+    }
+    logger.info(
+        f"Searching for record with query: {query_params}",
+    )
+    resp = hook.search_records(pid_type=LITERATURE_PID_TYPE, query_params=query_params)
+    logger.info(
+        f"Response from search: {resp}",
+    )
+    hits = resp.get("hits", {}).get("hits", [])
```
drjova, line 48: we could use `get_value`, just for consistency

workflows/plugins/include/utils/cds.py
```diff
@@ -0,0 +1,98 @@
...
+    for doi in dois:
+        clauses.add(f"dois.value:{doi}")
+    for report_number in report_numbers:
+        arxiv_candidate = report_number.lower().split("arxiv:")[-1]
+        if is_arxiv(arxiv_candidate):
+            clauses.add(f"arxiv:{arxiv_candidate}")
+        else:
+            clauses.add(f'report_numbers.value.fuzzy:"{report_number}"')
+
+    return " OR ".join(sorted(clauses))
```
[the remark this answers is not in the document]
drjova: sorted though it would not match the original order though

workflows/requirements.txt
```diff
@@ -1,2 +1,3 @@
 apache-airflow==2.9.3
 inspire-utils~=3.0.63
+idutils==1.4.3
```
drjova, line 3: do we need to hard pin?

workflows/tests/test_cds_tasks.py
```diff
@@ -0,0 +1,111 @@
...
+    dag = dagbag.get_dag("cds_harvest_dag")
+
+    @pytest.mark.vcr
+    def test_get_cds_data_vcr(self):
+        task = self.dag.get_task("get_cds_data")
+        res = task.execute(
+            context={"ds": "2025-05-23", "params": {"since": "2025-05-23"}}
+        )
+        # initial value 355, filtered to 109
+        assert len(res) == 109
```
drjova, line 22: I would avoid hardcoded number checking , maybe we could check few ids that are there?

drjova (changes requested):
Looks good, few more comments

PR #3589, opened 2025-06-11
cdsrdm: harvest new cds

ioannistsanaktsidis:
* ref https://github.com/cern-sis/issues-inspire/issues/929

### DIFF
--- a/workflows/dags/cds/cds_rdm_harvest.py
+++ b/workflows/dags/cds/cds_rdm_harvest.py
@@ -1,0 +1,129 @@
+import datetime
+import logging
+from urllib.parse import parse_qs, urlparse
+
+from airflow.decorators import dag, task, task_group
+from airflow.models.param import Param
+from airflow.providers.http.operators.http import HttpOperator
+from hooks.inspirehep.inspire_http_record_management_hook import (
+    InspireHTTPRecordManagementHook,
+)
+from include.utils.alerts import task_failure_alert
+from include.utils.cds import (
+    get_dois,
+    get_identifiers_for_scheme,
+    has_any_rdm_id,
+    retrieve_and_validate_record,
+    update_record,
+)
+from inspire_utils.record import get_value
+
+logger = logging.getLogger(__name__)
+
+
+def _pagination_fn(response):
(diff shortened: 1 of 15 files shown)


workflows/dags/cds/cds_rdm_harvest.py
```diff
@@ -0,0 +1,134 @@
...
+logger = logging.getLogger(__name__)
+
+
+@dag(
+    start_date=datetime.datetime(2025, 5, 22),
+    schedule="0 4 * * *",
+    catchup=False,
+    tags=["cds_rdm"],
+    params={
+        "since": Param(type=["null", "string"], default=""),
```
drjova, line 30: do we need `null`?

workflows/dags/cds/cds_rdm_harvest.py
```diff
@@ -0,0 +1,134 @@
...
+    def process_cds_rdm_response(cds_record):
+        @task(task_id="process_record")
+        def process_record(cds_record):
+            cds_id = cds_record.get("id")
+            identifiers = get_value(cds_record, "metadata.identifiers", [])
+            control_numbers = get_identifiers_for_scheme(identifiers, "inspire")
+            arxivs = get_identifiers_for_scheme(identifiers, "arxiv")
+            dois = force_list(
+                get_value(cds_record, "pids.doi.identifier", [])
+            ) + force_list(get_value(cds_record, "parent.pids.doi.identifier", []))
```
drjova, line 94: can we use the `has_any_rdm_id`?

workflows/dags/cds/cds_rdm_harvest.py
ioannistsanaktsidis: nope, removed

workflows/dags/cds/cds_rdm_harvest.py
ioannistsanaktsidis: we also need these values here. The has_any_rdm_id is just used for prefiltering. I have extracted dois get in a separate function

drjova (changes requested):

drjova (approved):
LGTM thanks @ioannistsanaktsidis 👍

PR #3596, opened 2025-06-17
workflows: cache venvs in celery workers

DonHaul:
* ref: cern-sis/issues-inspire/issues/933


ioannistsanaktsidis (approved):
Lets give it a try on dev first for some days!

PR #3600, opened 2025-06-19
backend: enrich records for cds

ioannistsanaktsidis:
* ref https://github.com/cern-sis/issues-inspire/issues/876


drjova (changes requested):
Thank you! Few minor comments

PR #3602, opened 2025-06-20
backend: fix linked books

ioannistsanaktsidis:
* ref https://github.com/cern-sis/issues-inspire/issues/950

### DIFF
--- a/backend/inspirehep/records/marshmallow/literature/utils.py
+++ b/backend/inspirehep/records/marshmallow/literature/utils.py
@@ -34,10 +34,23 @@ def get_pages(data):
 
 
 def get_parent_records(data):
-    book_records = InspireRecord.get_linked_records_from_dict_field(
-        data, "publication_info.parent_record"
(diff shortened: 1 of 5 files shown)


backend/inspirehep/records/marshmallow/literature/utils.py
```diff
@@ -34,10 +34,23 @@ def get_pages(data):
...
+    ]
+    book_records = list(
+        InspireRecord.get_linked_records_from_dict_field(
+            data, "publication_info.parent_record"
+        )
+    )
+
+    return sorted(
+        book_records,
+        key=lambda rec: linked_pids_order.index(str(rec["control_number"])),
```
drjova, line 52: what is the purpose of the sorting? because:

# Sort numerically (int)
[1, 2, 10]

# Sort (str)
[1, 10, 2]

drjova (changes requested):
just a question

drjova (approved):

PR #3604, opened 2025-06-24
cds: optimize query

ioannistsanaktsidis:
* ref https://github.com/cern-sis/issues-inspire/issues/876

### DIFF
--- a/backend/inspirehep/records/marshmallow/literature/cds.py
+++ b/backend/inspirehep/records/marshmallow/literature/cds.py
@@ -1,5 +1,4 @@
 from inspire_dojson.utils import get_recid_from_ref
-from invenio_pidstore.errors import PIDDoesNotExistError
 from marshmallow import fields
 
 from inspirehep.records.api import InspireRecord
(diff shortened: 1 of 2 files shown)


backend/inspirehep/records/marshmallow/literature/cds.py
```diff
@@ -28,24 +25,35 @@ def merge_orcids(author):
...
-    for institution in institutions:
-        ror = [
-            identifier
-            for identifier in institution.get("external_system_identifiers", [])
-            if identifier.get("schema") == "ROR"
-        ]
-        if ror:
-            ror_ids.extend(ror)
-    return ror_ids
+    if not has_ror:
```
drjova, line 35: I would invert to return early

backend/inspirehep/records/marshmallow/literature/cds.py
```diff
@@ -54,12 +62,21 @@ class LiteraturePublicSchemaCDS(LiteraturePublicSchema):
...
+            data, "authors.record"
+        )
+        institution_records = InspireRecord.get_linked_records_from_dict_field(
+            data, "authors.affiliations.record"
+        )
+        resolved_authors_by_id = {
+            record["control_number"]: record for record in author_records
+        }
+        resolved_institutions_by_id = {
+            str(record["control_number"]): record for record in institution_records
```
drjova, line 75: why we use here `str` and not for `authors`?

backend/inspirehep/records/marshmallow/literature/cds.py
```diff
@@ -28,24 +25,35 @@ def merge_orcids(author):
...
-            ror_ids.extend(ror)
-    return ror_ids
+    if not has_ror:
+        ror_ids = []
+        institutions_linked_pids = InspireRecord._get_linked_pids_from_field(
+            author, "affiliations.record"
+        )
+        for _, recid in institutions_linked_pids:
+            institution_record = resolved_institutions_by_id.get(recid)
+            if institution_record:
```
drjova, line 42: ditto

drjova (changes requested):
few minor comments

backend/inspirehep/records/marshmallow/literature/cds.py
ioannistsanaktsidis: Becuase for the institutions one later in the function we do 

https://github.com/inspirehep/inspirehep/pull/3604/files#diff-f616257fa08231f84a2654e38bef46c7648caae30e21598a86c13f8f9f49b361R37-R38

This returns a tuple of `[(pid_type, pid_value(string)]` :/ 

where for the authors we do 
https://github.com/inspirehep/inspirehep/pull/3604/files#diff-f616257fa08231f84a2654e38bef46c7648caae30e21598a86c13f8f9f49b361R11

which returns correctly the control number as int

ioannistsanaktsidis:
@drjova comments addressed as requested

PR #3623, opened 2025-07-04
gh wf: update workflow-tests

DonHaul:
`airflow db init`  has been removed on airflow 3
use `airflow db migrate` instead


ioannistsanaktsidis:
@DonHaul can we already merge this one before the upgrade to 3 is complete ?

DonHaul:
> @DonHaul can we already merge this one before the upgrade to 3 is complete ?

yes. its already valid since 2.7: https://airflow.apache.org/docs/apache-airflow/stable/release_notes.html#the-db-init-db-upgrade-commands-and-database-load-default-connections-configuration-options-are-deprecated-33136

PR #3624, opened 2025-07-04
fix: only put resolved authors in es

ioannistsanaktsidis:
* ref https://github.com/cern-sis/issues-inspire/issues/970


drjova (approved):
LGTM, thank you for the change!

PR #3626, opened 2025-07-07
backoffice: generalize documents/indexes

ioannistsanaktsidis:
* ref https://github.com/cern-sis/issues-inspire/issues/960


backoffice/backoffice/authors/tests/test_views.py
```diff
@@ -141,7 +142,10 @@ class TestAuthorWorkflowSearchViewSet(BaseTransactionTestCase):
     def setUp(self):
         super().setUp()
 
-        index = registry.get_indices().pop()
+        target_name = settings.OPENSEARCH_INDEX_NAMES[settings.AUTHORS_DOCUMENTS]
+        index = next(
+            (idx for idx in registry.get_indices() if idx._name == target_name), None
+        )
```
DonHaul, line 148: i would convert this into a function as we are already defining it in two places, and it will also be used on the `hep_tests`

backoffice/backoffice/hep/api/serializers.py
```diff
@@ -0,0 +1,108 @@
...
+        ),
+        OpenApiExample(
+            "Reject",
+            description="Hep Workflow Serializer",
+            value={"value": "reject", "create_ticket": False},
+        ),
+    ],
+)
+class HepResolutionSerializer(serializers.Serializer):
+    value = serializers.ChoiceField(choices=HEP_DECISION_CHOICES)
```
DonHaul, line 107: current `HEP_DECISION_CHOICES` may change in the future but ok for now

DonHaul (approved):
Small comment otherwise great  👍

backoffice/backoffice/authors/tests/test_views.py
ioannistsanaktsidis: Good catch, will convert it to a utils func.

backoffice/backoffice/common/utils.py
```diff
@@ -1,12 +1,22 @@
...
 
 
+def add_decision(workflow_id, user, action, serializer_class):
+    data = {"workflow": workflow_id, "user": user, "action": action}
+
+    serializer = serializer_class(data=data)
+    if serializer.is_valid(raise_exception=True):
+        serializer.save()
+        return serializer.data
+
```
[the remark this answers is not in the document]
ioannistsanaktsidis: yeap we can do that! Will change the implementation!

drjova (approved):
Just the comment, thanks!

PR #3630, opened 2025-07-14
workflows: use s3 for arxiv_harvest

DonHaul:
* ref: cern-sis/issues-inspire/issues/988


DonHaul:
assumptions:
- the bucket is called `data-store`

PR #3631, opened 2025-07-15
970 generalize views

ioannistsanaktsidis:


backoffice/backoffice/common/views.py
```diff
@@ -38,3 +51,67 @@ def create(self, request, *args, **kwargs):
...
         serializer.save()
         return Response(serializer.data, status=status.HTTP_201_CREATED)
+
+
+class BaseWorkflowViewSet(viewsets.ModelViewSet):
+    """
+    Base for any “Workflow” ModelViewSet.
+    Subclasses must set:
+      • schema_name
+    """
```
drjova, line 61: I would add a quick explanation what that means. i.e. the name should match with the inspire-schemas etc

drjova (approved):
LGTM just the small comment

PR #3632, opened 2025-07-15
workflows: add until option to cds-rdm-harvest

PascalEgn:
ref: https://github.com/cern-sis/issues-inspire/issues/952

### DIFF
--- a/workflows/dags/cds/cds_rdm_harvest.py
+++ b/workflows/dags/cds/cds_rdm_harvest.py
@@ -57,6 +57,7 @@ def _response_filter(responses):
     tags=["cds_rdm"],
     params={
         "since": Param(type=["string"], default=""),
+        "until": Param(type=["string"], default=datetime.datetime.now().isoformat()),
     },
     on_failure_callback=task_failure_alert,
 )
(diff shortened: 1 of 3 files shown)


workflows/dags/cds/cds_rdm_harvest.py
```diff
@@ -69,7 +70,7 @@ def cds_rdm_harvest_dag():
         method="GET",
         endpoint="/api/records",
         data={
-            "q": "{{ 'updated:>=' ~ (params.since or ds) }}",
+            "q": "updated:[{{ params.since or ds }} TO {{ params.until }}]",
```
drjova, line 73: Did we check with the CDS team that this is actually the recommended way?

DonHaul: ah cool didnt know this syntax was a possibility

PascalEgn: Using the timestamp is optional, for example:
https://repository.cern/api/records?q=updated%3A%5B2025-07-15+TO+2025-07-15T14%3A30%3A00.538472%5D&page=1&size=50&sort=newest

also works. (Which in this case is a scheduled dag run)

PR #3635, opened 2025-07-16
backoffice: hep views

ioannistsanaktsidis:
* ref https://github.com/cern-sis/issues-inspire/issues/960


backoffice/backoffice/authors/tests/test_search_filter_viewset.py
```diff
@@ -0,0 +1,151 @@
...
+
+        for item in response.json()["results"]:
+            assert item["status"] == AuthorStatusChoices.RUNNING
+
+    def test_filter_workflow_type(self):
+        self.api_client.force_authenticate(user=self.admin)
+
+        url = self.endpoint + f'?workflow_type="={AuthorWorkflowType.AUTHOR_CREATE}'
+
+        response = self.api_client.get(url)
```
drjova, line 111: generally we can do something like the following and not adding strings

```
self.api_client.get(
    self.endpoint,
    data={'workflow_type': AuthorWorkflowType.AUTHOR_CREATE},
)

```

backoffice/backoffice/authors/tests/test_search_filter_viewset.py
```diff
@@ -0,0 +1,151 @@
...
+
+        for item in response.json()["results"]:
+            assert item["workflow_type"] == AuthorWorkflowType.AUTHOR_CREATE
+
+    def test_ordering_updated_at(self):
```
drjova, line 116: we can use parametrize as in other tests

drjova (changes requested):
Thanks! It looks good, just few comments

PR #3636, opened 2025-07-17
Action development s3

DonHaul:


workflows/scripts/connections/connections.json
```diff
@@ -79,18 +79,14 @@
...
-    "description": "SMTP connection for sending emails",
-    "login": "",
-    "password": "",
-    "host": "email-server.inspire.ch",
-    "port": 25,
+    "s3_conn": {
+    "conn_type": "aws",
+    "description": "Used for arxiv harvesting temporary storage",
+    "login": "airflow",
+    "password": "airflow-inspire",
```
ioannistsanaktsidis, line 86: should we have them as vars and use them in the hook directly ?

DonHaul: by vars do you mean `environment variables`?

ioannistsanaktsidis: As we do it here https://github.com/inspirehep/inspirehep/blob/master/workflows/scripts/variables/variables.json. And then in the hook you can read it by `Variables.get`. That way we are not exposing them. Or am I missing sth ?

workflows/plugins/include/utils/s3.py
```diff
@@ -0,0 +1,50 @@
+import json
```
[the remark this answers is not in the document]
ioannistsanaktsidis: I think he means sth like https://github.com/inspirehep/inspirehep/blob/master/workflows/plugins/hooks/inspirehep/inspire_http_record_management_hook.py

workflows/scripts/connections/connections.json
DonHaul: I All our other connections are defined also `connections.json`  including the `inspire_db_connection` which also contains passwords. On `qa` and `prod` this values are different and also encrypted

drjova (approved):
LGTM

PR #3648, opened 2025-07-22
backoffice: check orcid exists on creation

DonHaul:
* ref: cern-sis/issues-inspire/issues/874


backend/inspirehep/submissions/views.py
```diff
@@ -109,7 +113,12 @@ def send_post_request_to_workflows(
                 endpoint=endpoint,
                 bearer_keyword=bearer_keyword,
             )
-            raise WorkflowStartError from e
+            match response.status_code:
```
drjova, line 116: might be a bit an overkill for 2 cases but nice!

PR #3653, opened 2025-07-29
submissions: fix error message

DonHaul:
* ref: cern-sis/issues-inspire/issues/874


backend/inspirehep/submissions/errors.py
```diff
@@ -20,7 +20,7 @@ class WorkflowStartError(RESTException):
 
 class WorkflowConflictError(RESTException):
```
drjova, line 21: shouldn't we be more explicit with the error name?

PR #3655, opened 2025-07-30
backend: remove cds endpoints from hep

DonHaul:
* ref: cern-sis/issues-inspire/issues/927


drjova:
Thanks! We should add also an alembic recipe to drop the table too

DonHaul:
> Thanks! We should add also an alembic recipe to drop the table too

added

ioannistsanaktsidis (approved):
LGTM! Lets run alembic on qa first

PR #3659, opened 2025-08-01
ui: fix see running dags button

DonHaul:
* ref: cern-sis/issues-inspire/issues/1019


DonHaul:
Now that I think about it the "See error details here" seems kinda redundant

PR #3663, opened 2025-08-04
workflows: error not being sent to bo

DonHaul:
* ref cern-sis/issues-inspire/issues/1012


workflows/dags/literature/arxiv_harvest.py
```diff
@@ -51,7 +51,7 @@
             ],
         ),
     },
-    on_failure_callback=task_failure_alert,
+    on_failure_callback=DagErrorNotifier(HEP),
```
ioannistsanaktsidis, line 54: lets do `collection=HEP` for consistency with the authors

workflows/plugins/include/utils/alerts.py
```diff
@@ -35,12 +37,17 @@ def task_failure_alert(context):
     )
 
 
-def dag_failure_callback(context):
-    """Callback function for DAG failure."""
-    try:
-        task_failure_alert(context)
-    except Exception as e:
-        logger.error(f"Error in sending alert: {e}")
+class DagErrorNotifier(BaseNotifier):
```
ioannistsanaktsidis, line 40: I would create 2 of them, one  that has the collection and also sets the status, like you do here,  and one simple which is just sending alerts without any collection. It would be easier in case we want to extend in the future

ioannistsanaktsidis (changes requested):
Small changes  otherwise looks good!

ioannistsanaktsidis (approved):

PR #3665, opened 2025-08-05
Basic hep decisions2

DonHaul:


backoffice/backoffice/hep/models.py
```diff
@@ -39,3 +39,4 @@ class HepDecision(BaseDecision):
         HepWorkflow, related_name="decisions", on_delete=models.CASCADE
     )
     action = models.CharField(max_length=30, choices=HEP_DECISION_CHOICES)
+    value = models.CharField(max_length=30, default="", blank=True)
```
drjova, line 42: hmm why do we need value?

DonHaul: this was before based on the author decisions where we only had the `approval` decision that could be either: `accept` `reject`. For `hep` we will have many kinds of decision `exact_match` `fuzzy_match` `merging_decisions` etc, this type of decision would be specified under `action`, and then value would correspond to the actual decision, say for `exact_match` this `value` would be the id of people we want to match for example

PR #3693, opened 2025-09-02
ui: fix for conference test and select field issue

karolina-siemieniuk-morawska:
ui: fix for conference test and select field issue

* ref: cern-sis/issues-inspire#983
* ref: cern-sis/issues-inspire#1003


ioannistsanaktsidis (approved):
Let's give it a try on QA and if all good we deploy Monday morning

PR #3701, opened 2025-09-04
workflows: hep normalize collaborations

DonHaul:
* ref: cern-sis/issues-inspire/issues/1050


workflows/dags/literature/hep_create_dag.py
```diff
@@ -119,7 +122,39 @@ def guess_coreness(**context):
...
+            response = inspire_http_hook.call_api(
+                endpoint="api/curation/literature/collaborations-normalization",
+                method="GET",
+                data={"collaborations": collaborations},
+            )
+            response.raise_for_status()
+            obj_accelerator_experiments = workflow_data.get(
+                "accelerator_experiments", []
+            )
+            normalized_accelerator_experiments = response.json()[
```
ioannistsanaktsidis, line 145: since we use 2 times response.json(), lets extract it in a variable before

ioannistsanaktsidis (changes requested):
Just a small improvement, otherwise all good!

ioannistsanaktsidis (approved):

PR #3702, opened 2025-09-05
Package download

DonHaul:


ioannistsanaktsidis:
Something more I forgot to add. I would also add the functionality of https://github.com/inspirehep/inspire-next/blob/be983b0ad4231315469d641dac955a5cefd1394f/inspirehep/modules/workflows/workflows/article.py#L182 as it is kind of related

DonHaul:
> Something more I forgot to add. I would also add the functionality of https://github.com/inspirehep/inspire-next/blob/be983b0ad4231315469d641dac955a5cefd1394f/inspirehep/modules/workflows/workflows/article.py#L182 as it is kind of related

if thats okay we implement it in  another pr

PR #3705, opened 2025-09-09
Arxiv plot extract

DonHaul:


ioannistsanaktsidis (commented):
Looks good, but it is missing some parts https://github.com/inspirehep/inspire-next/blob/master/inspirehep/modules/workflows/tasks/arxiv.py#L193-L204 . Is it because you didn't want to add it as a virtualenv ?

ioannistsanaktsidis (approved):
LGTM, we can have it like this for now, no problem

PR #3708, opened 2025-09-10
workflows: is_arxiv_paper

DonHaul:
*ref: cern-sis/issues-inspire/issues/1067


DonHaul:
in the end it got a bit bigger than expected:
- check_is_arxiv was implemented
- small rework on tests
- added arxiv_processing task group

PR #3713, opened 2025-09-18
workflows: GenericHttpHook.run refactor

DonHaul:
* ref: cern-sis/issues-inspire/issues/1065


DonHaul:
generic .run has to be kept due to 2 functionalities it contains:
- possiblity of changing the `method` with each call (in `HttpHook` this is define at the hook level
- possibility to reuse hook headers ? - maybe alreadt in default function

PR #3717, opened 2025-09-19
ieee harvest

DonHaul:


ioannistsanaktsidis:
Is there a way to disable verification ? It should be possible through connections, otherwise we need to override the hook

DonHaul:
> Is there a way to disable verification ? It should be possible through connections, otherwise we need to override the hook

Currently there's no way to disable it. I have opened an issue in airflow let's see what they say https://github.com/apache/airflow/issues/55885

Meanwhile I will override the hook

ioannistsanaktsidis (changes requested):
small  changes

PR #3722, opened 2025-09-25
workflows: add normalize_journal_titles step

PascalEgn:
ref: https://github.com/cern-sis/issues-inspire/issues/1046

### DIFF
--- a/backend/tests/integration/curation/test_curation_views.py
+++ b/backend/tests/integration/curation/test_curation_views.py
@@ -416,9 +416,9 @@ def test_normalize_journal_titles(inspire_app):
             data=orjson.dumps({"journal_titles_list": ["Technologies"]}),
         )
         assert response.status_code == 200
-        assert response.json == {
-            "normalized_journal_titles": {"Technologies": "Technol."}
-        }
+        assert (
+            response.json["normalized_journal_titles"].get("Technologies") == "Technol."
+        )
 
 
 def test_normalize_journal_titles_no_match(inspire_app):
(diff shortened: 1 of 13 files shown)


workflows/dags/literature/hep_create_dag.py
```diff
@@ -199,6 +199,94 @@ def count_reference_coreness(**context):
...
+                return
+
+            response = inspire_http_hook.call_api(
+                endpoint="api/curation/literature/normalize-journal-titles",
+                method="GET",
+                json={"journal_titles_list": titles_to_normalize},
+            )
+            response.raise_for_status()
+            normalized_titles_map = get_value(
+                response.json(), "normalized_journal_titles", []
```
ioannistsanaktsidis, line 239: since we repeat that, lets assign it in a var before

workflows/dags/literature/hep_create_dag.py
```diff
@@ -199,6 +199,94 @@ def count_reference_coreness(**context):
...
+            )
+            normalized_references_map = get_value(
+                response.json(), "normalized_journal_references", []
+            )
+            normalized_categories_map = get_value(
+                response.json(), "normalized_journal_categories", []
+            )
+
+            def process_entries(entries, title_getter):
```
ioannistsanaktsidis, line 248: lets extract this to a utility func, similar to what we do with guess_coreness.py etc

backend/tests/integration/curation/test_curation_views.py
```diff
@@ -416,9 +416,9 @@ def test_normalize_journal_titles(inspire_app):
             data=orjson.dumps({"journal_titles_list": ["Technologies"]}),
         )
         assert response.status_code == 200
-        assert response.json == {
-            "normalized_journal_titles": {"Technologies": "Technol."}
-        }
+        assert {"Technologies": "Technol."}.items() <= response.json[
```
ioannistsanaktsidis, line 419: not enough to just check directly 
```
assert response.json["normalized_journal_titles"].get("Technologies") == "Technol."
```

easier in my opinion but thats my 2 cents

ioannistsanaktsidis (changes requested):
small comments

PR #4149, opened 2026-04-08
workflows: hard-restart on restart workflow

DonHaul:
* ref: cern-sis/issues-inspire/issues/807 , cern-sis/issues-inspire/issues/1435


backoffice/backoffice/hep/api/views.py
```diff
@@ -229,6 +229,29 @@ def restart(self, request, pk=None):
...
+        if workflow.status == HepStatusChoices.COMPLETED:
+            return Response(
+                {"message": "Cannot restore a completed workflow."},
+                status=status.HTTP_400_BAD_REQUEST,
+            )
+        workflow.decisions.all().delete()
+        workflow.workflow_type = (
+            HepWorkflowType.HEP_CREATE
+            if workflow.workflow_type
+            in (HepWorkflowType.HEP_CREATE, HepWorkflowType.HEP_UPDATE)
```
ioannistsanaktsidis: What about SUBMISSION ?

ioannistsanaktsidis (changes requested):
Mostly the comment about submissions

backoffice/backoffice/hep/api/views.py
DonHaul: good catch

DonHaul:
all comments addressed

ioannistsanaktsidis (approved):
Some tests are failing, otherwise it looks good!

PR #4157, opened 2026-04-10
bump: inspire-dojson on hep

DonHaul:
* ref: cern-sis/issues-inspire/issues/1439


DonHaul:
Vars added in:
- https://github.com/cern-sis/kubernetes-inspire/commit/da9da4d855594e10359a0dd3114415694f119056
- https://github.com/cern-sis/kubernetes-inspire/commit/e1575c6d025a88d8d34de57d22448da9c0d4a0c5

I've realized that the env var `DOMAIN` and `SERVER_NAME` are the same:
https://github.com/cern-sis/kubernetes-inspire/blob/af41600aedb7f4f6ec99e3669677bdcbabd96e9f/next/base/config/inspirehep.cfg#L11
we cannot remove this definition has it's used for hep


Two things we can still do:
- if needed we can submit another release for `inspire-dojson` do use `DOMAIN` instead so we dont have the same var duplicated
- if needed we can rework `backend` to use the envvar instead of through `inspirehep.cfg`

DonHaul:
working in https://github.com/inspirehep/inspirehep/actions/runs/24340644369

ioannistsanaktsidis (changes requested):
pyproject.toml not pushed ? I would put it to 36 just for consistency

DonHaul:
> pyproject.toml not pushed ? I would put it to 36 just for consistency

you're right bumped it there

ioannistsanaktsidis (approved):
LGTM

PR #4169, opened 2026-04-14
workflows: rename s3_elsevier_conn to s3_publisher_conn

DonHaul:
* ref: cern-sis/issues-inspire/issues/1409

### DIFF
--- a/workflows/dags/literature/aps_harvest.py
+++ b/workflows/dags/literature/aps_harvest.py
@@ -37,7 +37,8 @@ logger = logging.getLogger(__name__)
 def aps_harvest_dag():
     """Harvest APS articles through the APS API and create HEP workflows."""
 
-    s3_store = S3JsonStore("s3_aps_conn")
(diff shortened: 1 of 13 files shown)


DonHaul:
might also make sense to "merge" `s3_aps_conn` into it, as its the exact same thing

workflows/dags/literature/aps_harvest.py
```diff
@@ -138,11 +139,14 @@ def process_articles(s3_harvest_key, **context):
                 logger.warning("Failed to process APS article %s: %s", doi, exc)
                 failed_records.append({"doi": doi, "error": str(exc)})
 
-        return s3_store.write_object({"failed_records": failed_records})
+        res = s3_store.write_object({"failed_records": failed_records})
+        return res
```
ioannistsanaktsidis: why not return directly ?

ioannistsanaktsidis (approved):
Just the comment

PR #4171, opened 2026-04-14
workflows: is_connection_url detects aws hosts

DonHaul:
* ref: cern-sis/issues-inspire/issues/1441

### DIFF
--- a/workflows/tests/conftest.py
+++ b/workflows/tests/conftest.py
@@ -75,7 +75,7 @@ def s3_desy_env(request):
 @pytest.fixture(scope="class")
 def hep_env(request):
     test_dict = {
-        "AIRFLOW_CONN_S3_CONN": "aws://airflow:airflow-inspire@s3:9000/"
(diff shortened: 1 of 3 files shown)


workflows/tests/conftest.py
```diff
@@ -75,7 +75,7 @@ def s3_desy_env(request):
 @pytest.fixture(scope="class")
 def hep_env(request):
     test_dict = {
-        "AIRFLOW_CONN_S3_CONN": "aws://airflow:airflow-inspire@s3:9000/"
+        "AIRFLOW_CONN_S3_CONN": "aws://airflow:airflow-inspire@/"
```
ioannistsanaktsidis, line 78: is the @ in the end needed ?

ioannistsanaktsidis (commented):
just a clarification

workflows/tests/conftest.py
DonHaul: yes, it separates `user:pw` from `host:port` which dont exist in this case

ioannistsanaktsidis (approved):
LGTM

PR #4173, opened 2026-04-15
workflows: multiple exact matches

ioannistsanaktsidis:
* ref https://github.com/cern-sis/issues-inspire/issues/1453

Search results
Screenshot 2026-04-15 at 14 57 05

Detail
Screenshot 2026-04-15 at 14 57 10

### DIFF
--- a/workflows/dags/literature/hep_create_dag.py
+++ b/workflows/dags/literature/hep_create_dag.py
@@ -77,6 +77,7 @@ from include.utils.constants import (
     STATUS_APPROVAL_MERGE,
     STATUS_BLOCKED,
     STATUS_COMPLETED,
+    STATUS_ERROR_MULTIPLE_EXACT_MATCHES,
     STATUS_MISSING_SUBJECT_FIELDS,
     STATUS_RUNNING,
     TICKET_HEP_CURATION_CORE,
@@ -248,7 +249,7 @@ def hep_create_dag():
         )
         return True
 
-    @task.branch
+    @task.branch(on_failure_callback=None)
     def check_for_exact_matches(**context):
         workflow_id = context["params"]["workflow_id"]
         workflow_data = s3_store.read_workflow(workflow_id)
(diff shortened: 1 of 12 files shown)


workflows/dags/literature/hep_create_dag.py
```diff
@@ -262,6 +263,17 @@ def check_for_exact_matches(**context):
         matches = response.json()["matched_ids"]
 
         if len(matches) >= 2:
+            workflow_data["status"] = STATUS_ERROR_MULTIPLE_EXACT_MATCHES
+            workflow_data["matches"] = get_value(workflow_data, "matches", {}) or {}
+            workflow_data["matches"]["exact"] = matches
```
DonHaul, line 268: does it make sense to set this when there is only 1 match as well?
the line `workflow_data["matches"]["exact"] = matches`

workflows/dags/literature/hep_create_dag.py
```diff
@@ -248,7 +249,7 @@ def check_for_blocking_workflows(**context):
         )
         return True
 
-    @task.branch
+    @task.branch(on_failure_callback=None)
```
DonHaul, line 252: I would say we keep don't remove the callback, we make the test go to skip with `raise AirflowSkipException` instead of `raise AirflowException` . this way we still track erros

DonHaul (commented):
A few comments

workflows/dags/literature/hep_create_dag.py
ioannistsanaktsidis: Yes it makes sense for debugging purposes. Will put it!

workflows/dags/literature/hep_create_dag.py
ioannistsanaktsidis: From what I saw AirflowException works better for our case  at this moment. We can think of the branching out solution but sth to be considered in the future.

ioannistsanaktsidis:
@DonHaul comments addressed!

DonHaul (approved):
LGTM

PR #4176, opened 2026-04-16
Refactor tests - part 2

DonHaul:


DonHaul:
From 11mins to 7mins

workflows/tests/test_cds_literature_harvest.py
```diff
@@ -1,25 +1,33 @@
...
-@pytest.mark.usefixtures("_s3_store")
+
+@pytest.mark.usefixtures("hep_env")
 class TestCDSLiteratureHarvest:
+    dag = dagbag.get_dag("cds_literature_harvest_dag")
+
     @patch(
         "hooks.backoffice.workflow_management_hook.WorkflowManagementHook.post_workflow"
     )
     def test_process_records(self, mock_post_workflow):
```
ioannistsanaktsidis, line 19: there is a cassette for this test. Dont forget to delete it + whatever else needs deletion

ioannistsanaktsidis (changes requested):
Just the validation with the cassettes that need deletion

workflows/tests/test_cds_literature_harvest.py
DonHaul: test was missing `@pytest.mark.vcr` it was accessing cds live. fixed.
Some cassettes have been renamed / slightly modified, otherwise I dont think nothing needs to be deleted

ioannistsanaktsidis (approved):
LGTM 🥇

PR #4177, opened 2026-04-16
workflows: cds single harvest by identifier

ioannistsanaktsidis:
* ref  https://github.com/cern-sis/issues-inspire/issues/1419

### DIFF
--- a/workflows/tests/test_cds_literature_single_harvest.py
+++ b/workflows/tests/test_cds_literature_single_harvest.py
@@ -1,0 +1,34 @@
+from unittest.mock import patch
+
+import pytest
+from airflow.models import DagBag
+
+from tests.test_utils import task_test2
+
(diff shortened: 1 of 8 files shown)


workflows/tests/test_cds_literature_single_harvest.py
```diff
@@ -0,0 +1,34 @@
...
+
+import pytest
+from airflow.models import DagBag
+
+from tests.test_utils import task_test2
+
+dagbag = DagBag()
+
+
+@pytest.mark.usefixtures("_s3_store", "hep_env")
```
DonHaul: we can ommit `_s3_store` as for simplicity its also added inside `hep_env`

DonHaul (approved):
Just the comment otherwise its good

PR #4185, opened 2026-04-20
backoffice(ui): withdrawn entry

ioannistsanaktsidis:


Makefile
```diff
@@ -47,6 +47,7 @@ django-setup:
 	docker compose exec backoffice-webserver python manage.py loaddata backoffice/authors/fixtures/workflows.json
 	docker compose exec backoffice-webserver python manage.py loaddata backoffice/hep/fixtures/workflows.json
 	docker compose exec backoffice-webserver python manage.py loaddata backoffice/hep/fixtures/decisions.json
+	docker compose exec backoffice-webserver python manage.py opensearch document index --force
```
DonHaul, line 50: is this required?

ioannistsanaktsidis: Yes, workflows are indexed in OpenSearch when workflows.json is loaded, before decisions exist, so the index has empty decisions for every workflow. This was not working locally. The re-index command runs after decisions.json is loaded to bring the OpenSearch documents in sync with the DB. In a real scenario this isn't needed because resolving a workflow always updates the workflow's status, which triggers a re-index of the workflow document, at that point the new decision is included automatically.

DonHaul (approved):

PR #4186, opened 2026-04-21
backend: remove self-curation ticket creation logic and associated te…

ioannistsanaktsidis:
…mplate

 * ref https://github.com/cern-sis/issues-inspire/issues/1459


DonHaul (approved):
Nice

PR #4187, opened 2026-04-21
backend: process airflow files correctly

DonHaul:
* ref: cern-sis/issues-inspire/issues/1420


backend/inspirehep/files/api/s3.py
```diff
@@ -53,6 +53,15 @@ def is_s3_url(url):
...
 
+    @staticmethod
+    def is_s3_url_from_airflow_bucket(url):
+        """Checks if the url is an S3 url and if contains `S3_AIRFLOW_BUCKET`.
+
+        :param url: the given url.
+        :return: boolean
+        """
+        return S3.is_s3_url(url) and current_app.config.get("S3_AIRFLOW_BUCKET") in url
+
```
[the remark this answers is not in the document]
ioannistsanaktsidis: this will need a default setup in the config or a fallback to False, otherwise `None in url` will raise a TypeError (mostly local problem)

backend/inspirehep/records/api/literature.py
```diff
@@ -455,7 +457,13 @@ def add_file(
                 )
                 return result
 
-            data = download_file_from_url(url, check_file_size=True)
+            if current_s3_instance.is_s3_url_from_airflow_bucket(url):
+                bucket = current_app.config.get("S3_AIRFLOW_BUCKET")
+                key = current_s3_instance.get_full_key_from_url(url)
+                response = current_s3_instance.client.get_object(Bucket=bucket, Key=key)
+                data = response["Body"].read()
```
ioannistsanaktsidis: This part here, does not check against the size limit  we have in the download url . I think this is important to check as we have a limit on the size we allow -> https://github.com/inspirehep/inspirehep/blob/master/backend/inspirehep/records/utils.py#L82-L86

backend/inspirehep/records/api/literature.py
```diff
@@ -433,9 +433,11 @@ def add_file(
...
         with app_context.app.app_context():
             result = {}
-            is_s3_or_public_url = current_s3_instance.is_s3_url_with_bucket_prefix(
-                url
-            ) or current_s3_instance.is_public_url(url)
+            is_s3_or_public_url = (
+                current_s3_instance.is_s3_url_with_bucket_prefix(url)
+                or current_s3_instance.is_public_url(url)
+            ) and not current_s3_instance.is_s3_url_from_airflow_bucket(url)
+
```
ioannistsanaktsidis, line 447: this is a bit hard to read, lets make it a bit nicer. For example , lets extract

```
 is_airflow_url = current_s3_instance.is_s3_url_from_airflow_bucket(url)
        is_s3_url = current_s3_instance.is_s3_url(url)
        is_public_url = current_s3_instance.is_public_url(url)
        is_s3_or_public_url = (current_s3_instance.is_s3_url_with_bucket_prefix(url) or is_public_url) and not is_airflow_url

```

and then reuse them accordingly

backend/inspirehep/records/api/literature.py
```diff
@@ -455,7 +457,13 @@ def add_file(
                 )
                 return result
 
-            data = download_file_from_url(url, check_file_size=True)
+            if current_s3_instance.is_s3_url_from_airflow_bucket(url):
+                bucket = current_app.config.get("S3_AIRFLOW_BUCKET")
+                key = current_s3_instance.get_full_key_from_url(url)
+                response = current_s3_instance.client.get_object(Bucket=bucket, Key=key)
```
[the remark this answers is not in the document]
ioannistsanaktsidis: This is actually valid , goes with my next comment

backend/inspirehep/records/api/literature.py
```diff
@@ -511,8 +519,8 @@ def add_file(
                 and not current_s3_instance.is_s3_url(url)
                 and not current_s3_instance.is_public_url(url)
                 and not original_url
-            ):
-                result["original_url"] = url
+            ) or current_s3_instance.is_s3_url_from_airflow_bucket(url):
+                result["original_url"] = original_url or url
```
ioannistsanaktsidis: that is a bit hard to understand, we can have it a bit nicer. sth like  this maybe 

```
is_http = url.startswith("http")
is_s3 = current_s3_instance.is_s3_url(url)
is_public = current_s3_instance.is_public_url(url)
is_airflow = current_s3_instance.is_s3_url_from_airflow_bucket(url)

should_set_original = (
    (is_http and not is_s3 and not is_public and not original_url)
    or is_airflow
)
```

also, for the airflow case this will always be set to the s3 part. Is that intended ?

ioannistsanaktsidis (changes requested):
Let's start with these comments first

backend/inspirehep/records/api/literature.py
DonHaul: @drjova wdyt about this.

a few questions he have:
- do we even need to set `original_url` here in a few of the cases we've seen its population in `article.py`
- if it's an airflow link say coming from publishers should we set this url as the `original_url`

DonHaul:
updated testing

backend/inspirehep/files/api/s3.py
```diff
@@ -231,6 +241,15 @@ def get_key_from_url(self, url):
             LOGGER.exception("S3 incorrect url.", url=url)
             raise
 
+    @staticmethod
+    def get_bucket_and_key_from_url(url):
+        try:
+            path_parts = urlsplit(url).path.split("/")
+            return path_parts[1], "/".join(path_parts[2:])
```
drjova, line 248: what about IndexError?

DonHaul: added

ioannistsanaktsidis (approved):
All good from my side. This will also need the env var on kubernetes-inspire so as it works on qa and eventually prod

drjova (approved):

PR #4195, opened 2026-04-24
workflows: revert withdrawn and fail is pdf not found

DonHaul:
* ref: cern-sis/issues-inspire/issues/1464


DonHaul:
the DECISION `withdrawn` still exists in the `backoffice`, to be deleted by the end on next week if we find that we cannot differentiate cases

ioannistsanaktsidis (approved):
LGTM! We will monitor usecases and act accordingly as you proposed!

PR #4197, opened 2026-04-24
workflows: save before await_decision_approval

DonHaul:
* ref: cern-sis/issues-inspire/issues/1472

### DIFF
--- a/workflows/dags/literature/hep_create_dag.py
+++ b/workflows/dags/literature/hep_create_dag.py
@@ -1411,7 +1411,7 @@ def hep_create_dag():
             ):
                 return (
                     "halt_for_approval_if_new_or_reject_if_not_relevant."
-                    "await_decision_approval"
+                    "save_workflow_before_approval"
                 )
 
             if is_auto_rejected(workflow_data):
@@ -1425,7 +1425,7 @@ def hep_create_dag():
 
             return (
                 "halt_for_approval_if_new_or_reject_if_not_relevant."
-                "await_decision_approval"
+                "save_workflow_before_approval"
             )
 
         @task.branch(trigger_rule=TriggerRule.NONE_FAILED_MIN_ONE_SUCCESS)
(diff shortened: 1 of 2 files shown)


DonHaul:
blocked by 
https://github.com/cern-sis/issues-inspire/issues/1469

workflows/dags/literature/hep_create_dag.py
```diff
@@ -1411,7 +1411,7 @@ def is_record_relevant(**context):
             ):
                 return (
                     "halt_for_approval_if_new_or_reject_if_not_relevant."
-                    "await_decision_approval"
+                    "save_workflow__1"
```
ioannistsanaktsidis: nice one, is this the new way of pointing to the task step so as it doesn't jump to a later save_workflow ?

DonHaul: its how airflow automatically names the task if it find identical ones. Its not new but the first time it appears in our wf
image

ioannistsanaktsidis: Another proposal would be to attach an id to this particular save_workflow, sth like `save_workflow_before_approval`, so as to differentiate and distinguish much better. WDYT ?

workflows/dags/literature/hep_create_dag.py
```diff
@@ -1411,7 +1411,7 @@ def is_record_relevant(**context):
...
                 )
```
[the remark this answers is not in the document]
ioannistsanaktsidis: cc @DonHaul

workflows/dags/literature/hep_create_dag.py
DonHaul: Sounds good, Just fixed it. should be okay now

DonHaul: Sounds good, Just fixed it. should be okay now

ioannistsanaktsidis (approved):
LGTM

PR #4202, opened 2026-04-28
backoffice(ui): fix filterDecisions function to exclude hidden decisi…

ioannistsanaktsidis:
…on actions

 * ref https://github.com/cern-sis/issues-inspire/issues/1470

### DIFF
--- a/ui/src/common/constants.ts
+++ b/ui/src/common/constants.ts
@@ -105,6 +105,15 @@ export enum WorkflowDecisions {
   HEP_ACCEPT = 'hep_accept',
   HEP_ACCEPT_CORE = 'hep_accept_core',
   HEP_REJECT = 'hep_reject',
+  MERGE_APPROVE = 'merge_approve',
+  MISSING_SUBJECT_FIELDS = 'missing_subject_fields',
   REJECT = 'reject',
   WITHDRAWN = 'withdrawn',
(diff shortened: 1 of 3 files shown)


ui/src/common/constants.ts
```diff
@@ -105,6 +105,15 @@ export enum WorkflowDecisions {
...
   HEP_REJECT = 'hep_reject',
+  MERGE_APPROVE = 'merge_approve',
+  MISSING_SUBJECT_FIELDS = 'missing_subject_fields',
   REJECT = 'reject',
   WITHDRAWN = 'withdrawn',
 }
+
+export const HIDDEN_DECISION_ACTIONS = new Set([
+  WorkflowDecisions.FUZZY_MATCH,
+  WorkflowDecisions.EXACT_MATCH,
```
DonHaul, line 116: I am wondering if there would be any usecase, where it would be useful to see in the UI what has this record been matched with

DonHaul (approved):
Just the comment

ui/src/common/constants.ts
ioannistsanaktsidis: That we can already do by the matches part (we save the exact match id, so we can have the link displayed already), no need to to have the automatic exact match decision displayed.

PR #4204, opened 2026-04-29
workflows: non atomic batch resolve

DonHaul:
* ref: cern-sis/issues-inspire/issues/1467

### DIFF
--- a/backoffice/backoffice/hep/api/views.py
+++ b/backoffice/backoffice/hep/api/views.py
@@ -3,6 +3,8 @@ import logging
 from rest_framework import status, viewsets
 from rest_framework.response import Response
 from django.shortcuts import get_object_or_404
+from django.utils.decorators import method_decorator
+from django.db import transaction
 from backoffice.hep.utils import (
     add_hep_decision,
     resolve_workflow,
@@ -181,27 +183,6 @@ class HepWorkflowViewSet(BaseWorkflowViewSet):
         workflow.save()
         return Response(status=status.HTTP_200_OK)
 
-    @action(
-        detail=False,
-        methods=["post"],
-        url_path="resolve",
-    )
-    def batch_resolve(self, request):
-        serializer = HepBatchResolutionSerializer(data=request.data)
-        serializer.is_valid(raise_exception=True)
-
-        results = []
-        for id in serializer.validated_data["ids"]:
-            try:
-                resolve_workflow(id, serializer.validated_data, request.user)
-                results.append({"id": id, "success": True})
-
-            except Exception as e:
-                logger.exception("Error resolving workflow %s", id)
-                results.append({"id": id, "success": False, "error": str(e)})
-
-        return Response({"results": results}, status=status.HTTP_200_OK)
-
     @action(detail=True, methods=["post"])
     def restart(self, request, pk=None):
         workflow = get_object_or_404(HepWorkflow, pk=pk)
@@ -250,6 +231,32 @@ class HepWorkflowViewSet(BaseWorkflowViewSet):
         workflow.save()
 
         return Response(self.get_serializer(workflow).data)
+
+
+@method_decorator(transaction.non_atomic_requests, name="dispatch")
+class HepWorkflowBatchResolveViewSet(viewsets.ViewSet):
+    def create(self, request):
+        serializer = HepBatchResolutionSerializer(data=request.data)
+        serializer.is_valid(raise_exception=True)
+
+        results = []
+        for wf_id in serializer.validated_data["ids"]:
+            try:
+                with transaction.atomic():
+                    resolve_workflow(wf_id, serializer.validated_data, request.user)
+                results.append({"id": wf_id, "success": True})
+            except Exception as e:
+                logger.exception("Error resolving workflow %s", wf_id)
+                results.append({"id": wf_id, "success": False, "error": str(e)})
+
+        has_failures = any(not result["success"] for result in results)
+        response_status = (
+            status.HTTP_500_INTERNAL_SERVER_ERROR
+            if has_failures
+            else status.HTTP_200_OK
+        )
+
+        return Response({"results": results}, status=response_status)
 
 
 @extend_schema_view(
--- a/backoffice/backoffice/hep/utils.py
+++ b/backoffice/backoffice/hep/utils.py
@@ -2,7 +2,6 @@ from backoffice.hep.api.serializers import (
     HepChangeStatusSerializer,
     HepDecisionSerializer,
 )
-from django.db import transaction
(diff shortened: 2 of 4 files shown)


DonHaul:
in the end the `@transaction.non_atomic_requests` as it would have to applied to the view as a whole. an alternative is to move `batch_resolve` to another view

backoffice/backoffice/hep/api/views.py
```diff
@@ -114,6 +115,13 @@ class HepWorkflowViewSet(BaseWorkflowViewSet):
     status_choices = HepStatusChoices
     schema_name = "hep"
 
+    @classmethod
+    def as_view(cls, actions=None, **initkwargs):
+        view = super().as_view(actions=actions, **initkwargs)
+        if actions == {"post": "batch_resolve"}:
+            return transaction.non_atomic_requests(view)
+        return view
+
```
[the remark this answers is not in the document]
DonHaul: using this would require separating it in a a separate view. which may not be the best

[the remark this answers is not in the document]
DonHaul: alternatively we can use `transaction.on_commit` and if the request fails we set status to errored. (decision would stay there)

backoffice/backoffice/hep/api/views.py
```diff
@@ -252,6 +234,24 @@ def restore(self, request, pk=None):
...
+@method_decorator(transaction.non_atomic_requests, name="dispatch")
+class HepWorkflowBatchResolveView(APIView):
+    def post(self, request):
+        serializer = HepBatchResolutionSerializer(data=request.data)
+        serializer.is_valid(raise_exception=True)
+
+        results = []
+        for id in serializer.validated_data["ids"]:
+            try:
+                resolve_workflow(id, serializer.validated_data, request.user)
```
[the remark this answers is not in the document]
DonHaul: makes sense to add so it rolls back decisions as the single resolve does

backoffice/backoffice/hep/api/views.py
```diff
@@ -252,6 +234,24 @@ def restore(self, request, pk=None):
...
+                results.append({"id": id, "success": True})
+            except Exception as e:
+                logger.exception("Error resolving workflow %s", id)
+                results.append({"id": id, "success": False, "error": str(e)})
+
+        return Response({"results": results}, status=status.HTTP_200_OK)
+
+
```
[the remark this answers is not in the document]
ioannistsanaktsidis: That would actually make sense no ?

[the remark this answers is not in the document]
DonHaul: yes, I will change it, is `400 BAD REQUEST` okay or is there a better status code for this usecase?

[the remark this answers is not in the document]
ioannistsanaktsidis: I would go for 500 , 400 means something the user sent is wrong. This is not the case as an error can occur if an upstream fails. No user involvement

backoffice/backoffice/hep/api/urls.py
```diff
@@ -0,0 +1,12 @@
+from django.urls import path
+
+from backoffice.hep.api.views import HepWorkflowBatchResolveView
+
+
+urlpatterns = [
```
drjova: shouldn't we register it here https://github.com/inspirehep/inspirehep/blob/307aa43b9d2389ca8996f96a945d7271e14d21d2/backoffice/config/api_router.py#L8 ?

drjova (changes requested):
Thanks @DonHaul! Just a comment

backoffice/backoffice/hep/api/views.py
```diff
@@ -252,6 +234,32 @@ def restore(self, request, pk=None):
...
 
 
+@method_decorator(transaction.non_atomic_requests, name="dispatch")
+class HepWorkflowBatchResolveView(APIView):
+    def post(self, request):
+        serializer = HepBatchResolutionSerializer(data=request.data)
+        serializer.is_valid(raise_exception=True)
+
+        results = []
+        for id in serializer.validated_data["ids"]:
```
ioannistsanaktsidis: this shadows default python id, lets rename it for consistency to _id or identifier

backoffice/backoffice/hep/api/views.py
```diff
@@ -252,6 +234,32 @@ def restore(self, request, pk=None):
         return Response(self.get_serializer(workflow).data)
 
 
+@method_decorator(transaction.non_atomic_requests, name="dispatch")
+class HepWorkflowBatchResolveView(APIView):
```
ioannistsanaktsidis: We now inherit against APIView. This will inherit global defaults from django. To verify if permissions etc apply the same style as before

backoffice/backoffice/hep/utils.py
```diff
@@ -55,14 +54,14 @@ def resolve_workflow(id, data, user):
...
-        )
-        logger.info("Decision has been added to the workflow %s", id)
+
+    add_hep_decision(
+        id,
+        user,
+        data["action"],
+        data.get("value"),
+    )
+    logger.info("Decision has been added to the workflow %s", id)
```
ioannistsanaktsidis, line 64: We need to check the other callers (single resolve) to verify the behavior. We can do it in another issue though.

ioannistsanaktsidis (changes requested):
Some comments to validate

backoffice/backoffice/hep/api/urls.py
DonHaul: we can't register it as its a `View` and not a `ViewSet`, alternatively we can more add it directly to the `urlpatterns` on  `api_router.py` although it would make the code less modular in my opinion.

Is that what you mean?

drjova: ok I see, thanks. Yes we can keep it here

backoffice/backoffice/hep/api/views.py
DonHaul: `BaseWorkflowViewSet` also doesn't have any checks for `permissions` etc, as long as the user has a token he can use the new api endpoint

backoffice/backoffice/hep/utils.py
DonHaul: Before it was simply keeping the decision there in case the call to airflow would fail.
Removing the transaction will make the decision not be there the when airflow call fails.
Both behaviours are valid

ioannistsanaktsidis (approved):
LGTM

backoffice/backoffice/hep/api/urls.py
DonHaul: @drjova can you approve?

drjova (approved):

PR #4222, opened 2026-05-08
workflows: manual merge

DonHaul:
* ref: cern-sis/issues-inspire/issues/1451


DonHaul:
requires manual and cicd testing

DonHaul:
there are some functionality ie. `save_and_complete_workflow` and `store_records` that is shared with `hep_create_dags`, to be consolidated into one in another PR

DonHaul:
no way to check if merge of these ids is already being processed since we don't store head id nor update id in the db

workflows/dags/literature/hep_manual_merge.py
```diff
@@ -0,0 +1,209 @@
...
+            ) < parser.parse(update_root["updated"]):
+                workflows.add_wf_record_source(
+                    record_uuid=head_uuid,
+                    source=update_root["source"],
+                    json=update_root["json"],
+                )
+            workflows.delete_wf_record_source(
+                record_uuid=update_uuid,
+                source=update_root["source"],
+            )
```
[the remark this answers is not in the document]
DonHaul: its the intended behaviour

ioannistsanaktsidis:
> there are some functionality ie. `save_and_complete_workflow` and `store_records` that is shared with `hep_create_dags`, to be consolidated into one in another PR

@DonHaul can we create a ticket for that ? Thanks!

backoffice/backoffice/hep/api/views.py
```diff
@@ -232,6 +233,38 @@ def restore(self, request, pk=None):
 
         return Response(self.get_serializer(workflow).data)
 
+    @action(detail=False, methods=["post"])
+    def manualmerge(self, request):
```
ioannistsanaktsidis: lets follow the pythonic way -> `manual_merge`

workflows/tests/test_workflow_utils.py
```diff
@@ -52,9 +52,55 @@ def test_read_wf_record_source_not_found(self):
...
+            record_uuid=record_uuid,
+            source="publisher",
+            json=record_json,
+        )
+        assert (
+            workflows.read_wf_record_source(record_uuid, "publisher")["json"]
+            == record_json
+        )
+
+        workflows.delete_wf_record_source(record_uuid=record_uuid, source="test_source")
```
ioannistsanaktsidis: test_source here is a bit irrelevant correct ? since it will always fallback to publisher no ? 

```
def get_source_for_root(source):
    """Source for the root workflow object.
    Args:
        source(str): the record source.
    Return:
        (str): the source for the root workflow object.
    Note:
        For the time being any workflow with ``acquisition_source.source``
        different than ``arxiv`` and ``submitter`` will be stored as
        ``publisher``.
    """
    return source if source in ["arxiv", "submitter"] else "publisher"


```

backoffice/backoffice/hep/api/views.py
```diff
@@ -232,6 +233,38 @@ def restore(self, request, pk=None):
...
+                "Triggering DAG %s with id  %s",
+                WORKFLOW_DAGS[workflow.workflow_type].initialize,
+                str(workflow.id),
+            )
+            airflow_utils.trigger_airflow_dag(
+                WORKFLOW_DAGS[workflow.workflow_type].initialize,
+                str(workflow.id),
+                **serializer.data,
+            )
+        except RequestException as e:
```
ioannistsanaktsidis, line 259: Will this fallback properly ? meaning if ariflow fails, will we have an orphan one in bakoffice ?

workflows/plugins/include/utils/workflows.py
```diff
@@ -209,25 +209,78 @@ def read_wf_record_source(record_uuid, source):
...
+        if "404:NOT FOUND" in str(e):
+            logger.warning(
+                f"Record source with uuid {record_uuid} and source {source} not found"
+                f" for deletion."
+            )
+            return
+        raise
+
+
+def add_wf_record_source(record_uuid, source, json):
```
ioannistsanaktsidis, line 275: lets also add a docistring here , liken the rest of the methods

ioannistsanaktsidis (changes requested):
few comments

workflows/tests/test_workflow_utils.py
DonHaul: indeed it made no sense

backoffice/backoffice/hep/api/views.py
DonHaul: it should leave an orphan one in backoffice yes. its the same pattern we have in other places we trigger Dags

ioannistsanaktsidis (approved):
Just the cassete recording update and we are good to go

backoffice/backoffice/hep/api/views.py
ioannistsanaktsidis: I see! We can leave i like that for now and see  how to tackle it in the future. (Ideally it shouldn't happen often)

PR #4223, opened 2026-05-08
backoffice(e2e): add backoffice e2e tests in ci

EliseRecejac:
- ref: https://github.com/cern-sis/issues-inspire/issues/1474

### DIFF
--- a/.github/workflows/e2e-tests.yml
+++ b/.github/workflows/e2e-tests.yml
@@ -19,6 +19,10 @@ on:
         description: The UI image
         type: string
         required: true
+      backoffice-image:
+        description: The backoffice image
+        type: string
+        required: true
 
 defaults:
   run:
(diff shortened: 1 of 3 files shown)


EliseRecejac:
For the moment all e2e tests are running for inspirehep PR as well as backoffice ones. Maybe we should only run the relevant ones

.github/workflows/e2e-tests.yml
```diff
@@ -93,6 +106,26 @@ jobs:
           APP_ENABLE_SECURE_HEADERS=False
           EOF
 
+      - name: create backoffice environment file
+        run: |
```
drjova, line 110: could we pass them as env for the job?

EliseRecejac: I think that if I do that I would have to call each of them each time I have `--env-file ./backoffice-env.list`, so this solution may be less verbose (it's also what Ben have done on the step above). But honestly it's not the part I'm the more comfortable with so we can totally discuss it if I miss something

drjova (approved):

.github/workflows/e2e-tests.yml
```diff
@@ -142,6 +175,36 @@ jobs:
...
+          --network host
+          --env-file ./backoffice-env.list
+          --name backoffice-worker
+          --entrypoint celery
+          ${{ inputs.backoffice-image }}
+          -A
+          config.celery_app
+          worker
+          -l
+          INFO
```
drjova, line 207: @EliseRecejac I think this is redundant, and it’s causing the error

PR #4228, opened 2026-05-11
backoffice(e2e): set up backoffice e2e tests and add tests to Authors collection

EliseRecejac:
- [ref: cern-sis/issues-inspire#1474](https://github.com/inspirehep/inspirehep/pull/4221)


e2e/cypress/e2e/backoffice/authors.test.js
```diff
@@ -0,0 +1,43 @@
+describe("New authors", () => {
+  it("should display new authors and enable access to the all authors list", () => {
+    cy.loginAsAdmin();
+    cy.loginToBackoffice();
```
ioannistsanaktsidis: instead of login in each test, we could simply do sth like

```
beforeEach(() => {
    cy.loginAsAdmin();
    cy.loginToBackoffice();
  });
```

ui/src/common/components/SearchResults.tsx
```diff
@@ -21,7 +21,11 @@ const SearchResults = ({
   const renderResults = (result: any, index: number) => {
     if (!isBackoffice) {
       return (
-        <div className="mv2" key={result.get('id')}>
+        <div
+          className="mv2"
+          key={result.get('id')}
+          data-testid="result-from-search"
```
ioannistsanaktsidis, line 27: We can leave that as is right now, but this made me realize we are calling this component in backoffice with `isBackoffice={false}`. Could you please create a ticket for validating this behavior ?

ui/public/config.js
```diff
@@ -38,5 +38,15 @@
...
 
+  const runtimeOverride =
+    typeof window !== 'undefined' && window.__RUNTIME_CONFIG__
+      ? window.__RUNTIME_CONFIG__
+      : {};
+
+  const CONFIG = {
+    ...DEFAULT_CONFIG,
+    ...runtimeOverride,
+  };
```
ioannistsanaktsidis: I would prefer reverting this part if it is only needed for local development as it is touching production code for only test related parts.

ioannistsanaktsidis (changes requested):

ui/public/config.js
EliseRecejac: my bad I was a bit confused, it was actually needed to avoid localhost to be called instead of host (the same way we do with the command `--env inspirehep_url=http://host.docker.internal:8080,backoffice_url=http://host.docker.internal:8001 ` in the makefile for the cypress config) but I managed to do something more restricted to the tests, let me know what you think of that :)

ui/src/common/components/SearchResults.tsx
EliseRecejac: yes sure

EliseRecejac: https://github.com/cern-sis/issues-inspire/issues/1495

drjova (approved):
LGTM

PR #4233, opened 2026-05-13
fix(ui): exact alignment between abstracts

EliseRecejac:
* ref: https://github.com/cern-sis/issues-inspire/issues/1492


EliseRecejac:
Capture d’écran 2026-05-13 à 10 24 42

michamos:
Isn’t it still wider? If you change the text to be the same in both cases, all line breaks should be in the same place

EliseRecejac:
> Isn’t it still wider? If you change the text to be the same in both cases, all line breaks should be in the same place

good catch it was 2px wider (because of the borders)

EliseRecejac:
Capture d’écran 2026-05-13 à 11 20 34
@michamos should be better now

drjova (approved):

PR #4236, opened 2026-05-15
backoffice(ui): increase density display of metadata

EliseRecejac:
- ref: https://github.com/cern-sis/issues-inspire/issues/1493

### DIFF
--- a/ui/src/backoffice/common/components/ResultItem/ResultItem.less
+++ b/ui/src/backoffice/common/components/ResultItem/ResultItem.less
@@ -6,6 +6,10 @@
     }
   }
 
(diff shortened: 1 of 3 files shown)


EliseRecejac:
Capture d’écran 2026-05-15 à 17 48 44

ioannistsanaktsidis (approved):
Just to validate if the comments from copilot make sense, otherwise looks good

EliseRecejac:
> Just to validate if the comments from copilot make sense, otherwise looks good

it doesn't make sense, we don't have th (I'm not even sure if it would be possible to have some and even if it was I'm not sure we would want to have the same design)

PR #4238, opened 2026-05-18
backoffice(ui): fix issue with timezone

EliseRecejac:
* ref: https://github.com/cern-sis/issues-inspire/issues/1487

### DIFF
--- a/ui/src/backoffice/utils/__tests__/utils.test.tsx
+++ b/ui/src/backoffice/utils/__tests__/utils.test.tsx
@@ -10,10 +10,14 @@ import {
 } from '../utils';
 import storage from '../../../common/storage';
 import { BACKOFFICE_LOGIN_API } from '../../../common/routes';
-import { WorkflowStatuses, WorkflowTypes } from '../../constants';
+import { WorkflowTypes } from '../../constants';
 import { WorkflowDecisions } from '../../../common/constants';
 
 jest.mock('../../../common/storage');
+jest.mock('../../../common/constants', () => ({
+  ...jest.requireActual('../../../common/constants'),
+  LOCAL_TIMEZONE: 'UTC',
+}));
 
 describe('refreshToken', () => {
   beforeEach(() => {
@@ -87,17 +91,24 @@ describe('filterByProperty', () => {
 
 describe('formatDateTime', () => {
   it('should return an object with formatted date and time for a valid date', () => {
-    const rawDateTime = '2024-02-11T15:30:00Z';
+    const rawDateTime = '2024-02-11T15:30:00.134496+00:00';
     const result = formatDateTime(rawDateTime);
 
     expect(result).toEqual({
       date: '2024-02-11',
-      time: '15:30',
+      time: '03:30 PM',
     });
   });
 
   it('should return undefined for an invalid date', () => {
     const rawDateTime = 'invalid-date-string';
+    const result = formatDateTime(rawDateTime);
+
+    expect(result).toBeUndefined();
+  });
+
+  it('should return undefined for an undefined date', () => {
+    const rawDateTime = undefined;
     const result = formatDateTime(rawDateTime);
 
     expect(result).toBeUndefined();
--- a/ui/src/backoffice/utils/utils.tsx
+++ b/ui/src/backoffice/utils/utils.tsx
@@ -10,6 +10,8 @@ import {
 } from '@ant-design/icons';
 import { Action, ActionCreator } from 'redux';
 
+import moment from 'moment-timezone';
+
 import storage from '../../common/storage';
 import { BACKOFFICE_LOGIN, BACKOFFICE_LOGIN_API } from '../../common/routes';
 import { searchQueryUpdate } from '../../actions/search';
@@ -21,6 +23,7 @@ import {
 import {
   WorkflowDecisions,
   HIDDEN_DECISION_ACTIONS,
+  LOCAL_TIMEZONE,
 } from '../../common/constants';
 
 export const COLLECTIONS = [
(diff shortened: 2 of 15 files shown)


backoffice/config/settings/base.py
```diff
@@ -28,7 +28,7 @@
 # http://en.wikipedia.org/wiki/List_of_tz_zones_by_name
 # though not all of them may be available with every OS.
 # In Windows, this must be set to your system time zone.
-TIME_ZONE = "UTC"
+TIME_ZONE = env("TIME_ZONE", default="Europe/Zurich")
```
drjova: I would leave the time zone on the app always in UTC and change it on the browser depending on the user's settings

workflows/plugins/include/utils/arxiv.py
```diff
@@ -22,7 +22,7 @@ def build_records(xml_records, submission_number):
             parsed_object.builder.add_acquisition_source(
                 source="arXiv",
                 method="hepcrawl",
-                date=datetime.datetime.now().isoformat(),
+                date=datetime.datetime.now(datetime.UTC).isoformat(),
```
EliseRecejac: @DonHaul why is it date and not datetime here ?

ui/src/backoffice/utils/__tests__/utils.test.tsx
```diff
@@ -87,12 +87,12 @@ describe('filterByProperty', () => {
 
 describe('formatDateTime', () => {
   it('should return an object with formatted date and time for a valid date', () => {
-    const rawDateTime = '2024-02-11T15:30:00Z';
+    const rawDateTime = '2024-02-11T15:30:00';
```
DonHaul: with the changes in django i.e.`datetime.datetime.now(datetime.UTC).isoformat()`
incoming format will be 
2026-05-19T11:28:22.134496+00:00

will it also work for such cases?

EliseRecejac: yes it works, I updated the test (it was a relicate from when I thought we would like to send zurich hour from backend so the result is now 1hour later)
I also updated the function to remove the seconds

DonHaul (commented):
LGTM

ui/src/backoffice/utils/__tests__/utils.test.tsx
```diff
@@ -87,12 +87,12 @@ describe('filterByProperty', () => {
...
+    const rawDateTime = '2024-02-11T15:30:00.134496+00:00';
     const result = formatDateTime(rawDateTime);
 
     expect(result).toEqual({
-      date: '2024-02-11',
-      time: '15:30',
+      date: '2/11/2024',
+      time: '03:30 PM',
     });
   });
```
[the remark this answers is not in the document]
EliseRecejac: since I configure the timezone in the jest config the test is deterministic

ui/craco.config.js
```diff
@@ -39,6 +39,7 @@ module.exports = {
   jest: {
     configure(config) {
       config.transformIgnorePatterns = ['/node_modules/(?!antd|rc-*)/.+\\.js$'];
+      config.globalSetup = '<rootDir>/src/jestGlobalSetup.js';
```
ioannistsanaktsidis: not enough to just set it for test ?  like     config.testEnvironmentOptions = { TZ: 'UTC' };
 ?

ioannistsanaktsidis (changes requested):
Image

can we make it not to break the line ? if needed we can remove "at " and just have a comma or sth.

workflows/plugins/include/utils/arxiv.py
```diff
@@ -22,7 +22,7 @@ def build_records(xml_records, submission_number):
             parsed_object.builder.add_acquisition_source(
                 source="arXiv",
                 method="hepcrawl",
-                date=datetime.datetime.now().isoformat(),
+                datetime=datetime.datetime.now(datetime.UTC).isoformat(),
```
ioannistsanaktsidis, line 25: good catch 🚀

workflows/tests/test_utils_desy.py
```diff
@@ -6,7 +6,7 @@
 
 class FixedDatetime:
     @classmethod
-    def now(cls):
+    def now(cls, _tz=None):
```
drjova, line 9: why do we need to change the signature in this function?

ui/src/backoffice/utils/utils.tsx
```diff
@@ -301,10 +301,18 @@ export const formatDateTime = (
...
+    if (Number.isNaN(isoString.getTime())) {
+      return undefined;
+    }
+    const [datePart, timePart] = [
+      isoString.toLocaleDateString(),
+      isoString.toLocaleTimeString([], {
+        hour: '2-digit',
+        minute: '2-digit',
+      }),
+    ];
```
drjova: I might miss some context, could we just use this or just moment to handle that? https://github.com/inspirehep/inspirehep/blob/4351068e11eda3cf845ff86e4a51b0f430a4e7ff/ui/src/common/components/RecordUpdateInfo.jsx#L9-L13

drjova (changes requested):

workflows/tests/test_utils_desy.py
EliseRecejac: because we are modifying `datetime.datetime.now(datetime.UTC).isoformat(),` adding the datetime.UTC to enforce a format that include timezone and avoid misunderstanding on the front side
without this argument some tests are failing because they cannot manage the argument in `now`

ui/craco.config.js
EliseRecejac: I think we do not support it with our jest version :( that was my first try

ui/src/backoffice/utils/utils.tsx
EliseRecejac: Indeed, let stay coherent with the existant

EliseRecejac: with this solution I would not even have to do changes in the backend however I think it's better to have unambigous dates

drjova: Also, we have already in place a way to freeze time in the tests.

ui/src/backoffice/utils/utils.tsx
```diff
@@ -298,16 +301,17 @@ export const filterByProperty = (
...
-    const formattedTimePart = timePart.slice(0, 5);
-    return { date: datePart, time: formattedTimePart };
-  } catch (error) {
+  if (!rawDateTime) {
     return undefined;
   }
+  const momentToLocalTimezone = moment.utc(rawDateTime).tz(LOCAL_TIMEZONE);
+  if (!momentToLocalTimezone.isValid()) return undefined;
+  return {
+    date: momentToLocalTimezone.format('MM-DD-YYYY'),
```
drjova: About this, do we have this requirement for date format?

drjova (commented):
Thank you @EliseRecejac, just the comment for the date format.

ioannistsanaktsidis (approved):
Just to validate the comment from @drjova , otherwise it looks good!

ui/src/backoffice/utils/utils.tsx
EliseRecejac: I checked with Micha we'll stick to the ISO format

ioannistsanaktsidis: Perfect, you can proceed merging then  after ci is done! Thanks !

drjova (approved):

PR #4255, opened 2026-05-20
backend: hep submission flag logic fix

DonHaul:
* ref: cern-sis/issues-inspire/issues/1460


ioannistsanaktsidis (approved):
LGTM ! Please create a ticket so as to cleanup the next logic after we have completed the migration!

PR #4259, opened 2026-05-22
backoffice(ui): add feature add/remove subjects

EliseRecejac:
- ref: https://github.com/cern-sis/issues-inspire/issues/1501


ui/src/backoffice/literature/containers/LiteratureDetailPageContainer.tsx
```diff
@@ -100,7 +104,12 @@ const LiteratureDetailPageContainer = ({
   const submissionContext: any = shouldShowSubmissionModal
     ? { email: acquisitionSourceEmail, title }
     : undefined;
-  const inspireCategories = data?.get('inspire_categories')?.toJS();
+  const inspireCategoriesImmutable = data?.get('inspire_categories');
+
+  const inspireCategories = useMemo(
+    () => inspireCategoriesImmutable?.toJS() ?? [],
+    [inspireCategoriesImmutable]
+  );
```
EliseRecejac, line 112: I'm not sure it's the right way to do that (combined with useEffect), I would maybe have rather an intermediate component with the whole page rendered once the data is fetched but it doesn't seem to be the way you handle it

ui/src/actions/backoffice.ts
```diff
@@ -432,9 +491,19 @@ export function clearBackofficeBatchSubmittedIds() {
...
+    status: WorkflowStatuses;
+  }
 ): (dispatch: ActionCreator<Action>) => Promise<void> {
   return async (dispatch) => {
+    if (subjectsPayload) {
+      const success = await dispatch(
+        updateLiteratureAction(id, subjectsPayload)
+      );
+      if (!success) return;
+    }
```
EliseRecejac: not sure it's the right pattern to chain api calls, happy to discuss if it's not :)

EliseRecejac:
https://github.com/user-attachments/assets/9ac18985-be3a-434c-96e1-70bbc8868ceb

ui/src/actions/backoffice.ts
```diff
@@ -383,7 +390,59 @@ export function fetchLiteratureRecord(
...
+  return {
+    type: BACKOFFICE_UPDATE_LITERATURE_SUCCESS,
+  };
+}
+
+function updateLiteratureActionError(errorPayload: { error: Error }) {
+  return {
+    type: BACKOFFICE_UPDATE_LITERATURE_ERROR,
+    payload: { ...errorPayload },
+  };
```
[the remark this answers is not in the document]
EliseRecejac: Not sure about this one, I used the same pattern as the other functions of this file

ui/src/actions/backoffice.ts
drjova: hmm why do we need this here?

ui/src/backoffice/literature/components/SubjectArea.tsx
drjova, line 1: The subjects here get copied out of redux into local component state. That leaves us with two sources of truth for the same data: the record in the store and the draft in the component. The effect is really there to paper over that gap, and that is usually where sync bugs sneak in later, for example if the record refreshes while a curator is partway through editing.

Since updateLiteratureAction already refetches the record on success, the store is effectively our source of truth, so I think we can read from it directly instead of mirroring it, an [example](https://github.com/inspirehep/inspirehep/blob/f2444a77aa773f9352bd9c25cfd04961db386c1d/ui/src/backoffice/literature/containers/LiteratureDetailPageContainer.tsx#L327) Two ways we could go:

- Drop the draft entirely and dispatch updateLiteratureAction on each add and remove, reading subjects straight from state.backoffice. Simplest and removes the copy, but it means one save plus a refetch per click.
- Keep the batched Save and Cancel, but move the working list into a small redux draft slice `BACKOFFICE_SET_SUBJECTS` action rather than local state, so the buffer lives in the redux store

Happy to discuss it further. The main thing I want to avoid is keeping a local mirror of the record so we are not maintaining the same data in two places.

ui/src/submissions/common/schemas/constants.js
```diff
@@ -200,6 +200,30 @@ export const inspireCategoryOptions = [
...
+  l: 'Lattice',
+  m: 'Math and Math Physics',
+  n: 'Theory-Nucl',
+  o: 'Other',
+  p: 'Phenomenology-HEP',
+  q: 'General Physics',
+  s: 'Data Analysis and Statistics',
+  t: 'Theory-HEP',
+  x: 'Experiment-Nucl',
+};
```
drjova, line 221: is it any specific reason that we sort them this way?

ui/src/backoffice/literature/containers/LiteratureDetailPageContainer.tsx
```diff
@@ -205,14 +249,15 @@ const LiteratureDetailPageContainer = ({
...
-                          pagination={false}
-                          size="small"
-                          rowKey={(record) =>
-                            `${record?.term}+${Math.random()}`
-                          }
+                        <SubjectArea
+                          hasBeenUpdated={hasBeenUpdated}
+                          status={status}
+                          handleCancel={() => setNewSubjects(inspireCategories)}
+                          handleUpdate={handleUpdate}
```
drjova: we could move the whole logic of the action inside the `SubjectArea` component, wdyt and it will work together with the redux store change

drjova (changes requested):
Thank you @EliseRecejac, few comments, happy to discuss

ui/src/submissions/common/schemas/constants.js
EliseRecejac: honestly I just copied the old one, indeed it may be better to stick to the same order as in the dropdown

drjova: We don't need a sort order other than alphabetical. @michamos, any ideas why it was implemented like that back then?

EliseRecejac: (it's alphabetical per shortcut, it wasn't displayed anywhere)

ui/src/actions/backoffice.ts
EliseRecejac: it's to save automaticcaly the subjects before sending the decision (it was requested by Micha)

drjova: sorry I wasn't clear, the `updateLiteratureAction` is inside the `resolveLiteratureAction`, it should be a separate action when the update the subjects and they click save (before the select any resolve action)

ui/src/submissions/common/schemas/constants.js
drjova: it doesn't seem alphabetically since 

 g: 'Gravitation and Cosmology',
 q: 'General Physics',

or 

  e: 'Experiment-HEP',
  x: 'Experiment-Nucl',

are far apart

EliseRecejac: I mean it's the shorcuts that are ordered, not their meaning

EliseRecejac: but yes it doesn't make sense, I think it was never displayed as a list before

drjova: Hmm, maybe it was used for keyboard shortcut, that we don't support anymore. We can drop them and use just the values for now. @michamos, is that okay?

EliseRecejac: sorry I wasn't clear, we still use the keyboard shortcuts, it was explicitly requested

ui/src/actions/backoffice.ts
```diff
@@ -383,7 +391,67 @@ export function fetchLiteratureRecord(
...
+    try {
+      const response = await httpClient.patch(
+        `${BACKOFFICE_API}/workflows/${LITERATURE_PID_TYPE}/${id}/`,
+        payload
+      );
+
+      dispatch(updateLiteratureActionSuccess());
+      notifyActionSuccess(WorkflowActions.UPDATE);
+      dispatch(fetchLiteratureRecordSuccess(response.data));
+      return true;
```
drjova: no need to return

ui/src/actions/backoffice.ts
```diff
@@ -383,7 +391,67 @@ export function fetchLiteratureRecord(
...
+    } catch (err) {
+      const { error } = httpErrorToActionPayload(err);
+      notifyActionError(
+        (typeof error?.error === 'string'
+          ? error?.error
+          : error?.error?.detail) || 'An error occurred'
+      );
+      dispatch(updateLiteratureActionError(error));
+      return false;
```
drjova: ditto

ui/src/backoffice/literature/containers/LiteratureDetailPageContainer.tsx
```diff
@@ -128,12 +140,27 @@ const LiteratureDetailPageContainer = ({
...
+          },
+          status:
+            status === WorkflowStatuses.MISSING_SUBJECT_FIELDS
+              ? WorkflowStatuses.APPROVAL
+              : status,
+        })
+      );
+      if (!success) return;
+    }
+    dispatch(resolveLiteratureAction(id, { action, value }));
```
drjova, line 169: we can move the whole logic inside the SubjectArea component

drjova (changes requested):
few small comments, thank you 😄

ui/src/backoffice/literature/containers/LiteratureDetailPageContainer.tsx
EliseRecejac: I don't think we can since it's used by LiteratureMatches and LiteratureDecisionBox

drjova: hmm, what I don't understand is why we are changing the `handleResolveAction` to handle the subject case? Resolve should be only for resolving actions not to update records.

ui/src/backoffice/literature/containers/LiteratureDetailPageContainer.tsx
```diff
@@ -328,6 +354,7 @@ const stateToProps = (state: RootStateOrAny) => ({
   loading: state.backoffice.get('loading'),
   restartActionInProgress: state.backoffice.get('restartActionInProgress'),
   isSuperUserLoggedIn: isSuperUser(state.user.getIn(['data', 'roles'])),
+  subjectsDraft: state.backoffice.get('subjectsDraft')?.toJS() ?? null,
```
drjova: .toJS() mints a new object each call, so this prop is never referentially equal across renders and the component rerenders on every dispatch. Keep it Immutable here and convert at the point of use, or use a memoized selector.

ui/src/backoffice/literature/containers/LiteratureDetailPageContainer.tsx
EliseRecejac: it's because Micha wants that, on resolving a workflow, if we have some pending subjects we should save them automatically before (to reduce the number of clicks)
I agree it's a weird thing to do but he said that they are used to this behaviour so we need to keep it

EliseRecejac:
I'm closing this one since we discussed with @ioannistsanaktsidis  and changed a bit how it works

PR #4262, opened 2026-05-26
 backoffice(ui): various UI improvements

EliseRecejac:
- ref: https://github.com/cern-sis/issues-inspire/issues/1502


EliseRecejac:
Capture d’écran 2026-05-26 à 13 43 37

EliseRecejac:
Capture d’écran 2026-05-26 à 14 11 30

EliseRecejac:
Capture d’écran 2026-05-26 à 14 12 02

ui/src/literature/components/ArxivEprintPdfLink.jsx
```diff
@@ -0,0 +1,33 @@
...
+class ArxivEprintPdfLink extends Component {
+  get arxivId() {
+    const { children } = this.props;
+    return children;
+  }
+
+  render() {
+    const { page } = this.props;
+
+    const href = `//arxiv.org/pdf/${this.arxivId}`;
```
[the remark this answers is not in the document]
EliseRecejac: it is actually what we want

ui/src/literature/components/ArxivEprintPdfLink.jsx
```diff
@@ -0,0 +1,33 @@
...
+      >
+        <LinkWithTargetBlank href={href}>PDF</LinkWithTargetBlank>
+      </EventTracker>
+    );
+  }
+}
+
+ArxivEprintPdfLink.propTypes = {
+  children: PropTypes.node.isRequired,
+};
```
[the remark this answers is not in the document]
EliseRecejac: It looks relevant but I copied ArxivEprintLink, so should I fix both or is it ok like this ?

ui/src/common/components/AggregationFilters.tsx
```diff
@@ -25,6 +25,7 @@ const AggregationFilters = ({
   embedded,
   page,
   namespace,
+  narrow,
```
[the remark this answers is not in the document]
EliseRecejac: again I followed the existing pattern (inline, embedded) but the comment seems right, should I update every props ?

ui/src/common/components/AggregationFilters.tsx
```diff
@@ -36,6 +37,7 @@ const AggregationFilters = ({
   embedded: boolean;
   page: number;
   namespace: string;
+  narrow: boolean;
```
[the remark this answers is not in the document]
EliseRecejac: false positive, with the defaultProps typescript is happy

ui/src/backoffice/authors/search/containers/SearchPageContainer.tsx
```diff
@@ -44,7 +44,10 @@ const AuthorsSearchPageContainer = ({
 }: BackofficeSearchPageProps) => {
   const renderAggregations = () => (
     <LoadingOrChildren loading={loadingAggregations}>
-      <AggregationFiltersContainer namespace={BACKOFFICE_AUTHORS_SEARCH_NS} />
+      <AggregationFiltersContainer
+        namespace={BACKOFFICE_AUTHORS_SEARCH_NS}
+        narrow
+      />
```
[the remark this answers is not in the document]
EliseRecejac: false positive

ui/src/backoffice/literature/search/containers/SearchPageContainer.tsx
```diff
@@ -226,6 +226,7 @@ const LiteratureSearchPageContainer = ({
     <LoadingOrChildren loading={loadingAggregations}>
       <AggregationFiltersContainer
         namespace={BACKOFFICE_LITERATURE_SEARCH_NS}
+        narrow
       />
```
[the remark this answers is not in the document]
EliseRecejac: false positive

ui/src/literature/components/ArxivEprintPdfLink.jsx
```diff
@@ -0,0 +1,33 @@
...
+    const { page } = this.props;
+
+    const href = `//arxiv.org/pdf/${this.arxivId}`;
+    return (
+      <EventTracker
+        eventCategory={page}
+        eventAction="link"
+        eventId="Arxiv eprint pdf link"
+      >
+        <LinkWithTargetBlank href={href}>PDF</LinkWithTargetBlank>
```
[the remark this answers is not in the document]
EliseRecejac: I also fix ArxivEprintLink

ui/src/backoffice/literature/components/LiteratureMatches.jsx
```diff
@@ -104,6 +105,13 @@ const LiteratureMatchItem = ({ match, selectedBestMatch, onSelect }) => {
             </Paragraph>
           </Descriptions.Item>
         )}
+        {reportNumbers.size > 0 && (
+          <Descriptions.Item label="Report numbers">
+            <Paragraph style={{ margin: 0, whiteSpace: 'pre-line' }}>
+              {reportNumbers.map((pn) => pn.get('value')).join('\n')}
+            </Paragraph>
+          </Descriptions.Item>
+        )}
```
drjova, line 113: Could we use this? https://github.com/inspirehep/inspirehep/blob/f2444a77aa773f9352bd9c25cfd04961db386c1d/ui/src/literature/components/ReportNumberList.jsx#L7 does it fit the UI needs for the curators?

ui/src/backoffice/literature/components/WorkflowResultItem.tsx
```diff
@@ -143,17 +145,37 @@ const WorkflowResultItem = ({
                   </Paragraph>
                 </Descriptions.Item>
               )}
+
+              {displayReportNumbers && (
+                <Descriptions.Item label="Report numbers">
+                  <Paragraph style={{ margin: 0, whiteSpace: 'pre-line' }}>
+                    {reportNumbers.map((pn: any) => pn.get('value')).join('\n')}
+                  </Paragraph>
```
drjova: ditto

ui/src/common/components/AggregationFilters.tsx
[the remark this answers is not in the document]
drjova: we can open a ticket for that, otherwise the PR will have a bigger scope, wdyt?

ui/src/common/components/AuthorList.jsx
```diff
@@ -50,12 +61,13 @@ class AuthorList extends Component {
...
             ? this.renderShowAllOrEtAl()
-            : null
+            : this.renderNumberOfAuthors()
         }
         extractKey={getAuthorName}
         renderItem={(author) => (
           <Author author={author} page={page} unlinked={unlinked} />
         )}
+        separator={SEPARATOR_SEMICOLON}
       />
```
drjova, line 71: we have this which is also taking care the author number of authors, can this be used here? https://github.com/inspirehep/inspirehep/blob/f2444a77aa773f9352bd9c25cfd04961db386c1d/ui/src/common/components/AuthorsAndCollaborations.jsx#L8

ui/src/literature/components/ArxivEprintPdfLink.jsx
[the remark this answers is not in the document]
drjova: This is a different way from what we are doing on hep. We do create URLs in the serializer and we are using it like this:
- [this](https://github.com/inspirehep/inspirehep/blob/f2444a77aa773f9352bd9c25cfd04961db386c1d/ui/src/literature/containers/DetailPageContainer/DetailPageContainer.tsx#L207-L216) is how we use it for a similar case on the UI
- https://github.com/inspirehep/inspirehep/blob/f2444a77aa773f9352bd9c25cfd04961db386c1d/backend/inspirehep/records/marshmallow/literature/ui.py#L156-L166

Indeed, we are handling more cases on hep, but can we check if we can reuse this component? 


Otherwise, we can create a new component but better move it inside the backoffice since it will be only related to there. 

Since this is a new component, it would be a good opportunity to start using functional components. We will likely need to refactor the existing ones as well at some point.

Also, we can use directly the https://github.com/inspirehep/inspirehep/blob/master/ui/src/literature/components/UrlsAction.tsx#L30 which wraps the event.

ui/src/literature/components/ReportNumberList.jsx
```diff
@@ -9,7 +9,7 @@ class ReportNumberList extends Component {
     const { reportNumbers } = this.props;
     return (
       <InlineDataList
-        label="Report number"
+        label="Report numbers"
```
drjova: That's a good catch! I will double-check this with @michamos. I don't remember if it was in the purpose singular.

drjova (changes requested):
Thank you @EliseRecejac few comments :)

ui/src/literature/components/ReportNumberList.jsx
EliseRecejac: I checked with him it should indeed be plural

ui/src/common/components/AuthorList.jsx
EliseRecejac: are you sure it displays the number of authors ? looking at the component and how it is used in https://inspirebeta.net/literature?sort=mostrecent&size=25&page=3 it doesn't seem to be the case but I may miss something

ui/src/common/components/AggregationFilters.tsx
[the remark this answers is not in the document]
EliseRecejac: https://github.com/cern-sis/issues-inspire/issues/1511

backend/inspirehep/matcher/serializers.py
```diff
@@ -19,6 +19,7 @@ class LiteratureSummary(Schema):
     number_of_pages = fields.Raw(dump_only=True)
     earliest_date = fields.Raw(dump_only=True)
     public_notes = fields.Method("get_public_notes", dump_only=True)
+    report_numbers = fields.Method("get_report_numbers", dump_only=True)
```
drjova, line 22: This should be enough, could you please check?
```suggestion
    report_numbers = fields.Field("get_report_numbers", dump_only=True, dump_default=missing)
```

backoffice/backoffice/hep/api/serializers.py
```diff
@@ -162,16 +187,21 @@ class Meta:
         fields = "__all__"
 
 
-class HepBackofficeSearchUISerializer(BaseBackofficeSearchUISerializer):
+class HepBackofficeSearchUISerializer(
+    FulltextLinksMixin, BaseBackofficeSearchUISerializer
+):
     def get_hit_representation(self, item):
         hit = super().get_hit_representation(item)
+        data = item.get("data")
```
drjova: data should be already be part of the [hit](https://github.com/inspirehep/inspirehep/blob/b11e414e1cf4df6e1149a05cfa161a42c5b7acd0/backoffice/backoffice/common/serializers.py#L33), I would move it directly there since we already have similar changes too

ui/src/backoffice/literature/components/LiteratureMainInfo.jsx
```diff
@@ -97,6 +101,18 @@ const LiteratureMainInfo = ({ data, isLiteratureUpdate }) => {
         {acceleratorExperiment && (
           <div className="mb2">Experiment: {acceleratorExperiment}</div>
         )}
+        {hasFullTextLinks && (
+          <div style={{ paddingLeft: '-15px' }}>
+            <UrlsAction
+              urls={fullTextLinks}
+              icon={<FilePdfOutlined />}
+              text="pdf"
+              trackerEventId="Pdf download"
```
drjova:
```suggestion
              trackerEventId="Backoffice pdf download"
```

ui/src/backoffice/literature/components/WorkflowResultItem.tsx
```diff
@@ -90,7 +99,20 @@ const WorkflowResultItem = ({
...
         <Col className="col-details">
-          <ResultItem>
+          <ResultItem
+            leftActions={
+              hasFullTextLinks && (
+                <UrlsAction
+                  urls={fullTextLinks as List<any>}
+                  icon={<FilePdfOutlined />}
+                  text="pdf"
+                  trackerEventId="Pdf download"
```
drjova:
```suggestion
                  trackerEventId="Backend search pdf download"
```

ui/src/common/components/AuthorList.jsx
drjova: it should yes, underlaying is using https://github.com/inspirehep/inspirehep/blob/master/ui/src/common/components/AuthorList.jsx#L36 which has the number of authors on the show all button (https://inspirebeta.net/literature/3152106). Do I miss something?

ui/src/literature/components/ArxivEprint.jsx
```diff
@@ -7,9 +7,10 @@ import ArxivEprintLink from './ArxivEprintLink';
 class ArxivEprint extends Component {
   render() {
     const { eprint } = this.props;
+    const arxivId = eprint.get('value');
     return (
       <span>
-        <ArxivEprintLink>{eprint.get('value')}</ArxivEprintLink>
+        <ArxivEprintLink>{arxivId}</ArxivEprintLink>
```
drjova: do we need that just for readability?

ui/src/literature/components/ReportNumberList.tsx
```diff
@@ -0,0 +1,28 @@
...
+import { List } from 'immutable';
+
+import InlineDataList from '../../common/components/InlineList';
+
+interface ReportNumberListProps {
+  reportNumbers: List<any>;
+  hideLabel?: boolean;
+}
+
+const ReportNumberList = ({
```
drjova, line 12: can we reuse it https://github.com/inspirehep/inspirehep/blob/b11e414e1cf4df6e1149a05cfa161a42c5b7acd0/ui/src/literature/components/ReportNumberList.jsx#L7 and add the label flag?

drjova (changes requested):
Thank you @EliseRecejac! few more comments :)

ui/src/literature/components/ArxivEprint.jsx
EliseRecejac: ah yes indeed it makes less sense now

ui/src/literature/components/ReportNumberList.tsx
EliseRecejac: it's actually the same one, I just updated the file to use functional component and move to typescript (so the file name has changed)

ui/src/common/components/AuthorList.jsx
EliseRecejac: Ah yes, but we want to add the number of authors when we do not display the show all button 
Capture d’écran 2026-05-29 à 08 50 09

drjova: Could we handle this case instead of duplicating part of the code? Alternatively, we could wrap the common logic to avoid the duplication.

drjova (dismissed):
Thank you @EliseRecejac, just the comment if we could wrap the code for this case

ui/src/common/components/AuthorList.jsx
EliseRecejac: I'm sorry I'm not sure to see which part exactly is duplicated ? is it that we should do something like 
```
  renderShowAllOrEtAlOrNumberOfAuthors() {
    const { enableShowAll, authors, limit } = this.props;
    if (authors.size > limit) {
      if (enableShowAll) {
        return (
          <div className="di pl1">
            <SecondaryButton onClick={this.onModalOpen}>
              Show All({authors.size})
            </SecondaryButton>
          </div>
        );
      }
      return <span> et al.</span>;
    }
    return (
      <span>
        {' '}
        ({authors.size} {pluralizeUnlessSingle('author', authors.size)})
      </span>
    );
  }

  renderAuthorList(authorsToDisplay, displayShowAll = true) {
    const { wrapperClassName, page, unlinked } = this.props;
    return (
      <InlineDataList
        wrapperClassName={wrapperClassName}
        items={authorsToDisplay}
        suffix={displayShowAll && this.renderShowAllOrEtAlOrNumberOfAuthors()}
        extractKey={getAuthorName}
        renderItem={(author) => (
          <Author author={author} page={page} unlinked={unlinked} />
        )}
        separator={SEPARATOR_SEMICOLON}
      />
    );
  }
  ```

File: backend/inspirehep/matcher/serializers.py [hunk 18]
```diff
@@ -19,6 +19,9 @@ class LiteratureSummary(Schema):
     number_of_pages = fields.Raw(dump_only=True)
     earliest_date = fields.Raw(dump_only=True)
     public_notes = fields.Method("get_public_notes", dump_only=True)
+    report_numbers = fields.Field(
```
ioannistsanaktsidis: this need to be a fields.Method

backoffice/backoffice/hep/api/serializers.py
```diff
@@ -68,7 +102,7 @@ class Meta:
         ),
     ],
 )
-class HepWorkflowSerializer(BaseWorkflowSerializer):
+class HepWorkflowSerializer(FulltextLinksMixin, BaseWorkflowSerializer):
```
ioannistsanaktsidis: ditto, no need of the fulltextlinksmixin

ui/src/backoffice/literature/components/LiteratureMainInfo.jsx
```diff
@@ -97,6 +101,18 @@ const LiteratureMainInfo = ({ data, isLiteratureUpdate }) => {
         {acceleratorExperiment && (
           <div className="mb2">Experiment: {acceleratorExperiment}</div>
         )}
+        {hasFullTextLinks && (
```
ioannistsanaktsidis: no need for this, we just need to the pdf link next to the e-print section.

ui/src/backoffice/literature/components/WorkflowResultItem.tsx
```diff
@@ -90,7 +99,20 @@ const WorkflowResultItem = ({
     >
       <Row justify="start" wrap={false}>
         <Col className="col-details">
-          <ResultItem>
+          <ResultItem
```
ioannistsanaktsidis, line 105: No need for that, pdf link is only needed in the detail page

ui/src/common/components/AuthorList.jsx
ioannistsanaktsidis: As this will have effect on hep too, we need to validate before proceeding.

ui/src/common/components/AuthorList.jsx
```diff
@@ -50,12 +61,13 @@ class AuthorList extends Component {
...
           authors.size > limit && displayShowAll
             ? this.renderShowAllOrEtAl()
-            : null
+            : this.renderNumberOfAuthors()
         }
         extractKey={getAuthorName}
         renderItem={(author) => (
           <Author author={author} page={page} unlinked={unlinked} />
         )}
+        separator={SEPARATOR_SEMICOLON}
```
ioannistsanaktsidis, line 70: This also need validation as it will start separating authors on hep with semicolon

ioannistsanaktsidis (changes requested):
Just a few comments

backend/inspirehep/matcher/serializers.py
EliseRecejac: @ioannistsanaktsidis

ioannistsanaktsidis:
Also this test here https://github.com/inspirehep/inspirehep/blob/master/ui/src/backoffice/literature/components/__tests__/WorkflowResultItem.test.tsx#L42 needs some prop updating (missing page prop or sth )

ui/src/backoffice/literature/components/__tests__/AutomaticDecision.test.jsx
```diff
@@ -5,24 +5,9 @@ import { Map } from 'immutable';
...
-        relevancePrediction={relevancePrediction}
-      />
-    );
-    expect(container.firstChild).toBeNull();
-  });
-
   test('renders null when relevancePrediction is missing', () => {
     const { container } = render(
-      <AutomaticDecision hasInspireCategories relevancePrediction={null} />
+      <AutomaticDecision relevancePrediction={null} />
```
ioannistsanaktsidis, line 10: Check also the other components calling this one. LiteratureDecisionBox still passes the hasInspireCategories

ioannistsanaktsidis (changes requested):

ui/src/backoffice/literature/components/WorkflowResultItem.tsx
```diff
@@ -143,17 +165,35 @@ const WorkflowResultItem = ({
                   </Paragraph>
                 </Descriptions.Item>
               )}
+
+              {displayReportNumbers && (
+                <Descriptions.Item label="Report numbers">
+                  <ReportNumberList reportNumbers={reportNumbers} hideLabel />
+                </Descriptions.Item>
+              )}
```
ioannistsanaktsidis, line 176: I am not sure I understand that part ? ticket says 
`in match candidates, also display report numbers (helps avoid false positives)
`, why do we need that here ? isn't what we have here enough https://github.com/inspirehep/inspirehep/pull/4262/changes#diff-9c7965f2b7766b0e4c6002e9d897475528aa15b0b88ac977b9964e02efc17d68R109 ?

EliseRecejac: as discussed we need to compare the matches with the main one only for fuzzy matching, that's why

ui/src/backoffice/literature/components/LiteratureMainInfo.jsx
EliseRecejac: actually we want to reproduce the inspirehep ui and have it there

ui/src/backoffice/literature/components/WorkflowResultItem.tsx
EliseRecejac: actually I think we want it on both (waiting for Micha's confirmation on that)

ui/src/common/components/AuthorList.jsx
EliseRecejac: for that and the number of authors I think we want it on hep too (waiting for Micha's answer)

ui/src/backoffice/literature/components/LiteratureMainInfo.jsx
```diff
@@ -10,12 +13,14 @@ import CollaborationList from '../../../common/components/CollaborationList';
...
 
-const LiteratureMainInfo = ({ data, isLiteratureUpdate }) => {
+const LiteratureMainInfo = ({ data, isLiteratureUpdate, page }) => {
   const authors = data?.get('authors');
   const title = data?.getIn(['titles', 0]);
   const abstract = data?.getIn(['abstracts', 0]);
   const license = data?.getIn(['license', 0]);
   const arxivEprints = data?.get('arxiv_eprints');
+  const arxivId = arxivEprints && arxivEprints.first()?.get('value');
+  const pdfLink = arxivId ? `//arxiv.org/pdf/${arxivId}` : null;
```
ioannistsanaktsidis: instead of taking the first one, you can iterate in a small function and construct the object needed , then pass it to the urlsAction and we should be ok. This is how it was done in the old system for reference

https://github.com/inspirehep/inspirehep-search-js/blob/5e4ef28f8865b77f9ab6ad980c427b2abbb59291/src/inspirehep-holdingpen-js/templates/inspirehep-holdingpen-js/details/details.html#L140-L145

ui/src/backoffice/literature/components/WorkflowResultItem.tsx
```diff
@@ -72,16 +78,20 @@ const WorkflowResultItem = ({
   const publicationInfo = data?.get('publication_info');
   const hasPublicationInfoValue = hasPublicationInfo(publicationInfo);
 
-  const arxivEprints = data?.get('arxiv_eprints');
+  const arxivEprints = data?.get('arxiv_eprints') as List<Map<string, any>>;
   const hasArxivEprints = List.isList(arxivEprints) && arxivEprints.size > 0;
+  const arxivId = hasArxivEprints && arxivEprints.first()?.get('value');
+
+  const pdfLink = arxivId ? `//arxiv.org/pdf/${arxivId}` : null;
```
ioannistsanaktsidis: since we do that here too, you can create a small utils function and use it in both places

ioannistsanaktsidis (changes requested):
Just the small change

ioannistsanaktsidis (approved):
All good! Just the validation for the authors change to be clarified!

PR #4263, opened 2026-05-26
backoffice(ui): remove useless isBackoffice condition

EliseRecejac:
* ref: https://github.com/cern-sis/issues-inspire/issues/1495


ioannistsanaktsidis (approved):
LGTM

PR #4265, opened 2026-05-27
e2e: update packages

EliseRecejac:
* ref: https://github.com/cern-sis/issues-inspire/issues/1488

### DIFF
--- a/e2e/cypress.config.js
+++ b/e2e/cypress.config.js
@@ -1,16 +1,17 @@
-const { defineConfig } = require('cypress')
+const { defineConfig } = require("cypress");
 
 module.exports = defineConfig({
-  projectId: '439or1',
-  videosFolder: 'cypress/__videos__',
-  screenshotsFolder: 'cypress/__screenshots__',
+  projectId: "439or1",
+  videosFolder: "cypress/__videos__",
+  screenshotsFolder: "cypress/__screenshots__",
   defaultCommandTimeout: 10000,
   env: {
-    inspirehep_url: 'http://localhost:8080',
-    backoffice_url: 'http://localhost:8001',
+    inspirehep_url: "http://localhost:8080",
+    backoffice_url: "http://localhost:8001",
     mobile_viewport_width: 375,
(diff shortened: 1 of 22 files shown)


e2e/package.json
```diff
@@ -11,9 +11,8 @@
   "author": "",
   "license": "ISC",
   "dependencies": {
-    "@cypress/skip-test": "^2.5.0",
-    "cypress": "^14.4.0",
-    "cypress-image-snapshot": "^4.0.1",
-    "moment": "^2.24.0"
+    "cypress": "^15.15.0",
+    "@simonsmith/cypress-image-snapshot": "^10.0.4",
```
ioannistsanaktsidis, line 15: for now it is ok, we might consider removing snapshots in the future as they are all skipped

e2e/cypress.config.js
```diff
@@ -1,26 +1,27 @@
...
   defaultCommandTimeout: 10000,
   env: {
-    inspirehep_url: 'http://localhost:8080',
-    backoffice_url: 'http://localhost:8001',
+    inspirehep_url: "http://localhost:8080",
+    backoffice_url: "http://localhost:8001",
     mobile_viewport_width: 375,
     mobile_viewport_height: 667,
   },
+  allowCypressEnv: true, //cypress-image-snapshot is still using Cypress.env and do not plan to move before cypress 16 (https://github.com/simonsmith/cypress-image-snapshot/issues/86)
```
ioannistsanaktsidis, line 14: so we would have to change that when moving to version 16 correct ?

ioannistsanaktsidis (approved):
Just some comments for clarification

e2e/cypress.config.js
EliseRecejac: we don't "need" to change it but from cypress 15 they encourage you to set it at false as a safety mesure
So for the moment we get a warning because of it being true but when moving to 16 we will be able to pass it to false

ioannistsanaktsidis:
No furthe comments! Nice job!

PR #4267, opened 2026-05-28
workflows: refactor data

DonHaul:
* ref: cern-sis/issues-inspire/issues/1422


workflows/plugins/include/utils/data.py
```diff
@@ -0,0 +1,90 @@
...
+def download_record_versions(id):
+    """Download the versions of the record.
+
+    Args: id (int): The id of the record.
+    Returns: dict: The record versions.
+    """
+    hepdata_response = generic_http_hook.call_api(
+        endpoint=f"/record/ins{id}?format=json"
+    )
+    payload = hepdata_response.json()
```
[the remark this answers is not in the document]
DonHaul: done internally

workflows/dags/data/data_harvest.py
```diff
@@ -72,113 +67,77 @@ def collect_ids(**context):
...
-            Returns: dict: The built record.
-            """
-
-            inspire_http_record_management_hook = InspireHTTPRecordManagementHook()
-            inspire_url = inspire_http_record_management_hook.get_url()
+        logger.info(f"Processing {len(record_ids)} records.")
+        for record_id in record_ids:
+            try:
+                hepdata_records.append(download_record_versions(record_id))
+                time.sleep(1)  # TODO: Remove sleep after bulk import harvest
```
ioannistsanaktsidis, line 92: lets create a ticket for this

workflows/dags/data/data_harvest.py
```diff
@@ -72,113 +67,77 @@ def collect_ids(**context):
...
+            except Exception as e:
+                logger.error(
+                    f"Error occurred while downloading"
+                    f"versions for record {record_id}: {e}"
```
ioannistsanaktsidis, line 96: space missing

workflows/dags/data/data_harvest.py
```diff
@@ -38,10 +37,6 @@ def data_harvest_dag():
     """
     generic_http_hook = GenericHttpHook(http_conn_id="hepdata_connection")
```
ioannistsanaktsidis: should we put that inside the task ?

ioannistsanaktsidis (changes requested):
few comments

workflows/dags/data/data_harvest.py
```diff
@@ -72,113 +67,77 @@ def collect_ids(**context):
...
-                control_number=old_record["control_number"],
-                revision_id=revision_id + 1,
-            )
-            logger.info(
-                f"Data Record Updated: "
-                f"{response.json()['metadata']['self']['$ref']}"
-            )
-            return response.json()
+                failed_records["load_failed"].append(data_record["id"])
+
```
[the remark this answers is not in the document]
ioannistsanaktsidis: This is actually valid no ?

workflows/dags/data/data_harvest.py
```diff
@@ -72,113 +67,77 @@ def collect_ids(**context):
...
+                    f"Error occurred while building " f"record {hepdata_record_id}: {e}"
+                )
+                failed_records["build_failed"].append(hepdata_record_id)
+                continue
 
-            Args: record (dict): The record to normalize.
-            Returns: dict: The normalized record.
-            """
-            normalizations = workflows.normalize_collaborations(record)
+            normalizations = workflows.normalize_collaborations(data_record)
```
ioannistsanaktsidis: now that everything lies in one task, if this fails for some reasone the whole task will fail. Should we guard it ?

backoffice/backoffice/authors/api/views.py
```diff
@@ -267,8 +268,7 @@ def resolve(self, request, pk=None):
                 )
             except RequestException as e:
                 return handle_request_exception(
-                    "Error triggering Airflow DAG",
-                    e,
+                    "Error triggering Airflow DAG run with id %s", e, workflow.id
```
ioannistsanaktsidis, line 271: can we also put it to the hep ones ?

workflows/dags/data/data_harvest.py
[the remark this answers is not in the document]
DonHaul: indeed i will print the whole record instead

workflows/dags/data/data_harvest.py
```diff
@@ -72,113 +67,86 @@ def collect_ids(**context):
...
+                    data_record["accelerator_experiments"] = accelerator_experiments
+                    data_record["collaborations"] = normalized_collaborations
+
+                logger.info(f"Normalized collaborations for record {data_record}.")
+            except Exception as e:
+                logger.error(
+                    f"Error occurred while normalizing collaborations "
+                    f"for record {data_record}: {e}"
+                )
+                failed_records["normalize_failed"].append(data_record["id"])
```
ioannistsanaktsidis: ditto

DonHaul: should be okay now

workflows/dags/data/data_harvest.py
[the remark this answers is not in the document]
ioannistsanaktsidis: That part here is still not resolved, no ? we need the whole record here correct ?

ioannistsanaktsidis (changes requested):
Just the remaining comment

workflows/dags/data/data_harvest.py
```diff
@@ -72,113 +67,84 @@ def collect_ids(**context):
...
+                    data_record["accelerator_experiments"] = accelerator_experiments
+                    data_record["collaborations"] = normalized_collaborations
+
+                logger.info(f"Normalized collaborations for record {data_record}.")
+            except Exception as e:
+                logger.error(
+                    f"Error occurred while normalizing collaborations "
+                    f"for record {data_record}: {e}"
+                )
+                failed_records["normalize_failed"].append(hepdata_record_id)
```
ioannistsanaktsidis, line 127: This key needs to be added to the failed_records dict as it will raise a KeyError

workflows/dags/data/data_harvest.py
```diff
@@ -72,113 +67,84 @@ def collect_ids(**context):
...
-                accelerator_experiments, normalized_collaborations = normalizations
-                record["accelerator_experiments"] = accelerator_experiments
-                record["collaborations"] = normalized_collaborations
-            return record
+        data_records = []
+        for hepdata_record in hepdata_records:
+            try:
+                data_record = build_record(data_schema, hepdata_record)
+            except Exception as e:
+                hepdata_record_id = hepdata_record["base"]["record"]["id"]
```
ioannistsanaktsidis: this should be computed and the beginning of the loop as if build succeeds and normalze fails it will throw an unbound variable error

ioannistsanaktsidis (changes requested):
Just the 2 comments

ioannistsanaktsidis (approved):
LGTM

PR #4276, opened 2026-06-01
ui: from craco to vite

EliseRecejac:
* ref: https://github.com/cern-sis/issues-inspire/issues/534


ui/src/common/components/SanitizedHTML.tsx
```diff
@@ -0,0 +1,53 @@
...
+  const result = String(DOMPurify.sanitize(html ?? '', config));
+
+  if (allowedAttributes) {
+    DOMPurify.removeHook('uponSanitizeAttribute');
+  }
+
+  return result;
+}
+
+const SanitizedHTML = ({
```
EliseRecejac, line 111: we use this component + dompurify instead of the module react-sanitized-html which doesn't work well with vite

ui/src/setupProxy.js
EliseRecejac, line 1: the proxy config is now directly in vite config

EliseRecejac:
Vite is now the standard builder for react app, it has a mature ecosystem and the cost of migration is not too high with a simple configuration.

The cost of the migration was mainly due to the use of vitest to run the tests which had impact on the snapshots (the representation is not exactly the same as with jest), the mocks in the test, and the need to use polyfills to minimize the changes on the tests.

Risks:
- process.env.NODE_ENV is normally supported by vite but we will have to check that after deployement we do not have issues
- bundle size issues => test:size is green so it should be ok

ui/index.html
```diff
@@ -0,0 +1,30 @@
+<!DOCTYPE html>
+<html lang="en">
+  <head>
+    <meta charset="utf-8" />
+    <meta name="viewport" content="width=device-width, initial-scale=1, shrink-to-fit=no" />
+    <meta name="theme-color" content="#000000" />
+    <link rel="manifest" href="/manifest.json" />
+    <link rel="shortcut icon" href="/favicon.ico" />
+    <!-- piwik url should be changed incase CERN piwik url modified (in config.REACT_APP_PIWIK_URL) -->
+    <link rel="preconnect" href="https://piwik.web.cern.ch" />
```
michamos: this should be updated https://github.com/cern-sis/kubernetes-inspire/blob/2e170468a4ad45f5ff9bc65cc267ef155f5d899a/ui/environments/prod/config.js#L3

EliseRecejac: good catch !

ioannistsanaktsidis:
@EliseRecejac please rebase and resolve conflicts so as to check also that one! Thank you!

ui/src/index.jsx
```diff
@@ -21,7 +19,7 @@ import { userInactive } from './actions/user';
 
 Sentry.init({
   dsn: getConfigFor('REACT_APP_SENTRY_DSN'),
-  release: process.env.REACT_APP_VERSION,
+  release: import.meta.env.VITE_APP_VERSION,
```
ioannistsanaktsidis, line 22: I still see that we set the REACT_APP_VERSION in many places https://github.com/search?q=repo%3Ainspirehep%2Finspirehep%20REACT_APP_VERSION&type=code . This needs to be renamed accordingly no ?

ui/package.json
```diff
@@ -54,7 +50,6 @@
     "lodash": "^4.17.21",
     "math-interval-2": "^1.1.0",
     "moment-timezone": "^0.5.28",
-    "moment-timezone-data-webpack-plugin": "^1.3.0",
```
ioannistsanaktsidis, line 57: Could this be replaced by

```
export default defineConfig({
  resolve: {
    alias: {
      'moment-timezone': 'moment-timezone/builds/moment-timezone-with-data-10-year-range',
    },
  },
});

```

in vite-config ? Would it achieve the same instead of loading everything ?

ui/src/common/components/SanitizedHTML.tsx
```diff
@@ -0,0 +1,50 @@
...
+import DOMPurify from 'dompurify';
+
+interface Props {
+  html: string;
+  allowedTags?: string[];
+  allowedAttributes?: Record<string, string[]>;
+  className?: string;
+}
+
+function sanitizeHtml(
```
ioannistsanaktsidis, line 83: SanitizedHTML is more permissive than the react-sanitized-html it replaces. With no allowedTags/allowedAttributes, config stays {} and we inherit DOMPurify's wide default instead of sanitize-html's strict one. NewsAndUpdates (external WordPress HTML) and FulltextSnippet now keep , class, style, id that were previously stripped. Not a security issue but it means that weird things can now render. Should we consider define allowed tags and attributes so as to match the old behavior ?

ui/package.json
```diff
@@ -12,19 +12,17 @@
...
-    "build": "NODE_OPTIONS=--openssl-legacy-provider craco build",
-    "test:unit": "craco test --resetMocks=false --silent",
-    "test:size": "NODE_OPTIONS=--openssl-legacy-provider yarn build && bundlemon",
-    "test": "yarn lint && yarn test:unit --watchAll=false --resetMocks=false && yarn test:size",
+    "start": "vite",
+    "build": "vite build",
+    "test:unit": "vitest run --silent",
+    "test:size": "yarn build && bundlemon",
+    "test": "yarn lint && yarn test:unit && yarn test:size",
     "analyze": "source-map-explorer 'build/static/js/*.js'",
```
ioannistsanaktsidis: that is not working anymore (not sure we need it though )

ui/src/index.jsx
```diff
@@ -1,11 +1,9 @@
-import 'core-js/es/object/entries';
-import 'core-js/es/array/includes';
```
ioannistsanaktsidis, line 2: Two options , either we support what is defined in package.json 

```
"browserslist": [
    ">0.2%",
    "not dead",
    "not ie <= 11",
    "not op_mini all"
  ]
```
and we add 

```
import legacy from '@vitejs/plugin-legacy';
plugins: [react(), svgr(), legacy({ targets: ['defaults'] })],
```
or we fall into the defaults of vite and delete browserslist. 

@drjova  any ideas on that one ?

ioannistsanaktsidis (changes requested):
few comments/clarifications/proposals

ui/src/index.jsx
ioannistsanaktsidis: A report from Matomo provided by Micha regarding browser versions used the last month 

[Export _ Browser version _ May 2026.csv](https://github.com/user-attachments/files/29051893/Export._.Browser.version._.May.2026.csv)

ui/package.json
EliseRecejac: it wasn't working before either... but I'll fix it

ui/src/index.jsx
EliseRecejac: Analyzing this file (with Claude's help) it looks like only 0.1% of the users are on unsupported browsers, some of them may be bots, so the impacted population (real users) is more likely 0.07% (900-1000 users) however the majority of them seems quite engaged, so I think it's better to support it 
let me know if you think it's useless

michamos: any drawbacks to using the legacy plugin?

EliseRecejac: the bundle is larger, I'm checking if it introduces performance issues, otherwise I think it's ok

EliseRecejac: it seems ok

michamos: I would be in favor of going for it then, but I’ll let @ioannistsanaktsidis @drjova give a technically-informed opinion too.

ioannistsanaktsidis: Lets go for it, no big harm!

ui/src/common/components/SanitizedHTML.tsx
```diff
@@ -0,0 +1,125 @@
...
+  allowedTags?: string[],
+  allowedAttributes?: Record<string, string[]>
+): string {
+  const config: DOMPurify.Config = {
+    ALLOWED_TAGS: allowedTags ?? DEFAULT_ALLOWED_TAGS,
+  };
+  if (allowedAttributes) {
+    config.ALLOWED_ATTR = [...new Set(Object.values(allowedAttributes).flat())];
+    DOMPurify.addHook('uponSanitizeAttribute', (node, data) => {
+      const tag = node.tagName.toLowerCase();
```
ioannistsanaktsidis: lets guard that in case tagName is not present

ioannistsanaktsidis (approved):
LGTM ! 🚀 🤞 🙏

PR #4286, opened 2026-06-04
backoffice: hepworkflow async trigger dag

DonHaul:
* ref: cern-sis/issues-inspire/issues/1513


ioannistsanaktsidis (commented):
It looks good, the only thing I am concerned is that now if the airflow trigger fails, for whatever reason, workflow is stuck in a weird processing state and the error is silently dropped. It's more of a question of what do we want to do in such cases. One alternative is to add an on_failure event on celery which sets the workflow status to error if the task failed to trigger. Let's discuss IRL before proceeding.

ioannistsanaktsidis (approved):

PR #4296, opened 2026-06-11
workflows: gracefull fail on arxiv,cds

ioannistsanaktsidis:
* ref https://github.com/cern-sis/issues-inspire/issues/1528


DonHaul (approved):
Looks Good 🚀

PR #4301, opened 2026-06-12
backoffice(ui): add control number when it exists

EliseRecejac:
* ref: https://github.com/cern-sis/issues-inspire/issues/1519


ioannistsanaktsidis (approved):
LGTM 🚀

PR #4303, opened 2026-06-12
backoffice(ui): fix header issue

EliseRecejac:
* ref: https://github.com/cern-sis/issues-inspire/issues/1087

### DIFF
--- a/ui/src/common/layouts/Header/Header.less
+++ b/ui/src/common/layouts/Header/Header.less
@@ -18,10 +18,18 @@
 
     .header {
       padding: @header-padding;
+    }
 
+    .header-backoffice {
+      height: fit-content;
+      padding: @header-padding-backoffice;
+    }
+
+    .header,
+    .header-backoffice {
       @media (max-width: @screen-xs-max) {
         line-height: 56px;
-        height: auto; // unset height for mobile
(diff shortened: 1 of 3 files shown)


ioannistsanaktsidis:
Logo seems smaller with the change, can we check that ?

now
Screenshot 2026-06-15 at 16 53 35

before
Screenshot 2026-06-15 at 16 53 39

EliseRecejac:
> Logo seems smaller with the change, can we check that ?
> 
> now Screenshot 2026-06-15 at 16 53 35
> 
> before Screenshot 2026-06-15 at 16 53 39

what is your browser ? I don't think I had the issue on chrome

ioannistsanaktsidis:
interesting, i am using also chrome

EliseRecejac:
> interesting, i am using also chrome

hm let me check again

ioannistsanaktsidis (changes requested):
Problem lies in missing this class 

https://github.com/inspirehep/inspirehep/blob/master/ui/src/common/components/Logo/Logo.less#L8-L10

By removing the {backoffice: ... the style does not apply

ioannistsanaktsidis (approved):
LGTM!

PR #4305, opened 2026-06-15
backoffice: add command to retrigger wf by id

DonHaul:
* ref: cern-sis/issues-inspire/issues/1535


ioannistsanaktsidis (approved):
Just copilots beautifications and we are good!

PR #4306, opened 2026-06-15
backoffice: maintenance script for restarting workflows

ioannistsanaktsidis:
* ref https://github.com/cern-sis/issues-inspire/issues/1535


DonHaul (approved):
tested in qa. works well

PR #4308, opened 2026-06-16
backoffice: async batch resolve

ioannistsanaktsidis:
* ref https://github.com/cern-sis/issues-inspire/issues/1496

### DIFF
--- a/backoffice/.envs/local/.django
+++ b/backoffice/.envs/local/.django
@@ -10,7 +10,6 @@ REDIS_URL=redis://cache:6379/4
 # Celery
 CELERY_BROKER_URL=amqp://guest:guest@backoffice-mq:5672/
 CELERY_RESULT_BACKEND=redis://cache:6379/5
-CELERY_TASK_ALWAYS_EAGER=True
 # Flower
 CELERY_FLOWER_USER=debug
 CELERY_FLOWER_PASSWORD=debug
--- a/backoffice/backoffice/hep/api/views.py
+++ b/backoffice/backoffice/hep/api/views.py
@@ -4,9 +4,9 @@ import os
 import requests
 from rest_framework import status, viewsets
 from rest_framework.response import Response
-from django.shortcuts import get_object_or_404
-from django.utils.decorators import method_decorator
 from django.db import transaction
+from django.shortcuts import get_object_or_404
+from backoffice.hep.tasks import batch_resolve_workflows
 from backoffice.hep.utils import (
     add_hep_decision,
     resolve_workflow,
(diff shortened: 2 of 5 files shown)


backoffice/backoffice/hep/tasks.py
```diff
@@ -22,3 +27,23 @@ def trigger_hep_workflow_initialization(workflow_id, workflow_type):
...
+    Airflow DAG task is cleared. Do NOT wrap ``resolve_workflow`` in ``transaction.atomic()``,
+    that would defer the decision commit past the Airflow clear and reintroduce the race.
+    """
+    user = get_user_model().objects.get(pk=user_id)
+    for wf_id in data["ids"]:
+        try:
+            resolve_workflow(wf_id, data, user)
+        except Exception:
+            logger.exception("Error resolving workflow %s", wf_id)
+            HepWorkflow.objects.filter(pk=wf_id).update(status=HepStatusChoices.ERROR)
```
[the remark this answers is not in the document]
ioannistsanaktsidis: valid

backoffice/backoffice/hep/api/views.py
```diff
@@ -280,30 +279,19 @@ def manual_merge(self, request):
...
-
-        has_failures = any(not result["success"] for result in results)
-        response_status = (
-            status.HTTP_500_INTERNAL_SERVER_ERROR
-            if has_failures
-            else status.HTTP_200_OK
-        )
+        data = serializer.validated_data
+        data["ids"] = [str(wf_id) for wf_id in data["ids"]]
+        batch_resolve_workflows.delay(data, request.user.id)
```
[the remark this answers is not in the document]
ioannistsanaktsidis: also valid, checking

ioannistsanaktsidis:
Comments from copilot addressed!

backoffice/.envs/local/.django
```diff
@@ -10,7 +10,6 @@ REDIS_URL=redis://cache:6379/4
 # Celery
 CELERY_BROKER_URL=amqp://guest:guest@backoffice-mq:5672/
 CELERY_RESULT_BACKEND=redis://cache:6379/5
-CELERY_TASK_ALWAYS_EAGER=True
```
drjova, line 13: For testing, I think we should leave that, otherwise the tests will be slower. On production we should not run it in eager mode.

backoffice/backoffice/hep/tasks.py
```diff
@@ -22,3 +29,27 @@ def trigger_hep_workflow_initialization(workflow_id, workflow_type):
         workflow_id,
     )
     return airflow_utils.trigger_airflow_dag(dag_id, workflow_id)
+
+
+@celery_app.task(soft_time_limit=10 * 60, time_limit=11 * 60)
```
drjova, line 34: `reject_on_worker_lost=True` and `acks_late=True` would be helpful here, and maybe autoretry on certain errors

backoffice/.envs/local/.django
ioannistsanaktsidis: discussed irl, this var is redundant as it is not referenced anywhere

backoffice/backoffice/hep/tasks.py
ioannistsanaktsidis: For these 2 we could add them considering the following risks

* a batch that hits the hard time_limit gets redelivered and retries forever. (most probably never gonna happen as we allow up to 250 batch) 
* the whole batch restarts from scratch (no resume), so already-done work is reprocessed. This is safe for add_hep_decision and workflow save but it might retrigger mutliple times the airflow task 


So not sure if we should apply them or let it fail and check if any limit is hit 

wdyt ?

drjova: I see, we can skip them then

drjova (approved):

PR #4318, opened 2026-06-17
workflows: speed-up auto-rejects

DonHaul:
* ref: cern-sis/issues-inspire/issues/1378

### DIFF
--- a/workflows/dags/literature/hep_create_dag.py
+++ b/workflows/dags/literature/hep_create_dag.py
@@ -221,11 +221,9 @@ def hep_create_dag():
                 workflow_id=workflow_id,
                 decision_data={"action": DECISION_DISCARD},
             )
-            workflow_management_hook.set_workflow_status(
-                status_name=STATUS_COMPLETED,
-                workflow_id=workflow_id,
-            )
-            return "run_next_if_necessary"
+            workflow_data["status"] = STATUS_COMPLETED
+            workflow_management_hook.update_workflow(workflow_id, workflow_data)
+            return "early_run_next_if_necessary"
 
         return "check_for_blocking_workflows"
 
@@ -426,7 +424,7 @@ def hep_create_dag():
 
         s3_store.write_workflow(workflow_data)
 
-    @task.branch
+    @task
     def check_if_previously_rejected(**context):
         """Equivalent to first IF of PROCESS_HOLDINGPEN_MATCH_HARVEST"""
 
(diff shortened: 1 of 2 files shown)


DonHaul:
Speed up auto-rejects:
`skip_if_auto_reject` will make the whole pipeline skip

Image

workflows/dags/literature/hep_create_dag.py
```diff
@@ -2040,19 +2050,25 @@ def run_next_if_necessary(**context):
...
-        >> [check_auto_approve_task, save_and_complete_workflow_task]
+        >> [check_auto_approve_task, early_save_workflow_task]
     )
+
     (
         check_auto_approve_task
-        >> check_if_previously_rejected()
-        >> [preprocessing_group, save_and_complete_workflow_task]
+        >> check_if_previously_rejected_task
+        >> [skip_if_auto_rejected_task, early_save_workflow_task]
```
ioannistsanaktsidis: This is ungated now, meaning that it always runs  no matter what. I.e 

* check_if_previously_rejected changed from @task.branch → plain @task, so it no longer skips any downstream.
* skip_if_auto_rejected gates preprocessing correctly, but early_save_workflow_task is ungated and fires on every run 


This causes a normal harvest to complete early and show as completed while the preprocessing still runs. 

Image

ioannistsanaktsidis (changes requested):

workflows/dags/literature/hep_create_dag.py
DonHaul: indeed i had forgotten about this scenario

workflows/dags/literature/hep_create_dag.py
```diff
@@ -2040,17 +2064,28 @@ def run_next_if_necessary(**context):
...
+    early_save_workflow_task = save_and_complete_workflow.override(
+        task_id="save_and_complete_workflow_early"
+    )()
 
     (
         check_for_fuzzy_matches_task
         >> await_decision_fuzzy_match()
         >> stop_if_existing_submission_notify_and_close_task
-        >> [check_auto_approve_task, save_and_complete_workflow_task]
+        >> [check_auto_approve_task, early_save_workflow_task]
```
ioannistsanaktsidis, line 2079: now that this is wired to early_save, shouldn't we also return the correct one here https://github.com/inspirehep/inspirehep/blob/master/workflows/dags/literature/hep_create_dag.py#L405 ? i.e return save_and_complete_workflow_early ?

workflows/dags/literature/hep_create_dag.py
```diff
@@ -2076,7 +2111,7 @@ def run_next_if_necessary(**context):
         >> validate_record()
         >> set_workflow_status_to_running()
         >> discard_older_wfs_w_same_source()
-        >> [check_for_blocking_workflows_task, run_next_if_necessary_task]
+        >> [check_for_blocking_workflows_task, early_save_workflow_task]
```
ioannistsanaktsidis, line 2117: same here, now that we rewired it should this https://github.com/inspirehep/inspirehep/blob/master/workflows/dags/literature/hep_create_dag.py#L228 branch return save_and_complete_workflow_early ?

workflows/dags/literature/hep_create_dag.py
```diff
@@ -2040,17 +2064,28 @@ def run_next_if_necessary(**context):
...
-        >> check_if_previously_rejected()
-        >> [preprocessing_group, save_and_complete_workflow_task]
+        >> check_if_previously_rejected_task
+        >> [skip_if_auto_rejected_task, continue_if_auto_rejected_task]
+    )
+
+    (
+        continue_if_auto_rejected_task
+        >> early_save_workflow_task
+        >> run_next_if_necessary()
```
ioannistsanaktsidis: now that you introduce 2 instances of this task, it would be better to create a run_next_if_necessary_early, so as to follow the same patter as the others

ioannistsanaktsidis (changes requested):
few comments

workflows/dags/literature/hep_create_dag.py
DonHaul: ndeed the task before (`stop_if_existing_submission_notify_and_close_task`) does require a correction as it was returning `save_and_complete_workflow` instead of `save_and_complete_workflow_early`

Did it understand this correctly?

workflows/dags/literature/hep_create_dag.py
DonHaul: ditto

PR #4319, opened 2026-06-18
workflows: discard even if 404

DonHaul:
* ref: cern-sis/issues-inspire/issues/1529


DonHaul:
In general this wont happen unless we delete the airflow workflows so really up to us if we should merge this pr or not

ioannistsanaktsidis:
I would say that we can skip this pr completely. It should never be the case in a normal scenario , right ?

ioannistsanaktsidis:
I will close that one as it should never happen. We can think about it in case we face this scenario more often.

PR #4321, opened 2026-06-18
workflows: allow local execution ticket skipping

DonHaul:
* ref: cern-sis/issues-inspire/issues/1530

### DIFF
--- a/workflows/plugins/include/utils/tickets.py
+++ b/workflows/plugins/include/utils/tickets.py
@@ -1,5 +1,10 @@
 import logging
 
+from airflow.sdk import Variable
+from hooks.backoffice.workflow_management_hook import HEP
+from hooks.backoffice.workflow_ticket_management_hook import (
+    BaseWorkflowTicketManagementHook,
+)
 from hooks.inspirehep.inspire_http_hook import (
     LITERATURE_ARXIV_CURATION_FUNCTIONAL_CATEGORY,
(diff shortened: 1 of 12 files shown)


ioannistsanaktsidis (commented):
What about the close_ticket part ?

DonHaul:
Refactor, so the author dags can also be executed locally

workflows/plugins/include/utils/tickets.py
```diff
@@ -84,3 +89,49 @@ def get_functional_category_and_ticket_type_from_publisher(workflow):
...
+        response = inspire_http_hook.create_ticket(
+            functional_category,
+            template_name,
+            subject,
+            email,
+            curation_context,
+        )
+        ticket_id = response.json()["ticket_id"]
+    else:
+        ticket_id = f"local-{ticket_type}"
```
ioannistsanaktsidis: this will produce a line longer than 32 chars, actually it is already 34 for hep_create_publisher_curation_core, meaning a save in the next step will actually raise. See https://github.com/inspirehep/inspirehep/blob/master/backoffice/backoffice/common/models.py#L7 . We should bump the model

ioannistsanaktsidis (changes requested):
What about https://github.com/inspirehep/inspirehep/blob/master/workflows/dags/author/author_update/author_update.py#L73-L87 ? We should guard this part too no ?

ioannistsanaktsidis (approved):
LGTM

PR #4323, opened 2026-06-19
bakcoffice: independent of s3

DonHaul:
* ref: cern-sis/issues-inspire/issues/1498

### DIFF
--- a/backoffice/config/settings/base.py
+++ b/backoffice/config/settings/base.py
@@ -163,6 +163,7 @@ AUTH_PASSWORD_VALIDATORS = [
 MIDDLEWARE = [
     "django_prometheus.middleware.PrometheusBeforeMiddleware",
     "django.middleware.security.SecurityMiddleware",
+    "whitenoise.middleware.WhiteNoiseMiddleware",
     "corsheaders.middleware.CorsMiddleware",
     "django.contrib.sessions.middleware.SessionMiddleware",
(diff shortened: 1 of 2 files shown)


ioannistsanaktsidis (commented):
2 remarks 

Since now [this](https://github.com/inspirehep/inspirehep/blob/master/backoffice/config/settings/base.py#L176) is exercised, we should consider moving it right after the SecurityMiddleware to avoid overhead on the request of the static file. WDYT ? 


Second, we should make sure that we run collectstatic command in every deployment no ? 
`python manage.py collectstatic --noinput` so as to regenerate the static files correctly

DonHaul:
> 2 remarks
> 
> Since now [this](https://github.com/inspirehep/inspirehep/blob/master/backoffice/config/settings/base.py#L176) is exercised, we should consider moving it right after the SecurityMiddleware to avoid overhead on the request of the static file. WDYT ?
> 
> Second, we should make sure that we run collectstatic command in every deployment no ? `python manage.py collectstatic --noinput` so as to regenerate the static files correctly

- moved `whitenoise` middleware right after `SecurityMiddleware`
- regarding the second point, we already have pvc setup, which we would use for this.  so we would only need to run it once, as it was done when we were using s3.  for our usecase i dont think we need to collect on every build as we only use the backed functionality of django

ioannistsanaktsidis (approved):
LGTM !

PR #4324, opened 2026-06-19
ui: replace react-loadable and react-quill

EliseRecejac:
* ref: https://github.com/cern-sis/issues-inspire/issues/1555


ui/src/submissions/__tests__/index.test.jsx
```diff
@@ -27,50 +26,46 @@ describe('Submissions', () => {
...
       route: '/submissions/authors/1',
     });
-    await Loadable.preloadAll();
 
     expect(asFragment()).toMatchSnapshot();
     expect(getByTestId('author-update-submission-page')).toBeInTheDocument();
   });
 
-  it.only('navigates to SubmissionSuccessPage when /submissions/authors/new/success', async () => {
+  it.only('navigates to SubmissionSuccessPage when /submissions/authors/new/success', () => {
```
ioannistsanaktsidis: Not related to your PR, but I just realized that this is actually the only test running on this file 😞 Could we take a look ?

ui/src/App.jsx
```diff
@@ -91,31 +81,33 @@ function App({ userRoles, dispatch, guideModalVisibility }) {
...
-          <Route path={JOURNALS} component={Journals} />
-          <Route path={DATA} component={Data} />
-          <PrivateRoute path={SUBMISSIONS} component={Submissions$} />
-          <Route
-            path={BIBLIOGRAPHY_GENERATOR}
-            component={BibliographyGeneratorPageContainer}
-          />
-          <Route path={ERRORS} component={Errors} />
-        </SafeSwitch>
+        <Suspense fallback={<Loading />}>
```
ioannistsanaktsidis, line 84: With that change there's a subtle behavior change around lazy-route load failures.

With react-loadable, a failed dynamic import re-rendered the loading component (and our Loading.jsx ignores the error prop, so it just showed a spinner), degraded, but the app shell (Header/Footer/nav) stayed mounted. 

With React.lazy, a rejected import() throws during render, and since <Suspense> only catches the pending state, it propagates to the only error boundary we have at the top-level  which replaces the entire app with ErrorAppCrash.

Realistically this can only happen the moment we deploy, which makes it a low priority. Not sure we should act on it now, but worth validating any other potential impact.

ioannistsanaktsidis (changes requested):
few comments

ui/src/App.jsx
EliseRecejac: old way
Capture d’écran 2026-06-24 à 16 10 54

new way
Capture d’écran 2026-06-24 à 16 26 34

I think the new way is better, with a link to go home if necessary

ui/src/submissions/__tests__/index.test.jsx
EliseRecejac: I think it's only something someone forgot, with a bit of update of the snapshots they pass just fine

ui/src/App.jsx
ioannistsanaktsidis: @drjova  @michamos you agree with this "new" behavior ? Very low priority/probability and it should never actually happen.

ioannistsanaktsidis: Discussed IRL with @michamos ! good to go!

ioannistsanaktsidis (approved):

ui/src/App.jsx
michamos: as discussed IRL, I agree :)

michamos: does the `go back home` link on the error page trigger a full page load, so it will eventually work if the user clicks it?

drjova: > @drjova @michamos you agree with this "new" behavior ? Very low priority/probability and it should never actually happen.

lgtm it’s actually the expected behavior

PR #524, opened 2019-10-23
worker: expose prometheus metrics on worker.

pazembrz:
INSPIR-2794

PR #4330, opened 2026-06-22
fix: transition to unittest.mock

gabrielbarroca:
Closes #524

## What changed
- Replaced all `import mock` and `from mock import` with 
  `from unittest import mock` and `from unittest.mock import` 
  across 84 test files
- Added Flask app context fixture in `tests/unit/conftest.py` 
  to fix `RuntimeError: Working outside of application context`
- Removed `UP026` ignore rule from `ruff.toml` since the 
  migration is now complete

## Tests
646 unit tests passing


backend/tests/unit/pidstore/test_providers/test_bai_provider.py
```diff
@@ -3,16 +3,20 @@
...
 import pytest
+from inspirehep.factory import create_app
 from inspirehep.pidstore.errors import PIDAlreadyExistsError
 from inspirehep.pidstore.providers.bai import InspireBAIProvider
 from invenio_pidstore.models import PersistentIdentifier, PIDStatus
 from sqlalchemy.exc import IntegrityError
 
 
 def test_bai_create_fails_after_retrying_when_bai_changes():
+    app = create_app(TESTING=True, SERVER_NAME="localhost:5000")
```
drjova: instead of creating an app everytime, we can create a fixture that wraps the tests that we need :)

drjova (changes requested):
Thank you, @gabrielbarroca! Just a minor comment for the future to avoid making the same change in multiple files where possible.

drjova:
@DonHaul @ioannistsanaktsidis any objections?

workflows/tests/test_workflow_utils.py
```diff
@@ -675,9 +675,18 @@ def test_normalize_collaborations(self):
...
+        with mock.patch(
+            "include.utils.workflows.normalize_collaborations"
+        ) as mock_normalize:
+            mock_normalize.return_value = (
+                [{"legacy_name": "LATTICE-ETM"}],
+                [{"value": "ETM", "record": "some_record"}],
+            )
+            accelerator_experiments, normalized_collaborations = (
+                workflows.normalize_collaborations(record)
+            )
```
ioannistsanaktsidis, line 689: why this part was needed ?

ioannistsanaktsidis (commented):
just the comment

backend/tests/integration/snow/test_snow.py
```diff
@@ -5,7 +5,9 @@
 # the terms of the MIT License; see LICENSE file for more details.
 
 
-import mock
+from unittest import mock
+from unittest.mock import patch
```
DonHaul, line 8: why not, just import patch ? mock is only being used for `mock.patch`

backend/tests/integration/hal/test_core_sword.py
```diff
@@ -20,8 +20,9 @@
 # or submit itself to any jurisdiction.
 
 import contextlib
+from unittest import mock
```
DonHaul: ditto

ioannistsanaktsidis:
@gabrielbarroca please rebase against latest master and re-push, after that the ci should be ok

ioannistsanaktsidis (approved):
LGTM! @drjova if you can approve as you requested changes before!

workflows/tests/test_workflow_utils.py
gabrielbarroca: The cassette kept showing up as modified after each test run, which suggested the recorded request was no longer matching exactly and VCR was trying to re-record it. Rather than spending time tracking down the exact source of the mismatch, I mocked normalize_collaborations directly to remove the dependency on that fragile recording. If you'd prefer to keep the VCR-based test, I'm happy to investigate the mismatch and restore that approach.

ioannistsanaktsidis: should be fine with the patch

drjova (approved):

PR #4336, opened 2026-06-25
backoffice: fine-grained source facet for desy

EliseRecejac:
* ref: https://github.com/cern-sis/issues-inspire/issues/1506

### DIFF
--- a/backoffice/backoffice/hep/api/views.py
+++ b/backoffice/backoffice/hep/api/views.py
@@ -403,7 +403,7 @@ class HepWorkflowDocumentView(BaseDocumentViewSet):
         "subject": "data.inspire_categories.term",
         "journal": "data.publication_info.journal_title.raw",
         "decision": "relevance_prediction.decision",
-        "source": "data.acquisition_source.source",
+        "source": "data.source",
         "data.arxiv_eprints.value": "data.arxiv_eprints.value",
         "data.dois.value": "data.dois.value",
     }
@@ -462,7 +462,7 @@ class HepWorkflowDocumentView(BaseDocumentViewSet):
             "enabled": True,
         },
         "source": {
-            "field": "data.acquisition_source.source",
+            "field": "data.source",
             "facet": TermsFacet,
             "options": {
                 "size": 20,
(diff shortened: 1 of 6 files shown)


EliseRecejac:
Capture d’écran 2026-06-25 à 10 24 45

backoffice/backoffice/hep/documents.py
```diff
@@ -226,6 +227,12 @@ def prepare_data(self, instance):
         data = instance.data or {}
         data.setdefault("public_notes", [])
         data.setdefault("arxiv_eprints", [])
+        acquisition_source = (data.get("acquisition_source") or {}).get("source")
+        if acquisition_source is not None and acquisition_source.lower() == "desy":
+            titles = data.get("titles") or []
+            data["source"] = f"desy/{titles[0].get('source')}" if titles else "desy"
+        else:
+            data["source"] = acquisition_source
```
ioannistsanaktsidis: We can make that a static method instead of computing it here. Sth like

`def _compute_source ...`


also we can invert the logic so as to return early -> 

```python
 acquisition_source = (data.get("acquisition_source") or {}).get("source")
 if acquisition_source is None or acquisition_source.lower() != "desy":
        return acquisition_source
.....
```

backoffice/backoffice/hep/api/views.py
```diff
@@ -462,7 +462,7 @@ def __init__(self, *args, **kwargs):
             "enabled": True,
         },
         "source": {
-            "field": "data.acquisition_source.source",
+            "field": "data.source",
```
ioannistsanaktsidis, line 465: Now that we will have a facet-per source, could you validate with Micha how many of these facets are expected  ? As we are capping the facets to 20 items per facet, we might need to check that

ioannistsanaktsidis (changes requested):
Some styling comments + one confirmation with Product, otherwise it looks good!

backoffice/backoffice/hep/api/views.py
EliseRecejac: discussed with @michamos, total more than 20 but not in a single state where curators need to act

ioannistsanaktsidis (approved):
All good!

PR #4338, opened 2026-06-25
gh-actions: bump actions

ioannistsanaktsidis:
* ref https://github.com/cern-sis/issues-inspire/issues/1478

### DIFF
--- a/.github/actions/combine-durations/action.yml
+++ b/.github/actions/combine-durations/action.yml
@@ -5,8 +5,8 @@ inputs:
   durations-path:
     description: The path to the durations file (must match `--durations-path` arg to pytest)
     required: true
-  split-prefix:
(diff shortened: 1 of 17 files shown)


ioannistsanaktsidis:
https://github.com/inspirehep/inspirehep/actions/runs/28176712379 successful run on action-dev branch

PascalEgn:
@ioannistsanaktsidis I create the `docker-build@v7.1.0` release which disables the additional artifact upload per default, so just bumping the version should be enough incase you wanna drop the artifact changes in the PR

ioannistsanaktsidis:
@PascalEgn  I updated the action to 7.1.0. I would prefer to keep the change of the artifacts since I think it is a cleaner  approach

PascalEgn (approved):

PR #4339, opened 2026-06-26
ui: replace react-vis by recharts in CitationPerYearGraph

EliseRecejac:
* ref: https://github.com/cern-sis/issues-inspire/issues/1559

### DIFF
--- a/ui/src/common/components/CitationsByYearGraph/CitationsByYearGraph.tsx
+++ b/ui/src/common/components/CitationsByYearGraph/CitationsByYearGraph.tsx
@@ -1,20 +1,15 @@
-import React, { useState, useEffect } from 'react';
-import {
-  LineMarkSeries,
-  MarkSeries,
-  LineSeries,
-  FlexibleWidthXYPlot,
-  YAxis,
-  XAxis,
-  Hint,
-  MarkSeriesPoint,
-  LineMarkSeriesPoint,
-  RVNearestXEventHandler,
-} from 'react-vis';
-import 'react-vis/dist/style.css';
-import maxBy from 'lodash/maxBy';
+import React, { useMemo } from 'react';
 import { Map } from 'immutable';
 
+import {
+  Line,
+  LineChart,
+  ResponsiveContainer,
+  Tooltip,
+  TooltipContentProps,
+  XAxis,
+  YAxis,
+} from 'recharts';
 import styleVariables from '../../../styleVariables';
 import LoadingOrChildren from '../LoadingOrChildren';
 import ErrorAlertOrChildren from '../ErrorAlertOrChildren';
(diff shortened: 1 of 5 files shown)


EliseRecejac:
Image 25-06-2026 à 17 10

drjova:
> Image 25-06-2026 à 17 10

thank you! How does it look like with 1000+ citations? Is it displayed like 1K?

michamos:
is there still detailed info about a data point displayed on hover? 
image

EliseRecejac:
Capture d’écran 2026-06-26 à 17 32 40
@drjova, @michamos everything as expected (normally)

michamos:
looks great!

ui/src/literature/containers/LiteratureSearchContainer.jsx
```diff
@@ -184,7 +193,9 @@ function LiteratureSearch({
             {enableCitationSummary && isCitationSummaryVisible && (
               <Row className="mt2">
                 <Col span={24}>
-                  <CitationSummaryBox namespace={namespace} />
+                  <Suspense fallback={null}>
+                    <CitationSummaryBox namespace={namespace} />
+                  </Suspense>
```
drjova: do we need this because of the library change, or in general for graceful handling? Can we push it down to CitationSummaryBox?

EliseRecejac: it was needed at some point because of the library change (to avoid bundle size tests to fail : recharts is bigger than react-vis) but with the other package updates it seems useless now, I revert it

ui/src/common/components/CitationsByYearGraph/CitationsByYearGraph.tsx
```diff
@@ -28,18 +23,53 @@ import EmptyOrChildren from '../EmptyOrChildren';
...
+  if (active && payload && payload.length) {
+    const firstPayload = payload[0];
+    if (firstPayload == null) {
+      return null;
+    }
+    const entry = firstPayload.payload;
+    const citations = entry.pastYearsY ?? entry.currentYearY;
+    return (
+      <div
+        style={{
```
EliseRecejac, line 54: I made it more reusable in this PR [here](https://github.com/inspirehep/inspirehep/pull/4340/changes#diff-ea33cb3ff268632a686510de6a78ecd47832e67c9ae21d47d314c16b8164ef50)

ui/src/common/components/CitationsByYearGraph/CitationsByYearGraph.tsx
```diff
@@ -115,91 +120,66 @@ function CitationsByYearGraph({
...
-  function renderCitationsGraph() {
-    const currentYear = new Date().getFullYear();
-    const yearOfLastCitation = seriesData?.slice(-1)[0]?.x;
-    const currentYearSeries = seriesData.slice(
-      seriesData.length - 2,
-      seriesData.length
-    );
-    const pastYearsSeries = seriesData.slice(0, seriesData.length - 1);
-
-    if (yearOfLastCitation === currentYear) {
```
drjova, line 146: @EliseRecejac where do we handle this case now? I see that with the new library we always have this case and not the LineMarkSeries equivalent. Have we moved this logic somewhere else?

EliseRecejac: Capture d’écran 2026-06-30 à 11 09 27
it's working just fine, since we use only one dataset for the two lines, if there is no defined currentYear data the dashed line will just never be displayed

drjova (approved):
lgtm

PR #4345, opened 2026-07-01
ui: fix label display in citation summary graph

EliseRecejac:
* ref: https://github.com/cern-sis/issues-inspire/issues/1568


EliseRecejac:
Capture d’écran 2026-07-01 à 13 08 54

drjova:
@EliseRecejac can we double check that doesn't break the other browsers on mobile too?

EliseRecejac:
> @EliseRecejac can we double check that doesn't break the other browsers on mobile too?

I already checked :
Chrome
Capture d’écran 2026-07-01 à 13 34 51
Firefox
Capture d’écran 2026-07-01 à 13 35 41


I also checked the desktop view, everything's fine

drjova (approved):
LGTM

PR #4346, opened 2026-07-01
backoffice: oncommit indexing

DonHaul:
* ref: cern-sis/issues-inspire/issues/1560

### DIFF
--- a/backoffice/backoffice/common/signals.py
+++ b/backoffice/backoffice/common/signals.py
@@ -1,0 +1,38 @@
+from copy import copy
+
+from django_opensearch_dsl.signals import RealTimeSignalProcessor
+from django_opensearch_dsl.registries import registry
+
+
+from functools import partial
+from typing import Any
+from django.db import models, transaction
+
+
+def update_registry_after_commit(instance):
+    registry.update(instance)
+    registry.update_related(instance)
+
+
+def delete_from_registry_after_commit(instance):
+    registry.delete(instance, raise_on_error=False)
+    registry.delete_related(instance, raise_on_error=False)
+
+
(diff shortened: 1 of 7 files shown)


DonHaul:
tests were initially failing due to data mutation issue:
https://github.com/inspirehep/inspirehep/actions/runs/28514280628/job/84522637503

drjova (approved):
LGTM 👍

backoffice/backoffice/common/signals.py
```diff
@@ -0,0 +1,38 @@
...
+
+
+def update_registry_after_commit(instance):
+    registry.update(instance)
+    registry.update_related(instance)
+
+
+def delete_from_registry_after_commit(instance):
+    registry.delete(instance)
+    registry.delete_related(instance)
```
ioannistsanaktsidis: delete_from_registry_after_commit calls registry.delete(instance) / registry.delete_related(instance) without raise_on_error=False. 

The base RealTimeSignalProcessor we are overriding passed that flag initially.  Dropping it means the kwarg defaults to True all the way down at opensearch's streaming_bulk. 

So deleting a workflow whose OS doc is missing for whatever reason returns a 404, which bulk() treats as a failure and raises BulkIndexError. Can we check if  we should add raise_on_error=False to both calls to match the base behavior or it was intended added like this ?

ioannistsanaktsidis (commented):
Just a comment to check

backoffice/backoffice/common/signals.py
DonHaul: good catch, I've added back `raise_on_error=False` so we know that the behaviour stays the same as before

ioannistsanaktsidis (approved):
LGTM

PR #4355, opened 2026-07-06
ui: remove antd/rc-select

EliseRecejac:
* ref: https://github.com/cern-sis/issues-inspire/issues/1561


EliseRecejac:
second commit only, the first will be removed when the respective PR will be merged

PR #4368, opened 2026-07-09
backoffice: improve batch request time

DonHaul:
* ref: cern-sis/issues-inspire/issues/1582


DonHaul:
it seems that this cannot be improved, as the `opensearch` signals would not be triggered if we used something like
```
HepWorkflow.objects.filter(pk__in=data["ids"]) ).update(status=HepStatusChoices.RUNNING)
```

PR #4377, opened 2026-07-14
backoffice: reduce response size for literature search

EliseRecejac:
* ref: https://github.com/cern-sis/issues-inspire/issues/1582


backoffice/backoffice/hep/documents.py
```diff
@@ -249,7 +249,11 @@ def prepare_relevance_prediction(self, instance):
         return self._prepare_optional_json_object(instance.relevance_prediction)
 
     def prepare_reference_count(self, instance):
-        return self._prepare_optional_json_object(instance.reference_count)
+        reference_count = self._prepare_optional_json_object(instance.reference_count)
+        if reference_count is None:
+            return None
+        total = len((instance.data or {}).get("references") or [])
+        return {**reference_count, "total": total}
```
EliseRecejac, line 256: needs a reindexation in inspirebeta

ui/src/backoffice/literature/components/LiteratureReferenceCount.jsx
```diff
@@ -1,21 +1,25 @@
 import React from 'react';
 
-const LiteratureReferenceCount = ({ referenceCount, totalReferences }) => {
+const LiteratureReferenceCount = ({
+  referenceCount,
+  totalReferences = undefined,
```
ioannistsanaktsidis: instead of keeping the default like that, we can convert this component to tsx, wdyt ?

backoffice/backoffice/hep/tests/test_documents.py
```diff
@@ -158,3 +155,33 @@ def test_prepare_optional_json_fields_preserve_non_empty_values(
     result = getattr(document, f"prepare_{field_name}")(instance)
 
     assert result == value
+
+
+def test_prepare_reference_count_returns_none_when_missing(document):
```
ioannistsanaktsidis: can we parametrize this one so as we cover the {} case as it was covered before ?

ioannistsanaktsidis (changes requested):
small changes

ioannistsanaktsidis (approved):
LGTM! As you mentioned, you will need to run a reindex of the docs in a backoffice dev worker in argo for the changes to appear 

`python manage.py opensearch document index -i inspire-backoffice-backend-dev-hep-workflows
`

PR #4378, opened 2026-07-14
backoffice: moving batchresolve into view

DonHaul:
We had initially moved it out of the view in order to take advantages of the different atomicity settings. 
Since we moved the airflow triggering functionality into celery, there's no advantage to having it isolated it. 
so in order to simplify the code lets move it back to its original view


ioannistsanaktsidis (approved):
Looks good! I leave it up to you if you would like to merge and deploy before the release!

DonHaul:
okay i deploy it tomorrow then

PR #4384, opened 2026-07-16
ui: update eslint config

EliseRecejac:
* ref: https://github.com/cern-sis/issues-inspire/issues/1557

What I did:
- move eslint config from eslintrc to flat config (standard with eslint 9)
- remove deprecated airbnb config and replace it with recommended from plugins + add the other rules
- align ts/tsx rules with js/jsx (more restrictive) + fix corresponding errors (lot of modified files but most of them is to remove the now useless react import, or fixing the arrow no body rules, or the defaultProps in typescript, for the last one no new errors from tsc)


EliseRecejac:
2nd commit only, first will disappear on rebase after merging https://github.com/inspirehep/inspirehep/pull/4354

ui/eslint.config.mjs
```diff
@@ -0,0 +1,535 @@
...
+  'no-lone-blocks': 'error',
+  'no-lonely-if': 'error',
+  'no-loop-func': 'error',
+  'no-multi-assign': 'error',
+  'no-multi-str': 'error',
+  'no-nested-ternary': 'error',
+  'no-new': 'error',
+  'no-new-func': 'error',
+  'no-new-object': 'error',
+  'no-new-symbol': 'error',
```
ioannistsanaktsidis: Reviewing with claude gave me that comment 

drop the 8 deprecated rules. no-new-symbol and jsx-a11y/label-has-for are pure dead weight (replacements already enabled via extends); the other six (no-new-object, no-return-await, global-require, lines-around-directive, lines-between-class-members, spaced-comment) will warn on ESLint 10.


could we check it ? Otherwise the rest looks good

ioannistsanaktsidis (commented):
Just the comment to check. Otherwise I don;t see an issue.

ioannistsanaktsidis (approved):
LGTM 🏅

PR #4385, opened 2026-07-16
ui: remove deprecated default props in functional components

EliseRecejac:
* ref: https://github.com/cern-sis/issues-inspire/issues/1571

### DIFF
--- a/ui/src/common/components/AddressList.jsx
+++ b/ui/src/common/components/AddressList.jsx
@@ -24,8 +24,4 @@ AddressList.propTypes = {
   addresses: PropTypes.instanceOf(List),
 };
 
-AddressList.defaultProps = {
-  addresses: null,
-};
-
 export default AddressList;
--- a/ui/src/authors/components/AssignAction.tsx
+++ b/ui/src/authors/components/AssignAction.tsx
@@ -13,7 +13,7 @@ function AssignAction({
   onAssign,
   onUnassign,
   disabled,
-  numberOfSelected,
+  numberOfSelected = 1,
 }: {
   onAssignToAnotherAuthor: Function;
   onAssign: Function;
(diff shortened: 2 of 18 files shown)


EliseRecejac:
3rd commit only, the two others are different PRs

ui/src/common/components/AddressList.jsx
```diff
@@ -24,8 +24,4 @@ AddressList.propTypes = {
   addresses: PropTypes.instanceOf(List),
 };
 
-AddressList.defaultProps = {
-  addresses: null,
```
ioannistsanaktsidis, line 28: Do you prefer to remove the null or keep it for now so as to keep parity ? Same in the other components changed here InstitutionAddressList, SupervisorList, RangeAggregation. WDYT ?

ioannistsanaktsidis (commented):
Not related to this PR but since you touched the AssignAction component, I saw that 
https://github.com/inspirehep/inspirehep/blob/master/ui/src/authors/containers/AssignAllActionContainer.tsx#L15 is passing a prop that does not exist on this component. Lets create an issue for that .

ui/src/common/components/AddressList.jsx
EliseRecejac: In each case null is handled the same way as undefined so I don't really see the point of defaulting them to null

EliseRecejac:
> Not related to this PR but since you touched the AssignAction component, I saw that https://github.com/inspirehep/inspirehep/blob/master/ui/src/authors/containers/AssignAllActionContainer.tsx#L15 is passing a prop that does not exist on this component. Lets create an issue for that .

since it's a pretty small thing I will handle it in this PR it's simpler

ioannistsanaktsidis (approved):
LGTM, just the issue creation for the tooltip that it is not related to this PR

ioannistsanaktsidis (approved):
LGTM

PR #4387, opened 2026-07-16
workflows: refresh docker compose

DonHaul:
* ref: cern-sis/issues-inspire/issues/1512

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -19,7 +19,7 @@ start-inspirehep:
 
 start-airflow:
 	echo -e "\033[0;32m Starting Airflow. \033[0m"	
-	docker compose up -d airflow-init airflow-worker airflow-api-server airflow-dag-processor airflow-triggerer airflow-scheduler
+	docker compose up -d airflow-init airflow-worker airflow-apiserver airflow-dag-processor airflow-triggerer airflow-scheduler
 	echo -e "\033[0;32m Airflow Started. \033[0m"
 
 start-backoffice:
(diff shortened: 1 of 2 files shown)


Makefile
ioannistsanaktsidis, line 27: not sure this will work with the new auth introduced

docker-compose.yml
```diff
@@ -62,47 +62,51 @@
...
     # Use simple http server on scheduler for health checks
     # See https://airflow.apache.org/docs/apache-airflow/stable/administration-and-deployment/logging-monitoring/check-health.html#scheduler-health-check-server
     # yamllint enable rule:line-length
-    AIRFLOW__SCHEDULER__ENABLE_HEALTH_CHECK: "true"
+    AIRFLOW__SCHEDULER__ENABLE_HEALTH_CHECK: 'true'
     # WARNING: Use _PIP_ADDITIONAL_REQUIREMENTS option ONLY for a quick checks
     # for other purpose (development, test and especially production usage) build/extend Airflow image.
     _PIP_ADDITIONAL_REQUIREMENTS: ${_PIP_ADDITIONAL_REQUIREMENTS:-}
+    # The following line can be used to set a custom config file, stored in the local config folder
+    AIRFLOW_CONFIG: '/opt/airflow/config/airflow.cfg'
```
ioannistsanaktsidis, line 99: This needs to be git ignored

docker-compose.yml
```diff
@@ -62,47 +62,51 @@
...
+  # In order to add custom dependencies or upgrade provider distributions you can use your extended image.
   # Comment the image line, place your Dockerfile in the directory where you placed the docker-compose.yaml
-  # and uncomment the "build" line below, Then run `docker compose build` to build the images.
+  # and uncomment the "build" line below, Then run `docker-compose build` to build the images.
   build:
     context: workflows
     dockerfile: Dockerfile
     target: test
-  environment: &airflow-common-env
-    AIRFLOW__API__BASE_URL: "http://airflow-api-server:8080"
```
ioannistsanaktsidis, line 74: is everything pointing correctly now that this is dropped ? because if it falls back to `localhost:8080` , this is the hep service

ioannistsanaktsidis (changes requested):
There are parts of the app that still point to airflow-api-server, these ones should be checked. E.g. https://github.com/inspirehep/inspirehep/blob/master/backoffice/.envs/local/.django#L22

ioannistsanaktsidis (approved):
LGTM

DonHaul:
removed  from `docker-compose` - `airflow-init`

```
        chown -R "${AIRFLOW_UID:-50000}:0" /opt/airflow/
```
as it was only added for some airflow internal tests
https://github.com/apache/airflow/commit/67ce62214dcba0a481faf22be31b8922aa5a73c4

PR #4388, opened 2026-07-20
ui: replace quill by tiptap

EliseRecejac:
* ref: https://github.com/cern-sis/issues-inspire/issues/1576


EliseRecejac:
3rd commit only, the two others are different PRs

EliseRecejac:
For this one we are back on the default behaviour: each block is a `<p>` and not a `<div>` so there is a 1rem bottom margin for each of them
I think it's a standard way of displaying things so it shouldn't be much of an issue, but to be discussed

PR #4399, opened 2026-07-23
ui: upgrade react-router-dom

EliseRecejac:
* ref: https://github.com/cern-sis/issues-inspire/issues/1570


ui/src/literature/index.tsx
```diff
@@ -1,31 +1,28 @@
...
-      <SafeSwitch>
-        <Route exact path={LITERATURE} component={SearchPageContainer} />
+      <RoutesWithFallback>
+        <Route index element={<SearchPageContainer />} />
+        <Route path=":id" element={<DetailPageContainer />} />
         <Route
-          exact
-          path={`${LITERATURE}/:id`}
-          component={DetailPageContainer}
+          path=":id/diff/:diff"
```
EliseRecejac, line 18: it looks like this one is a dead one, @ioannistsanaktsidis do you confirm ?

EliseRecejac: @michamos do you know ?

michamos: we don’t use this any longer (it was a page to check reference edits by users) so it’s ok to remove

ioannistsanaktsidis: Thanks @michamos , we will proceed remove the  deprecated code !

ioannistsanaktsidis (approved):
LGTM

PR #4403, opened 2026-07-28
backoffice(ui): replace submission source by submission email in detail page

EliseRecejac:
* ref: https://github.com/cern-sis/issues-inspire/issues/1604


EliseRecejac:
Capture d’écran 2026-07-28 à 09 09 32

ui/src/backoffice/literature/containers/LiteratureDetailPageContainer.tsx
```diff
@@ -296,7 +295,7 @@ const LiteratureDetailPageContainer = ({
                     >
                       <>
                         Harvested on <b>{acquisitionSourceDateTime}</b> from{' '}
-                        <b>{acquisitionSourceSource}</b> using{' '}
+                        <b>{acquisitionSourceEmail}</b> using{' '}
```
ioannistsanaktsidis: This will actually only use the email. What we want is that when we are on submission and there is an email to show that, in every other case it should be as before

ioannistsanaktsidis (changes requested):

ioannistsanaktsidis (approved):

PR #4409, opened 2026-07-28
backend: curation ticket fix

DonHaul:
* ref: cern-sis/issues-inspire/issues/1608

### DIFF
--- a/backend/inspirehep/snow/templates/snow/curation_core.html
+++ b/backend/inspirehep/snow/templates/snow/curation_core.html
@@ -19,7 +19,7 @@
 New submission from {{ email }}
 
 Curate record here:
-Record editor: https://{{ server_name }}/editor/record/literature/{{ recid }}
+Record editor: {{ server_name }}/editor/record/literature/{{ recid }}
 {% if link_to_pdf %}
 Link to pdf: {{ link_to_pdf|safe }}
 {% endif %}


backend/inspirehep/snow/templates/snow/curation_core.html
```diff
@@ -19,7 +19,7 @@
 New submission from {{ email }}
 
 Curate record here:
-Record editor: https://{{ server_name }}/editor/record/literature/{{ recid }}
+Record editor: {{ server_name }}/editor/record/literature/{{ recid }}
```
ioannistsanaktsidis, line 22: any ideas why it was working before ? meaning in next?

ioannistsanaktsidis (approved):
Just to validate that this https://github.com/inspirehep/inspirehep/blob/master/backend/inspirehep/snow/templates/snow/curation_jlab.html#L21 is not used (so no change) , and same for https://github.com/inspirehep/inspirehep/blob/master/backend/inspirehep/snow/templates/snow/conflicts-ticket-template.html 

these 2 templates most probably are dead code

backend/inspirehep/snow/templates/snow/curation_core.html
ioannistsanaktsidis: ok i see, we are now creating the url using https://github.com/inspirehep/inspirehep/blob/master/workflows/plugins/include/utils/workflows.py#L398 , so no need of the prefix

DonHaul:
for the jlab i remember micha mentioning they are no longer required do you confirm @michamos   - even though they are in next right now 
https://github.com/inspirehep/inspire-next/blob/master/inspirehep/modules/workflows/workflows/article.py#L366-L374

for the conflicts-ticket - i see it being used on the merge articles on next. we might have missed this one
https://github.com/inspirehep/inspire-next/blob/master/inspirehep/modules/workflows/workflows/article.py#L441-L446

ioannistsanaktsidis:
I remember discussing that it is not need to create a ticket when merge conflict arise. Please correct me if I am wrong

michamos:
you’re right @ioannistsanaktsidis

PR #4411, opened 2026-07-29
curation: match affiliations by ROR in affiliations_identifiers

ANNNA106:
Summary:
Extends `normalize_affiliations` to match affiliations via ROR (`affiliations_identifiers`) before falling back to the existing raw-text matching.

- Only non-obsolete institutions are matched (`ICN` excludes `"obsolete"`).
- 1:1 case (equal counts of `raw_affiliations` and ROR ids): prefer ROR match; fall back to raw-text matching per entry on miss.
- ROR-only case (no `raw_affiliations`): match on ROR only, no fallback.
- Everything else (mismatched counts): ROR matching skipped, original raw-text-only behavior unchanged.

Implementation:

- `curation/utils.py`: new `find_institution_by_ror(ror)` — queries `InstitutionsSearch` for a non-obsolete institution with a matching ROR.
- `curation/api.py`: `normalize_affiliations` branches into the three cases above, using cached lookups for both ROR and raw-text matching.

Testing:

Added integration tests for all branches (ROR match, obsolete exclusion, no-match, 1:1 fallback, ROR-only, mismatched counts). Full `test_api.py` suite passes (25/25), no regressions.

### DIFF
--- a/backend/inspirehep/curation/api.py
+++ b/backend/inspirehep/curation/api.py
@@ -13,7 +13,7 @@ from inspirehep.curation.utils import (
     create_accelerator_experiment_from_collaboration_match,
     enhance_collaboration_data_with_collaboration_match,
     find_collaboration_in_multisearch_response,
-    find_institution_by_ror,  # <-- new
+    find_institution_by_ror,
     find_unambiguous_affiliation,
     match_lit_author_affiliation,
 )
--- a/backend/inspirehep/curation/api.py
+++ b/backend/inspirehep/curation/api.py
@@ -13,6 +13,7 @@ from inspirehep.curation.utils import (
     create_accelerator_experiment_from_collaboration_match,
     enhance_collaboration_data_with_collaboration_match,
     find_collaboration_in_multisearch_response,
+    find_institution_by_ror,  # <-- new
     find_unambiguous_affiliation,
     match_lit_author_affiliation,
 )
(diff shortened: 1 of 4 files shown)


backend/inspirehep/curation/api.py
```diff
@@ -13,6 +13,7 @@
     create_accelerator_experiment_from_collaboration_match,
     enhance_collaboration_data_with_collaboration_match,
     find_collaboration_in_multisearch_response,
+    find_institution_by_ror,  # <-- new
```
michamos:
```suggestion
    find_institution_by_ror,
```

backend/inspirehep/curation/api.py
```diff
@@ -78,28 +103,53 @@ def normalize_affiliations(authors, workflow_id=None, **kwargs):
...
-                matched_affiliations[raw_aff] = matched_author_affiliations
-                author_affiliations.extend(matched_author_affiliations)
-            else:
-                ambiguous_affiliations.append(raw_aff)
+        aff_ids = get_value(author, "affiliations_identifiers", [])
+        rors = [aff_id["value"] for aff_id in aff_ids if aff_id.get("schema") == "ROR"]
+
+        if raw_affs and len(raw_affs) == len(rors):
+            # 1:1 correspondence: prefer ROR, fall back to raw text per-index
+            for raw_aff, ror in zip(raw_affs, rors, strict=False):
```
michamos, line 122: `strict=False` is the default, and you have just checked that they have the same length, so that flag isn’t really useful

backend/inspirehep/curation/api.py
```diff
@@ -66,9 +67,33 @@ def normalize_collaborations(collaborations, wf_id=None):
...
+        cache[raw_aff] = matched_aff
+    return matched_aff
+
+
+def _match_by_ror(ror, cache):
+    if ror in cache:
+        return cache[ror]
+    institution = find_institution_by_ror(ror)
+    matched_aff = (
+        [{"value": institution["legacy_ICN"], "record": institution["self"]}]
```
michamos, line 85: why have a list with one element here, instead of the element itself? are there cases where this would be a list with several items? raw affiliations can be one to many with normalized affiliations, but RORs are one to one.

backend/inspirehep/curation/api.py
```diff
@@ -66,9 +67,33 @@ def normalize_collaborations(collaborations, wf_id=None):
     }
 
 
+def _match_by_raw_affiliation(raw_aff, cache, workflow_id):
```
michamos, line 70: (not for @ANNNA106) I was suprised a `workflow_id` is needed here, turns out it’s for additional logging in https://github.com/inspirehep/inspirehep/blob/f882c5bc6f663e4cfb52477e5bf038bcfba600b4/backend/inspirehep/curation/utils.py#L267-L273. That should probably be refactored to use proper structured logging, including [binding the worflow id to the context ](https://www.structlog.org/en/stable/api.html#structlog.contextvars.bind_contextvars) wherever it gets received instead of drilling down manually to the logger call.

backend/inspirehep/curation/utils.py
```diff
@@ -303,3 +303,32 @@ def assign_institution(matched_affiliation):
...
+    )
+    result = (
+        InstitutionsSearch()
+        .query(query)
+        .source(["self", "legacy_ICN"])
+        .params(size=1)
+        .execute()
+    )
+    if result:
+        return result.hits[0].to_dict()
```
michamos, line 334: please check for more than one result here. If there are several, don’t return anything and log that the ROR match is ambiguous.

michamos (commented):
The logic seems fine except for the case of ambiguous ROR in institutions that should be handled. I’ve made a few smaller comments, and I will let @drjova and @ioannistsanaktsidis comment further on the code.

PR #4415, opened 2026-07-29
workflows: set url if already exists

DonHaul:
* ref: cern-sis/issues-inspire/issues/1614


ioannistsanaktsidis (approved):
LGTM!  Cannot really see any problem

PR #4418, opened 2026-07-30
backoffice: remove signup feature

EliseRecejac:
* ref: https://github.com/cern-sis/issues-inspire/issues/1623


EliseRecejac:
Capture d’écran 2026-07-30 à 16 48 26
Capture d’écran 2026-07-30 à 16 48 40

drjova:
hey @EliseRecejac, thank you for the PR, I would suggest removing this https://github.com/inspirehep/inspirehep/blob/master/backoffice/config/urls.py#L24 completely and this https://github.com/inspirehep/inspirehep/blob/f882c5bc6f663e4cfb52477e5bf038bcfba600b4/backoffice/config/urls.py#L15 and I will move this `"DJANGO_ACCOUNT_ALLOW_REGISTRATION` to k8s config

drjova:
Also, let's add to the settings something like that, so we can configure it with envs

```
SOCIALACCOUNT_ONLY = env.bool("DJANGO_SOCIALACCOUNT_ONLY",  False)
```

like we have here https://github.com/inspirehep/inspirehep/blob/f882c5bc6f663e4cfb52477e5bf038bcfba600b4/backoffice/config/settings/base.py#L331

drjova:
@EliseRecejac with `SOCIALACCOUNT_ONLY` shouldn't we have hidden login form?

EliseRecejac:
> @EliseRecejac with `SOCIALACCOUNT_ONLY` shouldn't we have hidden login form?

where ?

drjova:
> > @EliseRecejac with `SOCIALACCOUNT_ONLY` shouldn't we have hidden login form?
> 
> where ?

from here 
image

workflows/dags/literature/desy_harvest.py
drjova, line 79: why do we need to change that?

drjova (changes requested):

workflows/dags/literature/desy_harvest.py
EliseRecejac: it's a rebase issue I'm on it

drjova (approved):

PR #4439, opened 2026-08-03
worfklows: add fermilab report step

ioannistsanaktsidis:
* ref https://github.com/cern-sis/issues-inspire/issues/1630


workflows/plugins/include/utils/constants.py
```diff
@@ -53,6 +53,8 @@
 ]
 COMPLETED_STATUSES = [STATUS_COMPLETED]
 
+FERMILAB_COLLECTION = "Fermilab"
```
drjova: should it be in title case?

drjova:
```suggestion
FERMILAB_COLLECTION_NAME = "Fermilab"
```

drjova (changes requested):
few comments

ioannistsanaktsidis:
@drjova comments addressed

drjova (approved):

PR #4447, opened 2026-08-04
backoffice: skip signal handlers for unregistered models

drjova:
backoffice: skip signal handlers for unregistered models

* ref: https://github.com/cern-sis/issues-inspire/issues/1636


ioannistsanaktsidis (approved):
All good from my side! Waiting co-pilots review to finish and then we can merge! Thanks @drjova !

PR #4448, opened 2026-08-04
ai: add streaming response and loading texts

PascalEgn:


backend/inspirehep/search/views.py
```diff
@@ -56,6 +68,40 @@ def assistant_search():
...
+        return jsonify({"message": AI_SEARCH_ERROR_MESSAGE}), 502
+
+    def generate():
+        yield _server_sent_event({"type": "status", "stage": "connecting"})
+        try:
+            for event in events:
+                if event["type"] == "done":
+                    event = {**event, "records_api_url": records_api_url}
+                yield _server_sent_event(event)
+        except AiSearchError:
```
drjova: If it fails for another reason, it will explode. We can catch everything else too, to handle it more gracefully

ui/src/actions/aiSearch.ts
```diff
@@ -122,20 +128,77 @@ export function fetchAiSearchResults(
...
+            },
+            controller.signal
+          );
+        } finally {
+          if (activeStreamsByNamespace.get(namespace) === controller) {
+            activeStreamsByNamespace.delete(namespace);
+          }
+        }
+
+        if (streamError) {
```
drjova, line 203: what about the connection drops before done arrives? I would maybe treat this as an error too

drjova (approved):

PR #4449, opened 2026-08-04
workflows: submissions do not get auto-rejected if old arxives

DonHaul:
* ref: cern-sis/issues-inspire/issues/1632


DonHaul:
as discussed with Micha, if a previous submissions was rejected, the current submission should now be autorejected. - Default behaviour no changes required

ioannistsanaktsidis (approved):
LGTM!

PR #4452, opened 2026-08-04
workflows: classify_paper called with_author_keywords=true

DonHaul:
* ref: cern-sis/issues-inspire/issues/1639


DonHaul:
not required at all

PR #4457, opened 2026-08-05
workflows: guard if no notify submission ticket exists

DonHaul:
* ref: cern-sis/issues-inspire/issues/1612


DonHaul:
idea is marinating. to be re-checked later

DonHaul:
superceeded by https://github.com/inspirehep/inspirehep/pull/4486

PR #4463, opened 2026-08-06
workflows: only allow restart current of workflows in error

DonHaul:
* ref: cern-sis/issues-inspire/issues/1647


ioannistsanaktsidis (approved):
LGTM

PR #4464, opened 2026-08-06
backoffice: inscrease size of source facet

EliseRecejac:
* ref: https://github.com/cern-sis/issues-inspire/issues/1635

### DIFF
--- a/backoffice/backoffice/hep/constants.py
+++ b/backoffice/backoffice/hep/constants.py
@@ -1,5 +1,8 @@
 from django.db import models
 
+FACET_SIZE_MEDIUM = 20
+FACET_SIZE_LARGE = 300
+
 HEP_TICKET_TYPES = (
     ("hep_create_curation", "HEP create curation"),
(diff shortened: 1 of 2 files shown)


backoffice/backoffice/hep/constants.py
```diff
@@ -1,5 +1,8 @@
 from django.db import models
 
+FACET_SIZE_MEDIUM = 20
+FACET_SIZE_LARGE = 100
```
ioannistsanaktsidis: What is the approximate size we are waiting ? Can we validate with Florian ?

EliseRecejac: yes I asked him but he didn't reply yet

EliseRecejac: I thought that above 100 it would be a problem anyway

ioannistsanaktsidis: Agreed, IMO even 100 is too much, but lets see!

ioannistsanaktsidis (approved):

PR #4465, opened 2026-08-06
workflows: mark withdrawn papers

DonHaul:
* ref: cern-sis/issues-inspire/issues/1464


DonHaul:
Mechanism required adding an addition 2 tasks
- buffer_not_arxiv - due to ariflow limitation regarding branches
- check_if_is_withdrawn - to check if the workflow was withdrawn in previous steps and finish it

image


An alternative implementation would to check already in the `check_is_arxiv_paper()` task if its withdrawn and skip to the end. although this would require two extras requests to arxiv

PR #4466, opened 2026-08-06
backoffice: add fallback if celery unavailable

DonHaul:
* ref: cern-sis/issues-inspire/issues/1649

### DIFF
--- a/backoffice/backoffice/hep/api/views.py
+++ b/backoffice/backoffice/hep/api/views.py
@@ -166,10 +166,22 @@ class HepWorkflowViewSet(BaseWorkflowViewSet):
             )
 
         workflow = serializer.save()
-        logger.info("Creating workflow with id: %s", str(workflow.id))
-        trigger_hep_workflow_initialization.delay(
-            str(workflow.id), workflow.workflow_type
-        )
+
+        def _initialize():
+            try:
+                trigger_hep_workflow_initialization.delay(
+                    str(workflow.id), workflow.workflow_type
+                )
+            except Exception:
+                logger.exception(
+                    "Broker unavailable, triggering DAG inline for workflow %s",
+                    workflow.id,
+                )
+                trigger_hep_workflow_initialization(
+                    str(workflow.id), workflow.workflow_type
+                )
+
+        transaction.on_commit(_initialize)
         return Response(serializer.data, status=status.HTTP_201_CREATED)
 
     @action(detail=True, methods=["post"])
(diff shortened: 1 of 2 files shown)


backoffice/backoffice/hep/api/views.py
```diff
@@ -167,9 +167,20 @@ def create(self, request):
...
         workflow = serializer.save()
         logger.info("Creating workflow with id: %s", str(workflow.id))
-        trigger_hep_workflow_initialization.delay(
-            str(workflow.id), workflow.workflow_type
-        )
+        try:
+            trigger_hep_workflow_initialization.delay(
+                str(workflow.id), workflow.workflow_type
+            )
+        except ConnectionError as e:
```
ioannistsanaktsidis: Not sure this is the case even if sentry says so, I think the exception raised underneath is operational error and this will never be caught

backoffice/backoffice/hep/api/views.py
```diff
@@ -167,9 +167,20 @@ def create(self, request):
 
         workflow = serializer.save()
         logger.info("Creating workflow with id: %s", str(workflow.id))
-        trigger_hep_workflow_initialization.delay(
-            str(workflow.id), workflow.workflow_type
-        )
+        try:
+            trigger_hep_workflow_initialization.delay(
```
ioannistsanaktsidis: I think even with what we have the workflow will still rollback due to the atomicity.

I believe to survive the outage we would need sth like 
```
workflow = serializer.save()

def _initialize():
    try:
        trigger_hep_workflow_initialization.delay(str(workflow.id), workflow.workflow_type)
    except OperationalError:
        logger.exception("Broker unavailable, triggering DAG inline for workflow %s", workflow.id)
        try:
            trigger_hep_workflow_initialization(str(workflow.id), workflow.workflow_type)
        except RequestException:
            logger.exception("Inline fallback failed for workflow %s", workflow.id)

transaction.on_commit(_initialize)
return Response(serializer.data, status=status.HTTP_201_CREATED)
```

no ? Could we check that a bit ?

ioannistsanaktsidis (changes requested):
Some things to be validated

backoffice/backoffice/hep/api/views.py
DonHaul: from what I see sentry mentions it being a `Connection Error` for this exception. Where do you see the top exception is `OperationalError`?
I can convert it to a catch-all except if we are not sure

backoffice/backoffice/hep/api/views.py
DonHaul: Indeed incase the in-request airflow trigger fails we would need to wrap it aswell if its done within the request.
If its done inside `on_commit`, the transaction is already commited so it shouldnt be rolled back even when it fails. 
I will check

DonHaul: I confirm it if any thing fails in `on_commit` the transaction is not rolled back, so no need for the nested try/except. changes made

ioannistsanaktsidis (approved):

PR #4468, opened 2026-08-06
backoffice: sanitize and try/except is_another_hep_submission_running

DonHaul:
* ref: cern-sis/issues-inspire/issues/1650

### DIFF
--- a/backoffice/backoffice/hep/utils.py
+++ b/backoffice/backoffice/hep/utils.py
@@ -119,7 +119,11 @@ def is_another_hep_submission_running(workflow):
     """Check whether a matching HEP submission has not completed."""
 
     index_name = settings.OPENSEARCH_INDEX_NAMES.get(settings.HEP_DOCUMENTS)
-    arxiv_eprints_values = get_value(workflow, "data.arxiv_eprints.value", [])
+    arxiv_eprints_values = [
+        value
+        for value in get_value(workflow, "data.arxiv_eprints.value", []) or []
+        if value
+    ]
     dois_values = get_value(workflow, "data.dois.value", [])
 
     if not arxiv_eprints_values and not dois_values:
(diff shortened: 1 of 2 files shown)


backoffice/backoffice/hep/utils.py
```diff
@@ -119,7 +119,11 @@ def is_another_hep_submission_running(workflow):
     """Check whether a matching HEP submission has not completed."""
 
     index_name = settings.OPENSEARCH_INDEX_NAMES.get(settings.HEP_DOCUMENTS)
-    arxiv_eprints_values = get_value(workflow, "data.arxiv_eprints.value", [])
+    arxiv_eprints_values = [
+        value
+        for value in get_value(workflow, "data.arxiv_eprints.value", []) or []
+        if value is not None
+    ]
     dois_values = get_value(workflow, "data.dois.value", [])
```
ioannistsanaktsidis, line 127: should we do the same for this one ?

backoffice/backoffice/hep/utils.py
```diff
@@ -119,7 +119,11 @@ def is_another_hep_submission_running(workflow):
     """Check whether a matching HEP submission has not completed."""
 
     index_name = settings.OPENSEARCH_INDEX_NAMES.get(settings.HEP_DOCUMENTS)
-    arxiv_eprints_values = get_value(workflow, "data.arxiv_eprints.value", [])
+    arxiv_eprints_values = [
+        value
+        for value in get_value(workflow, "data.arxiv_eprints.value", []) or []
+        if value is not None
```
ioannistsanaktsidis: why is not None and not just if value ?

ioannistsanaktsidis (changes requested):
few comments

backoffice/backoffice/hep/utils.py
DonHaul: not sure if it can happen. will mess around a bit with the form to see if it occurs as well. 
But I would be in favour of not adding it if we're not sure

DonHaul: dois are checked against a regex so this should never happen
image

ioannistsanaktsidis (approved):
LGTM

PR #4479, opened 2026-08-18
workflows: submissions should never be autorejected

DonHaul:
* ref: cern-sis/issues-inspire/issues/1653


ioannistsanaktsidis (approved):
LGTM

PR #4480, opened 2026-08-18
backoffice: force restart on latest dag version

DonHaul:
* ref: cern-sis/issues-inspire/issues/1658

### DIFF
--- a/backoffice/backoffice/common/airflow_utils.py
+++ b/backoffice/backoffice/common/airflow_utils.py
@@ -62,6 +62,7 @@ def clear_airflow_dag_tasks(dag_id, workflow_id, tasks=None):
         "only_failed": False,
         "reset_dag_runs": True,
         "include_downstream": True,
+        "run_on_latest_version": True,
     }
 
     if tasks:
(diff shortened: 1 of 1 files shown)


ioannistsanaktsidis (approved):
LGTM! 

##NOTE##

With `run_on_latest_version` on, resolving a parked workflow bumps the run to the latest DAG version, and verify_integrity creates a TI for every task the run lacks, upstream ones included. So a task added to preprocessing after a workflow parked will still run on resume.

DonHaul:
Checking `apache/airflow` when calling clear task instances
- [`post_clear_task_instances`](https://github.com/apache/airflow/blob/04145448135cfcc53644e8ca1a15fa188b129851/airflow-core/src/airflow/api_fastapi/core_api/routes/public/task_instances.py#L864) gets called
- [`dag.clear`](https://github.com/apache/airflow/blob/04145448135cfcc53644e8ca1a15fa188b129851/airflow-core/src/airflow/serialization/definitions/dag.py#L1346) is called and subsequently `clear_task_instances` which seems to simply set the dag version for the ti being created - [here](https://github.com/apache/airflow/blob/main/airflow-core/src/airflow/models/taskinstance.py#L451)

PR #4488, opened 2026-08-19
ui: upgrade antd to 5.29.3

EliseRecejac:
* ref: https://github.com/cern-sis/issues-inspire/issues/1655


EliseRecejac:
5th commit only

ui/src/common/components/EventStartDateFilter.jsx
```diff
@@ -53,10 +54,10 @@ function EventStartDateFilter({ onChange, selection = '', switchTitle }) {
               <Switch
                 checked={isUpcoming}
                 onChange={onUpcomingSwitchChange}
-                // if onChange is called before animation, it slows down & freezes the animation
+                // if onChange is called before the toggle transition, it slows down & freezes it
                 // because onChange() triggers search requests and eventually rendering aggs/results
                 // which is already a lot of events/task that have priority over animation in browsers
-                onAnimationEnd={onUpcomingSwitchAnimationEnd}
+                onTransitionEnd={onUpcomingSwitchTransitionEnd}
```
ioannistsanaktsidis, line 62: This now fires a search on Hover. 

https://github.com/user-attachments/assets/9acbbf70-23ec-4531-b359-e497508dbfd7

ioannistsanaktsidis (changes requested):
Few comments to start with, then we will re-review

This https://github.com/inspirehep/inspirehep/blob/master/ui/src/backoffice/authors/containers/AuthorDetailPageContainer.tsx#L185 should now me !!( .... ) so as to force to boolean.

ui/src/backoffice/common/components/Breadcrumbs/Breadcrumbs.less
```diff
@@ -15,3 +18,7 @@
 .search-bar-small:focus-within {
   width: 750px;
 }
+
+.ant-select {
```
ioannistsanaktsidis: lets scope this only in the backoffice, someone visiting the backoffice and then hep will get this evaluated

ui/src/common/components/RouterLinkButton.tsx
```diff
@@ -1,18 +1,28 @@
...
+  ...rest
+}: RouterLinkButtonProps) {
+  const handleClick = useLinkClickHandler<HTMLElement>(to);
+
   return (
-    <Link
-      className={classNames('ant-btn ant-btn-primary', className)}
-      {...linkProps}
-    />
+    <Button {...rest} type={type} block={block} onClick={handleClick}>
```
ioannistsanaktsidis, line 23: Maybe here we could also add href=useHref(to) so as to keep the middle button behavior or the right click open in a new tab ? can we test that ?

ui/package.json
```diff
@@ -108,7 +109,9 @@
   },
   "resolutions": {
     "moment-timezone/moment": "2.29.4",
```
ioannistsanaktsidis: with the upgrade moment is also kinda lost by default now correct? Should we pin it as a dependency so as to not be depending on hoisting ?

ioannistsanaktsidis (changes requested):
few comments

ui/src/common/components/RouterLinkButton.tsx
EliseRecejac: it's used at two places : the submit shortcuts at the bottom of home page and the backoffice dashboard, adding href only breaks the first one so I added it on the second one

ioannistsanaktsidis (approved):
No further comments by my side! LGTM

PR #4490, opened 2026-08-19
backend: lift submission flag for next

DonHaul:
* ref: cern-sis/issues-inspire/issues/1503

### DIFF
--- a/backend/tests/integration/submissions/test_submissions_views.py
+++ b/backend/tests/integration/submissions/test_submissions_views.py
@@ -607,9 +607,9 @@ def test_update_author_creates_new_workflow(inspire_app, override_config):
 
 
 @freeze_time("2019-06-17")
-def test_new_literature_submit_no_merge_next(inspire_app, requests_mock):
+def test_new_literature_submit_no_merge_backoffice(inspire_app, requests_mock):
     requests_mock.post(
-        f"{current_app.config['INSPIRE_NEXT_URL']}/workflows/literature",
+        f"{current_app.config['INSPIRE_BACKOFFICE_URL']}/api/workflows/literature/",
         json={"workflow_object_id": 30},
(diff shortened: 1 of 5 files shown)


backend/tests/integration/submissions/test_submissions_views.py
```diff
@@ -607,9 +607,9 @@ def test_update_author_creates_new_workflow(inspire_app, override_config):
 
 
 @freeze_time("2019-06-17")
-def test_new_literature_submit_no_merge_next(inspire_app, requests_mock):
+def test_new_literature_submit_no_merge_backoffice(inspire_app, requests_mock):
     requests_mock.post(
-        f"{current_app.config['INSPIRE_NEXT_URL']}/workflows/literature",
```
ioannistsanaktsidis, line 612: SInce we are removing this , should we also remove the config var ?

ioannistsanaktsidis (commented):
Also as a note, https://github.com/inspirehep/inspirehep/blob/abbddc9504278ab83bd430fc0c9a513a48ded78a/backend/inspirehep/config.py#L47 is pointing by default in beta. Any reason for that ?

DonHaul:
> Also as a note,
> 
> https://github.com/inspirehep/inspirehep/blob/abbddc9504278ab83bd430fc0c9a513a48ded78a/backend/inspirehep/config.py#L47
> 
> is pointing by default in beta. Any reason for that ?

no reason I've changed it for `http://localhost:8001` although it gets overriden to 'INVENIO_INSPIRE_BACKOFFICE_URL=http://host.docker.internal:8001' in `docker-compose.services.yml`

ioannistsanaktsidis (approved):
LGTM! Lets also remove this vars from kubernetes-inspire

PR #4494, opened 2026-08-20
workflows: remove logs in merge_articles

DonHaul:
* ref: cern-sis/issues-inspire/issues/1675


workflows/dags/literature/hep_create_dag.py
```diff
@@ -1393,17 +1395,15 @@ def merge_articles(matched_control_number, **context):
...
+                "Merged record with conflicts: %s",
+                len(conflicts),
             )
             workflow_data["data"] = merged
 
             workflow_data["merge_details"] = {
                 "head_uuid": str(head_uuid),
                 "head_version_id": head_version_id,
                 "merger_head_revision": head_revision_id,
-                "merger_original_root": head_root,
```
ioannistsanaktsidis, line 1406: no need of that one ?

workflows/dags/literature/hep_create_dag.py
ioannistsanaktsidis: What about this one ?

ioannistsanaktsidis (commented):
2 comments for clarification. Also, do you think we should have the result on s3 ?

ioannistsanaktsidis (approved):
LGTM

PR #4497, opened 2026-08-24
ui: get rid of remaining console warnings

EliseRecejac:
* ref: https://github.com/cern-sis/issues-inspire/issues/1668


ui/src/submissions/common/withFormItem.jsx
```diff
@@ -51,26 +51,30 @@ export default function withFormItem(FormInputComponent) {
...
         onlyChild,
+        addonBefore,
         ...props
       } = this.props;
       const renderedError = this.renderError();
       return (
         <Form.Item
           className={classNames({ 'mb2-important': onlyChild })}
-          hasFeedback={renderedError != null}
+          hasFeedback
```
ioannistsanaktsidis, line 61: this will always be true now no ?

ui/src/authors/components/PositionsTimeline.tsx
```diff
@@ -14,25 +14,28 @@ const PositionsTimeline = ({ positions }: { positions: List<any> }) => {
...
-      <Timeline.Item key={`#${displayDate}@${institution}`}>
-        <div>{displayDate}</div>
-        <div>
-          {rank && <strong>{rank}, </strong>}
-          <Affiliation affiliation={position} />
-        </div>
-      </Timeline.Item>
-    );
+    return {
+      children: (
```
ioannistsanaktsidis, line 24: Item still accepts a key. Should we add it back ?

ioannistsanaktsidis (commented):
2 comments for clarification

ui/src/submissions/common/withFormItem.jsx
EliseRecejac: in fact, if I'm not mistaken, renderedError can never be null so it was already always true

ui/src/authors/components/PositionsTimeline.tsx
EliseRecejac: indeed, it's strange that it didn't generate any error, maybe because it's wrapped in a function

ui/src/submissions/common/withFormItem.jsx
ioannistsanaktsidis: Not sure about that, note that 

```
 renderError() {
      const { field, form } = this.props;
      const { errors } = form;
      const { name } = field;
      const errorMessage = getIn(errors, name);
      return (
        errorMessage && (
          <span data-test-id={`${field.name}-error`}>{errorMessage}</span>
        )
      );
    }
```

this can return undefined. and `!= null` in js checks against null or undefined

EliseRecejac: indeed I forgot that !

EliseRecejac: my mistake actually I put it always at true to get rid of this error
```
Warning: [antd: Input] When Input is focused, dynamic add or remove prefix / suffix will make it lose focus caused by dom structure change. Read more: https://ant.design/components/input/#FAQ Error Component Stack
    at antd.js?v=34b2d107:53722:16
    at CompactItem (antd.js?v=34b2d107:8139:5)
    at div (<anonymous>)
    at Compact (antd.js?v=34b2d107:8149:15)
    at StatusProvider (antd.js?v=34b2d107:62195:3)
    at div (<anonymous>)
    at div (<anonymous>)
    at div (<anonymous>)
    at antd.js?v=34b2d107:49975:16
    at FormItemInput (antd.js?v=34b2d107:61990:5)
    at div (<anonymous>)
    at antd.js?v=34b2d107:50117:16
    at div (<anonymous>)
    at ItemHolder (antd.js?v=34b2d107:62266:5)
    at InternalFormItem (antd.js?v=34b2d107:62426:5)
    at WithFormItem (withFormItem.jsx?t=1789476329516:15:3)
    ```

Comment (EliseRecejac) [reply ui/src/submissions/common/withFormItem.jsx:61]: in fact it doesn't change anything visually

File: ui/src/authors/components/PositionsTimeline.tsx (hunk 2 above)
Comment (ioannistsanaktsidis) [reply ui/src/authors/components/PositionsTimeline.tsx:24]: it fallsback internally to the index , but it is better to keep the key

File: ui/src/submissions/common/withFormItem.jsx (hunk 1 above)
Comment (EliseRecejac) [reply ui/src/submissions/common/withFormItem.jsx:61]: because if there is no error validateStatus is empty

Comment (EliseRecejac) [reply ui/src/submissions/common/withFormItem.jsx:61]: Sorry wasn't clear I needed to rememorate things :
hadFeedback displays an icon is validateStatus is not empty, since it's filled only if there is an error the result is the same but setting it as always true enables to get rid of the console warning

Comment (ioannistsanaktsidis) [reply ui/src/submissions/common/withFormItem.jsx:61]: I see , cool!

File: ui/src/authors/components/PositionsTimeline.tsx [hunk 3]
```diff
@@ -14,25 +14,29 @@ const PositionsTimeline = ({ positions }: { positions: List<any> }) => {
...
     const rank = position.get('rank');
     const displayDate = position.get('display_date');
 
-    return (
-      <Timeline.Item key={`#${displayDate}@${institution}`}>
-        <div>{displayDate}</div>
-        <div>
-          {rank && <strong>{rank}, </strong>}
-          <Affiliation affiliation={position} />
+    return {
```
ioannistsanaktsidis, line 22: i think the key needs to be on top level, next to the children 
{
key: 
children:
}

ioannistsanaktsidis (approved):
LGTM

PR #4502, opened 2026-08-25
ui: replace moment by dayjs

EliseRecejac:
* ref: https://github.com/cern-sis/issues-inspire/issues/1663


EliseRecejac:
2nd commit only

ui/src/seminars/components/__tests__/SeminarTimezone.test.jsx
```diff
@@ -6,7 +6,7 @@ describe('SeminarTimezone', () => {
   it('renders with timezone', () => {
     advanceTo('2020-09-10');
     const { getByText } = render(<SeminarTimezone timezone="Europe/Zurich" />);
-    expect(getByText('Times in Europe/Zurich (CEST)')).toBeInTheDocument();
+    expect(getByText('Times in Europe/Zurich (GMT+2)')).toBeInTheDocument();
```
ioannistsanaktsidis, line 9: So basically now this renders based on GMT. Could we check with Micha if thats ok ?

EliseRecejac: @michamos is it ok ?

ui/src/submissions/common/schemas/date.js
```diff
@@ -1,8 +1,22 @@
...
+
+dayjs.extend(customParseFormat);
+
+const FREE_FORM_DATE_FORMATS = [
+  'YYYY-MM-DD',
+  'YYYY-MM',
+  'YYYY',
+  'D MMMM YYYY',
+  'MMMM YYYY',
+];
```
ioannistsanaktsidis, line 13: This is now restricting to these formats. Theoritically it is ok, as it only applies to the free text dates (e.g thesis info). 2 questions to validate with product. 

1. Why is not a picker used there? 
2. Is this list enough ? placeholder has only YYYY-MM-DD, YYYY-MM or YYYY . Difference to this one is that before date() with no format called moment(value, undefined, false).isValid(). With no format, moment tries ISO 8601, then RFC 2822, then falls back to new Date(value). Anything the browser's Date parser accepts passed. Not really an issue, but we need to validate if we need the strict formats. If yes, then what you added is actually making the system better

ui/src/submissions/common/schemas/date.js
```diff
@@ -1,8 +1,22 @@
...
 
 function getDateValidatorFor(format) {
-  return (value) => moment(value, format, Boolean(format)).isValid();
+  const formats = format ? [format] : FREE_FORM_DATE_FORMATS;
+  return (value) =>
+    value == null ||
```
ioannistsanaktsidis, line 18: The value == null guard is fine on its own, but it changes one caller: deadline_date in submissions/jobs/schemas/job.js has no .required() on the status: 'closed' branch. On master that branch still rejected an empty deadline only because moment(undefined, format, true) is invalid; with the guard it now passes.

Repro: new job form → status closed → fill everything except Deadline → submit. Master shows Does not match with YYYY-MM-DD inline; this branch submits and the backend returns 400 Validation error. (deadline_date is required in jobs.json).

Possible fix in job.js to update the "then" part to `date(DATE_RANGE_FORMAT).required().label('Deadline')`,
Closed jobs still accept a past deadline; the existing job schema tests pass with this change.

Could you pls check that too ?

ui/src/seminars/components/__tests__/SeminarTimezone.test.jsx
EliseRecejac: checked with him it's ok

ui/src/submissions/common/schemas/date.js
EliseRecejac: the list matches what is currently tested, I checked with Micha we could restrict it to match the placeholder but it's not an issue to keep as it is, so I think we could just keep that

ioannistsanaktsidis (approved):
LGTM

PR #4504, opened 2026-08-26
ui: fix connectToHistory for piwik

EliseRecejac:
* ref: https://github.com/cern-sis/issues-inspire/issues/1687

### DIFF
--- a/ui/src/tracker.js
+++ b/ui/src/tracker.js
@@ -18,7 +18,25 @@ export function injectTrackerToHistory(history) {
       url: PIWIK_URL,
       siteId: Number(PIWIK_SITE_ID),
     });
-    return piwik.connectToHistory(history);
+    // react-piwik@1.12.0's published build forwards the raw `listen` argument
+    // straight to `track()`, which reads `pathname`/`search` off it. But
+    // redux-first-history (history v5) emits `{ location, action }`, so Piwik
+    // would compute `undefined + undefined` and crash on `.replace`. Hand it a
+    // history whose `listen` unwraps that object into a bare location.
+    const historyForTracker = {
+      get location() {
+        return history.location;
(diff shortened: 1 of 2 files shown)


ui/src/tracker.js
```diff
@@ -18,7 +18,18 @@ export function injectTrackerToHistory(history) {
       url: PIWIK_URL,
       siteId: Number(PIWIK_SITE_ID),
     });
-    return piwik.connectToHistory(history);
+    // react-piwik@1.12.0's published build forwards the raw `listen` argument
+    // straight to `track()`, which reads `pathname`/`search` off it. But
+    // redux-first-history (history v5) emits `{ location, action }`, so Piwik
+    // would compute `undefined + undefined` and crash on `.replace`. Hand it a
+    // history whose `listen` unwraps that object into a bare location.
+    const historyForTracker = {
```
ioannistsanaktsidis, line 26: for the location maybe a getter is better. sth like 

```
get location() {
        return history.location;
      },
```

ui/src/tracker.js
```diff
@@ -18,7 +18,18 @@ export function injectTrackerToHistory(history) {
...
+      listen: (callback) =>
+        history.listen((update) =>
+          callback(update && update.location ? update.location : update)
+        ),
+    };
```
ioannistsanaktsidis, line 38: I would say this is better to wrpa in a try catch (if for some reason the tracker throws the whole page will freeze (bug we had before) 
```
try {
            callback(update && update.location ? update.location : update);
     } catch {

          }
```

ioannistsanaktsidis (changes requested):
2  small comments to check

ioannistsanaktsidis (approved):
LGTM

PR #4506, opened 2026-08-27
ui: redispatch actions on loggedIn change

EliseRecejac:
* ref: https://github.com/cern-sis/issues-inspire/issues/1688


ioannistsanaktsidis (commented):
It looks good, do you think we should a couple of tests for that one ? Currently there are none

ioannistsanaktsidis (approved):

PR #4509, opened 2026-08-28
ui: handle bar click on safari

EliseRecejac:
* ref: https://github.com/cern-sis/issues-inspire/issues/1699

### DIFF
--- a/ui/eslint.config.mjs
+++ b/ui/eslint.config.mjs
@@ -61,7 +61,7 @@ const coreRules = {
   'no-alert': 'warn',
   'no-await-in-loop': 'error',
   'no-bitwise': 'error',
-  'no-console': 'warn',
+  'no-console': ['error', { allow: ['error'] }],
   'no-continue': 'error',
   'no-else-return': ['error', { allowElseIf: false }],
   'no-eval': 'error',
--- a/ui/src/common/components/CitationSummaryGraph/CitationSummaryGraph.less
+++ b/ui/src/common/components/CitationSummaryGraph/CitationSummaryGraph.less
@@ -1,4 +1,12 @@
 .__CitationSummaryGraph__ {
+  .citeable-bar rect:hover {
+    fill: @blue-7;
+  }
+
+  .published-bar rect:hover {
+    fill: @orange-7;
+  }
+
   .recharts-cartesian-axis-tick-value {
     fill: @text-color;
   }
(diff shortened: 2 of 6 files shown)


ioannistsanaktsidis (changes requested):
I would actually follow a different approach into that one. 

The cause is that the hover colour comes from React state, so every hover re-renders <Bar> and recharts rebuilds the bar <rect>s. The rect that got mousedown is detached by mouseup, and WebKit then fires no click at all. Some browsers tolerate that. some not

Why don't we  make it match RangeAggregation, which already does this correctly:

In CitationSummaryGraph.tsx we can delete hoveredBar state, the three hover handlers, onMouseEnter/onMouseLeave on both `<Bar>`s, and the HOVERED_* exports. 

Then we can handle everything in .less like 
 .citeable-bar rect:hover { fill: @ blue-7 }, .published-bar rect:hover { fill: @ orange-7 }.

ui/eslint.config.mjs
```diff
@@ -61,7 +61,7 @@ const coreRules = {
   'no-alert': 'warn',
   'no-await-in-loop': 'error',
   'no-bitwise': 'error',
-  'no-console': 'warn',
+  'no-console': ['error', { allow: ['error'] }],
```
EliseRecejac, line 64: this one is a bit of a side quest which I don't think deserve its own issue, it's just to get rid of some unwanted warnings

ui/src/common/components/CitationSummaryGraph/CitationSummaryGraph.less
```diff
@@ -1,4 +1,12 @@
 .__CitationSummaryGraph__ {
+  .citeable-bar rect:hover {
+    fill: #096dd9 !important;
+  }
+
+  .published-bar rect:hover {
+    fill: #d46b08 !important;
```
ioannistsanaktsidis: Do we actually need !important for that one ? I think it is ok without it. Can we check ?

EliseRecejac: no you are right

ui/src/common/components/CitationSummaryGraph/CitationSummaryGraph.less
```diff
@@ -1,4 +1,12 @@
 .__CitationSummaryGraph__ {
+  .citeable-bar rect:hover {
+    fill: #096dd9 !important;
```
ioannistsanaktsidis: we can also use the vars instead no blue-7 , orange-7 ?

ioannistsanaktsidis (approved):
LGTM

PR #4513, opened 2026-08-31
backoffice: raise  limit on celery only for batch

ioannistsanaktsidis:
* ref https://github.com/cern-sis/issues-inspire/issues/1672


EliseRecejac (approved):
LGTM

PR #4514, opened 2026-08-31
backend: decode correctly arxiv response

EliseRecejac:
* ref: https://github.com/cern-sis/issues-inspire/issues/1640

### DIFF
--- a/backend/inspirehep/records/api/literature.py
+++ b/backend/inspirehep/records/api/literature.py
@@ -705,6 +705,7 @@ def import_arxiv(arxiv_id):
         return {}
 
     try:
+        response.encoding = "utf-8"
         parser = ArxivParser(response.text)
         return parser.parse()
     except Exception as exc:


backend/inspirehep/records/api/literature.py
```diff
@@ -705,6 +705,7 @@ def import_arxiv(arxiv_id):
         return {}
 
     try:
+        response.encoding = "utf-8"
```
ioannistsanaktsidis, line 708: Can we also have a regression test based on the example provided in the ticket ?

ioannistsanaktsidis (changes requested):

backend/tests/integration/records/api/test_api_literature.py
```diff
@@ -549,6 +553,18 @@ def test_import_article_doi_already_in_inspire(inspire_app):
         import_article(doi_value)
 
 
+def test_import_arxiv_decodes_response_as_utf8(inspire_app):
+    arxiv_id = "2607.20148"
+
+    result = import_arxiv(arxiv_id)
```
ioannistsanaktsidis: aas this is calling directly the arxiv api, we would need to record a cassette. You can check how similar tests are doing it. Also better to use import_article as the rest .

ioannistsanaktsidis (changes requested):

ioannistsanaktsidis (approved):
LGTM

PR #4517, opened 2026-09-01
e2e: make references tests more robust

EliseRecejac:
* ref: https://github.com/cern-sis/issues-inspire/issues/1703


ioannistsanaktsidis (changes requested):
Should we apply the same for the other three siblings too ?

EliseRecejac:
> Should we apply the same for the other three siblings too ?

we could but I'm not sure it's really relevant since only this one was failing

ioannistsanaktsidis:
Ok , will approve as is then and we will monitor

ioannistsanaktsidis (approved):
LGTM

PR #4551, opened 2026-09-09
ui: add safeguard to tests and fix potential errors

EliseRecejac:
* ref: https://github.com/cern-sis/issues-inspire/issues/1710


ui/src/fixtures/render.tsx
```diff
@@ -5,6 +5,11 @@ import { Provider } from 'react-redux';
 
 import { getStore } from './store';
 
+const routerFutureFlags = {
+  v7_startTransition: true,
+  v7_relativeSplatPath: true,
```
ioannistsanaktsidis, line 10: what are these ones ?

ui/src/common/components/CheckboxAggregation.jsx
```diff
@@ -167,7 +167,10 @@ class CheckboxAggregation extends Component {
     const { name, buckets } = this.props;
     return (
       <AggregationBox name={name}>
-        {buckets.take(maxBucketCountToDisplay).map(this.renderBucket)}
+        {buckets
+          .filter((bucket) => bucket.get('key') != null)
```
ioannistsanaktsidis: I would fix the fixture and not the component here.

ui/src/common/components/RangeAggregation/RangeAggregation.jsx
```diff
@@ -112,6 +112,10 @@ function getSanitizedEndpointsFromSelections(selections, minMaxPair) {
 }
 
 function getSliderMarks([lower, upper], [min, max]) {
+  if (!Number.isFinite(lower) || !Number.isFinite(upper)) {
+    return {};
+  }
```
ioannistsanaktsidis: If this is added only to silence the test part, I would prefer fixing it on the fixture level

ioannistsanaktsidis (changes requested):
Some commnets/clarifications

ui/src/fixtures/render.tsx
EliseRecejac: it enables to get rid of these warnings :
```
⚠️ React Router Future Flag Warning: React Router will begin wrapping state updates in `React.startTransition` in v7. You can use the `v7_startTransition` future flag to opt-in early. For more information, see https://reactrouter.com/v6/upgrading/future#v7_starttransition.
⚠️ React Router Future Flag Warning: Relative route resolution within Splat routes is changing in v7. You can use the `v7_relativeSplatPath` future flag to opt-in early. For more information, see https://reactrouter.com/v6/upgrading/future#v7_relativesplatpath.
```
I think the goal is to ease the future transition to v7, also it's only used in the tests because it's the only place where we use MemoryRouter

ui/src/common/components/__tests__/AggregationFilters.test.jsx
```diff
@@ -10,6 +10,7 @@ describe('AggregationFilters', () => {
         buckets: [
           {
             key: 'foo',
+            key_as_string: '2012',
```
ioannistsanaktsidis, line 13: should we apply this to the initialaggregations also ?

ioannistsanaktsidis (approved):
LGTM

PR #4560, opened 2026-09-15
e2e: remove snapshots from e2e tests

EliseRecejac:
* ref: https://github.com/cern-sis/issues-inspire/issues/1711


e2e/cypress.config.js
```diff
@@ -11,16 +11,13 @@ module.exports = defineConfig({
     mobile_viewport_width: 375,
     mobile_viewport_height: 667,
```
ioannistsanaktsidis: these 2 are now dead no ?

ioannistsanaktsidis (changes requested):
We will need to regenerate the lock file for the e2e also

ioannistsanaktsidis (approved):
LGTM

PR #4565, opened 2026-09-17
e2e: remove cy.wait and force true

EliseRecejac:
* ref: https://github.com/cern-sis/issues-inspire/issues/1727


e2e/cypress/e2e/references.test.js
```diff
@@ -165,7 +164,6 @@ describe("Reference container", () => {
         .find('[data-test-id="edit-reference"]')
```
ioannistsanaktsidis: Should this match the button and not the wrapper ?

disabled={!loggedIn} is on the inner <button class="ant-btn">, not the data-test-id="edit-reference" wrapper div.

loggedIn only becomes true after /api/accounts/me returns; the page renders (and waitForLoading passes) before that.

this line still aliases the wrapper div, so should("not.be.disabled") passes instantly and the click hits a disabled button → drawer never opens → next cy.get times out.

The old cy.wait(3000) was the only thing that let /accounts/me finish first. The other three sites got the .ant-btn fix; this one just lost the sleep.
Fix: .find('[data-test-id="edit-reference"] .ant-btn').

ioannistsanaktsidis (changes requested):
Can we check this one ?

e2e/cypress/e2e/literatureRelations.test.js
```diff
@@ -51,58 +51,57 @@ describe('Literature and Conferences', () => {
...
-      cy.get('@literature-title').should('be.oneOf', titles);
+      cy.get("@literature-title").should("be.oneOf", titles);
     });
   });
 });
 
-describe('Export to CDS', () => {
-  it('check if record has CDS:true', () => {
-    cy.login('admin');
+describe("Export to CDS", () => {
```
ioannistsanaktsidis, line 67: I am also wondering how this works correctly. Export to CDS is an async task, that sometimes might take a bit of time to finish (since it runs on the background) . Can we validate it ?

EliseRecejac: good catch, there are no real assertions in the test... (they pass anyway) 🤦‍♀️

e2e/cypress/e2e/literatureRelations.test.js
```diff
@@ -1,45 +1,45 @@
-import _ from 'lodash';
+import _ from "lodash";
```
ioannistsanaktsidis: no need for that anymore right ?

e2e/cypress/e2e/seminars.test.js
```diff
@@ -10,7 +10,7 @@ describe("Seminar Submission", () => {
     const endDateMoment = moment("2020-05-06 14:30");
     const formData = {
       name: "The Cool Seminar",
-      timezone: "Europe/Zurich",
+      timezone: "Europe/Vatican",
```
ioannistsanaktsidis, line 13: why this was changed ?

ioannistsanaktsidis (commented):
2 small comments

e2e/cypress/e2e/seminars.test.js
EliseRecejac: for this one I'm not sure why it was working before, because the problem is that Europe/Zurich is the default value so you cannot select it in the dropdown (I mean you can but it doesn't make sense + you have two values Europe/Zurich in the dom so there is an error), by changing it I can really check the selection, the other solution would be to remove it completely from formData

ioannistsanaktsidis (approved):
LGTM

PR #4567, opened 2026-09-18
backoffice: restore clears merge_details

EliseRecejac:
* ref: https://github.com/cern-sis/issues-inspire/issues/1717


backoffice/backoffice/hep/api/views.py
```diff
@@ -311,6 +311,7 @@ def restore(self, request, pk=None):
                 status=status.HTTP_400_BAD_REQUEST,
             )
         workflow.decisions.all().delete()
+        workflow.merge_details = None
```
ioannistsanaktsidis: Not in scope of this PR(my bad) , but maybe it is actually good to clear everything on restore. e.g 

```
workflow.matches = None
workflow.merge_details = None
workflow.classifier_results = None
workflow.relevance_prediction = None
workflow.reference_count = None
workflow.journal_coverage = None
```

@DonHaul  any opinions on this ?

DonHaul: makes sense so it become a true hard restart

ioannistsanaktsidis (approved):
LGTM

PR #4570, opened 2026-09-21
e2e: fix config for celery in hep-worker

EliseRecejac:
* ref: https://github.com/cern-sis/issues-inspire/issues/1731


ioannistsanaktsidis (approved):
LGTM

